PR #147, opened 2018-02-02
Migrate to SimpleMultiSig

elenadimitrova:
- Remove `multisig-wallet` ethpm package as that repo no longer contains the source we use. Contract we use has been moved to https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol 
- Replace MultiSig implementation
Considering Gnosis MultiSigWallet has unnecessary (for us) implementation complexity we chose to switch to  a simpler, MIT licensed MultiSig from `simple-multisig` repo https://github.com/christianlundkvist/simple-multisig 

Relevant EIPs to MultiSig standards:
https://github.com/ethereum/EIPs/issues/763
https://github.com/ethereum/EIPs/issues/191

SimpleMultiSig release article
https://medium.com/@ChrisLundkvist/exploring-simpler-ethereum-multisig-contracts-b71020c19037


area:
Yeah, Elena is right - I was mistaken in thinking that the threshold for adding/removing signatories was different from that required to do other things. So this applies with either implementation.

area:
Other differences between the two implementations: 

* With simplemultisig, the gas costs are placed on one person (though they are lower in total, I suspect)
* It will be (I believe) technically more difficult to use. We might need to cook up a (very) simple client.

elenadimitrova:
Some concrete gas costs
`SimpleMultiSig` only ever needs one transaction (which contains all signatures)
```
multisig.execute -> 75,413 gas
```

`MultiSigWallet` needs an initial transaction submission and then as many confirmations as required, i.e. the `confirmTransaction` 50,110 gas cost should be multiplied by the required number of approvals -1.
```
multisig.submitTransaction -> 167,202 gas
multisig.confirmTransaction -> 50,110 gas (not enough signatures gathered)
multisig.confirmTransaction -> 79,857 gas (enough signatures gathered, execute the tx)
```

elenadimitrova:
After a discussion between @area @jakzilla and myself we decided to shelve this and stay on the Gnosis MultiSig for now at least. Considerations include:
- gas costs aren't a concern as this will be used for the occasional contract upgrades only
 - using a proven to be (so far) secure contract on the foundation network
In future I believe **EIP 763** will adopt the SimpleMultisig design as a standard which will drive adoption and encourage more work on formal proof and official security reviews. At which point we're likely to reconsider it.

PR #155, opened 2018-02-13
Return funds when a task is cancelled

### LINKED ISSUE #22: Task cancellation
### 4.1 Tasks
If no worker has been assigned to a task, the manager has the ability to cancel the task entirely.

Any funds that have already been assigned to the task are assigned to the domain that the task was created in.

sekl:
Implemented refunding task budget to domain part of #22

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -330,9 +330,13 @@ contract ColonyTask is ColonyStorage, DSMath {
     uint256 taskTokenBalance = taskPot.balance[token];
     uint256 taskEtherBalance = taskPot.balance[0x0];
 
-    // Return task funds to the domain
-    IColony(this).moveFundsBetweenPots(taskPotId, domainPotId, taskTokenBalance, token);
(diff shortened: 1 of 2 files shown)


contracts/ColonyTask.sol
```diff
@@ -321,6 +321,22 @@ contract ColonyTask is ColonyStorage, DSMath {
...
+
+    // Get task and domain pots so we can return the funds
+    uint256 taskPotId = tasks[_id].potId;
+    Pot storage taskPot = pots[taskPotId];
+    uint256 domainId = tasks[_id].domains[0];
+    uint256 domainPotId = domains[domainId].potId;
+    uint256 taskTokenBalance = taskPot.balance[token];
+    uint256 taskEtherBalance = taskPot.balance[0x0];
+
+    // Return task funds to the domain if a pot is not empty
```
elenadimitrova: The task can actually be funded by a number of different (whitelisted) tokens so this approach won't scale well. Instead, after a task is cancelled, you should call `moveFundsBetweenPots` explicitly for as many types of funding there're are e.g this is a change that probably doesn't need any contract updates just a test to prove it works.

Issue #50, opened 2017-09-26
Dealing with false submissions - same hash, different number of leaves


elenadimitrova:
### 7.4 Submission of a new root hash
_Verifying a submission_
If only one state is submitted by the end of the submission period, then the new state is accepted, and proposals of the next state can begin to be made. This is expected to be the most common occurrence.
If more than one state has been submitted, then either someone has made a mistake, or there is a malicious entity trying to introduce a fraudulent reputation change. In this event, the a challenge-response protocol can establish which state is incorrect (see Section 7.5)

### 7.5.1 The Challenge-Response Protocol
We assume that the correct hash is one of the submitted hashes. This is a reasonable assumption, as only one out of all the miners is required to make a correct submission, and there is an incentive for them to do so (the reward defined in Section 7.8). Thus our task is not to validate the correct hash but to invalidate the false ones.
We must prove all but one submission incorrect by having each submission navigate a series of challenges. These challenges refer to events that happened in the colony network within the last update cycle that have an effect on reputation. The responses to the challenges are Merkle proofs that the corresponding reputation update was properly handled. Anyone is able to respond to a challenge, regardless of who submitted the original hash; this should ensure that the correct state is always defended.
We begin with the scenario where only two submissions are made, and one is correct.
First we consider the case in which the same hash has been submitted twice, but with a disagreement about the number of leaves it contained. In this situation, the users are required to submit a Merkle proof for the last leaf in their tree. We are able to tell if a submitted proof corresponds to the last leaf in the tree, as all sibling hashes must be hashed on the left. Noting the positions in the Merkle proof in which a hash is hashed with itself also allows us to tell the index of this last leaf; if it corresponds to either of the submitted states, that one is correct.

![screen shot 2017-09-26 at 11 01 27](https://user-images.githubusercontent.com/703848/30849148-11bc6032-a2aa-11e7-883d-78a54cf61b74.png)

The more complicated case where two different hashes have been submitted is considered in #49

Issue #49, opened 2017-09-26
Dealing with false submissions - different hashes


elenadimitrova:
### 7.5.1 The Challenge-Response Protocol
We now consider the more complicated case where two different hashes have been submitted.
1. The Justification Tree
The first step is for both parties to upload a justification of their submitted hash. This justification consists of a second Merkle root and two Merkle proofs.
The Merkle root in questions is the `JustificationRootHash` (`JRH`); it is the root of the ‘Justification Tree’ – a tree where each leaf represents a complete reputation state i.e. each leaf is a `ReputationRootHash`.
The left-most leaf of the Justification Tree (`RH0`) is the final accepted reputation state from the last update. Both parties must submit a proof that their tree does indeed start at `RH0`. Note that in a Merkle proof for a left-most leaf, all siblings are hashed on the right.
The right-most leaf of the Justification Tree (`RHn`) is the `ReputationRootHash` they originally submitted. Both parties must submit a Merkle proof that their tree does indeed end with this hash. 
Note that in a Merkle proof for a right-most leaf, all siblings are hashes on the left; furthermore, by noting the steps at which a hash is hashed with itself, we can determine the index of the last leaf.
This index is required to be `n` — the number of reputation updates in the log.
The intermediate leaves of the Justification Tree represent the evolution of the reputation state, with `RHi` corresponding to the reputation state after the first `i` reputation updates in this cycle
have been applied.
An example of such a tree is shown in Figure 4. Note that in the first stage of the tree, every neighbouring pair of data leaves is hashed, and so any pair has a shared Merkle proof to the root. We do this so that each element in the first row of hashes (indicated by `h0;1`; `h0;2` : : : `hn-1;n` in Figure 4) represents a transition between two states. It is these transitions that are the subject of the mining dispute resolution process.
Since any two differing submitted states agree on the first leaf `RH0` (which is the `ReputationRootHash` accepted at the end of the previous iteration of the mining process), and disagree on the last leaf `RHn` (the hash they submitted), there must be a hash (`hi;i+1`) that represents the transition from `RHi` to `RHi+1`) where they agree on the starting state but disagree on the result. This transition is meant to be the effect of a single reputation update (the `i-th` - We start counting at zero, i.e. the transition from `RH0` to `RH1` is the ‘zeroth’.), and this is the reputation update we will calculate on-chain to establish which submission is incorrect.
First, however, we must establish where the two submissions first differ, see #48

Issue #47, opened 2017-09-26
Submission of a new root hash - eligibility calculation


elenadimitrova:
### 7.4 Submission of a new root hash
*Who can submit a new root hash?*
All CLNY token holders are eligible to become miners and participate in the reputation update process, but since any user can calculate the correct root hash locally, it would be possible for any
miner to submit the hash to the contract. 
It is however undesirable to have too many submissions for every update. We propose a mechanism
that only allows some miners to submit results to begin with. To participate in the mining process, Common Colony Token holders must stake some of their tokens to become ‘reputation miners’. A submission will only be accepted from a miner if 
```
keccak256(address, N, hash) < target
```
At the beginning of the submission window, the target is set to `0` and slowly increases to `2^256 - 1` after `150` blocks. We limit the total number of miners allowed to submit a specific hash to `12`. The variable `N` that goes into the hash is some integer greater than `0` and less than the number
of tokens the Common Colony Token holder address has staked divided by `10^15`, meaning that users with a large stake have a higher chance of qualifying to submit a hash sooner than smaller stake holders. The factor of `10^15` is introduced to ensure that all hashes a user is eligible to submit can be calculated in a few seconds by the client. It also effectively creates a minimum number of tokens that must be staked to submit a hash. This puts a tangible cost on any attacks revolving around spamming known false submissions (see Section 7.7).
Miners will also be required to have staked their tokens for a few update cycles before they are eligible to submit or support a hash.

elenadimitrova:
@area for colony contribute launch considering the limit on CLNY transfers we probably don't need the remaining logic defined here?, e.g.:
>Miners will also be required to have staked their tokens for a few update cycles before they are eligible to submit or support a hash.

area:
In the cold light of nearly a year later, we can't think of a reason why you need to have staked for a few cycles, just since before the start of the current cycle (to avoid people withdrawing and depositing in a new address that they may be able to submit from earlier). This is much easier to implement, so unless we come up with a reason why you need to have staked for multiple cycles, we're just going to do that.

kronosapiens:
Hi, it is not entirely clear to me what work remains to be done here -- the entry submission code already uses the stake to determine eligibility. If we are omitting the need for a minimum staking period then... ?

kronosapiens:
Noting: this will be implemented by adding a `lastStaked` timestamp to the token locking contract, and updating it whenever a deposit (but not a withdrawal) is made.

[...]

Issue #43, opened 2017-09-20
Reputation update log


elenadimitrova:
### 6.4 Reputation update log
Whenever an event that causes one or more users to have their reputation updated in a colony occurs, a corresponding entry is recorded in a log in the `ColonyNetwork` contract. Each entry in the log contains:
 - The user suffering the reputation loss or gain.
 - The amount of reputation to be lost or gained.
 - The colony the update has occurred in.
 - How many reputation entries will need to be updated (including parent, child and colony-wide total reputations). This is the motivation for storing `n_parents` and `n_children` for each skill and domain, as described in Section 6.3.
 - How many total updates to reputations have occurred before this one in this cycle, including
decays and updates to parents and children.

If the reputation update is the result of a dispute being resolved (as outlined in Section 6.2.2), then instead of these first three properties, there is a reference to the dispute-specific record of stakes in the relevant colony. For the structure of this log, and an explanation of the way that it allows individual updates to be extracted in constant gas, see Appendix A.

This log exists to define an ordering of all reputation updates in a reputation update cycle that is accessible on-chain. In the event of a dispute during the reputation mining protocol (described in Section 7), the `ColonyNetwork` contract can use this record to establish whether an update has been included correctly.

kronosapiens:
Already implemented

PR #156, opened 2018-02-14
Introduce staking and submission of Reputation Root Hashes, with skeleton of dispute mechanism

### LINKED ISSUE #56: Costs and rewards of Mining
### 7.4 Submission of a new root hash
_Mining Rewards_
When a state is accepted, a number of (newly minted) Common Colony Tokens are made available for the users who submitted the correct state to claim as a reward. They also receive a corresponding amount of reputation in the Common Colony (in a special mining skill, which only users in the Common Colony can earn by performing this task). This reputation update is no different from any other, aside from the limitations of who is able to earn it, and will be included in the subsequent reputation update cycle. The size of the rewards and their distribution are described in Section 7.8.

area:
Closes #56. The functionality removed from #56 can be found in #164. 

Adds the skeleton for #49 and #50 - currently dispute resolution is mocked by a simple function call `invalidateHash`.
* Implements everything, I think, from #47, save for
>  Miners will also be required to have staked their tokens for a few update cycles before they are eligible to submit or support a hash.
* Reimplements some of #43 (see below)

This PR introduces the concept of staking colony tokens, which then allows their owners to take part in the reputation mining process. I include a skeleton of the dispute process, though the disputes are not actually validated - as mentioned, a call to `invalidateHash` is used to 'time out' an opponent.

The first change to existing functionality is that there are now two reputation update logs. Only one of these is active at a time, where the active reputation log is appended to if new tasks are completed or reputation will need to change for any reason. When a new reputation mining cycle begins, this reputation log is frozen so that it is not a 'moving target' for the reputation miners, and the other reputation log is cleared from the previous reputation mining cycle and is set to active. All functions that interact with the reputation logs directly have a `activeLog` parameter, that determines whether the active log is to be read from or the inactive log is to be read from. Any functions that wrote to the reputation update log always write to the active log. 

The new storage variables introduced to ColonyNetwork are
```
uint256 activeReputationUpdateLog; // Tracks whether updateLog 0 or 1 is the active log
bytes32 reputationRootHash; // The reputation root hash of the reputation state tree accepted at the end of the last completed update cycle
mapping (address => uint) stakedBalances; // Mapping containing how much hash been staked by each user
address reputationMiningCycle; // Address of the currently active reputation mining cycle contract
uint256 reputationRootHashNNodes; // The number of nodes in the reputation state tree that was accepted at the end of the last mining cycle
```

The new functionality for the `ColonyNetwork` contract is implemented in `ColonyNetworkStaking.sol`. Note that this also contains a *separate* contract, called `ReputationMiningCycle`. A new instance of this is deployed at the start of each mining cycle by the `ColonyNetwork` contract. This was primarily motivated for gas cost reasons; we do not need to keep the history of all the reputation mining cycles, and so by having this in a separate contract we can `selfdestruct` at the end of each mining cycle to cleanly free all the storage that has been used. 

Miners interact with the `ReputationMiningCycle` contract to submit their own hashes or object to others. A submission takes the form of a hash and the number of nodes the tree that the hash is the root of contains. The contract then adds a `lastResponseTimestamp` and a `challengeStepCompleted` property to make the `Submission` struct. In the below explanation, S_n represents the n*th* submission.

If only one submission is made, S0, then when the submission window expires `confirmNewHash(0)` can be called by any user to set the `ReputationRootHash` property on `ColonyNetwork` to S0.hash, similarly set `ReputationRootHashNNodes`, reward the users that backed that hash with newly minted tokens and reputation (though strictly they will only have their reputation included in the next cycle), deploy a new instance of `ReputationMiningCycle` and destroy the current one. 

Note that the parameter passed to `confirmNewHash` is *not* the index of the submission, but the index of the dispute round that resulted in one hash being left standing. In the case of one submission, it is the first dispute round. Dispute rounds only matter when multiple hashes are submitted, and a worked example is show below.

In the case of multiple hashes being submitted, hashes are paired off against each other. A future PR will introduce challenges involving merkle proofs, but for now `respondToChallenge` is used, which just assumes that a challenge has been passed. For a hash that has not passed its challenge in time, `invalidateHash` is used, which will pass that hash's opponent on to the next round.

Consider five submissions have been made, and `disputeRounds` looks like
```
0: [S0, S1, S2, S3, S4]
```
`S0` will face off against `S1`, `S2` v `S3`. `S4` has no opponent, and so someone (likely someone who submitted `S4`) can call `invalidateHash(5)` to get a 'bye' to the next round. `disputeRounds` now looks like
```
0: [S0, S1, S2, S3, ]
1: [S4]
```
Let's now say that S1 fails a challenge. S0 passes to the next round
```
0: [, S1, S2, S3, ]
1: [S4, S0]
```
Part of the `invalidateHash` call notes that there is now a pairing of `S4` vs `S0` that must be sorted. Their timeout starts immediately, even though `S2` vs `S3` in round 0 hasn't yet resolved. We don't need to wait for them. `S0` fails this challenge.

```
0: [, S1, S2, S3, ]
1: [, S0]
2: [S4]
```
From this point, let's consider two possibilities. Firstly, the case where `S2` and `S3` time out, which increases the `nInvalidatedHashes` counter by two. As this is now one less than `nSubmittedHashes`, we can call `confirmHash(2)` in order to confirm S4 as the new reputation hash.

If instead, `S3` fails, but `S2` passes, we move to 
```
0: [, S1, , S3 , ]
1: [, S0, S2]
2: [S4]
```
S0, S1 and S3 already failed challenges, and `S2` isn't going to get another opponent in round 1. `invalideHash(1, 3)` passes `S2` on to the next round.

```
0: [, S1, , S3 , ]
1: [, S0, ]
2: [S4, S2]
```
`S4` and `S2` face off, with `S2` timing out. `S4` moves on to round 3, and `confirmHash(3)` can now be called to end this reputation cycle and start the next.

This all assumes that one hash submitted is correct that can always pass an on-chain challenge i.e. there is one honest miner. 

**Non-functional changes**

Web3.js 1.0 is removing support for synchronous calls, and for a while with the mining client (not in this PR) I was trying to use web3.js 1.0, which is used by the new ganache beta. `testHelper.currentBlockTime()` was the only sync call that we had, so I've turned it in to an async call and `await`ed where appropriate. 

**Issues**

There are a lot of TODOs floating around, but a big one is that the hash submission window never closes here, which is a problem, and requires that
```
    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
    // require((now-reputationMiningWindowOpenTimestamp) <= 3600);
```

be fixed while maintaining the deterministic nature of the tests. I am open to suggestions on that front!

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -67,6 +67,13 @@ contract Colony is ColonyStorage {
     return token.mint(_wad);
   }
 
+  function mintTokensForColonyNetwork(uint _wad) public {
+    require(msg.sender == colonyNetworkAddress); // Only the colony Network can call this function
+    require(this == IColonyNetwork(colonyNetworkAddress).getColony("Common Colony")); // Function only valid on the Common Colony
+    token.mint(_wad);
+    token.transfer(colonyNetworkAddress, _wad);
+  }
+
   //TODO: Secure this function
   function addGlobalSkill(uint _parentSkillId) public
   returns (uint256)
--- a/contracts/ColonyNetworkStaking.sol
+++ b/contracts/ColonyNetworkStaking.sol
@@ -67,6 +67,7 @@ contract ColonyNetworkStaking is ColonyNetworkStorage {
     require(msg.sender == reputationMiningCycle);
     for (uint256 i = 0; i < stakers.length; i++) {
       // This is pretty harsh! Are we happy with this?
+      // Alternative: lose more than they would have gained for backing the right hash.
       stakedBalances[stakers[i]] = 0;
     }
     // TODO: Where do these staked tokens go?
@@ -117,6 +118,7 @@ contract ReputationMiningCycle {
     //Check the ticket is an eligible one for them to claim
     require(entry <= IColonyNetwork(colonyNetworkAddress).getStakedBalance(msg.sender) / 10**15);
     require(entry > 0);
+    // TODO: Require minimum stake, that is (much) more than the cost required to defend the valid submission.
     //Check the ticket is a winning one.
     // require((now-reputationMiningWindowOpenTimestamp) < 3600);
     // x = floor(uint((2**256 - 1) / 3600)
@@ -145,7 +147,6 @@ contract ReputationMiningCycle {
     submittedHashes[newHash][nNodes].push(msg.sender);
     //Note that they submitted it.
     submittedEntries[newHash][msg.sender][entry] = true;
-
   }
 
   function confirmNewHash(uint256 roundNumber) public {
@@ -160,18 +161,19 @@ contract ReputationMiningCycle {
   function invalidateHash(uint256 round, uint256 idx) public {
     // TODO: Require that it has failed a challenge, or failed to respond in time.
     // Move its opponent on to the next stage.
-    uint256 opponentIdx = (idx % 2 == 1 ? idx-1 : idx+1);
+    uint256 opponentIdx = (idx % 2 == 1 ? idx-1 : idx + 1);
     // TODO: Check opponent is good to move on - we're assuming both haven't timed out here.
 
     // We require that we actually had an opponent - can't invalidate the last hash.
     // If we try, then the next require should catch it.
+    require(disputeRounds[round].length > opponentIdx);
     require(disputeRounds[round][opponentIdx].hash!="");
     disputeRounds[round+1].push(disputeRounds[round][opponentIdx]);
     delete disputeRounds[round][opponentIdx];
     nInvalidatedHashes += 1;
 
     // Punish the people who proposed this
-    IColonyNetwork(colonyNetworkAddress).punishStakers(submittedHashes[disputeRounds[round][0].hash][disputeRounds[round][0].nNodes]);
+    IColonyNetwork(colonyNetworkAddress).punishStakers(submittedHashes[disputeRounds[round][idx].hash][disputeRounds[round][idx].nNodes]);
 
     //TODO: Can we do some deleting to make calling this as cheap as possible for people?
   }
--- a/contracts/ColonyNetworkStaking.sol
+++ b/contracts/ColonyNetworkStaking.sol
@@ -46,6 +46,9 @@ contract ColonyNetworkStaking is ColonyNetworkStorage {
     require(msg.sender == reputationMiningCycle);
     reputationRootHash = newHash;
     reputationRootHashNNodes = newNNodes;
+    // Clear out the reputation log. We're setting a new root hash, so we're done with it.
+    delete ReputationUpdateLog;
+    // Reward stakers
     rewardStakers(stakers);
     reputationMiningCycle = 0x0;
     startNextCycle();
@@ -79,9 +82,32 @@ contract ColonyNetworkStaking is ColonyNetworkStorage {
     // TODO: Actually think about this function
     // Passing an array so that we don't incur the EtherRouter overhead for each staker if we looped over
     // it in ReputationMiningCycle.invalidateHash;
-    // for (uint256 i = 0; i < stakers.length; i++) {
-    //   //We need to... do something. They get newly minted tokens, right?
-    // }
+    address commonColonyAddress = _colonies["Common Colony"];
+    uint256 reward = 10**18; //TODO: Actually work out how much reputation they earn, based on activity elsewhere in the colony.
+    if (reward >= uint256(int256(-1))/2) {
+      reward = uint256(int256(-1))/2;
+    }
+    // TODO: We need to be able to prove that this will never happen, otherwise we're locked out of reputation mining.
+    // Something like the above cap is an adequate short-term solution, but at the very least need to double check the limits
+    // (which I've fingered-in-the-air, but could easily have an OBOE hiding inside).
+    assert(reward < uint256(int256(-1))); // We do a cast later, so make sure we don't overflow.
+    IColony(commonColonyAddress).mintTokensForColonyNetwork(stakers.length * reward); // This should be the total amount of new tokens we're awarding.
+    for (uint256 i = 0; i < stakers.length; i++) {
+      // We *know* we're the first entries in this reputation update log, so we don't need all the bookkeeping in
+      // the AppendReputationUpdateLog function
+      ReputationUpdateLog.push(ReputationLogEntry(
+        stakers[i], //The staker getting the reward
+        int256(reward),
+        0, //TODO: Work out what skill this should be. This should be a special 'mining' skill.
+        commonColonyAddress, // They earn this reputation in the common colony.
+        4, // Updates the user's skill, and the colony's skill, both globally and for the special 'mining' skill
+        i*4)//We're zero indexed, so this is the number of updates that came before in the reputation log.
+      );
+
+      // Also give them some newly minted tokens.
+      // We reinvest here as it's much easier (gas-wise).
+      stakedBalances[stakers[i]] += reward;
+    }
   }
 }
 
--- a/contracts/ColonyNetworkStaking.sol
+++ b/contracts/ColonyNetworkStaking.sol
@@ -11,11 +11,9 @@ import "./ColonyNetworkStorage.sol";
 import "./IColonyNetwork.sol";
 
 
-contract ColonyNetworkStaking is ColonyNetworkStorage {
+contract ColonyNetworkStaking is ColonyNetworkStorage, DSMath {
 
-  event Address(address _a);
-  event Bool(bool _b);
-  function deposit(uint _amount) public {
+  function deposit(uint256 _amount) public {
     // Get CLNY address
     Token clny = Token(IColony(_colonies["Common Colony"]).getToken());
     uint256 networkBalance = clny.balanceOf(this);
@@ -24,14 +22,14 @@ contract ColonyNetworkStaking is ColonyNetworkStorage {
     // Check it actually transferred
     assert(clny.balanceOf(this)-networkBalance==_amount);
     // Note who it belongs to.
-    stakedBalances[msg.sender] += _amount;
+    stakedBalances[msg.sender] = add(stakedBalances[msg.sender], _amount);
   }
 
-  function withdraw(uint _amount) public {
+  function withdraw(uint256 _amount) public {
     uint256 balance = stakedBalances[msg.sender];
     require(balance >= _amount);
     bytes32 submittedHash;
-    (submittedHash, ) = ReputationMiningCycle(reputationMiningCycle).hasSubmitted(msg.sender);
+    (submittedHash, ) = ReputationMiningCycle(reputationMiningCycle).reputationHashSubmissions(msg.sender);
     bool hasRequesterSubmitted = submittedHash == 0x0 ? false : true;
     require(hasRequesterSubmitted==false);
     stakedBalances[msg.sender] -= _amount;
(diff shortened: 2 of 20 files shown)


contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+import "./EtherRouter.sol";
+import "./Token.sol";
+import "./ColonyNetworkStorage.sol";
+import "./IColonyNetwork.sol";
+
+
+contract ColonyNetworkStaking is ColonyNetworkStorage {
+
+  event Address(address _a);
+  event Bool(bool _b);
```
elenadimitrova: Are these events left over from debugging? Neither is used.

helpers/test-data-generator.js
```diff
@@ -18,15 +18,11 @@ import {
 import testHelper from "../helpers/test-helper";
 
 module.exports = {
-  async setupAssignedTask(
-    colonyNetwork,
-    colony,
-    dueDate = testHelper.currentBlockTime(),
```
elenadimitrova: Can't we do the default `dueDate` setting here by just awaiting the `testHelper` call?

test/colony-network-staking.js
```diff
@@ -0,0 +1,774 @@
...
+
+const EtherRouter = artifacts.require("EtherRouter");
+const IColony = artifacts.require("IColony");
+const IColonyNetwork = artifacts.require("IColonyNetwork");
+const Token = artifacts.require("Token");
+const ReputationMiningCycle = artifacts.require("ReputationMiningCycle");
+
+const BN = require("bn.js");
+
+contract("ColonyNetwork", accounts => {
```
elenadimitrova: Rename the tested contract `ColonyNetwork ` to `ColonyNetworkStaking` to differentiate.

contracts/Colony.sol
```diff
@@ -66,6 +66,13 @@ contract Colony is ColonyStorage {
     return token.mint(_wad);
   }
 
+  function mintTokensForColonyNetwork(uint _wad) public {
```
elenadimitrova: Why not just call `mint` on the CLNY token from `ColonyNetwork`?

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    assert(clny.balanceOf(this)-networkBalance==_amount);
+    // Note who it belongs to.
+    stakedBalances[msg.sender] += _amount;
+  }
+
+  function withdraw(uint _amount) public {
+    uint256 balance = stakedBalances[msg.sender];
+    require(balance >= _amount);
+    bytes32 submittedHash;
+    (submittedHash, ) = ReputationMiningCycle(reputationMiningCycle).hasSubmitted(msg.sender);
```
elenadimitrova: This is a public getter on hasSubmitted mapping being used meaning it doesn't go through EtherRouter?

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    }
+  }
+}
+
+// TODO: Can we handle a dispute regarding the very first hash that should be set?
+
+
+contract ReputationMiningCycle {
+  address colonyNetworkAddress;
+  // TODO: Do we need both these mappings?
```
elenadimitrova: I couldn't work out a way to do without them but maybe we can get rid of `submittedEntries` if we can accommodate `entry` in `submittedHashes` e.g.
```
struct Entry {
        address reputationMiner;
        uint N;
    }
mapping (bytes32 => mapping( uint256 => Entry[])) public submittedHashes;
```

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    // TODO: Where do these staked tokens go? Maybe split between the person who did the 'invalidate' transaction
+    // and the colony network?
+    // TODO: Lose rep?
+  }
+
+  function rewardStakers(address[] stakers) internal {
+    // Internal unlike punish, because it's only ever called from setReputationRootHash
+
+    // TODO: Actually think about this function
+    // Passing an array so that we don't incur the EtherRouter overhead for each staker if we looped over
```
elenadimitrova: Copy and paste leftover to remove

contracts/ColonyNetworkStaking.sol
area: Yup, they are. Good catch.

helpers/test-data-generator.js
area: Pretty sure I tried this, and you can't do it.

contracts/Colony.sol
area: The permission requirements are different, and I thought it would have been less clear if we had to accommodate the default 'auth' modifier, though not impossible. I thought it was clearer to have two separate functions.

`mint` can be called by any authorized user in the colony (i.e. an admin).
`mintTokensForColonyNetwork` can only be called by the `colonyNetwork`, and only on the Common Colony.

We could add `colonyNetwork` to the list of users authorized to call `mint` for the Common Colony, but then there is the possibility that address could be removed from the authority which would break the reputation mining...

contracts/ColonyNetworkStaking.sol
area: Correct. `ReputationMiningCycle` does not use EtherRouter, as a new `ReputationMiningCycle` is deployed every mining cycle. If we want to upgrade `ReputationMiningCycle`, we upgrade `ColonyNetworkStaking` (which deploys it) and the new `ReputationMiningCycle` contract will be used for the next cycle.

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+  event Address(address _a);
+  event Bool(bool _b);
+  function deposit(uint _amount) public {
+    // Get CLNY address
+    Token clny = Token(IColony(_colonies["Common Colony"]).getToken());
+    uint256 networkBalance = clny.balanceOf(this);
+    // Move some over.
+    clny.transferFrom(msg.sender, this, _amount);
+    // Check it actually transferred
+    assert(clny.balanceOf(this)-networkBalance==_amount);
```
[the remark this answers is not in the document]
area: This was me guarding the `ColonyNetworkStaking` contract from an upgraded CLNY contract that had a flaw in it that allowed `transferFrom` to happen but didn't actually transfer the balance over. Likely overcautious. Will remove and check just `transferFrom`'s value if you agree that's too much.

test/colony-network-staking.js
```diff
@@ -0,0 +1,774 @@
...
+      assert.equal(userBalance.toNumber(), 0);
+    });
+
+    // it('should allow a new cycle to start if there is none currently', async function(){
+    //   let addr = await colonyNetwork.getReputationMiningCycle.call();
+    //   assert(addr==0x0);
+    //   await colonyNetwork.startNextCycle();
+    //   addr = await colonyNetwork.getReputationMiningCycle.call();
+    //   assert(addr!=0x0);
+    // })
```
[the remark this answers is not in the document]
area: Ah, yeah. Basically, this is now implicitly tested by our 'before' when the tests first start. Will remove.

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    }
+    // TODO: Require minimum stake, that is (much) more than the cost required to defend the valid submission.
+    // Check the ticket is a winning one.
+    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
+    // require((now-reputationMiningWindowOpenTimestamp) <= 3600);
+    // x = floor(uint((2**256 - 1) / 3600)
+    if (now-reputationMiningWindowOpenTimestamp <= 3600) {
+      uint x = 32164469232587832062103051391302196625908329073789045566515995557753647122;
+      uint target = (now - reputationMiningWindowOpenTimestamp ) * x;
+      require(uint256(keccak256(msg.sender, entry, newHash)) < target);
```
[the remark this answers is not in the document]
area: I don't think so; we could never directly compare a `bytes32` to a `uint256`, which is what the whitepaper is asserting we do. On an 'algorithmic' level, it is correct, but I'm happy to amend the whitepaper if we want to be strictly correct (the best form of correct!).

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    clny.transferFrom(msg.sender, this, _amount);
+    // Check it actually transferred
+    assert(clny.balanceOf(this)-networkBalance==_amount);
+    // Note who it belongs to.
+    stakedBalances[msg.sender] += _amount;
+  }
+
+  function withdraw(uint _amount) public {
+    uint256 balance = stakedBalances[msg.sender];
+    require(balance >= _amount);
```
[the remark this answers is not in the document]
area: Another of our classic disagreements! I only like modifiers if they're going to be reused...

contracts/ColonyNetworkStaking.sol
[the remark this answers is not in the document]
elenadimitrova: Valid point which depends on whether the Token is upgradable ;)

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// TODO: Can we handle a dispute regarding the very first hash that should be set?
+
+
+contract ReputationMiningCycle {
+  address colonyNetworkAddress;
+  // TODO: Do we need both these mappings?
+  mapping (bytes32 => mapping( uint256 => address[])) public submittedHashes;
+  mapping (address => Submission) public hasSubmitted;
```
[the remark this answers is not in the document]
area: I agree. I'm pretty sure I thought this was going to be a mapping of Booleans, but it expanded from there and didn't get renamed!

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,306 @@
...
+    require(entry <= IColonyNetwork(colonyNetworkAddress).getStakedBalance(msg.sender) / 10**15);
+    require(entry > 0);
+    if (hasSubmitted[msg.sender].hash != 0x0) {           // If this user has submitted before during this round...
+      require(newHash == hasSubmitted[msg.sender].hash);  // ...require that they are submitting the same hash ...
+      require(nNodes == hasSubmitted[msg.sender].nNodes); // ...require that they are submitting the same number of nodes for that hash ...
+      require (submittedEntries[newHash][msg.sender][entry] == false); // ... but not this exact entry
+    }
+    // TODO: Require minimum stake, that is (much) more than the cost required to defend the valid submission.
+    // Check the ticket is a winning one.
+    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
```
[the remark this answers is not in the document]
area: If I forward time by 3600, and this is in place, no submissions will be accepted (because some time will have elapsed between `forwardTime` and the submission being made.

Conversely, if I forward time by e.g. 3500, sometimes the tests will randomly fail because the hash will happen to not be small enough.

contracts/ColonyNetworkStaking.sol
area: Though maybe not as much removed as you're expecting.

contracts/ColonyNetworkStaking.sol
```diff
@@ -0,0 +1,305 @@
+pragma solidity ^0.4.17;
+pragma experimental "v0.5.0";
+pragma experimental "ABIEncoderV2";
+
+import "../lib/dappsys/auth.sol";
+import "./Authority.sol";
+import "./IColony.sol";
+import "./EtherRouter.sol";
+import "./Token.sol";
```
elenadimitrova: We can use `ERC20Extended` instead of `Token` here

test/colony-network-staking.js
```diff
@@ -0,0 +1,762 @@
...
+    const repCycle = ReputationMiningCycle.at(reputationMiningCycleAddress);
+    if (respondToChallenge) {
+      const oppIdx = idx % 2 === 1 ? idx - 1 : idx + 1;
+      await repCycle.respondToChallenge(round, oppIdx);
+      await testHelper.forwardTime(600, test);
+    }
+    return repCycle.invalidateHash(round, idx);
+  }
+
+  async function giveUserCLNYTokens(address, _amount) {
```
elenadimitrova: I could really use this in the `testHelper` for testing #97

Issue #38, opened 2017-09-20
Earning reputation from contributing to a task by Evaluator


elenadimitrova:
### 6.2.1 Reputation change from contributing to a task
Upon completion of a task, the evaluator also earns reputation based on their token reward. 

For the purposes of calculating earned reputation the rating of the evaluator is assumed to be the same as the maximum of that of a manager, i.e. 5.

There is no explicit rating of the evaluator, but as with all other payments and rewards, an objection
can be raised before a payout occurs; if the evaluation is changed by such an objection, the evaluator’s reward is reduced or turns in to a loss of reputation.

kronosapiens:
This functionality has been implemented (see #158)

Issue #154, opened 2018-02-12
Move to three point grading scale.


area:
As part of our feedback from Daniel Kronovet, he suggested migrating from a five-point grading scale to a three-point grading scale, denoting inadequate, adequate, and excellent. This is something we would like to implement.

Currently, a six-point scale is used (0-5, inclusive). The new scale should be 1-3 inclusive. For the worker, a '1' indicates a failure to complete the task, and should result in loss of reputation equal in magnitude to that they would have gained if they were rated a '2'. A '2' should be seen as the normal and expected rating for some work, and the worker gains reputation equal to the token payout received. A '3' indicates going above and beyond the call of duty, and should result in a reputation gain equal to 1.5 that which they would have gained if rated '2'.

Unlike the current implementation, these same factors should be used for the rating of the manager (-1, 1, 1.5).

This scale is now not linear, and so the half-point penalization for not committing or revealing the rating they were meant to submit should be removed, and replaced by a direct reputation penalty, equal to half the reputation they would have gained if they were rated a '1'.

**Optional:** The [whitepaper](https://circleci.com/gh/JoinColony/Colony-Whitepaper) will need to be amended to described this new implementation. This can be done by the same person who implements this on the contract-side, for bonus points, or not.

Have I missed anything, @elenadimitrova, or would you suggest any changes to the above? This is appropriate for 'good first issue' once we've agreed on the definition of the work.

elenadimitrova:
Sounds good. Just to clarify the case of missing rating penalties:
- What do we assume the missing rating to be? In the current solution we assume 5 for both manager and worker. Here will it be a '3'?
- Clarifying the penalty amount from this:
>reputation penalty, equal to half the reputation they would have gained if they were rated a '1'
The reputation they gain if they were rated a '1' would be '-1' and the penalty therefore '-0.5'? That's the same as the current implementation right?

area:
> What do we assume the missing rating to be? In the current solution we assume 5 for both manager and worker. Here will it be a '3'?

Yeah, I guess so. Benefit-of-the-doubt and all that. If people don't want others getting credit for work they don't do, they should reveal their vote!

> The reputation they gain if they were rated a '1' would be '-1' and the penalty therefore '-0.5'? That's the same as the current implementation right?

[...]

PR #158, opened 2018-02-15
Evaluator reputation update

lazoviccorp:
Implements evaluator reputation update part for #38 

- Setting rating of the evaluator when reputation is calculated.
- Updating reputation for every role using loop.
- Introduced payout for evaluator and updated tests accordingly.
- Split `setTaskPayout` to 3 functions, each for every role. Reviews are needed for `setTaskEvaluatorPayout` and `setTaskWorkerPayout`.
- `setTaskManagerPayout` can be only called by a manager of the task.
- Updated gas costs.


contracts/ColonyTask.sol
```diff
@@ -292,27 +294,18 @@ contract ColonyTask is ColonyStorage, DSMath {
...
-    colonyNetworkContract.appendReputationUpdateLog(managerRole.user, managerReputation, skillId);
+    updateReputation(_id, MANAGER, skillId);
 
     // NOTE: reputation change amount is hereby limited to MAXINT/30
-    int workerReputation = SafeMath.mulInt(int(workerPayout), (int(workerRole.rating)*2 - 50)) / 30;
-    colonyNetworkContract.appendReputationUpdateLog(workerRole.user, workerReputation, skillId);
+    updateReputation(_id, WORKER, skillId);
     // TODO Reputation changes for other relevant roles, domains.
+
+    updateReputation(_id, EVALUATOR, skillId);
```
elenadimitrova: I don't see the benefit of calling another function 3 times from here. This incurs some unnecessary extra reads from storage. We can refactor out the internal function in cases where the logic lies in multiple functions but here it's all within `finalizeTask`. Plus with the suggestion from above of balancing gas costs and design you best keep it all in this function.

helpers/test-data-generator.js
```diff
@@ -81,6 +85,12 @@ module.exports = {
     await colony.proposeTaskChange(txData, 0, MANAGER_ROLE);
     transactionId = await colony.getTransactionCount.call();
     await colony.approveTaskChange(transactionId, WORKER_ROLE, { from: worker });
+
+    txData = await colony.contract.setTaskPayout.getData(taskId, EVALUATOR_ROLE, tokenAddress, evaluatorPayout.toString());
+    await colony.proposeTaskChange(txData, 0, MANAGER_ROLE);
+    transactionId = await colony.getTransactionCount.call();
+    await colony.approveTaskChange(transactionId, WORKER_ROLE, { from: worker });
```
elenadimitrova: Note that `setTaskPayout` transaction reviewers are currently the task Manager and Worker. However when setting the Evaluator the reviewers logically have to be the Manager and Evaluator. With the current `ColonyTransactionReviewer` design this is not possible. So we have to either:
a) Split the `setTaskPayout` in two functions, e.g. `setTaskWorkerPayout` and `setTaskEvaluatorPayout`. The disadvantage here is that for additional roles we may want to support in future, i.e. in addition to manager, evaluator and worker, we'll have to keep introducing new functions so we lose 1 level of abstraction.

b) We refactor the `ColonyTransactionReviewer` to support this model. 

I haven't explored option b) so feel free to experiment. Overall I'd prefer option a) for simplicity but I want to hear what @area thinks about it.

contracts/ColonyTask.sol
elenadimitrova: A small note here to switch the order in which `appendReputationUpdateLog` is called to match the order of role ids, i.e. manager (roleId = 0), evaluator (roleId = 1), worker (roleId = 2). Just so when getting the respective reputation log entries, it's more intuitive.

lazoviccorp: I was getting the 'stack too deep' error as i mentioned in description, because of too much local variables (16 is max i think).

I will try to bring back everything into finalizeTask, but i will probably remove some local variables like [this one](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L303) or [this one](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L296) and access them directly [here](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L308).
What do you think?

contracts/ColonyTask.sol
```diff
@@ -232,10 +232,12 @@ contract ColonyTask is ColonyStorage, DSMath {
   taskExists(_id)
   taskNotFinalized(_id)
   {
+    uint8 initialRating = (_role == EVALUATOR) ? 50 : 0;
+
     tasks[_id].roles[_role] = Role({
       user: _user,
       rated: false,
-      rating: 0
+      rating: initialRating
```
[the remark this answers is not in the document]
lazoviccorp: Yes, initially it was done just before calculating the reputation.

I don't know exactly how objections will work, but the reason i decided to do it this way is because maybe we would want to allow objections to the evaluator rating to be made before actual calculation of the reputation.

contracts/ColonyTask.sol
elenadimitrova: I had problems when I tried to do that if I remember correctly. But otherwise yes, try cutting down some of the local variables.

elenadimitrova: `skillId` does not need to be explicit as well.

elenadimitrova:
@lazoviccorp This works for the `taskFinalize` logic I think 
```

    for (uint8 roleId = 0; roleId <= 2; roleId ++) {
      uint payout = task.payouts[roleId][token];
      Role storage role = task.roles[roleId];
      uint8 rating = (roleId == EVALUATOR) ? 50 : role.rating;
      uint divider = (roleId == WORKER) ? 30 : 50;
      int reputation = SafeMath.mulInt(int(payout), (int(rating)*2 - 50)) / divider;
      //TODO: Change the skillId to represent the task domain
      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.skills[0]);
    }

    task.finalized = true;
```
Note that this doesn't implement the todo fix

elenadimitrova:
@lazoviccorp This works for the `taskFinalize` logic I think
```    

    for (uint8 roleId = 0; roleId <= 2; roleId ++) {
      uint payout = task.payouts[roleId][token];
      Role storage role = task.roles[roleId];
      uint8 rating = (roleId == EVALUATOR) ? 50 : role.rating;
      int divider = (roleId == WORKER) ? 30 : 50;
      int reputation = SafeMath.mulInt(int(payout), (int(rating)*2 - 50)) / divider;
      //TODO: Change the skillId to represent the task domain
      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.skills[0]);
    }

    task.finalized = true;
```
Note that this doesn't implement the todo fix

contracts/ColonyTask.sol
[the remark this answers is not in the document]
elenadimitrova: You're right, objections will be done before reputation calculation but the evaluator rating won't be needed there still.

helpers/test-data-generator.js
lazoviccorp: For now i created separate functions for every role. I will try to find better solution after this PR is merged, because these changes are required for #154.

contracts/Colony.sol
```diff
@@ -48,7 +48,9 @@ contract Colony is ColonyStorage {
     // Initialise the task update reviewers
     IColony(this).setFunctionReviewers(0xda4db249, 0, 2); // setTaskBrief => manager, worker
     IColony(this).setFunctionReviewers(0xcae960fe, 0, 2); // setTaskDueDate => manager, worker
-    IColony(this).setFunctionReviewers(0xbe2320af, 0, 2); // setTaskPayout => manager, worker
+    IColony(this).setFunctionReviewers(0x6aaf4a77, 0, 2); // setTaskManagerPayout => manager, worker
```
elenadimitrova: It doesn't make sense to get the worker to agree to what the manager payout should be. Just don't permission the `setTaskManagerPayout` function for now.

PR #163, opened 2018-02-19
Disable requiring flowtype annotations

area:
We don't use flow, but we want to keep using eslint-config-colony.
So we're disabling it for all javascript files.

Builds weren't broken (this is only a warning), but disabling it stops everyone's client looking like this:

screen shot 2018-02-19 at 10 02 43

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,9 +12,10 @@
   "rules": {
     "import/no-extraneous-dependencies": [2, {"devDependencies": ["**/*.js"]}],
     "max-len": [2, { "code": 150, "ignoreComments": true }],
-    "prettier/prettier": ["error", {"printWidth": 150}]
+    "prettier/prettier": ["error", {"printWidth": 150}],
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -15,6 +15,14 @@
...
   "parserOptions": {
-   "ecmaVersion": 2017
- }
+    "ecmaVersion": 2017
+  },
+  "overrides": [
+    {
+      "files": "*.js",
+        "rules": {
+          "flowtype/require-valid-file-annotation": "off"
```
chmanie: You could just put it into the general "rules" section. eslint just applies for .js files anyways :)

area: Here I was dutifully following what was in the Changelog, but yeah, that's much better :)

chmanie: Yeah in the changelog that was more targeting the situation where you want to disable it for certain files.

Issue #24, opened 2017-09-18
Task completion and rating


elenadimitrova:
### 4.1 Tasks

After the task has been assigned, the worker may make a ‘final submission’, which includes some
evidence that the work has been completed.
Once the due date has passed or the worker has made their submission, the evaluator may rate
the work. Regardless of whether the rating is positive or not, the task enters a state in which
objections to the final state of the task can be made and disputes can be initiated (see Section
9). Once three days have elapsed, no more objections or disputes can be raised. Once all pending
disputes related to the task are resolved, the parties involved get punished or rewarded based on
the final state of the task.
If the evaluator’s rating for the work is changed as a result of a dispute, they get a reduced payout
based on the discrepancy between their original score and the score that their peers determined to
be more appropriate. If an objection has determined the manager or evaluator should be punished,
they lose their stake, otherwise the stake is able to be reclaimed. The worker is paid based on the
final score awarded to their work, taking into account the result of any disputes.

elenadimitrova:
@lazoviccorp working on adding the task payouts logic to do with task work being rejected or accepted based on the rating of the worker

kronosapiens:
Closing -- this functionality was completed in #166

PR #166, opened 2018-02-24
24 Task completion and rating

lazoviccorp:
Implements worker payout logic based on rating part for #24 

- Setting the worker payout to 0 if work rating is less or equal to 2.

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -297,13 +297,21 @@ contract ColonyTask is ColonyStorage, DSMath {
     for (uint8 roleId = 0; roleId <= 2; roleId++) {
       uint payout = task.payouts[roleId][token];
       Role storage role = task.roles[roleId];
+
       uint8 rating = (roleId == EVALUATOR) ? 50 : role.rating;
       int divider = (roleId == WORKER) ? 30 : 50;
+
       int reputation = SafeMath.mulInt(int(payout), (int(rating)*2 - 50)) / divider;
+      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.domains[0]);
+
       if (roleId == WORKER) {
         colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.skills[0]);
+
+        if (rating <= 20) {
+          task.payouts[roleId][token] = 0;
+          task.totalPayouts[token] = sub(task.totalPayouts[token], payout);
+        }
       }
-      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.domains[0]);
     }
 
     task.finalized = true;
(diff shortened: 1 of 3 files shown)


contracts/ColonyFunding.sol
```diff
@@ -41,15 +41,15 @@ contract ColonyFunding is ColonyStorage, DSMath {
   }
 
   function setTaskManagerPayout(uint256 _id, address _token, uint256 _amount) public isManager(_id) {
-    setTaskPayout(_id, 0, _token, _amount);
+    setTaskPayout(_id, MANAGER, _token, _amount);
```
elenadimitrova: 👍

contracts/ColonyFunding.sol
```diff
@@ -60,11 +60,26 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+
+    moveFundsBetweenPots(task.potId, 1, payout, _token);
+  }
+
   function claimPayout(uint256 _id, uint256 _role, address _token) public
   taskFinalized(_id)
   {
     Task storage task = tasks[_id];
     require(task.roles[_role].user == msg.sender);
+    if (_role == WORKER) {
```
elenadimitrova: The above solution means we won't need this check here as well, as the worker payout is `0`.

contracts/ColonyFunding.sol
```diff
@@ -60,11 +60,26 @@ contract ColonyFunding is ColonyStorage, DSMath {
     return task.payouts[_role][_token];
   }
 
+  function moveFailedTaskFunds(uint256 _id, address _token) public
```
elenadimitrova: Instead of introducing a new function specifically for this, can we simply set the Worker payout to `0` in `finalizeTask`? Then we can use `moveFundsBetweenPots` to refund anything left after everyone's claimed?

lazoviccorp: Ok, setting the worker rating to 0 after reputation update

contracts/ColonyTask.sol
```diff
@@ -297,13 +297,21 @@ contract ColonyTask is ColonyStorage, DSMath {
...
       int divider = (roleId == WORKER) ? 30 : 50;
+
       int reputation = SafeMath.mulInt(int(payout), (int(rating)*2 - 50)) / divider;
+      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.domains[0]);
+
       if (roleId == WORKER) {
         colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.skills[0]);
+
+        if (rating <= 20) {
+          task.payouts[roleId][token] = 0;
```
elenadimitrova: 👍

contracts/ColonyTask.sol
```diff
@@ -297,13 +297,21 @@ contract ColonyTask is ColonyStorage, DSMath {
...
+          task.totalPayouts[token] = sub(task.totalPayouts[token], payout);
+        }
       }
-      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.domains[0]);
```
elenadimitrova: Any reason you moved this? It won't make a difference to the end result but you're reordering the reputation log entries which can break tests.

lazoviccorp: Just to make it more readable, no other reason

PR #168, opened 2018-02-25
Colony network run Token auctions

### LINKED ISSUE #97: The token auction
### 10.3.2 The token auction
The colony tokens collected are auctioned off by the Colony Network Contract, with the auctions denominated in Common Colony Tokens, the proceeds of which are given to the Common Colony as revenue. These auctions — one for each type of token collected — occur on a regular basis of once a month.
We believe such behaviour would be beneficial for the Common Colony Token holders (whose Common Colony Tokens gain value by having an explicit use beyond reputation mining) and the Common Colony itself (which turns the fees into Common Colony Tokens which it can then use to fund further development of the network without explicitly buying them back or creating more, diluting existing holders). It also provides an immediate mechanism of price discovery for the colony tokens, which are unlikely to be traded on third-party exchanges until much later in the lifetime of the colony. By auctioning off the collected tokens, we also prevent the Common Colony collecting a large number of different tokens that it has to manage, which would prove cumbersome and annoying for the colony.

elenadimitrova:
Closes #97 

Implements Dutch auction model for auctioning tokens earned by the Common Colony in exchange for CLNY. The auction model is inspired by a version of that used at [Raiden](https://github.com/raiden-network/raiden-token/blob/master/contracts/auction.sol) and earlier at [Gnosis](https://github.com/miladmostavi/gnosis-contracts/blob/master/contracts/solidity/DAO/DAODutchAuction.sol) ICOs.

The following auction rules apply:
- A new Auction is started via the `colonyNetwork.startTokenAuction` function which creates a new 
`DutchAuction` instance for a given Token X. Note that **ALL Token X** owned by the ColonyNetwork is transferred to the Auction contract here for selling.
- Auction is explicitly started via the `start` function where the `startTime`, `quantity` of token sold are set.
- Users can submit no-limit bids and multiple bids per account are allowed
- Before a bid is submitted the user has to have approved the `DutchAuction` contract address to transfer their bid CLNY amount. The bid will fail if this is not met.
- The auction is complete when it has received enough CLNY to sell its token quantity at the current price. No more bids are accepted after this point and `finalize` can be called.
- Current price can be gotten via the `price` function while the auction is open for bids. After closing the `finalPrice` should be queried. The current price is calculated as 
```
uint daysOpen = duration / 86400;
uint r = duration % 86400;
10^(36-daysOpen) * (864000 - 9r) / 864000;
```
We also limit the price to never go below 1 with 
```
    if (duration > 3110400) {
      return 1;
    }
```
Visualised `log(price) / elapsed time` graph
screen shot 2018-03-18 at 10 37 19

- After all tokens have been claimed, the token contract can be killed and storage refunded via `close`

### DIFF
--- a/contracts/ColonyNetworkAuction.sol
+++ b/contracts/ColonyNetworkAuction.sol
@@ -1,0 +1,218 @@
+/*
+  This file is part of The Colony Network.
+
+  The Colony Network is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  The Colony Network is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity ^0.4.17;
+pragma experimental "v0.5.0";
+pragma experimental "ABIEncoderV2";
+
+import "../lib/dappsys/math.sol";
+import "./ColonyNetworkStorage.sol";
+import "./ERC20Extended.sol";
+import "./IColony.sol";
+
+
+contract ColonyNetworkAuction is ColonyNetworkStorage {
+  event AuctionCreated(address auction, address token, uint256 quantity);
+
+  function startTokenAuction(address _token) public {
+    address commonColony = _colonies["Common Colony"];
+    address clny = IColony(commonColony).getToken();
+    DutchAuction auction = new DutchAuction(clny, _token);
+    uint availableTokens = ERC20Extended(_token).balanceOf(this);
+    ERC20Extended(_token).transfer(auction, availableTokens);
+    AuctionCreated(address(auction), _token, availableTokens);
+  }
+}
+
+
+contract DutchAuction is DSMath {
+  address public colonyNetwork;
+  ERC20Extended public clnyToken;
+  ERC20Extended public token;
+
+  // Total number of auctioned tokens
+  uint public quantity;
+  bool public started;
+  uint public startTime;
+  uint public endTime;
+  uint public minPrice;
+  uint public constant TOKEN_MULTIPLIER = 10 ** 18;
+  
+  // Keep track of all CLNY wei received
+  uint public receivedTotal;
+  uint public bidCount;
+  uint public claimCount;
+
+  // Final price in CLNY per 10**18 Tokens (min 1, max 1e18)
+  uint public finalPrice;
+  bool public finalized;
+  
+  mapping (address => uint256) public bids;
+
+  modifier auctionNotStarted {
+    require(startTime == 0);
+    require(!started);
+    _;
+  }
+
+  modifier auctionStartedAndOpen {
+    require(started);
+    require(startTime > 0);
+    require(endTime == 0);
+    _;
+  }
+
+  modifier auctionClosed {
+    require(endTime > 0);
+    _;
+  }
+
+  modifier auctionNotFinalized() {
+    require(!finalized);
+    _;
+  }
+
+  modifier auctionFinalized {
+    require(finalized);
+    _;
+  } 
+
+  modifier allBidsClaimed  {
+    require(claimCount == bidCount);
+    _;
+  }
+
+  event AuctionBid(address indexed _sender, uint _amount, uint _missingFunds);
+  event AuctionClaim(address indexed _recipient, uint _sentAmount);
+  event AuctionFinalized(uint _finalPrice);
+
+  function DutchAuction(address _clnyToken, address _token) public {
+    colonyNetwork = msg.sender;
+    require(_clnyToken != 0x0 && _token != 0x0);
+    assert(_token != _clnyToken);
+    clnyToken = ERC20Extended(_clnyToken);
+    token = ERC20Extended(_token);
+  }
+
+  function start() public
+  auctionNotStarted
+  {
+    quantity = token.balanceOf(this);
+    assert(quantity > 0);
+
+    // Set the minimum price as such that it doesn't cause the finalPrice to be 0
+    minPrice = (quantity >= TOKEN_MULTIPLIER) ? 1 : TOKEN_MULTIPLIER / quantity;
+
+    startTime = now;
+    started = true;
+  }
+
+  function totalToEndAuction() public view 
+  auctionStartedAndOpen
+  returns (uint)
+  {
+    return mul(quantity, price()) / TOKEN_MULTIPLIER;
+  }
+
+  // Get the price in CLNY per 10**18 Tokens (min 1 max 1e36)
+  // Starting price is 10**36, after 1 day price is 10**35, after 2 days price is 10**34 and so on
+  function price() public view
+  auctionStartedAndOpen
+  returns (uint)
+  {
+    uint duration = sub(now, startTime);
+    uint daysOpen = duration / 86400;
+    uint r = duration % 86400;
+    uint p = mul(10**sub(36, daysOpen), sub(864000, mul(9,r))) / 864000;
+    p = p < minPrice ? minPrice : p;
+    return p;
+  }
+
+  function bid(uint256 _amount) public
+  auctionStartedAndOpen
+  {
+    require(_amount > 0);
+    uint _totalToEndAuction = totalToEndAuction();
+    uint remainingToEndAuction = sub(_totalToEndAuction, receivedTotal);
+
+    // Adjust the amount for final bid in case that takes us over the offered quantity at current price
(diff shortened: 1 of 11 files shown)


contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+import "./IColony.sol";
+
+
+contract ColonyNetworkAuction is ColonyNetworkStorage {
+  event AuctionCreated(address auction, address token, uint256 quantity, uint256 startPrice);
+
+  function startTokenAuction(
+    address _token,
+    uint256 _quantity,
+    uint256 _startPrice)
```
area: `_startPrice` should not be a parameter of the auction. Otherwise, I can just set it at 1, and win everything for ~nothing. This should always be something absurdly high - worst case scenario, each auction will have to run for at least a few days, which lets people spot that it's happening and decide if they want to be involved.

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+import "./ERC20Extended.sol";
+import "./IColony.sol";
+
+
+contract ColonyNetworkAuction is ColonyNetworkStorage {
+  event AuctionCreated(address auction, address token, uint256 quantity, uint256 startPrice);
+
+  function startTokenAuction(
+    address _token,
+    uint256 _quantity,
```
area: Should it not just be for all tokens currently held?

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+    uint256 _startPrice)
+    public 
+    {
+    ERC20Extended clny = ERC20Extended(IColony(_colonies["Common Colony"]).getToken());
```
area: Is there a reason I'm not spotting to cast to `ERC20Extended` only to cast it to `address` below?

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+
+    // Update the total amount of funds for which tokens have been claimed
+    tokensClaimed += amount;
+
+    // Set receiver bid to 0 before transferring the tokens
+    bids[msg.sender] = 0;
+    require(token.transfer(msg.sender, tokens));
+    AuctionClaim(msg.sender, tokens);
+
+    assert(token.balanceOf(msg.sender) >= tokens);
```
area: Given we're talking about BYOT, what do you think about enforcing that these token balances are correct?

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+  uint public endBlock;
+  // Keep track of all CLNY wei received
+  uint public receivedTotal;
+  // Keep track of cumulative CLNY funds for which the tokens have been claimed
+  uint public tokensClaimed;
+  // Total number of auctioned tokens
+  uint public quantity;
+  // CLNY Wei per Token wei
+  uint public finalPrice;
+  uint public constant MULTIPLIER = 10 ** 18;
```
area: The comment above says start price startPrice is in CLNY wei and yet we're still multiplying by this factor in e.g. `price()`? The tests make it look like startPrice is CLNY. I *strongly* disagree with this `MULTIPLIER` being present at all - everything on-chain should be done in the indivisible unit, and clients deal with showing it in a 'human' form if desired.

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+
+  function totalToEndAuction() public view returns (uint) {
+    return (quantity * price()) / MULTIPLIER;
+  }
+
+  // Get the current Token price in CLNY wei
+  // If the end block is set, i.e. auction is closed for bids, use the endBlock in calculation
+  function price() public view returns (uint) {
+    uint lastBlock = endBlock == 0 ? block.number : endBlock;
+    return startPrice * MULTIPLIER / (1 + lastBlock - startBlock);
```
area: I think a linear price decay function is too slow, and it's definitely the case if we start with an absurdly high start price, as I'd prefer.

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,203 @@
...
+  function claim() public 
+  auctionFinalized
+  returns (bool)
+  {
+    uint amount = bids[msg.sender];
+    uint tokens = mul(amount, MULTIPLIER) / finalPrice;
+
+    // Due to finalPrice floor rounding, the number of assigned tokens may be higher
+    // than expected. Therefore, the number of remaining unassigned auction tokens
+    // may be smaller than the number of tokens needed for the last claimTokens call
```
area: This is unfair to the last claimant, though I agree only a little bit. Instead, when finalizing the auction, we could add `1` to the price after the division, and keep track of the number of people that have claimed, rather than the number of tokens that have been claimed. Leftovers could be returned to the `ColonyNetwork`.

contracts/ColonyNetworkAuction.sol
elenadimitrova: Sure, I'm up for simplifying it.

contracts/ColonyNetworkAuction.sol
elenadimitrova: Is `1e18` absurd enough? Considering this is the price of 1 Wei of the actioned token as paid in CLNY Wei. So that is a starting price of 1e18 times higher than CLNY.

contracts/ColonyNetworkAuction.sol
elenadimitrova: No good reason. Missed that when cleaning up

contracts/ColonyNetworkAuction.sol
elenadimitrova: `MULTIPLIER` was meant to aid division precision not cater for wei. But since the start price will very high anyway, i.e. more than that multiplier catered for, it should be sufficient.

contracts/ColonyNetworkAuction.sol
elenadimitrova: This is slow? It falls off a cliff in 10 blocks
screen shot 2018-02-28 at 17 29 31

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,216 @@
...
+
+  function totalToEndAuction() public view 
+  auctionStartedAndOpen
+  returns (uint) 
+  {
+    return mul(quantity, price()) / TOKEN_DECIMALS;
+  }
+
+  // Get the price in CLNY per 10**18 Tokens (min 1, max 1e18)
+  // Starting price is 10**36, after 1 day price is 10**36, after 2 days price is 10**35 and so on
```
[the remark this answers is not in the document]
elenadimitrova: I actually had the starting price at 10^37 and only on Sunday noticed that made little sense

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,216 @@
...
+import "./ColonyNetworkStorage.sol";
+import "./ERC20Extended.sol";
+import "./IColony.sol";
+
+
+contract ColonyNetworkAuction is ColonyNetworkStorage {
+  event AuctionCreated(address auction, address token, uint256 quantity);
+
+  function startTokenAuction(address _token) public {
+    address commonColony = _colonies["Common Colony"];
```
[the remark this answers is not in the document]
elenadimitrova: Good point. I'll add a check

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,216 @@
...
+    uint amount = bids[msg.sender];
+    uint tokens = mul(amount, TOKEN_DECIMALS) / finalPrice;
+    claimCount += 1;
+    
+    // Set receiver bid to 0 before transferring the tokens
+    bids[msg.sender] = 0;
+    uint beforeClaimBalance = token.balanceOf(msg.sender);
+    require(token.transfer(msg.sender, tokens));
+    assert(token.balanceOf(msg.sender) == add(beforeClaimBalance, tokens));
+    assert(bids[msg.sender] == 0);
```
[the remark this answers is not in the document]
elenadimitrova: oh I don't know, I'm just double making sure because of unknown unknowns..

contracts/ColonyNetworkAuction.sol
```diff
@@ -0,0 +1,216 @@
...
+    require(_clnyToken != 0x0 && _token != 0x0);
+    clnyToken = ERC20Extended(_clnyToken);
+    token = ERC20Extended(_token);
+  }
+
+  function start() public
+  auctionNotStarted
+  {
+    quantity = token.balanceOf(this);
+    // Quantity has to be a minimum of 1e18 tokens
```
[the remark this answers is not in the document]
elenadimitrova: On 1) it's because we open `totalToEndAuction` to be 0 in cases where `quantity(of tokens sold) * price(clny per 1e18 tokens) < 1e18`, e.g. if `price = 1` and `quantity < 1e18`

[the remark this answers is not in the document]
elenadimitrova: ok so what I'll do instead is maintain a variable (per auction) minimum price = `1e18/quantity`

PR #171, opened 2018-02-28
ES6 exports, storage constants, helper functions arguments

lazoviccorp:
- Moved role constants to ColonyStorage.
- Using es6 exports.
- Passing object instead of list to `setupAssignedTask `, `setupFundedTask `, `setupRatedTask `..

### DIFF
--- a/contracts/ColonyStorage.sol
+++ b/contracts/ColonyStorage.sol
@@ -67,6 +67,10 @@ contract ColonyStorage is DSAuth {
   uint256 potCount;
   uint256 domainCount;
 
+  uint8 constant MANAGER = 0;
+  uint8 constant EVALUATOR = 1;
+  uint8 constant WORKER = 2;
+
   struct Task {
     bytes32 specificationHash;
     bytes32 deliverableHash;
--- a/test/colony-network-staking.js
+++ b/test/colony-network-staking.js
@@ -1,6 +1,6 @@
 /* globals artifacts */
-import testHelper from "../helpers/test-helper";
-import testDataGenerator from "../helpers/test-data-generator";
+import { forwardTime, checkErrorRevert } from "../helpers/test-helper";
+import { giveUserCLNYTokens, setupRatedTask, fundColonyWithTokens } from "../helpers/test-data-generator";
 
 const EtherRouter = artifacts.require("EtherRouter");
 const IColony = artifacts.require("IColony");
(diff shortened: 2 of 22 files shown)


contracts/ColonyStorage.sol
```diff
@@ -29,6 +29,10 @@ contract ColonyStorage is DSAuth {
   // this one will have the getters. Make custom getters in the contract that seems most appropriate,
   // and add it to IColony.sol
 
+  uint8 constant MANAGER = 0;
+  uint8 constant EVALUATOR = 1;
+  uint8 constant WORKER = 2;
+
```
elenadimitrova: Because of the `EtherRouter` design the ordering of storage variables here shouldn't be interrupted and the `resolver` address property has to always be the first one. Please append these new storage constants to the end of variable declarations at line https://github.com/JoinColony/colonyNetwork/pull/171/files#diff-f4e9c4137ebcc35fead90e9b07c10f4dR73

test/colony-network-staking.js
```diff
@@ -682,30 +682,30 @@ contract("ColonyNetworkStaking", accounts => {
...
       await clny.approve(colonyNetwork.address, "1000000000000000000");
       await colonyNetwork.deposit("1000000000000000000");
       const addr = await colonyNetwork.getReputationMiningCycle.call();
-      await testHelper.forwardTime(3600, this);
+      await forwardTime(3600, this);
       const repCycle = ReputationMiningCycle.at(addr);
       await repCycle.submitNewHash("0x12345678", 10, 10);
-      await testDataGenerator.fundColonyWithTokens(commonColony, clny, "300000000000000000000");
-      const taskId1 = await testDataGenerator.setupRatedTask(colonyNetwork, commonColony);
+      await fundColonyWithTokens(commonColony, clny, "350000000000000000000");
```
elenadimitrova: Why did this value change? Not that it makes a difference in the test but wondering?

contracts/ColonyStorage.sol
elenadimitrova: Maybe it's not even needed we move these to Storage contract

lazoviccorp: Right now they are used in ColonyFunding and ColonyTask, but since functions in ColonyFunding that are using these variables are probably going to be removed, i can revert this part if you want

test/colony-network-staking.js
lazoviccorp: The test fails on [this](https://github.com/JoinColony/colonyNetwork/blob/develop/helpers/test-data-generator.js#L119) line i think, since there are not enough tokens to move. I don't know how this was not failing earlier. Maybe i'm missing something

contracts/ColonyStorage.sol
elenadimitrova: No, it's fine. Let's leave them there

test/colony-network-staking.js
elenadimitrova: Just debugged this to see what's happening and found it's due to the unit test actually inheriting commonColony funds from previous test. The totalPayouts are correctly `350000000000000000000`. 
Since the common colony funding is not under testing here and tests are very slow already, I'm inclined to accept a little dirty test harness setup.

PR #176, opened 2018-03-08
Refactor task update reviews

elenadimitrova:
Update the design of task properties update based on mutual agreement between task role users. i.e. updating the `task.dueDate` property once approved by both the task manager and worker. This refactoring is inspired by the [SimpleMultiSig design by Chris Lundkvist](https://github.com/christianlundkvist/simple-multisig)

SimpleMultiSig release article
https://medium.com/@ChrisLundkvist/exploring-simpler-ethereum-multisig-contracts-b71020c19037 

Task updates are now considerably cheaper at ~98,000 gas as compared to previously incurring ~300,000 gas (`proposeTaskChange` 179235 + `approveTaskChange` 121274)


test/colony.js
```diff
@@ -72,6 +81,11 @@ contract("Colony", () => {
       assert.equal(taskCount, 0);
     });
 
+    it("should return zero for taskChangeNonce", async () => {
+      const taskChangeNonce = await colony.getTaskChangeNonce.call();
+      assert.equal(taskChangeNonce, 0);
```
area: We should check in a test that it increments correctly.

contracts/ColonyTask.sol
```diff
@@ -137,34 +137,50 @@ contract ColonyTask is ColonyStorage, DSMath {
...
-    require(task.roles[_role].user == msg.sender);
     require(!task.finalized);
-
+    
     uint8[2] storage _reviewers = reviewers[sig];
-    require(_reviewers[0] != 0 || _reviewers[1] != 0);
-    require(_reviewers[0] == _role || _reviewers[1] == _role);
+    uint8 r1 = _reviewers[0];
+    uint8 r2 = _reviewers[1];
+    require(r1 != 0 || r2 != 0);
```
area: Took me a while to realise that this is `||` and not `&&` because `0` represents the manager. Can we add a comment noting that?

contracts/ColonyTask.sol
```diff
@@ -137,34 +137,50 @@ contract ColonyTask is ColonyStorage, DSMath {
...
 
-  function approveTaskChange(uint256 _transactionId, uint8 _role) public {
-    Transaction storage _transaction = transactions[_transactionId];
-    bytes memory _data = _transaction.data;
-    var (sig, taskId) = deconstructCall(_data);
+  function executeTaskChange(
+    uint8[] _sigV,
+    bytes32[] _sigR,
+    bytes32[] _sigS,
+    uint256 _value,
```
area: Surely value is always going to be zero in our case? I can't think of a reason it matters, given that we always call against `address(this)`, other than if they signed a transaction that moved more ETH than the colony had (which would cause the call to fail). I'm all for being extensible in the future without changing the interface, but for now, surely it's better to be more explicit about this with e.g. `require(_value==0)`?

contracts/ColonyTask.sol
```diff
@@ -137,34 +137,50 @@ contract ColonyTask is ColonyStorage, DSMath {
...
+    uint8 r1 = _reviewers[0];
+    uint8 r2 = _reviewers[1];
+    require(r1 != 0 || r2 != 0);
+    
+    // Follows ERC191 signature scheme: https://github.com/ethereum/EIPs/issues/191
+    bytes32 txHash = keccak256(
+      byte(0x19),
+      byte(0),
+      this,
+      address(this),
```
area: Why is this line `address(this)` and the previous line not?

test/colony-task-work-rating.js
```diff
@@ -217,10 +217,9 @@ contract("Colony Task Work Rating", () => {
     it("should assign rating 5 to manager and penalise worker by 0.5, when they haven't submitted rating on time", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
       await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: EVALUATOR });
-      await forwardTime(SECONDS_PER_DAY * 5, this);
+      await forwardTime(SECONDS_PER_DAY * 6, this);
```
area: I'd prefer `SECONDS_PER_DAY * 5 + 1` to `SECONDS_PER_DAY*6` to make it clearer the boundary that is being skipped over here.

test/colony-task-work-rating.js
```diff
@@ -272,7 +271,7 @@ contract("Colony Task Work Rating", () => {
 
     it("should assign rating 5 to manager and 4.5 to worker when no one has submitted any ratings", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
-      await forwardTime(SECONDS_PER_DAY * 10, this);
+      await forwardTime(SECONDS_PER_DAY * 11, this);
```
area: Again, would prefer `SECONDS_PER_DAY * 10 + 1`

test/colony-task-work-rating.js
```diff
@@ -235,11 +234,11 @@ contract("Colony Task Work Rating", () => {
     it("should assign rating 5 to manager and 0 to worker, when ratings not submitted on time and their own rating is below 5", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
 
-      const RATING_SECRET_2 = web3Utils.soliditySha3(RATING_2_SALT, 4);
+      const RATING_SECRET_2 = await web3Utils.soliditySha3(RATING_2_SALT, 4);
       await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_SECRET_2, { from: EVALUATOR });
-      await forwardTime(SECONDS_PER_DAY * 5 + 2, this);
+      await forwardTime(SECONDS_PER_DAY * 5 + 20, this);
```
area: If `+1` is not adequate here, we should really work out why.

test/colony-task-work-rating.js
```diff
@@ -214,10 +217,9 @@ contract("Colony Task Work Rating", () => {
     it("should assign rating 5 to manager and penalise worker by 0.5, when they haven't submitted rating on time", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
       await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: EVALUATOR });
-      await forwardTime(SECONDS_PER_DAY * 5, this);
+      await forwardTime(SECONDS_PER_DAY * 6, this);
```
area: Would prefer `SECONDS_PER_DAY*5 + 1` to make it clear the boundary that is being tested here.

test/colony-task-work-rating.js
```diff
@@ -232,11 +234,11 @@ contract("Colony Task Work Rating", () => {
     it("should assign rating 5 to manager and 0 to worker, when ratings not submitted on time and their own rating is below 5", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
 
-      const RATING_SECRET_2 = web3Utils.soliditySha3(RATING_2_SALT, 4);
+      const RATING_SECRET_2 = await web3Utils.soliditySha3(RATING_2_SALT, 4);
       await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_SECRET_2, { from: EVALUATOR });
-      await forwardTime(SECONDS_PER_DAY * 5 + 2, this);
+      await forwardTime(SECONDS_PER_DAY * 5 + 20, this);
```
area: If `+1` is not adequate here/below, we should work out why.

test/colony-task-work-rating.js
```diff
@@ -269,7 +271,7 @@ contract("Colony Task Work Rating", () => {
 
     it("should assign rating 5 to manager and 4.5 to worker when no one has submitted any ratings", async () => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony });
-      await forwardTime(SECONDS_PER_DAY * 10, this);
+      await forwardTime(SECONDS_PER_DAY * 11, this);
```
area: Again, would prefer `SECONDS_PER_DAY*10 + 1`

contracts/ColonyTask.sol
area: I think we should also comment this is the line stopping someone calling an arbitrary function on the contract.

area: Also that this line is the only thing stopping someone calling an arbitrary function on the contract.

contracts/ColonyTask.sol
elenadimitrova: This was for future extensibility yes but we can restrict tx value to 0 for now

PR #177, opened 2018-03-09
Gas cost tests refactoring to use eth-gas-reporter

elenadimitrova:
Gas cost tests were switched to `ganache-cli` as time forward functionality is not available in parity yet. Additionally private key extraction of test accounts is easier. 

We also make use of  `eth-gas-reporter` to report on gas costs as this provides better reporting (screenshot below) and simplifies our tests. We may consider running it against the entire test suite under `/test`, rather than just the limited `gasCosts` tests in future when the performance is improved. Currently this is prohibitive.  

screen shot 2018-03-09 at 10 40 01

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -28,7 +28,7 @@ module.exports = {
     reporter: "mocha-circleci-reporter",
     reporterOptions: {
       currency: "USD",
-      gasPrice: 20,
+      gasPrice: 5,
       onlyCalledMethods: true
     }
   },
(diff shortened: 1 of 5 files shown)


truffle.js
```diff
@@ -25,7 +25,12 @@ module.exports = {
     }
   },
   mocha: {
-    reporter: "mocha-circleci-reporter"
+    reporter: "mocha-circleci-reporter",
+    reporterOptions: {
+      currency: "USD",
+      gasPrice: 20,
```
area: Current 'Safe Low' gas price (according to ETH Gas Station) is 3. How about making this 5?

PR #178, opened 2018-03-10
Allow colonies to be created with existing tokens

### LINKED ISSUE #160: Allow colonies to be created with existing tokens
As it stands, when a colony is created, a new token is deployed alongside the colony. This should be optional, with the alternative being that the creator provides the address of (what is presumed to be) an ERC20 token, that is then used as the colony's own token. If the address passed to the `createColony` function is `0x0`, a new token contract is deployed, as happens currently.

If the supplied contract is not an ERC20 token, things will break in fun and exciting ways - that is unavoidable, and mitigating that should be considered outside the scope of the PR that implements this issue.

lazoviccorp:
Closes #160

- Token is no longer created inside `createColony` function. Token address is passed instead.
Checking if token has ERC20 interface is not implemented in this task.

### DIFF
--- a/test/colony-network.js
+++ b/test/colony-network.js
@@ -14,6 +14,7 @@ import { setupColonyVersionResolver } from "../helpers/upgradable-contracts";
 
 const EtherRouter = artifacts.require("EtherRouter");
 const Colony = artifacts.require("Colony");
+const Token = artifacts.require("Token");
 const ColonyFunding = artifacts.require("ColonyFunding");
 const ColonyTask = artifacts.require("ColonyTask");
 const ColonyTransactionReviewer = artifacts.require("ColonyTransactionReviewer");
@@ -94,7 +95,8 @@ contract("ColonyNetwork", accounts => {
 
   describe("when creating new colonies", () => {
     it("should allow users to create new colonies", async () => {
-      await colonyNetwork.createColony(COLONY_KEY, ...TOKEN_ARGS);
+      const token = await Token.new(...TOKEN_ARGS);
+      await colonyNetwork.createColony(COLONY_KEY, token.address);
       const address = await colonyNetwork.getColony.call(COLONY_KEY);
       const colonyCount = await colonyNetwork.getColonyCount.call();
       assert.notEqual(address, 0x0);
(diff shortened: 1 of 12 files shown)


test/colony-network.js
```diff
@@ -256,20 +270,23 @@ contract("ColonyNetwork", accounts => {
 
   describe("when adding a skill", () => {
     beforeEach(async () => {
-      await colonyNetwork.createColony("Common Colony", ...TOKEN_ARGS);
+      const token = await Token.new(...TOKEN_ARGS);
+      await colonyNetwork.createColony(COLONY_KEY, token.address);
+      const colony = await colonyNetwork.getColony.call(COLONY_KEY);
+      await token.setOwner(colony);
```
elenadimitrova: Why isn't this happening in the `createColony` call when the token is assigned to the colony?

test/colony.js
```diff
@@ -44,13 +44,13 @@ contract("Colony", () => {
   beforeEach(async () => {
     COLONY_KEY = getRandomString(7);
     const tokenArgs = getTokenArgs();
-    await colonyNetwork.createColony(COLONY_KEY, ...tokenArgs);
+    token = await Token.new(...tokenArgs);
+    await colonyNetwork.createColony(COLONY_KEY, token.address);
     const address = await colonyNetwork.getColony.call(COLONY_KEY);
+    await token.setOwner(address);
```
elenadimitrova: Same as above

test/reputation-update.js
```diff
@@ -39,11 +39,11 @@ contract("Colony Reputation Updates", () => {
     colonyNetwork = await IColonyNetwork.at(etherRouter.address);
     await setupColonyVersionResolver(colony, colonyTask, colonyFunding, colonyTransactionReviewer, resolver, colonyNetwork);
     const tokenArgs = getTokenArgs();
-    await colonyNetwork.createColony("Common Colony", ...tokenArgs);
+    colonyToken = await Token.new(...tokenArgs);
+    await colonyNetwork.createColony("Common Colony", colonyToken.address);
     const commonColonyAddress = await colonyNetwork.getColony.call("Common Colony");
+    await colonyToken.setOwner(commonColonyAddress);
```
elenadimitrova: Same as above

test/colony-network.js
lazoviccorp: Because ColonyNetwork is not the owner of the token, so it cant assign another owner. It think it cant be done with `delegatecall` solidity function either, since i need to change the storage of the `DSAuth` contract, but i need to preserve the `msg.sender`. Do you have a suggestion what is the best way to do this?

elenadimitrova: After a discussion with @area he made a good point that ERC20 is not `ownable` so let's leave it as you've implemented and rely on Token owners to set the Token ownership correctly.

PR #179, opened 2018-03-13
Add reputation staking in gas cost test

### LINKED ISSUE #173: Evaluate gas costs for reputation staking
Following the introduction of [staking](https://github.com/JoinColony/colonyNetwork/pull/156), we want to track the corresponding gas costs.

Actions to be measured:

1. staking tokens
2. submit new hashes
3. challenging & resolving the new reputation hash
4. withdrawing tokens

lsenta:
Closes #173


elenadimitrova:
Change all `stacking` to `staking`, they mean different things in English.

lsenta:
Typo fixed, thanks for the note.

I added the measure for the hash submission process.

elenadimitrova:
There seems to be an issue with gas cost reports for `ReputationMiningCycle` contract, logged here
https://github.com/cgewecke/eth-gas-reporter/issues/64

elenadimitrova:
@lsenta can you have a look at the suggested fix in issue above and see if you can coerce `ReputationMiningCycle` to show its gas costs?

lsenta:
@elenadimitrova I implemented the following fixes and it's working:

- give the contract it's own file,
- require it during migration,
- exclude it from the check-storage script.

The last one was a "just get it done" fix, let me know if we want to refactor the smart contract instead (I'll open a ticket).

contracts/ColonyNetworkStaking.sol
```diff
@@ -122,195 +123,4 @@ contract ColonyNetworkStaking is ColonyNetworkStorage, DSMath {
...
-      IColonyNetwork(colonyNetworkAddress).punishStakers(submittedHashes[disputeRounds[round][idx].hash][disputeRounds[round][idx].nNodes]);
-
-    }
-    //TODO: Can we do some deleting to make calling this as cheap as possible for people?
-  }
-
-
-
-}
+// TODO: Can we handle a dispute regarding the very first hash that should be set?
```
elenadimitrova: This comment is in `ReputationMiningCycle` contract already. Does it need to be here too?

migrations/2_deploy_contracts.js
```diff
@@ -16,4 +17,11 @@ module.exports = (deployer, network) => {
   deployer.deploy([ColonyNetworkStaking]);
   deployer.deploy([EtherRouter]);
   deployer.deploy([Resolver]);
+
+  // We `require` the ReputationMiningCycle object to make sure
+  // it is injected in the `artifacts` variables during test
+  // preparation. We need this for the eth-gas-reporter.
+  // See https://github.com/cgewecke/eth-gas-reporter/issues/64
+  // This log prevents "variable not used" warnings.
+  console.log(ReputationMiningCycle);
```
elenadimitrova: Can we disable "variable not used" here instead of logging the contract. That will be a very messy console output.

scripts/check-storage.js
```diff
@@ -12,7 +12,8 @@ fs.readdirSync("./contracts/").forEach(contractName => {
       "EtherRouter.sol",
       "Migrations.sol",
       "Resolver.sol",
-      "Token.sol"
+      "Token.sol",
+      "ReputationMiningCycle.sol"
```
elenadimitrova: No need to include it here as this is not an `EternalStorage`-based contract

lsenta: The build broke without it:
https://circleci.com/gh/JoinColony/colonyNetwork/1119

From what I understand:
We want to avoid state defined in contract that inherit from ColonyNetworkStorage. That's a way to centralize all the state in the ColonyNetworkStorage definition,
right?

If that's the case we might need to check the inheritance hierarchy in the `check-storage.js` script.

elenadimitrova: Sorry the diff was hiding the part that this is an exclusion list. You've correctly excluded it, my bad.

contracts/ColonyNetworkStaking.sol
lsenta: Bad copy and paste, I'll put it at the top of the network staking contract since it's feature related.

migrations/2_deploy_contracts.js
lsenta: Nice catch, we don't even need a variable... force of habit!

PR #182, opened 2018-03-18
Bootstrapping reputation

### LINKED ISSUE #40: Bootstrapping reputation (new Colonies)
### 6.2.3 Bootstrapping reputation
Since a colony’s decision making procedure rests on reputation weighted voting, we are presented with a bootstrapping problem for new colonies. When a colony is new, no-one has yet completed any work in it and so nobody will have earned any reputation. Consequently, no objections can be raised and no disputes can be resolved as no-one is able to vote. Then, once the first task is successfully completed, that user has a dictatorship over decisions in the same domains or skills until another user earns similar types of reputation.
To prevent this, when a colony is created, the creator can choose addresses to have initial reputation assigned to them in the colony-wide domain to allow the colony to bootstrap itself. The reputation assigned to each a user will be equal to the number of tokens received, i.e. if a member receives ten tokens, they also receive ten reputation in the colony-wide domain. Given that reputation decays over time, this initial bootstrapping will not have an impact on the longterm operation of the colony. This is the only time that reputation can be created without an associated task being paid out. Users receiving reputation are presumably the colony creator and their colleagues, and this starting reputation should be seen as a representation of the existing trust which exists within the team.

filiplazovic:
Closes #40

Added function for bootstrapping the colony.
- Function expects array of addresses and amount of reputation/tokens for every address.
- Function can be called:
  - Only by the authorized address
  - When the colony is in 'bootstrap phase' i.e. when `taskCount` is `0`.
- Reputation is assigned in the root domain.
- It is assumed that there is enough tokens to be transferred.

***Updated**

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -61,6 +61,20 @@ contract Colony is ColonyStorage {
     });
   }
 
+  function bootstrapColony(address[] _users, int[] _amounts) public
+  auth
+  isInBootstrapPhase
+  {
+    require(_users.length == _amounts.length);
+
+    for (uint i = 0; i < _users.length; i++) {
+      require(_amounts[i] >= 0);
+
+      token.transfer(_users[i], uint(_amounts[i]));
+      IColonyNetwork(colonyNetworkAddress).appendReputationUpdateLog(_users[i], _amounts[i], domains[1].skillId);
+    }
+  }
+
   function mintTokens(uint _wad) public
   auth
   {
--- a/test/colony.js
+++ b/test/colony.js
@@ -1,4 +1,7 @@
 /* globals artifacts */
+
+import { toBN } from "web3-utils";
+
 import {
   MANAGER,
   EVALUATOR,
@@ -31,13 +34,19 @@ import {
 } from "../helpers/test-helper";
 import { fundColonyWithTokens, setupRatedTask, setupAssignedTask, setupFundedTask } from "../helpers/test-data-generator";
 
+import { setupColonyVersionResolver } from "../helpers/upgradable-contracts";
+
+const Colony = artifacts.require("Colony");
+const Resolver = artifacts.require("Resolver");
 const EtherRouter = artifacts.require("EtherRouter");
 const IColony = artifacts.require("IColony");
 const IColonyNetwork = artifacts.require("IColonyNetwork");
 const Token = artifacts.require("Token");
 const Authority = artifacts.require("Authority");
+const ColonyFunding = artifacts.require("ColonyFunding");
+const ColonyTask = artifacts.require("ColonyTask");
 
-contract("Colony", () => {
+contract("Colony", addresses => {
   let COLONY_KEY;
   let colony;
   let token;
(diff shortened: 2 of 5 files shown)


contracts/Colony.sol
```diff
@@ -61,6 +61,23 @@ contract Colony is ColonyStorage {
     });
   }
 
+  function bootstrapColony(uint _amountPerUser, address[] _users) public
```
area: I suspect that people will very commonly want to bootstrap their colony such that users have different amounts of starting reputation from each other. Why not take an array of `uint`s as well? In that scenario, don't worry about figuring out at the start whether there are enough tokens - get `balanceOf(this)` at the start, and track what it is as you sent tokens out; if not enough would remain for the next transfer, then `mint` some.

Alternatively, it would be more gas efficient to just ignore minting in this function at all, and we expect them to call `mint` before calling this function if they see that it's going to be necessary. I am aware this would be contrary to what I told you originally via Gitter - apologies for that. I hadn't properly thought about whether one could efficiently implement what I was asking for. I think I would prefer that implementation, overall.

contracts/Colony.sol
```diff
@@ -61,6 +61,23 @@ contract Colony is ColonyStorage {
...
+  auth
+  isInBootstrapPhase
+  {
+    colonyBootstrapped = true;
```
area: I don't see a need for this to be its own variable. We can simply say that `taskCount==0` means that we're in the bootstrap phase. Note that this doesn't mean that we replace this line with `taskCount+=1`. There is no inherent problem with calling `bootstrapColony` multiple times (as it is, at least). Indeed, if someone wanted to give *a lot* of people starting reputation and balances, they might need to call the function multiple times if the list was so long that the transaction that contained all of them required more `gas` than the block limit!

Once the first task is created as normal, this function becomes unable to be called because of the condition in `isInBootstrapPhase`.

contracts/Colony.sol
```diff
@@ -61,6 +61,23 @@ contract Colony is ColonyStorage {
...
+
+    uint diff = (_amountPerUser * _users.length) - token.balanceOf(this);
+    if (diff > 0) {
+      token.mint(diff);
+    }
+
+    for (uint i = 0; i < _users.length; i++) {
+      token.transfer(_users[i], _amountPerUser);
+      IColonyNetwork(colonyNetworkAddress).appendReputationUpdateLog(_users[i], int(_amountPerUser), 1);
```
area: The skill that this reputation is being added to should be the skill related to the colony's top-level domain, which will be stored in `domains[1].skillId` as you said.

This cast to `int` is required, because the reputationUpdateLog requires an `int` for that parameter. However, there is no guard to check that it is not going to underflow. We should insert one, or require the input to be of type `int` and then require that it is not below zero (and then cast it to `uint` for the token transfer).

contracts/Colony.sol
filiplazovic: Thanks, i agree that tokens should be minted outside of this function.

contracts/Colony.sol
filiplazovic: Yea, beginner mistake with the `int`s.

test/colony.js
```diff
@@ -180,6 +199,86 @@ contract("Colony", () => {
...
+      const amount = toBN(10 * 1e18).toString();
+      await colony.mintTokens(amount);
+      colony.bootstrapColony([INITIAL_ADDRESSES[0]], [INITIAL_REPUTATIONS[0]]);
+      colony.bootstrapColony([INITIAL_ADDRESSES[0]], [INITIAL_REPUTATIONS[0]]);
+
+      const balance = await token.balanceOf(INITIAL_ADDRESSES[0]);
+      assert.equal(balance.toString(), amount);
+    });
+
+    it("should throw if length of inputs is not equal", async () => {
```
area: Strictly, this test is only checking it throws if `INITIAL_ADDRESSES` is longer then `INITIAL REPUTATIONS`. If the reverse is true, it's not going to throw. Either add a check for the inputs being the same length or reword this test's description. I think I have a slight preference for the former; if the inputs are not the same, it's almost certainly a mistake on the user's part and so we shouldn't take action that they can't revert.

filiplazovic: Agreed, thank you.

PR #183, opened 2018-03-22
Use EtherRouter for functions with dynamic return data size

elenadimitrova:
- Use `returndatasize` and `returndatacopy` opcodes introduced in Byzantium hard fork which give us access to the size of the return data for a call
http://solidity.readthedocs.io/en/v0.4.21/assembly.html#opcodes 
This also simplifies the `Resolver` requirements as we no longer need to track return data size for functions.

- Byzantium EIPs are allowed in prity config

- Since the above is now possible we've extended the `getTask` function to return the task domains and skills too instead of relying on recursive calls to `getTaskDomain` and `getTaskSkill` both of which have been removed. 

- Task model has been switched to use a single domain only.

- We remove leftover storage variables from `ColonyStorage` related to the old method of reviewing task updates.

### DIFF
--- a/test/token.js
+++ b/test/token.js
@@ -1,5 +1,5 @@
 /* globals artifacts */
-import { getTokenArgs, expectEvent, checkErrorRevert, web3GetBalance, checkErrorNonPayableFunction } from "../helpers/test-helper";
+import { getTokenArgs, expectEvent, checkErrorAssert, checkErrorRevert, web3GetBalance, checkErrorNonPayableFunction } from "../helpers/test-helper";
 import { setupUpgradableToken } from "../helpers/upgradable-contracts";
 
 const EtherRouter = artifacts.require("EtherRouter");
@@ -61,7 +61,7 @@ contract("Token", accounts => {
     });
 
     it("should NOT be able to transfer more tokens than they have", async () => {
-      await checkErrorRevert(etherRouterToken.transfer(ACCOUNT_TWO, 1500001));
+      await checkErrorAssert(etherRouterToken.transfer(ACCOUNT_TWO, 1500001));
       const balanceAccount2 = await etherRouterToken.balanceOf.call(ACCOUNT_TWO);
       assert.equal(0, balanceAccount2.toNumber());
     });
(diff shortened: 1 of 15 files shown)


test/token.js
```diff
@@ -61,7 +61,7 @@ contract("Token", accounts => {
     });
 
     it("should NOT be able to transfer more tokens than they have", async () => {
-      await checkErrorRevert(etherRouterToken.transfer(ACCOUNT_TWO, 1500001));
+      await checkErrorAssert(etherRouterToken.transfer(ACCOUNT_TWO, 1500001));
```
area: I don't think I exactly understand why these have gone from `revert`s to `assert`s?

package.json
```diff
@@ -15,7 +15,7 @@
     "start:blockchain:client": "bash ./scripts/start-blockchain-client.sh",
     "stop:blockchain:client": "bash ./scripts/stop-blockchain-client.sh",
     "test:contracts": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle test --network development",
-    "test:contracts:upgrade": "npm run start:blockchain:client parity & npm run generate:test:contracts && truffle migrate --reset --compile-all && truffle test ./upgrade-test/* --network integration",
+    "test:contracts:upgrade": "npm run start:blockchain:client & npm run generate:test:contracts && truffle migrate --reset --compile-all && truffle test ./upgrade-test/* --network integration",
```
area: Do the upgrade tests no longer work on Parity? I can't spot a time advancement anywhere - is there another reason?

test/token.js
elenadimitrova: Because the old `EtherRouter` did a `require` check on the success of the `delegatecall` success. If that failed, even for an `assert` we still got a `require` failure

area: So the require check has been replaced by a `switch`. If it fails, it then does a `revert`, which is what a `require` does, right, and so the behaviour should be the same?

elenadimitrova: I take that bit back, you're right. I am still not 100% sure what is going on here but the history is that when the ganache version switched, this somehow broke the way failed asserts behaved when executed via `EtherRouter` which necessitated these changes: https://github.com/JoinColony/colonyNetwork/commit/780227a65c84a18a2e7a59ef3472fc86acb9ecd7#diff-6af24f2316d75cbe66c8a65032a385ba

elenadimitrova: And before that when we switched the EtherRouter from using an `assert` to `require` those `checkErrorAssert`s on Token test worked https://github.com/JoinColony/colonyNetwork/pull/129/files#diff-3776618b80cc8f554676f40c34ae0c9f

package.json
elenadimitrova: It turns out you have to explicitly enable the Hard fork EIPs for Parity to work with them
https://github.com/paritytech/parity/commit/07c63c14f37a5be4f214dfe6acff033f3ec6f1c0

PR #195, opened 2018-04-13
Claiming Rewards from the Reward Pot

### LINKED ISSUE #95: Claiming Rewards from the Reward Pot
### 10.2 Claiming Rewards from the Reward Pot
Rewards accumulate in the rewards pot. To trigger an actual payout to users (i.e. to make rewards claimable) a special type of proposal is made, proposing that all users should receive a payout based on the reward pot’s holdings.
This reward payout proposal includes the specific currency that should be paid, and only one currency is handled at a time. In the event that the proposal is approved by vote of reputation, then all user’s tokens are locked until they claim their payout. Locking is necessary, because the token balance of each account factors into the rewards formula of Equation (7). Locking is triggered by incrementing the colony’s ‘most recent payout’ counter.
Our currency contract contains a locking mechanism ensuring that a user cannot move tokens while they have (token-weighted) votes to reveal; we use the same mechanism here to ensure that a user cannot move tokens after a payout is approved by the members of the colony but before the user has claimed their rewards. The colony has a counter for each user that is incremented whenever they claim a payout; they can also waive their claim to a payout that will increment this counter.
While it is of course up to the members of each individual colony to decide, it is advisable that these payout proposals should only be accepted sporadically to keep the gas costs low for the users claiming their payouts, as well as simply to not be a nuisance to the users continually finding their tokens locked.
**Rewards are only available to accounts that hold both tokens and reputation**, and the amount claimable by each account depends on both token balance and reputation (see equation (7) below). Therefore we need to have a similar behaviour to ‘lock’ the reputation of the users for the payout. When a payout is activated, the current state of the reputation tree is recorded in the payout itself. Users are paid out according to their reputation in this state, rather than the most recent state, to ensure all users get an appropriate payout and to avoid exploiting the system (which might otherwise be possible via e.g. delaying reward collection until after completing a task, increasing their reputation).

**The Rewards Formula**
The amount that each user (ui) of a colony (C) is entitled to claim (pi) is a function of their colony
token holdings (ti) and their total reputation in the colony (ri):
screen shot 2017-10-03 at 13 21 14
This is a (normalised) geometric average of the user’s token holdings and reputation. We note that this is very unlikely to payout all the tokens set aside for a payout — the only way it would do so is if everyone had the same proportion of reputation in the colony as they did proportion of tokens in the colony. However, the geometric average is the natural way to fairly capture the influence of two variables with different ranges, and ensures that large token holders must earn large amounts of reputation to get the most from the payouts. The total reputation and user reputation in the colony are all provable on-chain at claim time via a Merkle proof that the `ReputationRootHash` (Section 7) contains some values claimed by the user; the user’s balance of colony tokens and the total number of tokens issued is trivial to lookup.
After some sufficiently long period of time (500000 blocks), all unclaimed tokens can be reclaimed on behalf of the colony by a user, and the payout closed. Any users that have not claimed their payout by that point will still have their tokens locked, and they will remain locked until they issue a transaction waiving their claim to the payout (indeed, they already passively did this by not claiming it in a timely fashion). Unclaimed tokens are returned to the reward pot and become part of the next reward cycle.

filiplazovic:
Closes #95 

# Reward payout

Storage variables:
- Global reward payout counter - `globalRewardPayoutCount`
- Active reward payouts - `activeRewardPayouts`
- User reward payout counter mapping - `userRewardPayoutCount`
- Reward payout cycles mapping - `rewardPayoutCycles`

Implemented functions for:
- starting the payout cycle - `startNextRewardPayout`
- claiming reward payout - `claimRewardPayout`
- waiving reward - `waiveRewardPayout`
- moving remaining funds after payout - `moveRemainingRewardPayoutFunds`
- getting the info about payout - `getRewardPayoutInfo`

## Starting the payout cycle

Takes one argument: token address

Saving all the necessary data to `rewardPayoutCycles` (current root hash, current total tokens, reward pot balance, token, current block number).
Incrementing `globalRewardPayoutCount` and using it for mapping key.
Setting payout token to active, which will not allow creation of new payouts with the same token until the current one finishes.

## Claiming the reward payout

Takes three arguments: payout id, user reputation, colony-wide reputation

Can be called only if reward payout is active i.e if its created less than 60 days ago.
Can be called only if payout is not claimed, i.e if `payoutId - userRewardPayoutCount = 1` (To force the user to claim the payout that is next in queue, otherwise there would be a lot of problems).

Calculating amount to be send to user according to formula defined in the whitepaper.
Formula
Incrementing `userRewardPayoutCount`.
Sending the user the calculated amount.

## Waiving reward payout

Takes one argument: number of payouts to waive

Can only be called if user has not claimed that reward i.e if `payoutId - userRewardPayoutCount = 1`.

Incrementing `userRewardPayoutCount`.

## Moving remaining funds after payout

Can be called only if payout is finished i.e if its created more than 60 days ago.
Setting payout token to inactive. This will allow creation of reward payout with that token.

Moves remaining funds to pot 0.

## Getting the info about the payout
Takes one argument: payout id

Getting the useful information about the payout for UI.

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -19,12 +19,16 @@ pragma solidity ^0.4.23;
 pragma experimental "v0.5.0";
 
 import "../lib/dappsys/math.sol";
+import "./SafeMath.sol";
 import "./ERC20Extended.sol";
 import "./IColonyNetwork.sol";
 import "./ColonyStorage.sol";
 
 
 contract ColonyFunding is ColonyStorage, DSMath {
+  event RewardPayoutCycleStarted(uint256 indexed id);
+  event RewardPayoutCycleEnded(uint256 indexed id);
+
   function getFeeInverse() public pure returns (uint256) {
     // TODO: refer to ColonyNetwork
     return 100;
@@ -138,6 +142,105 @@ contract ColonyFunding is ColonyStorage, DSMath {
 
   function getNonRewardPotsTotal(address _token) public view returns (uint256) {
     return nonRewardPotsTotal[_token];
+  }
+
+  function getGlobalRewardPayoutCount() public view returns (uint256) {
+    return globalRewardPayoutCount;
+  }
+
+  function getUserRewardPayoutCount(address _user) public view returns (uint256) {
+    return userRewardPayoutCount[_user];
+  }
+
+  function startNextRewardPayout(address _token) public auth {
+    require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
+
+    uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
+    require(totalTokens > 0, "colony-reward-payout-invalid-total-tokens");
+
+    activeRewardPayouts[_token] = true;
+    globalRewardPayoutCount += 1;
+
+    //TODO: Lock everyones tokens
+
+    rewardPayoutCycles[globalRewardPayoutCount] = RewardPayoutCycle(
+      IColonyNetwork(colonyNetworkAddress).getReputationRootHash(),
+      totalTokens,
+      pots[0].balance[_token],
+      _token,
+      block.timestamp
+    );
+
+    emit RewardPayoutCycleStarted(globalRewardPayoutCount);
+  }
+
+  function claimRewardPayout(uint256 _payoutId, uint256[7] squareRoots, uint256 _userReputation, uint256 _totalReputation) public {
+    RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
+    require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
+    require(_payoutId - userRewardPayoutCount[msg.sender] == 1, "colony-reward-payout-bad-id");
+
+    //TODO: Prove that userReputation and totalReputation in reputationState are correct
+
+    uint256 userTokens = token.balanceOf(msg.sender);
+
+    require(_totalReputation > 0, "colony-reward-payout-invalid-total-reputation");
+    require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
+    require(_userReputation > 0, "colony-reward-payout-invalid-user-reputation");
+
+    // squareRoots[0] - square root of _userReputation
+    // squareRoots[1] - square root of userTokens
+    // squareRoots[2] - square root of _totalReputation
+    // squareRoots[3] - square root of totalTokens
+    // squareRoots[4] - square root of numerator
+    // squareRoots[5] - square root of denominator
+    // squareRoots[6] - square root of payout.amount
+
+    require(mul(squareRoots[0], squareRoots[0]) <= _userReputation, "colony-reward-payout-invalid-parametar-user-reputation");
+    require(mul(squareRoots[1], squareRoots[1]) <= userTokens, "colony-reward-payout-invalid-parametar-user-token");
+    require(mul(squareRoots[2], squareRoots[2]) <= _totalReputation, "colony-reward-payout-invalid-parametar-total-reputation");
+    require(mul(squareRoots[3], squareRoots[3]) <= payout.totalTokens, "colony-reward-payout-invalid-parametar-total-tokens");
+    require(mul(squareRoots[6], squareRoots[6]) <= payout.amount, "colony-reward-payout-invalid-parametar-amount");
+    uint256 numerator = mul(squareRoots[0], squareRoots[1]);
+    uint256 denominator = mul(squareRoots[2], squareRoots[3]);
+
+    require(mul(squareRoots[4], squareRoots[4]) <= numerator, "colony-reward-payout-invalid-parametar-numerator");
+    require(mul(squareRoots[5], squareRoots[5]) <= denominator, "colony-reward-payout-invalid-parametar-denominator");
+
+    uint256 reward = (mul(squareRoots[4], squareRoots[6]) / (squareRoots[5] + 1)) ** 2;
+
+    pots[0].balance[payout.tokenAddress] = sub(pots[0].balance[payout.tokenAddress], reward);
+
+    userRewardPayoutCount[msg.sender] += 1;
+
+    // TODO: Unlock user tokens
+
+    ERC20Extended(payout.tokenAddress).transfer(msg.sender, reward);
+  }
+
+  function waiveRewardPayouts(uint256 _numPayouts) public {
+    require(add(userRewardPayoutCount[msg.sender], _numPayouts) <= globalRewardPayoutCount, "colony-reward-payout-invalid-num-payouts");
+
+    userRewardPayoutCount[msg.sender] += _numPayouts;
+
+    //TODO unlock user tokens
+  }
+
+  function finalizeRewardPayout(uint256 _payoutId) public {
+    require(_payoutId <= globalRewardPayoutCount, "colony-reward-payout-not-found");
+
+    RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
+
+    require(activeRewardPayouts[payout.tokenAddress], "colony-reward-payout-token-not-active");
+    require(block.timestamp - payout.blockTimestamp > 60 days, "colony-reward-payout-active");
+
+    activeRewardPayouts[payout.tokenAddress] = false;
+
+    emit RewardPayoutCycleEnded(_payoutId);
+  }
+
+  function getRewardPayoutInfo(uint256 _payoutId) public view returns (bytes32, uint256, uint256, address, uint256) {
+    RewardPayoutCycle memory rewardPayoutInfo = rewardPayoutCycles[_payoutId];
+    return (rewardPayoutInfo.reputationState, rewardPayoutInfo.totalTokens, rewardPayoutInfo.amount, rewardPayoutInfo.tokenAddress, rewardPayoutInfo.blockTimestamp);
   }
 
   function updateTaskPayoutsWeCannotMakeAfterPotChange(uint256 _id, address _token, uint _prev) internal {
--- a/contracts/ColonyStorage.sol
+++ b/contracts/ColonyStorage.sol
@@ -35,7 +35,7 @@ contract ColonyStorage is DSAuth {
   // Mapping function signature to 2 task roles whose approval is needed to execute
   mapping (bytes4 => uint8[2]) reviewers;
   uint256 taskChangeNonce;
-  
+
   mapping (uint256 => Task) tasks;
 
   // Pots can be tied to tasks or domains, so giving them their own mapping.
@@ -43,6 +43,31 @@ contract ColonyStorage is DSAuth {
   // to anything yet, but has had some siphoned off in to the reward pot.
   // Pot 0 is the 'reward' pot containing funds that can be paid to holders of colony tokens in the future.
   mapping (uint256 => Pot) pots;
+
+  struct RewardPayoutCycle {
+    // Reputation root hash at the time of reward payout creation
+    bytes32 reputationState;
+    // Total tokens at the time of reward payout creation
+    uint256 totalTokens;
+    // Amount alocated for reward payout
+    uint256 amount;
+    // Token in which a reward is paid out with
+    address tokenAddress;
+    // Time of creation (in seconds)
+    uint256 blockTimestamp;
+  }
+
+  // Keeps track of all reward payout cycles
+  mapping (uint256 => RewardPayoutCycle) rewardPayoutCycles;
+  // Active payouts for particular token address. Assures that one token is used for only one active payout
+  mapping (address => bool) activeRewardPayouts;
+  // Incremented every time a reward payout is created. Used for comparing with `userRewardPayoutCount`
+  // to ensure that rewards are claimed in the right order.
+  // Can be used for iterating over all reward payouts (excluding index 0)
+  uint256 globalRewardPayoutCount;
+  // Keeps track of how many payouts are claimed by the particular user.
+  // Can be incremented either by waiving or claiming the reward.
+  mapping (address => uint256) userRewardPayoutCount;
 
   // This keeps track of how much of the colony's funds that it owns have been moved into pots other than pot 0,
   // which (by definition) have also had the reward amount siphoned off and put in to pot 0.
@@ -74,7 +99,7 @@ contract ColonyStorage is DSAuth {
     uint256[] skills;
 
     // TODO switch this mapping to a uint8 when all role instances are uint8-s specifically ColonyFunding source
-    mapping (uint256 => Role) roles; 
+    mapping (uint256 => Role) roles;
     // Maps a token to the sum of all payouts of it for this task
     mapping (address => uint256) totalPayouts;
     // Maps task role ids (0,1,2..) to a token amount to be paid on task completion
--- a/contracts/IColony.sol
+++ b/contracts/IColony.sol
@@ -26,7 +26,11 @@ contract IColony {
   /// @notice Event logged when a new task is added
   /// @param id The newly added task id
   event TaskAdded(uint256 indexed id);
-  
+
+  /// @notice Event logged when a new reward payout cycle has started
+  /// @param id Payout id
+  event RewardPayoutCycleStarted(uint256 indexed id);
+
   // Implemented in DSAuth.sol
   /// @notice Get the `Authority` for the colony
   /// @return The `Authority` contract address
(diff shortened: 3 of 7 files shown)


contracts/ColonyFunding.sol
```diff
@@ -141,6 +142,86 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+
+    //TODO: Unlock user tokens
+
+    rewardPayoutCycles[_payoutId].remaining = sub(rewardPayoutCycles[_payoutId].remaining, reward);
+
+    payoutToken.transfer(msg.sender, reward);
+  }
+
+  function waiveRewardPayout(uint8 _payoutId) public {
+    require(_payoutId - userRewardPayoutCount[msg.sender] == 1);
```
area: Arguably, it should be possible for a user to waive multiple payouts at once - if they've been away for a long time, maybe there are lots for them to waive.

contracts/ColonyStorage.sol
```diff
@@ -45,6 +45,18 @@ contract ColonyStorage is DSAuth {
   // Pot 0 is the 'reward' pot containing funds that can be paid to holders of colony tokens in the future.
   mapping (uint256 => Pot) pots;
 
+  struct RewardPayoutCycle {
```
area: Document the properties of this struct.

contracts/ColonyFunding.sol
```diff
@@ -141,6 +142,86 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+  function claimRewardPayout(uint8 _payoutId, int256 _userReputation, int256 _totalReputation) public {
+    require(block.number - rewardPayoutCycles[_payoutId].blockNumber < 500000);
+    require(_payoutId - userRewardPayoutCount[msg.sender] == 1);
+
+    //TODO: Prove that userReputation and totalReputation in reputationState are correct
+
+    address tokenAddress = rewardPayoutCycles[_payoutId].token;
+    ERC20Extended payoutToken = ERC20Extended(tokenAddress);
+    uint256 userTokens = payoutToken.balanceOf(msg.sender);
+    uint256 totalTokens = sub(payoutToken.totalSupply(), payoutToken.balanceOf(address(this)));
```
area: 1. The tokens used to work out the reward a user is entitled to should be the colony tokens they hold in the colony paying out the reward, not the tokens of the type being paid out that they already hold.

2. I hadn't considered removing the colony tokens held by the colony from consideration from this calculation. I think that this will give strictly better behaviour than what was spec'd (everyone should get slightly more rewards than they would have done before - i.e. there will be less left over and returned to the pot). Very nice!

area:
The approximations with the payout function is to be expected - even if the calculation could be done precisely on-chain, the payouts could not. I suspect what you've done is accurate enough, but maybe test with very small numbers for both the tokens and the reputation is in order just to see how bad it can be.

Parallel payouts with one token is an interesting one. I like that you've taken that possibility in to account so far; given that an additional mapping would be required, and we want to encourage payouts to be fairly infrequent, I think I'm okay with disallowing it. I will check with some others and confirm.

filiplazovic:
Playing with ganache was useful, although the code was rusty.

Thanks for the feedback. I'm just not sure what you meant by:
> even if the calculation could be done precisely on-chain, the payouts could not

area:
>Thanks for the feedback. I'm just not sure what you meant by:
>>    even if the calculation could be done precisely on-chain, the payouts could not

Just that there is a smallest indivisible token unit on-chain, so even if the calculation could be done precisely and required, for example, .1 tokens to be paid out, we couldn't do that.

contracts/IColony.sol
```diff
@@ -276,7 +276,48 @@ contract IColony {
...
-  
+
+  /// @notice Start next reward payout in `_token`. All funds in that are in reward pot at the time will become unavailable.
+  /// All tokens will be locked, and can be unlocked by calling `waiveRewardPayout` or `claimRewardPayout`.
+  /// @param _token Addess of the token used for reward payout
+  function startNextRewardPayout(address _token) public returns (uint256);
+
+  /// @notice Claim the reward payout at `_payoutId`. User needs to provide his reputation and colony-wide reputation
+  /// which will be proven via Merkle proof inside this function.
+  /// Can only be called if payout is active, i.e its created less than 500k blocks ago.
```
area: This comment should talk about time, not blocks.

contracts/IColony.sol
```diff
@@ -276,7 +276,48 @@ contract IColony {
...
+  /// @param _payoutId Id of the reward payout
+  /// @return Reputation root hash at the time of creation
+  /// @return Total amount of tokens taken aside for reward payout
+  /// @return Remaining (unclaimed) amount of tokens
+  /// @return Token address
+  /// @return Block number at the time of creation
+  function getRewardPayoutInfo(uint8 _payoutId) public view returns (bytes32, uint256, uint256, address, uint256);
+
+  /// @notice Moves remaining (unclaimed) tokens back to reward pot
+  /// Can only be called when reward payout cycle is finished i.e its created more than 500k blocks ago
```
area: Again, 60 days

contracts/ColonyFunding.sol
```diff
@@ -141,6 +142,89 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+      amount,
+      _token,
+      block.timestamp
+    );
+
+    return globalRewardPayoutCount;
+  }
+
+  function claimRewardPayout(uint8 _payoutId, int256 _userReputation, int256 _totalReputation) public {
+    require(block.timestamp - rewardPayoutCycles[_payoutId].blockTimestamp < 5184000000); // 60 days
```
area: `block.timestamp` is in seconds, not milliseconds. So this is actually 60000 days! While technically you won't need to change the tests (their very large numbers will still take us past the reward payout window once this is fixed), I'd prefer it if they were changed too.

contracts/ColonyFunding.sol
```diff
@@ -141,6 +142,89 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+    require(_payoutId - userRewardPayoutCount[msg.sender] == 1);
+
+    //TODO: Prove that userReputation and totalReputation in reputationState are correct
+
+    uint256 userTokens = token.balanceOf(msg.sender);
+    uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
+
+    require(userTokens > 0 && totalTokens > 0 && _userReputation > 0 && _totalReputation > 0);
+
+    uint256 diff = 76 - SafeMath.numDigits(totalTokens * uint256(_totalReputation));
```
area: `uint256` can hold up to a 78 digit number in base 10, so this can underflow to a very large value, which I'm not terribly happy with. I *think* the worse case scenario there is that `reward` ends up being zero (i.e. someone gets less than they should, rather than more than they should), but users would be pretty annoyed about that. Did this implementation come from anywhere in particular? Why did you choose 76?

test/colony-funding.js
```diff
@@ -403,4 +407,367 @@ contract("Colony Funding", () => {
...
+        totalReputation: toBN(Math.floor(Math.random() * 1e12) + 1e8),
+        userReputation: toBN(Math.floor(Math.random() * 1e8)),
+        delta: 10
+      },
+      {
+        totalReputation: toBN(Math.floor(Math.random() * 1e21) + 1e16).mul(toBN(Math.floor(Math.random() * 1e3))),
+        userReputation: toBN(Math.floor(Math.random() * 1e16)),
+        delta: 100
+      }
+    ];
```
area: I like these tests! Can we have more of them, for even more extreme values? 😃

contracts/ColonyFunding.sol
```diff
@@ -141,6 +142,89 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+    uint256 factor = (diff % 2 == 0) ? diff : sub(diff, 1);
+    uint256 sharePercentage = SafeMath.sqrt(SafeMath.percent(userTokens * uint256(_userReputation), totalTokens * uint256(_totalReputation), factor));
+    uint256 reward = rewardPayoutCycles[_payoutId].amount * sharePercentage / (10 ** (factor / 2));
```
area: It also feels like I would be able to cook up an overflow here.

test/colony-funding.js
```diff
@@ -403,4 +407,367 @@ contract("Colony Funding", () => {
...
+          .sqrt();
+        const reward = amount.mul(sharePercentage).round();
+
+        await newColony.claimRewardPayout(payoutId, data.userReputation.toString(), data.totalReputation.toString(), {
+          from: userAddress1
+        });
+
+        const rewardPayoutInfoAfterPayout = await newColony.getRewardPayoutInfo(payoutId);
+        const remaining = rewardPayoutInfoAfterPayout[2];
+
```
area: Can we print out how close we were explicitly for these tests?

contracts/ColonyFunding.sol
filiplazovic: Sorry, i should have checked the documentation.

contracts/ColonyFunding.sol
filiplazovic: I thought it was 77 (idiot me), and 1 will be added in `SafeMath.percent`. Which is not good, anyway.
About the underflows and overflows i was not so sure, thats why i didn't added the tests for more extreme values.
We can can continue this on Gitter, i just realized a few more things.

test/colony-funding.js
filiplazovic: Sure, i just deleted them because i was getting `0` reward (As you mentioned above)

filiplazovic:
@area Can you take a look at this?

test/colony-funding.js
```diff
@@ -403,4 +406,290 @@ contract("Colony Funding", () => {
...
+      await checkErrorRevert(colony.startNextRewardPayout(otherToken.address), "colony-reward-payout-token-active");
+    });
+
+    it("should be able to collect rewards from multiple payouts of different token", async () => {
+      const tokenArgs = getTokenArgs();
+      const newToken = await Token.new(...tokenArgs);
+      await fundColonyWithTokens(colony, newToken, initialFunding.toString());
+
+      const payoutId1 = await colony.startNextRewardPayout.call(newToken.address);
+      await colony.startNextRewardPayout(newToken.address);
```
elenadimitrova: I'd argue to not return anything out of that function and instead get the `payoutId` via emitting an `event` out of the transaction itself. See https://github.com/JoinColony/colonyNetwork/blob/develop/test/colony-network-auction.js#L92

area:
I like the choice of implementation 3 from the [repository we had on the side](https://github.com/area/payoutCalculationComparison); I think that implementation has the properties I'm happiest with. Two things about the tests are giving me pause though:

1. I like tests to be as deterministic as possible - no randomness! I also don't think we need this many; the important regimes to test are very small numbers (for accuracy) and very large ones (for overflow); I would expect to see tests using numbers like `1`, `10`, `1e76`, `1e77` and not much in between. 
2. The amount the solidity implementation is wrong by appears to be positive, which is definitely bad - if true, it'd mean that the last person to claim wouldn't be able to claim what was owed to them. That wasn't the case in the other repository, though, so something a little odd is going on here.

Finally, gas costs for this implementation are higher than I was hoping for. A way to reduce them is to have the user supply the values of the sqrt, which are then verified by squaring the passed claimed square root and checking that it is less than the original value. This calculation can be done automatically by our client, so it won't be more onerous for the user, and should reduce gas costs drastically.

filiplazovic:
1. Im ok with that.
2. I think you are wrong, otherwise the tests would fail. I explicitly added `sqrt(denomerator) + 1` [here](https://github.com/JoinColony/colonyNetwork/pull/195/files#diff-8920cdf7ab56c438c907df9bd37d2931R193) in effort to avoid that. It seems like its doing the job, unless you know a case when it wont work.

Thanks for the hint about gas costs

contracts/IColony.sol
```diff
@@ -275,7 +275,49 @@ contract IColony {
...
   /// @param _token Address of the token, `0x0` value indicates Ether
   function claimPayout(uint256 _id, uint256 _role, address _token) public;
-  
+
+  /// @notice Start next reward payout in `_token`. All funds in that are in reward pot at the time will become unavailable.
+  /// All tokens will be locked, and can be unlocked by calling `waiveRewardPayout` or `claimRewardPayout`.
+  /// @param _token Addess of the token used for reward payout
+  function startNextRewardPayout(address _token) public returns (uint256);
+
+  /// @notice Claim the reward payout at `_payoutId`. User needs to provide his reputation and colony-wide reputation
```
elenadimitrova: Just some notes on comments language:
- Pls use the non-gender specific `their` instead of `his` or `hers`. There may be women interacting with this contract you know :)

contracts/ColonyStorage.sol
```diff
@@ -44,6 +44,26 @@ contract ColonyStorage is DSAuth {
...
+    address tokenAddress;
+    // Unix timestamp of when the payout has started
+    uint256 blockTimestamp;
+  }
+
+  mapping (uint256 => RewardPayoutCycle) rewardPayoutCycles;
+  mapping (address => bool) activeRewardPayouts;
+  uint8 globalRewardPayoutCount;
+  mapping (address => uint256) userRewardPayoutCount;
+
```
elenadimitrova: Comment mappings, arrays, global variables, everything really that is storage. As you see all other items, not just structs in `ColonyStorage` are commented.
Our current stand on documentation is that `IColony` and `IColonyNetwork` are the two interfaces we document plus the two respective storage contracts: `ColonyStorage` and `ColonyNetworkStorage` 
In addition any other non-EtherRouter contracts (not covered by the above) are documented e.g. `ReputationMiningCycle`

contracts/IColony.sol
```diff
@@ -275,7 +275,49 @@ contract IColony {
...
+  /// @return Reputation root hash at the time of creation
+  /// @return Total colony tokens at the time of creation
+  /// @return Total amount of tokens taken aside for reward payout
+  /// @return Remaining (unclaimed) amount of tokens
+  /// @return Token address
+  /// @return Block number at the time of creation
+  function getRewardPayoutInfo(uint8 _payoutId) public view returns (bytes32, uint256, uint256, uint256, address, uint256);
+
+  /// @notice Moves remaining (unclaimed) tokens back to reward pot
+  /// Can only be called when reward payout cycle is finished i.e when 60 days passed from its creation
```
elenadimitrova: when 60 days passed -> when 60 days _have_ passed

contracts/IColony.sol
```diff
@@ -275,7 +275,49 @@ contract IColony {
...
+  /// @notice Claim the reward payout at `_payoutId`. User needs to provide his reputation and colony-wide reputation
+  /// which will be proven via Merkle proof inside this function.
+  /// Can only be called if payout is active, i.e when 60 days passed from its creation.
+  /// Can only be called if next in queue
+  /// @param _payoutId Id of the reward payout
+  /// @param _userReputation User reputation at the point of creation of reward payout cycle
+  /// @param _totalReputation Total reputation at the point of creation of reward payout cycle
+  function claimRewardPayout(uint8 _payoutId, int256 _userReputation, int256 _totalReputation) public;
+
+  /// @notice Waive reward payouts. This will unlock the senders tokens and increment users reward payout counter,
```
elenadimitrova: senders tokens -> sender's tokens

contracts/IColony.sol
```diff
@@ -275,7 +275,49 @@ contract IColony {
   /// @param _role Id of the role, as defined in `ColonyStorage` `MANAGER`, `EVALUATOR` and `WORKER` constants
   /// @param _token Address of the token, `0x0` value indicates Ether
   function claimPayout(uint256 _id, uint256 _role, address _token) public;
-  
+
+  /// @notice Start next reward payout in `_token`. All funds in that are in reward pot at the time will become unavailable.
```
elenadimitrova: "in token" implies in some Token contract, pls replace with "for" or just say "Start next `_token` reward payout"

elenadimitrova: Also here this reads weird
>All funds in that are in reward pot at the time 
Did you mean:
"All funds in the reward pot will become unavailable"

contracts/ColonyFunding.sol
```diff
@@ -140,6 +141,94 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+    //TODO: Prove that userReputation and totalReputation in reputationState are correct
+
+    uint256 userTokens = token.balanceOf(msg.sender);
+
+    require(_totalReputation > 0, "colony-reward-payout-invalid-total-reputation");
+    require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
+    require(_userReputation > 0, "colony-reward-payout-invalid-user-reputation");
+
+    uint256 numerator = mul(SafeMath.sqrt(uint256(_userReputation)), SafeMath.sqrt(userTokens));
+    uint256 denomerator = mul(SafeMath.sqrt(uint256(_totalReputation)), SafeMath.sqrt(payout.totalTokens));
```
elenadimitrova: I don't know what `denomerator` is in English did you mean `denominator`?

PR #200, opened 2018-05-01
Fix issue with moderately sized reputation logs

area:
Previously, we called delete on the array of reputation updates in ColonyNetwork once a mining cycle completed, but I was mistaken in thinking that that took constant gas - it actually explicitly assigns to every element in the whole array, which obviously runs in to issues with gas. Similarly, doing `array.length=0` also appears to explicitly assign to every element in the array. The only way to delete data with constant gas costs is `selfdestruct`. Fortunately, we already had a related contract (ReputationMiningCycle) that we `selfdestruct`. This commit moves the storage of the reputation log to the ReputationMiningCycle contract, allowing it to be cleaned up nicely.

I think there is a gas penalty for adding new entries to the logs for this implementation compared to the previous one, due to the increase in calls between contracts, but it's better than having reputation mining lock up once more than 20 reputation updates happen in an hour.

There has been a change to the functions that inspect the reputation log; `colonyNetwork` now has functions that tell you the address of the current `reputationMiningCycle` and the address of the next one. The 'current' `reputationMiningCycle` contract is the one that mining clients are currently 'arguing' over, and is having no more entries appended to the log. The 'next' `reputationMiningCycle` is the one which will have new entries added to the log. This is a reversal of what 'current' meant when we talked about 'current' logs, but makes more sense if we are talking about `reputationMiningCycle` contracts rather than logs.


contracts/ColonyNetworkStaking.sol
```diff
@@ -69,24 +69,29 @@ contract ColonyNetworkStaking is ColonyNetworkStorage, DSMath {
...
     startNextCycle();
+    rewardStakers(stakers);
   }
 
   function startNextCycle() public {
-    require(reputationMiningCycle == 0x0);
-    reputationMiningCycle = new ReputationMiningCycle();
+    require(currentReputationMiningCycle == 0x0);
+    currentReputationMiningCycle = nextReputationMiningCycle;
+    if (currentReputationMiningCycle == 0x0) {
```
elenadimitrova: But we `require` that to be `0x0` two lines up?

contracts/ColonyNetwork.sol
```diff
@@ -24,6 +24,7 @@ import "./IColony.sol";
 import "./EtherRouter.sol";
 import "./Token.sol";
 import "./ColonyNetworkStorage.sol";
+import "./ReputationMiningCycle.sol"; //TODO: Replace with interface
```
elenadimitrova: The interface would be nice to carry over all public function natspec comments too. That will nicely shrink the main contract length and improve readability. Is it too complex to do it here?

contracts/ColonyNetworkStorage.sol
```diff
@@ -59,20 +59,10 @@ contract ColonyNetworkStorage is DSAuth {
...
-  }
-
-  mapping (uint => ReputationLogEntry[]) reputationUpdateLogs;
-  // Tracks whether updateLog 0 or 1 is the active log
-  uint256 activeReputationUpdateLog;
   // Address of the currently active reputation mining cycle contract
-  address reputationMiningCycle;
+  address currentReputationMiningCycle;
+  // Address of the next active reputation mining cycle contract, which is where new reputation updates are put.
+  address nextReputationMiningCycle;
```
elenadimitrova: These names confused me. The "current" and "next" definitions differ depending on who you are, for a rep miner, the above makes sense, but for the network, the current log is the one which it appends updates to. I am struggling to come up with some alternatives but maybe `minerReputationMiningCycle` and `networkReputationMiningCycle`?

area: I went with the perspective of the reputation miner, as the contract in question is `ReputationMiningCycle`, which is definitely more concerned with the state of reputation mining than which logs are currently being appended to.

I agree this isn't perfect (which is why I called it out in the PR description), but I wasn't able to come up with anything better. I prefer this nomenclature to `minerReputationMiningCycle` and `networkReputationMiningCycle`, though am open to further suggestions.

contracts/ColonyNetworkStaking.sol
area: And we set it to `nextReputationMiningCycle` on the line in between, which will be `0x0` the very first time this function is run to kick off reputation mining, and never again.

contracts/ColonyNetwork.sol
```diff
@@ -245,30 +237,9 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
-  function getReputationUpdateLogLength(bool active) public view returns (uint) {
-    uint logIdx = activeReputationUpdateLog;
-    if (!active) {
-      logIdx = (logIdx + 1 ) % 2;
-    }
-    return reputationUpdateLogs[logIdx].length;
+    uint nParents = skills[_skillId].nParents;
+    // TODO: Is it cheaper to check if _amount is <0, and if not, just set nChildren to 0, because children won't be updated for such an update?
+    uint nChildren = skills[_skillId].nChildren;
+    ReputationMiningCycle(nextReputationMiningCycle).appendReputationUpdateLog(_user, _amount, _skillId, msg.sender, nParents, nChildren);
```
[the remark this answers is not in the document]
area: From `Colony`? There would still be a call to `ColonyNetwork` to make sure that the calling contract was a colony every time, so I don't think you gain much. If you want to save gas here, I think the approach to take is to pass arrays to `appendReputationUpdateLog` that are then iterated (but good iteration!) over in `ReputationMiningCycle`, so you only have one call for a block of updates. You would go from eight external calls to two, so that should save at least 120000 gas, right?

contracts/IColonyNetwork.sol
```diff
@@ -128,27 +128,20 @@ contract IColonyNetwork {
...
-  /// @return Reputation update log array length
-  function getReputationUpdateLogLength(bool activeLog) public view returns (uint256);
+  /// @notice Get the address of the reputation mining cycle that is current being processed by reputation miners.
+  /// @return address of ReputationMiningCycle currently being processed.
+  function getReputationMiningCycle() public view returns (address);
+
+  /// @notice Get the address of the reputation mining cycle that will be the next processed by reputation miners, and
+  /// is currently having log entries added to it.
+  /// @return address of ReputationMiningCycle next to be processed.
+  function getNextReputationMiningCycle() public view returns (address);
```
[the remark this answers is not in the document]
area: Yeah, though we should probably settle on what to name the two `ReputationMiningCycles` to determine what this function / parameter are best called 😄

contracts/ColonyNetwork.sol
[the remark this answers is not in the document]
elenadimitrova: Do we have a limit on the number of domains and skills on a task?

contracts/ColonyNetworkStorage.sol
elenadimitrova: Some more ideas then, in order of preference:
`activeReputationMiningCycle` / `inactiveReputationMiningCycle` or `passiveReputationMiningCycle`
`reputationMiningCycle` / `nextReputationMiningCycle`
`runningReputationMiningCycle` / `dormantReputationMiningCycle`

Also would prefer a single getter like before.

helpers/test-data-generator.js
```diff
@@ -15,17 +15,25 @@ import {
...
 } from "../helpers/constants";
 import { currentBlockTime, createSignatures } from "../helpers/test-helper";
 
 const IColony = artifacts.require("IColony");
 const Token = artifacts.require("Token");
 
 export async function setupAssignedTask({ colonyNetwork, colony, dueDate, domain = 1, skill = 0, evaluator = EVALUATOR, worker = WORKER }) {
-  await colony.makeTask(SPECIFICATION_HASH, domain);
+  // By using a random specification hash rather than SPECIFICATION_HASH, we can create tasks in parallel and pick up
+  // the correct taskID afterwards.
```
elenadimitrova: I don't get what or why we are doing this

test/colony-network-staking.js
```diff
@@ -258,6 +258,28 @@ contract("ColonyNetworkStaking", accounts => {
       const rootHashNNodes = await colonyNetwork.getReputationRootHashNNodes.call();
       assert(rootHashNNodes.equals(10));
     });
+
+    it("should not allow someone who is not ColonyNetwork to appendReputationUpdateLog", async () => {
```
elenadimitrova: Maybe tests should move to `reputation-update.js` for functionality moved from `ColonyNetwork` to `ReputationMiningCycle`

contracts/ColonyNetwork.sol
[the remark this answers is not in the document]
area: Currently only one of each. I don't see multiple domains happening, but I can see a desire for multiple skills in the future. But that introduces problems regardless of which implementation discussed here, as far as I can see?

helpers/test-data-generator.js
area: I want to make the client tests quicker to run, which currently have a lot of setup with multiple tasks being created to give users tokens and staking (anywhere between 3 and 11 times, I believe). If we can do that setup in parallel, I'm hoping it'll be quicker. This prepares for that; #203 is the remaining piece of the puzzle.

contracts/ColonyNetworkStorage.sol
area: I am happy with `active`/`inactive`, I think, so I'll switch to that.

contracts/ColonyNetwork.sol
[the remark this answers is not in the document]
elenadimitrova: I can't find a call for a limit on number of skills in the paper, maybe add it to the list of missing bits in that WP repo issue. And yes, it's a problem either way but it doesn't have to be solved here for the MVP.

test/colony-network-staking.js
area: I've ended up renaming `colony-network-staking.js` to `colony-network-mining.js`, which I think more accurately captures what the tests in that file have become, as well as the fact that they now span `ReputationMiningCycle` and `ColonyNetwork`.

helpers/test-data-generator.js
```diff
@@ -15,17 +15,25 @@ import {
...
+  const specificationHash = web3Utils.randomHex(32);
   let taskId = await colony.getTaskCount.call();
+  taskId = taskId.minus(1);
```
elenadimitrova: Remember that somehow magically our task ids have become 1-based ..

contracts/IReputationMiningCycle.sol
```diff
@@ -0,0 +1,144 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity ^0.4.23;
+pragma experimental "v0.5.0";
+
+
+contract IReputationMiningCycle {
```
elenadimitrova: Some gas golfing can be done with this interface to replace the import in `ColonyNetworkStaking` and also help decouple the mining client.

helpers/test-data-generator.js
elenadimitrova: Or is this because you are adding `1` to the `id` at the start of the loop after?

area: No, this was a mistake. Doesn't matter now though, because of the other, better way of using events you suggested 😄

contracts/IReputationMiningCycle.sol
area: Done for the mining client, but not for `ColonyNetworkStaking`.

PR #205, opened 2018-05-11
Sign messages in a way that real-world wallets expect

chmanie:
While testing the MultiSig implementation @JamesLefrere and I encountered that the ecsign method used in createSignatures (in test-helpers.js) works differently from the ones used by web3 or ethers (or any wallet implementation, really). The latter prepend a prefix `("\x19Ethereum Signed Message:\n" + len(message))` to the message hash which is to sign, while the former does not. The prefix can also be found in the ERC191 specification: https://github.com/ethereum/EIPs/issues/191.

In order to add support across all wallets (including MetaMask and Hardware wallets) we should use this method of signing as it's the only one those support.

The proposed change fixes the MultiSig (executeTaskChange) for us when using it in the browser (with an ethers wallet), although it breaks a whole lot of other tests. Maybe we're doing something stupid on the Solidity side of things so we'd like for someone more proficient in this to help us out (@elenadimitrova, @area?).

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -139,23 +139,30 @@ contract ColonyTask is ColonyStorage, DSMath {
     return taskChangeNonce;
   }
 
-  function getSignedMessageHash(uint256 _value, bytes _data) private returns (bytes32 txHash) {
-    return keccak256(
-      "\x19Ethereum Signed Message:\n32",
-      keccak256(
-        address(this),
-        address(this),
-        _value,
-        _data,
-        taskChangeNonce
-      )
+
+  function getSignedMessageHash(uint256 _value, bytes _data, uint8 _mode) private returns (bytes32 txHash) {
+    bytes32 hash = keccak256(
+      address(this),
+      address(this),
+      _value,
+      _data,
+      taskChangeNonce
     );
+    if (_mode==0) {
+      // 'Normal' mode - geth, etc.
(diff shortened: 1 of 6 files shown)


area:
So, EIP712 - once it gets support - is definitely the way to go. Much better from a UX perspective (it's less opaque what one is signing), and hopefully everyone will implement in the same way. Once it's accepted and begins to gain adoption, we can deprecate the implementation here, and force everyone to move over to it.

Until then, I have added the requested prefix to the messages being signed that the contract expects to receive. In order to make things easier, I have hashed the data being passed, then prepending `\x19Ethereum Signed Message:\n32"` to the hash, and then signing that message. This makes things even more opaque than they were (if a user inspects the message they are signing, it is 32 random bytes, and if they wish to independently verify they are signing what they intend to, they have to do this hashing themselves). However, this makes it *a lot* easier to verify the signature on-chain, because the length of the message having the prefix added is always 32 bytes. If this was not done, 

I have also added support for Trezor-style signing, where the message length is not an ASCII-string, but is encoded as a varint. This requires an extra parameter to be passed to `executeTaskChange` for each signature (the `_mode` parameter) but the correct value for this parameter for each signature can, in principle, be obtained by the client submitting the transaction (by running ECRecover against each possibility, and seeing if the recovered address is the manager or required counterparty).

Yes, this is all a much worse solution than we would have come up with in a vacuum, but I'm pretty sure it's the best we can do for now and maintain compatibility with the wallets we want to support.

PR #207, opened 2018-05-17
Remove Colony name from storage and rename "Common Colony" to "Meta Colony"

elenadimitrova:
Here we remove 1 of the 3 mappings of colony addresses in storage which held a colony bytes32 "key" mapping. This has been depreciated as a legacy feature in favour of the colony index which can be used instead to uniquely identify a colony, similarly to the task index.

Additionally the "Common Colony" has been renamed to "Meta Colony" to make clearer its purpose.

Also I've slipped in an update to the lockfile which I forgot in https://github.com/JoinColony/colonyNetwork/pull/204/

### DIFF
--- a/contracts/ColonyNetwork.sol
+++ b/contracts/ColonyNetwork.sol
@@ -105,8 +105,9 @@ contract ColonyNetwork is ColonyNetworkStorage {
     return reputationRootHashNNodes;
   }
 
-  // TODO secure
-  function createMetaColony(address _tokenAddress) public {
+  function createMetaColony(address _tokenAddress) public 
+  auth
+  {
     // Add the root global skill
     skillCount += 1;
     Skill memory rootGlobalSkill;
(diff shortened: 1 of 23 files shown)


contracts/ColonyNetwork.sol
```diff
@@ -107,9 +105,21 @@ contract ColonyNetwork is ColonyNetworkStorage {
     return reputationRootHashNNodes;
   }
 
-  function createColony(bytes32 _name, address _tokenAddress) public
-  colonyKeyUnique(_name)
+  function createMetaColony(address _tokenAddress) public 
+  auth
   {
+    // Add the root global skill
```
area: Let's put a `require(metaColony==0)` in here. Accidentally running this function twice would be... bad.

contracts/IColonyNetwork.sol
```diff
@@ -82,38 +82,42 @@ contract IColonyNetwork {
...
+  /// @param _id Id of the colony to get
   /// @return The colony address, if no colony was found, returns 0x0
-  function getColonyAt(uint256 _idx) public view returns (address);
+  function getColony(uint256 _id) public view returns (address);
 
   /// @notice Returns the latest Colony contract version. This is the version used to create all new colonies
   /// @return The current / latest Colony contract version
   function getCurrentColonyVersion() public view returns (uint256);
 
   /// @notice Upgrades a colony with key identifier: `_name` to a new Colony contract version `_newVersion`
```
area: We're not using `_name` any more.

PR #209, opened 2018-05-21
WIP: Updated tests to handle changes in EIP150

### LINKED ISSUE #194: Properly account for EIP150 in tests
Currently, in test-helper.js in `checkError` there is a hardcoded gas value. This value should not be hardcoded, but properly account for the expected behaviour introduced in [EIP150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md), where when a `CALL` is made only 63/64 on the remaining gas is passed on to to the called contract. For us, this primarily applies to the `DELEGATECALL` in EtherRouter.

To fix this issue:

* Remove the hardcoded value in `test-helper.js`, and replace it with a proper assertion to check the expected amount of gas is used - something like `receipt.gasUsed/63 >= transaction.gas/64`.
* Remove the exception for the `coverage` tests in the same function.
* Update EtherRouter, as there is no longer a need for `sub(gas, 700)` 
* Add `EIP150Transition` to `parity-genesis.template.json`
* Fix any unforseen issues arising from the above changes.

filiplazovic:
Closes #194


helpers/test-helper.js
```diff
@@ -90,14 +90,10 @@ async function checkError(promise, errMsg, isAssert) {
...
     const transaction = await web3GetTransaction(tx);
-    if (network !== "coverage") {
-      // When a transaction `throws`, all the gas sent is spent. So let's check that we spent all the gas that we sent.
-      // TODO: Properly account for EIP150 and passing only 63/64 gas on CALL.
-      // See https://github.com/JoinColony/colonyNetwork/issues/194 for details.
-      assert.closeTo(transaction.gas, receipt.gasUsed, 95000, "didnt fail - didn't throw and use all gas");
-    }
+    // When a transaction `throws`, all but one 64th of maximum amount is spent. So let's check that
+
+    assert.closeTo(receipt.gasUsed, transaction.gas - transaction.gas / 64, 110000, "didnt fail - didn't throw and use all gas");
```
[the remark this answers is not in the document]
filiplazovic: When i was testing a failing `assert`, used gas was not **exactly** `gas - gas / 64`. Should it be?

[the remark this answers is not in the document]
elenadimitrova: 63/64 of the gas remaining at the time the CALL is made so there will be some difference. Do you know what it is `110000` seems a lot?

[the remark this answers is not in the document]
filiplazovic: So there will be some gas spend in EtherRouter, then the 63/64 of remaining will be passed to `CALL`?
It is, but thats the result i was getting, i will double check.
So should i subtract whats used in EtherRouter, and then check for the exact number?

helpers/test-helper.js
```diff
@@ -90,14 +90,14 @@ async function checkError(promise, errMsg, isAssert) {
...
-      // See https://github.com/JoinColony/colonyNetwork/issues/194 for details.
-      assert.closeTo(transaction.gas, receipt.gasUsed, 95000, "didnt fail - didn't throw and use all gas");
-    }
+    // As of EIP150, when a CALL is made only 63/64 on the remaining gas is passed on to to the called contract.
+    // If an assert fails, that gas is expected to be spent.
+
+    const totalPassedToCall = (transaction.gas - receipt.gasUsed) * 64;
+    const etherRouterGas = transaction.gas - totalPassedToCall;
+    const remaining = transaction.gas - etherRouterGas;
+    assert.equal(receipt.gasUsed, transaction.gas - remaining / 64, "didnt fail - didn't throw and use all gas");
```
filiplazovic: Attempt one, will only work with 1 child `CALL`. More testing needed

helpers/test-helper.js
```diff
@@ -90,14 +90,14 @@ async function checkError(promise, errMsg, isAssert) {
...
-      // See https://github.com/JoinColony/colonyNetwork/issues/194 for details.
-      assert.closeTo(transaction.gas, receipt.gasUsed, 95000, "didnt fail - didn't throw and use all gas");
-    }
+    // As of EIP150, when a CALL is made only 63/64 on the remaining gas is passed on to to the called contract.
+    // If an assert fails, that gas is expected to be spent.
+
+    const remainingBeforeCall = (transaction.gas - receipt.gasUsed) * 64;
+    const etherRouterGas = transaction.gas - remainingBeforeCall;
+    const remaining = transaction.gas - etherRouterGas;
+    assert.equal(receipt.gasUsed, transaction.gas - remaining / 64, "didnt fail - didn't throw and use all gas");
```
filiplazovic: Attempt one, will only work with 1 child `CALL`. More testing needed

helpers/test-helper.js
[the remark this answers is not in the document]
elenadimitrova: Yes and yes :)

helpers/test-helper.js
```diff
@@ -90,14 +90,14 @@ async function checkError(promise, errMsg, isAssert) {
...
-    if (network !== "coverage") {
-      // When a transaction `throws`, all the gas sent is spent. So let's check that we spent all the gas that we sent.
-      // TODO: Properly account for EIP150 and passing only 63/64 gas on CALL.
-      // See https://github.com/JoinColony/colonyNetwork/issues/194 for details.
-      assert.closeTo(transaction.gas, receipt.gasUsed, 95000, "didnt fail - didn't throw and use all gas");
-    }
+    // As of EIP150, when a CALL is made only 63/64 on the remaining gas is passed on to to the called contract.
+    // If an assert fails, that gas is expected to be spent.
+
+    const remainingBeforeCall = (transaction.gas - receipt.gasUsed) * 64;
```
elenadimitrova: Sorry to keep picking at this but here I have no idea what's going on each line of gas calculation. Please make it obvious with the variables naming. For example, this is called `remainingBeforeCall` while the calculation is actually the total gas sent, minus the used so far multiplied by 64?!

helpers/test-helper.js
```diff
@@ -90,14 +90,14 @@ async function checkError(promise, errMsg, isAssert) {
...
+    const etherRouterGas = transaction.gas - remainingBeforeCall;
```
elenadimitrova: Is `etherRouterGas` this the same as `receipt.gasUsed`? And how is this `etherRouterGas`?

helpers/test-helper.js
filiplazovic: `receipt.gasUsed` will be gas used in `EtherRouter` + 63/64 of the remaining, so `transaction.gas - receipt.gasUsed` will be the remaining 1/64, then i multiply it by 64. So the result will be the remaining after `EtherRouter` has done its job.

This is not final solution, just me debugging and trying to extract the values i need. I accidentally pushed this. If i am wasting your time here, then i can call you when i finish.

elenadimitrova: I had a chat with @area about this and we might just remove the assert checks after all. (sorry) We will likely use asserts in code but should not be able to get to them while testing.

PR #214, opened 2018-05-28
Token Locking

### LINKED ISSUE #217: Token Locking
Create token locking mechanism that supports _per-user_ locking functionality. It should be a level higher than the colony, meaning that multiple colonies should be able to lock the same token, as it is possible that multiple colonies have the same native token, which was made possible by BYOT functionality. It should also be universal, meaning that in can be used for variety of purposes like reward payouts, token-weighed voting etc.

filiplazovic:
Closes #217 

Forth version. Same as the first one, except its modified so it can be used for by multiple colonies.
Works like this:
We have total and users lock count. Counters are separate for every token. If total count is greater than the users, then the users tokens are locked, otherwise, if they are equal, users token are unlocked. When colony increments total lock count, everyones tokens are locked, even if they don't have any deposited tokens. This is the simplest and cheapest way to avoid double spending by user.
If by any chance happens that users lock count is greater than total lock count, then this implementation is not used the way its meant to.

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -151,11 +151,10 @@ contract ColonyFunding is ColonyStorage, DSMath {
     return globalRewardPayoutCount;
   }
 
-  function getUserRewardPayoutCount(address _user) public view returns (uint256) {
-    return userRewardPayoutCount[_user];
-  }
-
   function startNextRewardPayout(address _token) public auth {
+    ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+    tokenLocking.lockToken(address(token));
+
     require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
 
     uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
@@ -163,8 +162,6 @@ contract ColonyFunding is ColonyStorage, DSMath {
 
     activeRewardPayouts[_token] = true;
     globalRewardPayoutCount += 1;
-
-    //TODO: Lock everyones tokens
 
     rewardPayoutCycles[globalRewardPayoutCount] = RewardPayoutCycle(
       IColonyNetwork(colonyNetworkAddress).getReputationRootHash(),
@@ -178,13 +175,20 @@ contract ColonyFunding is ColonyStorage, DSMath {
   }
 
   function claimRewardPayout(uint256 _payoutId, uint256[7] _squareRoots, uint256 _userReputation, uint256 _totalReputation) public {
+    require(!userRewardPayouts[msg.sender][_payoutId], "colony-reward-payout-already-claimed-or-waived");
+
     RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
+    // Checking if payout is active
     require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
-    require(_payoutId - userRewardPayoutCount[msg.sender] == 1, "colony-reward-payout-bad-id");
 
-    //TODO: Prove that userReputation and totalReputation in reputationState are correct
+    //TODO: Prove that userReputation and totalReputation in reputationState (reputation root has at the start of the payout) are correct
 
-    uint256 userTokens = token.balanceOf(msg.sender);
+    ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+
+    bool locked = tokenLocking.usersTokensLocked(address(token), msg.sender);
+    require(locked, "colony-reward-payout-invalid-token-lock");
+
+    uint256 userTokens = tokenLocking.getUserDepositedBalance(address(token), msg.sender);
 
     require(_totalReputation > 0, "colony-reward-payout-invalid-total-reputation");
     require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
@@ -213,19 +217,19 @@ contract ColonyFunding is ColonyStorage, DSMath {
 
     pots[0].balance[payout.tokenAddress] = sub(pots[0].balance[payout.tokenAddress], reward);
 
-    userRewardPayoutCount[msg.sender] += 1;
-
-    // TODO: Unlock user tokens
+    userRewardPayouts[msg.sender][_payoutId] = true;
+    tokenLocking.unlockTokenForUser(address(token), msg.sender);
 
     ERC20Extended(payout.tokenAddress).transfer(msg.sender, reward);
   }
 
-  function waiveRewardPayouts(uint256 _numPayouts) public {
-    require(add(userRewardPayoutCount[msg.sender], _numPayouts) <= globalRewardPayoutCount, "colony-reward-payout-invalid-num-payouts");
+  function waiveRewardPayout(uint256 _payoutId) public {
+    require(_payoutId <= globalRewardPayoutCount, "colony-reward-payout-invalid-id");
+    require(!userRewardPayouts[msg.sender][_payoutId], "colony-reward-payout-already-claimed-or-waived");
+    userRewardPayouts[msg.sender][_payoutId] = true;
 
-    userRewardPayoutCount[msg.sender] += _numPayouts;
-
-    //TODO unlock user tokens
+    ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+    tokenLocking.unlockTokenForUser(address(token), msg.sender);
   }
 
   function finalizeRewardPayout(uint256 _payoutId) public {
--- a/contracts/ColonyNetworkStorage.sol
+++ b/contracts/ColonyNetworkStorage.sol
@@ -35,6 +35,8 @@ contract ColonyNetworkStorage is DSAuth {
   uint256 currentColonyVersion;
   // Address of the Meta Colony
   address metaColony;
+  // Address of token locking contract
+  address tokenLocking;
   // Maps index to colony address
   mapping (uint256 => address) colonies;
   mapping (address => bool) _isColony;
--- a/contracts/ColonyStorage.sol
+++ b/contracts/ColonyStorage.sol
@@ -21,6 +21,7 @@ pragma experimental "v0.5.0";
 import "../lib/dappsys/auth.sol";
 import "./ERC20Extended.sol";
 import "./IColonyNetwork.sol";
+import "./ITokenLocking.sol";
 
 
 contract ColonyStorage is DSAuth {
@@ -61,13 +62,13 @@ contract ColonyStorage is DSAuth {
   mapping (uint256 => RewardPayoutCycle) rewardPayoutCycles;
   // Active payouts for particular token address. Assures that one token is used for only one active payout
   mapping (address => bool) activeRewardPayouts;
-  // Incremented every time a reward payout is created. Used for comparing with `userRewardPayoutCount`
-  // to ensure that rewards are claimed in the right order.
+  // Incremented every time a reward payout is created. Used for indexing the reward payout
   // Can be used for iterating over all reward payouts (excluding index 0)
   uint256 globalRewardPayoutCount;
-  // Keeps track of how many payouts are claimed by the particular user.
-  // Can be incremented either by waiving or claiming the reward.
-  mapping (address => uint256) userRewardPayoutCount;
+
+  // Keeps track if user claimed or waived specific reward payout.
+  // Used for ensuring that user cannot claim or waive the reward twice
+  mapping (address => mapping(uint256 => bool)) userRewardPayouts;
 
   // This keeps track of how much of the colony's funds that it owns have been moved into pots other than pot 0,
   // which (by definition) have also had the reward amount siphoned off and put in to pot 0.
--- a/contracts/ITokenLocking.sol
+++ b/contracts/ITokenLocking.sol
@@ -1,0 +1,54 @@
+pragma solidity ^0.4.23;
+
+
+contract ITokenLocking {
+
+  /// @notice Set the ColonyNetwork contract address
+  /// @dev ColonyNetwork is used for checking if sender is a colony created on colony network
+  /// @param _colonyNetwork Address of the ColonyNetwork
+  function setColonyNetwork(address _colonyNetwork) public;
+
+  /// @notice Get ColonyNetwork address
+  /// @return ColonyNetwork address
+  function getColonyNetwork() public view returns (address);
+
+  /// @notice Locks everyones' tokens on `_token` address
+  /// @param _token Address of the token we want to lock
+  function lockToken(address _token) public;
+
+  /// @notice Increments the lock counter for the `_user`. Can only be called by a colony
+  /// @param _token Address of the token we want to unlock
+  /// @param _user Address of the user
+  function unlockTokenForUser(address _token, address _user) public;
+
+  /// @notice Deposit `_amount` of colony tokens. Can only be called if user tokens are not locked
+  /// Before calling this function user has to allow that their tokens can be transferred by token locking contract
+  /// @param _amount Amount to deposit
+  function deposit(address _token, uint256 _amount) public;
+
+  /// @notice Withdraw `_amount` of deposited tokens. Can only be called if user tokens are not locked
+  /// @param _amount Amount to withdraw
+  function withdraw(address _token, uint256 _amount) public;
+
+  /// @notice Get deposited balance by `_user`
+  /// @param _user Address of the user
+  /// @return Users deposited amount
+  function getUserDepositedBalance(address _token, address _user) public view returns(uint256);
+
+  /// @notice Check if users tokens are locked
+  /// @param _token Address of the token
+  /// @param _user Address of the user
+  /// @return true if users tokens are locked, false otherwise
+  function usersTokensLocked(address _token, address _user) public view returns (bool);
+
+  /// @notice Get global token lock count
+  /// @param _token Address of the token
+  /// @return Global token lock count
+  function getTotalTokenLockCount(address _token) public view returns (uint256);
+
+  /// @notice Get user token lock count
+  /// @param _token Address of the token
+  /// @param _user Address of the user
+  /// @return User token lock count
+  function getUserTokenLockCount(address _token, address _user) public view returns (uint256);
+}
--- a/contracts/TokenLocking.sol
+++ b/contracts/TokenLocking.sol
@@ -1,0 +1,71 @@
+pragma solidity ^0.4.23;
+
+import "./ERC20Extended.sol";
+import "./IColonyNetwork.sol";
+import "./TokenLockingStorage.sol";
+import "../lib/dappsys/math.sol";
+
+contract TokenLocking is TokenLockingStorage, DSMath {
+  modifier tokenNotLocked(address _token) {
+    require(userTokenLocks[_token][msg.sender].count == totalTokenLockCount[_token]);
+    _;
+  }
+
+  modifier onlyColony() {
+    require(IColonyNetwork(colonyNetwork).isColony(msg.sender), "token-locking-sender-not-colony");
+    _;
+  }
+
+
+  function setColonyNetwork(address _colonyNetwork) public auth {
+    colonyNetwork = _colonyNetwork;
+  }
+
+  // Used for testing purposes
+  function getColonyNetwork() public view returns (address) {
+    return colonyNetwork;
+  }
+
+  function lockToken(address _token) public onlyColony {
+    totalTokenLockCount[_token] += 1;
+  }
+
+  function unlockTokenForUser(address _token, address _user) public onlyColony {
+    // This should never throw if you are using it right, so it can't be tested
+    assert(userTokenLocks[_token][_user].count < totalTokenLockCount[_token]);
+    userTokenLocks[_token][_user].count += 1;
+  }
+
+  function deposit(address _token, uint256 _amount) public
+  tokenNotLocked(_token)
+  {
+    // If we transfer before we increment `depositedBalances`, user won't be able to take advantage of re-entrance
+    require(ERC20Extended(_token).transferFrom(msg.sender, address(this), _amount));
+
+    userTokenLocks[_token][msg.sender].amount = add(userTokenLocks[_token][msg.sender].amount, _amount);
+  }
+
+  function withdraw(address _token, uint256 _amount) public
+  tokenNotLocked(_token)
+  {
+    userTokenLocks[_token][msg.sender].amount = sub(userTokenLocks[_token][msg.sender].amount, _amount);
+
+    require(ERC20Extended(_token).transfer(msg.sender, _amount));
+  }
+
+  function getUserDepositedBalance(address _token, address _user) public view returns (uint256) {
+    return userTokenLocks[_token][_user].amount;
+  }
+
+  function usersTokensLocked(address _token, address _user) public view returns (bool) {
+    return userTokenLocks[_token][_user].count < totalTokenLockCount[_token];
+  }
+
+  function getTotalTokenLockCount(address _token) public view returns (uint256) {
+    return totalTokenLockCount[_token];
+  }
+
+  function getUserTokenLockCount(address _token, address _user) public view returns (uint256) {
+    return userTokenLocks[_token][_user].count;
+  }
+}
--- a/contracts/TokenLockingStorage.sol
+++ b/contracts/TokenLockingStorage.sol
@@ -1,0 +1,26 @@
+pragma solidity ^0.4.23;
+
+import "../lib/dappsys/auth.sol";
+
+
+contract TokenLockingStorage is DSAuth {
+  address resolver;
+
+  // Address of ColonyNetwork contract
+  address colonyNetwork;
+
+  struct TokenLock {
+    // Amount of deposited tokens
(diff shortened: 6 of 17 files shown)


gasCosts/gasCosts.js
```diff
@@ -248,9 +251,9 @@ contract("All", accounts => {
     });
 
     it("when working with reward payouts", async () => {
-      const totalReputation = toBN(350 * 1e18);
-      const workerReputation = toBN(200 * 1e18);
-      const managerReputation = toBN(100 * 1e18);
+      const totalAmount = toBN(350 * 1e18);
+      const workerAmount = toBN(200 * 1e18);
+      const managerAmount = toBN(100 * 1e18);
```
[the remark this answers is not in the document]
filiplazovic: They represent both tokens and reputation. See https://github.com/JoinColony/colonyNetwork/pull/214/files#diff-16a74db0d2b552059a5414340b64eafdR271

test/colony-funding.js
```diff
@@ -415,21 +415,32 @@ contract("Colony Funding", addresses => {
...
+      await token.approve(colony.address, userReputation2.toString(), {
+        from: userAddress2
+      });
+
+      await colony.deposit(userReputation1.toString(), {
+        from: userAddress1
+      });
+      await colony.deposit(userReputation2.toString(), {
+        from: userAddress2
+      });
```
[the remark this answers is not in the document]
filiplazovic: You are right, no need to deposit from `userAddress2` here.
Not sure what you meant with
> Same for `userAddress3`

contracts/ColonyStaking.sol
```diff
@@ -0,0 +1,43 @@
...
+
+  function deposit(uint256 _amount) public tokenNotLocked {
+    stakerBalances[msg.sender] = add(stakerBalances[msg.sender], _amount);
+    totalStaked = add(totalStaked, _amount);
+    token.transferFrom(msg.sender, address(this), _amount);
+  }
+
+  function withdraw(uint256 _amount) public tokenNotLocked {
+    uint256 balance = stakerBalances[msg.sender];
+    require(_amount <= balance);
```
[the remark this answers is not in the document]
filiplazovic: I just realized, maybe this line is not needed. Will that be covered [here](https://github.com/JoinColony/colonyNetwork/pull/214/files#diff-391043a2d9d7b367e893aec67e3b04c2R26)?

gasCosts/gasCosts.js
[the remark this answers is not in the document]
filiplazovic: Whatever we decide, we should apply that [here](https://github.com/filiplazovic/colonyNetwork/blob/feature/token-locking/test/colony-funding.js#L413) also

contracts/ColonyFunding.sol
```diff
@@ -174,14 +172,23 @@ contract ColonyFunding is ColonyStorage, DSMath {
     emit RewardPayoutCycleStarted(globalRewardPayoutCount);
   }
 
-  function claimRewardPayout(uint256 _payoutId, uint256[7] _squareRoots, uint256 _userReputation, uint256 _totalReputation) public {
-    RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
+  function claimNextRewardPayout(uint256[7] _squareRoots, uint256 _userReputation, uint256 _totalReputation) public {
+    uint256 payoutId = add(userRewardPayoutCount[msg.sender], 1);
```
elenadimitrova: We discussed this a few times already but just to reiterate. Inferring the `payoutId` this way has a problem for any user who "misses" a reward payout for any reason. Examples:
- `payoutId: 9` has finished and a user earns their first Colony tokens and deposit/lock them. Then `payoutId: 10` starts and user calls `claimNextRewardPayout` which infers the `payoutId` as `1` not `10`
- Token holder X successfully claims the first 10 reward payouts which takes `userRewardPayoutCount[X] = 10`. Then they unlock their tokens and sell them all making them ineligible for reward payouts. Another 3 reward payouts go by and  user X earns some colony tokens again which they deposit/lock. The current payout Id is `14` but the `userRewardPayoutCount` derived here `10+1=11`. The claim fails.

elenadimitrova: You also have to add tests covering the above scenarios.

test/colony.js
```diff
@@ -723,4 +733,77 @@ contract("Colony", addresses => {
       await checkErrorRevert(colony.claimPayout(taskId, MANAGER_ROLE, token.address, { from: OTHER }));
     });
   });
+
+  describe("when locking tokens", async () => {
```
elenadimitrova: Take this out into its own test file. Token locking logic lives outside the `Colony` and ever outside the `ColonyNetwork`, similarly to the `ColonyNetworkAuction` and `ColonyNetworkMining`

contracts/TokenLocking.sol
```diff
@@ -0,0 +1,53 @@
...
+  function getUserTokens(address _user, address _token) public view returns (uint256) {
+    return userDepositedTokens[_user][_token];
+  }
+
+  function getUserLockInfo(address _user, address _token) public view returns (uint256, bool) {
+    Lock memory lock = userTokenLockChanges[_user][_token];
+    return (lock.timestamp, lock.isLocked);
+  }
+
+  function lock(address _token) public {
```
elenadimitrova: Why is `lock` separate to `deposit`?

elenadimitrova: Delegating the responsibility to the user for locking their own tokens depends on them knowing when to lock exactly, i.e. they all have to lock before the `startNextRewardPayout` is called and also relies on *all* of them locking their tokens otherwise the `RewardPayoutCycle.totalTokens` (which is calculated as all tokens owned by anyone but the colony itself) will be wrong.

contracts/ColonyFunding.sol
filiplazovic: Ok, so you mean, if the user was not eligible to claim first 9 payouts, **we** should increment the counter for him (in claimNextRewardPayout?), or you are simply stating that we should add tests for these scenarios?

contracts/ColonyFunding.sol
```diff
@@ -155,14 +155,12 @@ contract ColonyFunding is ColonyStorage, DSMath {
   function startNextRewardPayout(address _token) public auth {
     require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
 
-    uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
```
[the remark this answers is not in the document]
filiplazovic: I have no idea how that happened. Will revert that

contracts/ColonyFunding.sol
elenadimitrova: You cannot _just_ add tests for this scenario as this won't work with the current implementation. I am pointing out a problem that should be fixed and tested for.

contracts/TokenLocking.sol
filiplazovic: I think that users would like to have control of their tokens. We could help the on the front-end so they don't mess up.

You think that `RewardPayoutCycle.totalTokens` should not be `sub(token.totalSupply(), token.balanceOf(address(this)))`, but it should be sum of all deposited tokens?

filiplazovic: Also, yeah, i could merge `lock` and `deposit`

area: > I think that users would like to have control of their tokens.

The way that I have envisioned this happening, this should always be true. Ultimately, tokens will be able to be locked for two reasons: 

* A payout is live, and they've not claimed or waived their payout yet
* A token-weighted vote they have taken part in has reached the 'reveal' stage.

In both cases, it is completely in the user's control to unlock their tokens. In the first case, they claim their payout or waive their claim to the payout. In the second case, they reveal any votes they committed.

filiplazovic: In this implementation, users tokens will not be unlocked just by calling `claimNextRewardPayout`, they need to call `unlock` (in TokenLocking contract) themselves.

There are two scenarios we are discussing here:
1) Using counter, where we lock the users tokens and we unlock them, e.g:
- User deposits his tokens
- We call `startNextRewardPayout` (we will lock everyones tokens inside this function)
- User calls `claimNextRewardPayout` (we will unlock users tokens in this function)

2) Using this implementation:
- User deposits token
- **User** locks his tokens by calling `lock` in TokenLocking contract
- We call `startNextRewardPayout` (we dont touch users tokens here)
- User calls `claimNextRewardPayout` (we check if he locked his tokens before reward payout start)
- **User** unlocks his tokens

area: So what are the benefits of doing it this way? To me, it looks like it requires extra transactions by the user. I agree that in general this shouldn't prevent us from doing something - as you have mentioned, the UI can guide the user. But if we can avoid them, it seems to me like we should?

filiplazovic: Thats a valid point.
There is maybe one exception where this one would be cheaper vs the version with counters:
- when users tokens are locked multiple times, either by reward payouts, voting etc. he would need to unlock them multiple times (increase the counter by claiming, waiving, whatever), wheres here he would finish all the stuff he needs to finish, and at the end unlock the tokens once. Im not saying if this valid enough argument to choose this one over the other, but maybe worth to consider.

contracts/Colony.sol
```diff
@@ -37,6 +37,12 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     token = ERC20Extended(_token);
   }
 
+  function setTokenLocking(address _tokenLocking) public
```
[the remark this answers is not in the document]
filiplazovic: In this case external call to `ColonyNetwork` inside `claimNextRewardPayout` adds ~7000 gas, while setting the `setTokenLocking` costs 47458 gas. It will be either us spending 47k or user spending 7k more

filiplazovic:
I will keep the commit history as it is for now, just in case, and will squash it after review

test/token-locking.js
```diff
@@ -0,0 +1,179 @@
...
+      await tokenLocking.deposit(token.address, usersTokens);
+      const { logs } = await colony.startNextRewardPayout(otherToken.address);
+      const payoutId = logs[0].args.id;
+      await colony.waiveRewardPayout(payoutId);
+      await tokenLocking.withdraw(token.address, usersTokens);
+      const userDepositedBalance = await tokenLocking.getUserLockedBalance.call(token.address, addresses[0]);
+      assert.equal(userDepositedBalance, 0);
+    });
+
+    it("should not be able to lock tokens twice", async () => {
```
elenadimitrova: You are testing a success here but the title claims you're testing for a failure?

test/token-locking.js
```diff
@@ -0,0 +1,179 @@
...
+
+      await tokenLocking.setColonyNetwork(colonyNetwork.address);
+      colonyNetworkAddress = await tokenLocking.getColonyNetwork.call();
+      assert.equal(colonyNetworkAddress, colonyNetwork.address);
+    });
+
+    it("should correctly deposit tokens", async () => {
+      await token.approve(tokenLocking.address, usersTokens);
+      await tokenLocking.deposit(token.address, usersTokens);
+      const userDepositedBalance = await tokenLocking.getUserLockedBalance.call(token.address, addresses[0]);
```
elenadimitrova: Can you change the account used in all tests here as the user whose tokens get locked to be another account than the coinbase. This is the default account which owns most of the contracts you test with so it's risky to use it as a representative account for testing permissions especially. Shouldn't be that difficult to switch to `addresses[1]` for example.

test/token-locking.js
```diff
@@ -0,0 +1,179 @@
...
+    it("should change timestamp when user deposit their tokens", async () => {
+      await token.approve(tokenLocking.address, usersTokens);
+      let info = await tokenLocking.getUserTokenLockInfo.call(token.address, addresses[0]);
+      const lastDepositTimestampsBeforeDeposit = info[1];
+      await forwardTime(1000);
+      await tokenLocking.deposit(token.address, usersTokens);
+      info = await tokenLocking.getUserTokenLockInfo.call(token.address, addresses[0]);
+      const lastDepositTimestampsAfterDeposit = info[1];
+
+      assert(lastDepositTimestampsAfterDeposit > lastDepositTimestampsBeforeDeposit, "timestamp of deposit is not updated correctly");
```
elenadimitrova: You should really be getting the blocktime and testing you are close to that. See `testHelper.currentBlockTime` and `assert.closeTo` in `mocha`.

contracts/ColonyFunding.sol
```diff
@@ -178,13 +175,23 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
+    //TODO: Prove that userReputation and totalReputation in reputationState (reputation root has at the start of the payout) are correct
 
-    uint256 userTokens = token.balanceOf(msg.sender);
+    ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+
+    uint256 userLockCount;
+    uint256 lastDepositTimestamp;
+    (userLockCount, lastDepositTimestamp) = tokenLocking.getUserTokenLockInfo(address(token), msg.sender);
+    uint256 totalLockCount = tokenLocking.getTotalTokenLockCount(address(token));
+    require(userLockCount < totalLockCount && lastDepositTimestamp < payout.blockTimestamp, "colony-reward-payout-invalid-token-lock");
```
elenadimitrova: You say you've introduced `lastDepositTimestamp` to avoid being able to double spend. Can you explain a scenario where that would be possible?

contracts/ColonyStorage.sol
```diff
@@ -61,13 +62,13 @@ contract ColonyStorage is DSAuth {
...
+  // Incremented every time a reward payout is created. Used for indexing the reward payout
   // Can be used for iterating over all reward payouts (excluding index 0)
   uint256 globalRewardPayoutCount;
-  // Keeps track of how many payouts are claimed by the particular user.
-  // Can be incremented either by waiving or claiming the reward.
-  mapping (address => uint256) userRewardPayoutCount;
+
+  // Keeps track if user claimed or waived specific reward payout.
+  // Used for ensuring that user cannot claim or waive the reward twice
+  mapping (address => mapping(uint256 => bool)) userRewardPayouts;
```
elenadimitrova: I don't yet understand this change but it maybe related to the double spending you were talking about in justifying `lastDepositTimestamp`

contracts/Colony.sol
```diff
@@ -83,8 +83,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   }
 
   function mintTokensForColonyNetwork(uint _wad) public {
-    require(msg.sender == colonyNetworkAddress, "colony-access-denied-only-network-allowed"); // Only the colony Network can call this function
     require(this == IColonyNetwork(colonyNetworkAddress).getMetaColony(), "colony-access-denied-only-meta-colony-allowed"); // Function only valid on the Meta Colony
+    require(msg.sender == colonyNetworkAddress, "colony-access-denied-only-network-allowed"); // Only the colony Network can call this function
```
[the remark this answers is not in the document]
filiplazovic: So i can reach both lines in tests, just to increase coverage

contracts/TokenLocking.sol
```diff
@@ -0,0 +1,76 @@
...
+    // This should never throw if you are using it right, so it can't be tested
+    assert(userTokenLockCount[_token][_user] < totalTokenLockCount[_token]);
+    userTokenLockCount[_token][_user] += 1;
+  }
+
+  function deposit(address _token, uint256 _amount) public
+  tokenNotLocked(_token)
+  {
+    // Update only if needed
+    if (userTokenLockCount[_token][msg.sender] < totalTokenLockCount[_token]) {
```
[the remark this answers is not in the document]
filiplazovic: Ok, thanks, i didn't know that :)

contracts/ITokenLocking.sol
```diff
@@ -0,0 +1,46 @@
...
+contract ITokenLocking {
+
+  /// @notice Sets an address of ColonyNetwork contract
+  /// Used for checking if sender is a colony created on colony network
+  /// @param _colonyNetwork Address of the ColonyNetwork
+  function setColonyNetwork(address _colonyNetwork) public;
+
+  /// @notice Get ColonyNetwork address
+  /// @return ColonyNetwork address
+  function getColonyNetwork() public view returns (address);
```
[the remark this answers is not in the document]
filiplazovic: No, we are returning value from storage

test/token-locking.js
filiplazovic: Yes, the title is wrong

test/colony-funding.js
```diff
@@ -484,51 +495,60 @@ contract("Colony Funding", addresses => {
...
         from: userAddress1
       });
 
-      await colony.claimRewardPayout(payoutId2, initialSquareRoots1, userReputation1.toString(), totalReputation.toString(), {
+      await colony.claimRewardPayout(payoutId2.toString(), initialSquareRoots, userAmount1.toString(), totalAmount.toString(), {
         from: userAddress1
       });
     });
 
+    // Temporary test to increase coverage
```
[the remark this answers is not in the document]
filiplazovic: I don't remember why i put that there. The test will be different once we implement proofs, but its not temporary

contracts/Colony.sol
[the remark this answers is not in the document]
elenadimitrova: You don't have to make sacrifices like that for coverage. In theory we aim for that to be 100% but in practice when there is a valid case to lower it we can do that.  It is a sacrifice in a way as the second `require` is less restrictive than the first and protects us against cases where a malicious attacker calls in.

contracts/ColonyFunding.sol
filiplazovic: Ok i will try. So we are talking about the implementation without `lastDepositTimestamp`, so in this line above we would just have `require(userLockCount < totalLockCount)`.

So, this will be the timeline of events:
- there is `address1` which have reputation of 10 and 10 tokens, `address2` which also has reputation of 10 and 10 tokens, and `address3` with reputation 10 and 0 tokens
- `address1` and `address2` deposits all of their tokens to token locking contract
- **first** reward payout starts, locking all deposited tokens
- `address1` claims their payout, withdraws their tokens
- `address1` sends all of their tokens to `address3`
- `address3` deposits their tokens (they can because their `lockedBalance` is `0`)
- **second** reward payout starts, locking the tokens of `address3`
- `address3` is now eligable to claim the **first** reward payout since they have 10 reputation, 10 tokens, and their tokens are locked i.e `userLockCount < totalLockCount` is true.
- this can be repeated until funds from reward pot are drained, so `address2` would be the victim here

elenadimitrova: Ok I understand. It's a valid scenario but it feels cleaner if we just not allow any deposits while a token is locked by simply removing that check in `tokenNotLocked` modifier?

contracts/ColonyStorage.sol
elenadimitrova: This will be simply achieved with setting the user lock count to the global one `userLockCount = totalLockCount` won't it?

contracts/ColonyFunding.sol
filiplazovic: You mean the check `if (lockedBalances[_token][msg.sender] > 0)` in `tokenNotLocked `?

If we remove it, we would also remove `userTokenLockCount[_token][msg.sender] = totalTokenLockCount[_token]` since it will be useless.

That further means that `lastDepositTimestamps` is also not needed, and we will end up with the very first implementation.

contracts/ColonyStorage.sol
filiplazovic: No, the payout counters are different thing and needs to be handled separately.

contracts/ColonyFunding.sol
elenadimitrova: Yes that check in `tokenNotLocked` is what I meant. 

The very first implementation did not cater for locking the same token in multiple colonies and also assumed the value of `rewardPayoutId`

filiplazovic: Yeah, sorry. What i meant is that we will end up with something similar to first implementation. So locking the same token for multiple colonies would remain, but so will the characteristic that we didn't like, which is: tokens will be locked for everybody, regardless if they deposited or not

contracts/ColonyStorage.sol
elenadimitrova: ok yes you're right

contracts/ColonyFunding.sol
elenadimitrova: So the main disadvantage is that if someone's balance is `0` they are locked to deposit? Anything else we are sacrificing in this approach?

filiplazovic: For now, no.

Im just curious how will this approach behave in real world, when bunch of colonies start locking the same token.

I must say that my favourite one was the second one, where the user was responsible for locking, but the counters are also nice.

So, i would like to hear final decision from you and @area also?

elenadimitrova: About this:
>Or when some random colony locks the token that is used in several colonies

The colony only ever locks its _own_ token, even when payouts are triggered for other tokens it works with. For colonies that decide to reuse tokens as their own native currency this may need to be aided in the UI with sufficient gaps in between reward payouts

elenadimitrova: I liked the block timestamp logic in the second implementation too, but I think delegating the responsibility for locking to the user won't work in real life. What we have here, pending the fixes to last review is probably best.

contracts/ColonyNetworkStorage.sol
```diff
@@ -35,6 +35,8 @@ contract ColonyNetworkStorage is DSAuth {
   uint256 currentColonyVersion;
   // Address of the Meta Colony
   address metaColony;
+  // Address of token locking contract
+  address tokenLocking;
```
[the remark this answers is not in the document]
elenadimitrova: Please ignore this comment as the outcome of a meeting @area had yesterday is we won't yet support upgrades.

contracts/TokenLocking.sol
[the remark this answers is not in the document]
elenadimitrova: Where has this logic gone?

contracts/TokenLockingStorage.sol
```diff
@@ -0,0 +1,22 @@
...
+  address resolver;
+
+  // Address of ColonyNetwork contract
+  address colonyNetwork;
+
+  // Maps token to user to amount locked
+  mapping (address => mapping (address => uint256)) depositedBalances;
+
+  // Maps token to user to number of locks
+  mapping (address => mapping (address => uint256)) userTokenLockCount;
```
elenadimitrova: Sorry just noticed that if we merge these two mappings it'll be much cheaper gas-wise. So have a
```
struct TokenLock {
    uint256 amount;
    uint256 counter;
  }
mapping (address => mapping (address => TokenLock)) userTokenLocks;
```

contracts/TokenLocking.sol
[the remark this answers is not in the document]
filiplazovic: `userTokenLockCount < totalTokenLockCount` will never be `true`, since in `tokenNotLocked` modifier we require that `userTokenLockCount == totalTokenLockCount`

contracts/TokenLockingStorage.sol
filiplazovic: Awsome, thanks 👍

test/token-locking.js
```diff
@@ -0,0 +1,265 @@
...
+      });
+      await tokenLocking.deposit(token.address, usersTokens, {
+        from: userAddress
+      });
+      const userDepositedBalance = await tokenLocking.getUserDepositedBalance.call(token.address, userAddress);
+      assert.equal(userDepositedBalance.toNumber(), usersTokens);
+    });
+
+    it("should not be able todo deposit tokens if they are not approved", async () => {
+      await checkErrorRevert(
```
elenadimitrova: Do you have some editor settings that keeps formatting on multiple lines like this? Our current max len I think is 150 and I don't think you're reaching that.

test/token-locking.js
```diff
@@ -0,0 +1,265 @@
...
+        from: userAddress
+      });
+      await tokenLocking.deposit(token.address, usersTokens, {
+        from: userAddress
+      });
+      const userDepositedBalance = await tokenLocking.getUserDepositedBalance.call(token.address, userAddress);
+      assert.equal(userDepositedBalance.toNumber(), usersTokens);
+    });
+
+    it("should not be able todo deposit tokens if they are not approved", async () => {
```
elenadimitrova: todo -> to

contracts/ColonyFunding.sol
```diff
@@ -213,19 +217,19 @@ contract ColonyFunding is ColonyStorage, DSMath {
 
     pots[0].balance[payout.tokenAddress] = sub(pots[0].balance[payout.tokenAddress], reward);
 
-    userRewardPayoutCount[msg.sender] += 1;
-
-    // TODO: Unlock user tokens
+    userRewardPayouts[msg.sender][_payoutId] = true;
+    tokenLocking.unlockTokenForUser(address(token), msg.sender);
```
elenadimitrova: So this `unlockTokenForUser` call checks if the tokens are locked and throws otherwise. I don't think the `usersTokensLocked` call is needed therefore.

test/colony-funding.js
```diff
@@ -682,91 +704,202 @@ contract("Colony Funding", addresses => {
       await colony.startNextRewardPayout(otherToken.address);
     });
 
+    it("should not be able to finalize the payout if payout is not active", async () => {
```
elenadimitrova: This title is a little confusing as the finalization happens when the payout is not active, I think it should read `should not be able to finalize reward payout twice`

PR #220, opened 2018-05-31
Add Reputation Mining doc.

### LINKED ISSUE #219: Add Reputation Mining overview to docs
## Summary
As a developer (or curious individual) I want to read a single page that gives me an overview of how the reputation mining system works with the colonyNetwork contracts, and provides me with some next steps for experimenting with it.

## Context
For the hackathon, we have expected most projects to focus within the realm of colony creation, tasks, and perhaps skills. But the Reputation system is an essential part of the picture, and it should also have some materials for people to experiment with.

gichiba:
Closes #219

### DIFF
--- a/docs/_Docs_ReputationMining.md
+++ b/docs/_Docs_ReputationMining.md
@@ -29,7 +29,7 @@ Reputation within a specific colony can only be earned and lost by completing ta
 
 
 ### Reputation Root Hashes
-Using a [Patricia Tree](https://github.com/ethereum/wiki/wiki/Patricia-Tree), all updates to the global state of reputation are maintained on-chain with a single "Reputation Root Hash.
+Using a [Patricia Tree](https://github.com/ethereum/wiki/wiki/Patricia-Tree), all updates to the global state of reputation are maintained on-chain with a single "Reputation Root Hash".
 
 The Reputation Root Hash is effectively a fingerprint for the network state (of reputation). It must be calculated off-chain, but it represents the collection of all reputation-updating events that have occured on the network, and even a single discrepancy in the complete history of reputational updates will result in a different Root Hash.  
 
--- a/docs/_Docs_ReputationMining.md
+++ b/docs/_Docs_ReputationMining.md
@@ -9,7 +9,7 @@ This is an overview of the process by which Reputation scores in Colony are main
 For a complete description of the Reputation Mining process, see the [whitepaper](https://colony.io/whitepaper.pdf)
 
 # Rationale for off-chain Reputation
-Events in Colony that affect reputation are numerous and expected to be frequent. The calculations for global reputation updates would exceed Ethereum's gas limit, and cannot be performed on-chain. However, Reputation doesn't *need* to be on-chain, because reputation is entirely deterministic and calculated from on-chain events.
+Events in Colony that affect reputation are numerous and expected to be frequent. The calculations for global reputation updates would be unreasonable to perform on-chain. However, Reputation doesn't *need* to be on-chain, because reputation is entirely deterministic and calculated from on-chain events.
 
 So, rather than calculating reputation scores on-chain, users submit to the colony contract a transaction  which contains their score, together with a proof that the score is consistent with the global state of reputation.
 
--- a/docs/_Docs_ReputationMining.md
+++ b/docs/_Docs_ReputationMining.md
@@ -4,4 +4,55 @@ section: Docs
 order: 9
 ---
 
-Pending finalization of the reputation system implementation (see [GitHub](https://github.com/JoinColony/colonyNetwork) for more information).
+This is an overview of the process by which Reputation scores in Colony are maintained.
+
+For a complete description of the Reputation Mining process, see the [whitepaper](https://colony.io/whitepaper.pdf)
+
+# Rationale for off-chain Reputation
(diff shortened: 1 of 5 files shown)


gichiba:
@area If you could make sure I haven't said anything terribly wrong in this Rep Mining TL;DR, that'd be great... 

Combined with some minor other fixes to the docs; I was hoping we could merge this in today.

docs/_Docs_ReputationMining.md
```diff
@@ -4,4 +4,55 @@ section: Docs
...
+
+So, rather than calculating reputation scores on-chain, users submit to the colony contract a transaction  which contains their score, together with a proof that the score is consistent with the global state of reputation.
+
+## Reputation Mining Overview
+The process of maintaining consensus over the global state of reputation is called *Reputation Mining*.
+
+Reputation Mining resembles a proof-of-stake blockchain consensus protocol. Miners must put up a stake of the network token, CLNY, to participate in the reputation mining process, whereby the global state of reputation is calculated off-chain and submitted to the blockchain as a `reputationRootHash`. In the case of honest submissions, the miner is rewarded reputation in the Meta Colony tagged with a special 'mining' skill. In the event of malicious activity, a miner with a false submission will be punished by having her stake slashed.
```
area: 'Slashed' feels very Casper-specific to me. Can we use more general language?

docs/_Docs_ReputationMining.md
```diff
@@ -4,4 +4,55 @@ section: Docs
...
+
+The process works on an “innocent until proven guilty” principle: Although it’s impossible for a valid hash to be proven true on-chain, it’s relatively straightforward for an invalid hash to be proven false. In the event that two submitted hashes differ, the false one can be fleshed out through an automated challenge process *on-chain* that will eventually zero in on the exact on-chain events that differ between the two submitted hashes.
```
area: Not impossible, just very expensive. 

I'd also make it clear somewhere in this page that the narrative it supplies is assuming two things:

* Exactly two reputation hashes are submitted
* Exactly one of those hashes is wrong, and the other is correct.

In reality, an unlimited number of hashes can be submitted, and none of them are guaranteed to be right.
We are assuming that the correct hash is one of the (potentially many) submitted. It is possible that two incorrect hashes will be played off against each other, and one will 'win' that challenge. That doesn't mean that it's correct hash, just that it made a mistake in its reputation calculation later than the losing hash.

docs/_Docs_ReputationMining.md
```diff
@@ -4,4 +4,55 @@ section: Docs
...
+
+## The Reputation Mining Cycle
+
+
+### Reputation Root Hashes
+Using a [Patricia Tree](https://github.com/ethereum/wiki/wiki/Patricia-Tree), all updates to the global state of reputation are maintained on-chain with a single "Reputation Root Hash".
+
+The Reputation Root Hash is effectively a fingerprint for the network state (of reputation). It must be calculated off-chain, but it represents the collection of all reputation-updating events that have occured on the network, and even a single discrepancy in the complete history of reputational updates will result in a different Root Hash.  
+
+Miners are incentivized to maintain consensus over the state of reputation by repeatedly calculating the Reputation Root Hash and submitting it to the network.
```
area: This sentence starts 'miners are incentivized' and doesn't end with a description of the incentive which you naturally expect.

docs/_Docs_ReputationMining.md
```diff
@@ -4,4 +4,55 @@ section: Docs
...
+### Challenges
+The challenge-response protocol is described in section 7.5.1 of the [Colony whitepaper](https://colony.io/whitepaper.pdf).
+
+Leaving the exact mechanism aside, the general process can be summarized as follows:
+
+If two different root hashes are submitted, one must be false, and the miner that submitted an invalid hash must be punished.
+
+In whichever submission is false, there must be one or more reputation update that is not consistent with the global state of reputation up until the current submission window.
+
+Miners can submit a justification to the network that show how a reputation update results in a Root Hash consistent with the network state right before this submission window, and they can do so for each mining cycle going back to the network's beginning.
```
area: 'that shows'

The justification hash essentially commits the miner to not change their 'working' during the dispute resolution process in a way that can be verified on chain. It only represents the 'working' during the current mining window; it contains no proof for any previous windows. They could in theory be produced, but there's no real reason to.

docs/_Docs_ReputationMining.md
```diff
@@ -4,4 +4,55 @@ section: Docs
 order: 9
 ---
 
-Pending finalization of the reputation system implementation (see [GitHub](https://github.com/JoinColony/colonyNetwork) for more information).
+This is an overview of the process by which Reputation scores in Colony are maintained.
+
+For a complete description of the Reputation Mining process, see the [whitepaper](https://colony.io/whitepaper.pdf)
+
+# Rationale for off-chain Reputation
+Events in Colony that affect reputation are numerous and expected to be frequent. The calculations for global reputation updates would exceed Ethereum's gas limit, and cannot be performed on-chain. However, Reputation doesn't *need* to be on-chain, because reputation is entirely deterministic and calculated from on-chain events.
```
area: 'exceed Ethereum's gas limit' is a weird thing to claim; given enough time, you have an unlimited amount of gas on the blockchain. Just dial it down to something like 'the calculations for global reputation updates are so numerous that it would be impractical and preposterously expensive'.

gichiba: hmmm. I do wonder if there's a way to justify the claim... something like 
1) reputation update calc for an on-chain event costs higher than the gas limit per block 
2) the number of reputation update events per block is non-zero
3) if reputation operates on-chain, the gas required to complete calculations will increase at a rate greater than the additional gas available with each new block

PR #222, opened 2018-06-04
Feature/useful mining client

area:
# Do not merge in current state

Currently depends on #212, but I'm going to drop those commits. I'll also fill in this PR properly, but just getting it going so Chris can leave any comments.

To get a (complete?) dev environment going, including reputation mining:

* `bash ./scripts/start-blockchain-client.sh` (leave running)
* `./node_modules/.bin/truffle migrate --reset`
* `node ./client.js --file ./reputations.json --colonyNetworkAddress 0x76d508fa65654654ffdb334a3023353587112e09  --minerAddress 0xb77d57f4959eafa0339424b83fcfaf9c15407461` (leave running)

Note that a redeployment is not necessary, and if your `colonyNetwork` contract is elsewhere, or you've set up reputation mining for another account, you should change those parameters.

It will mine a new reputation state every hour, but you can skip time via

```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id": 1}' localhost:8545
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[]}' localhost:8545
```
And it should mine a new state within ten seconds
Finally, you have to mine three new states before you'll see a nonzero reputation at http://127.0.0.1:3000/0xdb8fe93a3a9c97f04f5c862f52a84f992bd331df/3/0xb77d57f4959eafa0339424b83fcfaf9c15407461  Any tasks completed will have their reputational consequences shown by the oracle after two reputation updates have occurred; these delays are a result of how the reputation mining system works.

PR #223, opened 2018-06-05
Add mining client usable for local development

area:
This is (almost) the same functionality as #222, but without the dependency on #212 . As a result, parent reputations and colony-wide reputation totals are not present, but can be merged sooner rather than later for the benefit of hackathon participants.

To get a (complete?) dev environment going, including reputation mining:

* `bash ./scripts/start-blockchain-client.sh` (leave running)
* `./node_modules/.bin/truffle migrate --reset`
* `cd client && yarn`
* `node ./client.js --file ./reputations.json --colonyNetworkAddress 0x76d508fa65654654ffdb334a3023353587112e09  --minerAddress 0xb77d57f4959eafa0339424b83fcfaf9c15407461` (leave running)

Note that a redeployment is not necessary, and if your `colonyNetwork` contract is elsewhere, or you've set up reputation mining for another account, you should change those parameters.

It will mine a new reputation state every hour, but you can skip time via

```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id": 1}' localhost:8545
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[]}' localhost:8545
```
And it should mine a new state within ten seconds
Finally, you have to mine three new states before you'll see a nonzero reputation at http://127.0.0.1:3000/0xdb8fe93a3a9c97f04f5c862f52a84f992bd331df/0/0xb77d57f4959eafa0339424b83fcfaf9c15407461  Any tasks completed will have their reputational consequences shown by the oracle after two reputation updates have occurred; these delays are a result of how the reputation mining system works.

`startNextCycle` is removed from some of the test setups, because the migration now does it. The changes to the migration are made for the benefit of development locally, at the cost of making a 'real' deployment harder, but I think we were moving away from using Truffle for that anyway...


client/main.js
```diff
@@ -101,6 +91,7 @@ class ReputationMiningClient {
     const tx = await this.ganacheWallet.sendTransaction(patriciaTreeDeployTx);
     this.reputationTree = new ethers.Contract(ethers.utils.getContractAddress(tx), PatriciaTreeJSON.abi, this.ganacheWallet);
     this.nReputations = 0;
+    this.reputations = {};
```
chmanie: This secret key, is it ever going to change? Is it ok for it to be hardcoded here just like that?
https://github.com/JoinColony/colonyNetwork/pull/223/files#diff-096103d8490b885f9d5037baea6aea80R16

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -0,0 +1,143 @@
...
+      jsonfile.writeFileSync(file, this.reputations);
+    } else {
+      for (let i = 0; i < Object.keys(this.reputations).length; i += 1) {
+        const key = Object.keys(this.reputations)[i];
+        await this.reputationTree.insert(key, this.reputations[key], { gasLimit: 4000000 }); // eslint-disable-line no-await-in-loop
+      }
+    }
+
+    console.log("🏁 Initialised");
+    const that = this;
```
chmanie: A cool feature of the new ES6 fat arrow is that it automatically binds the function to the outer context of it. So no need for `that` or `self` anymore. Hurray!

client/main.js
area: I can't see why it would ever change. This is the key used by the internal ganache instance used to host the Patricia Tree. It's never used for anything outward facing.

chmanie:
I made some changes to the ReputationMinerClient:

* Renamed it
* Attached the miner as a class instance variable instead of subclassing it
* Separated out the console related things (as far as I could for now)
* Used the fs-loader to load the contracts from a path relative to the cwd
* The file.json is now written and read relative to the cwd
* Made it compatible for an npm release (not released yet)

The way you start the miner now has changed. 

OLD:
```
node ./client.js --file ./reputations.json --colonyNetworkAddress 0x76d508fa65654654ffdb334a3023353587112e09 --minerAddress 0xb77d57f4959eafa0339424b83fcfaf9c15407461
```

NEW:
```
// From the colonyNetwork folder
node packages/reputation-miner/bin/index.js --file ./reputations.json --colonyNetworkAddress 0x76d508fa65654654ffdb334a3023353587112e09 --minerAddress 0xb77d57f4959eafa0339424b83fcfaf9c15407461
```

I have a couple of more things on my list which are less pressing. From my side this is ok to be reviewed and merged, for the rest I can open a new issue.

area:
I'm happy with Chris' changes, he's happy with mine. Elena is happy with the changes to the migration and the tests, so I'm going to merge this despite the fact no-one's 'officially' reviewed it...

elenadimitrova:
Everyone's happy with everyone! How lovely!

PR #224, opened 2018-06-05
Roles and permissions

filiplazovic:
- `ownerRole` - One person. Creator of the colony. Can call `setToken`, `bootstrapColony`, `mintTokens`, `addGlobalSkill`, `removeAdminRole`, plus all functions that admin can call.
- `adminRole` - has permission to call `setAdmin`, `moveFundsBetweenPots`, `addDomain`, `makeTask`, `cancelTask`, `startNextRewardPayout`.
- Member - can only call public functions that don't require auth.

Task level permission (as discussed with Pat, we should be the one to decide what those should be):

- `setTaskSkill` - Since worker is going to receive the reputation in this skill, should be agreed with manager and worker.
- `setTaskDomain` - Manager.
- `setTaskBrief` - Agreed between manager and worker.
- `setTaskDueDate` - Agreed between manager and worker.
- `finalizeTask` - Can only be called by a manager.
- `setTaskManagerPayout` - Can only be called by manager.
- `setTaskEvaluatorPayout` - Agreed between manager and evaluator.
- `setTaskWorkerPayout` - Agreed between manager and worker.

Assigning roles for the first time:

- Manager role - set automatically to whoever created a task
- Evaluator role - Signed by manager and address we want to assign role to
- Worker role - Signed by manager and address we want to assign role to

Unassigning roles:

- Manager role - Can’t be unassigned
- Evaluator role - Signed by manager and currently assigned evaluator
- Worker role - Signed by manager and currently assigned worker

Changing roles:

- Evaluator and worker - We would have 2 transactions: Unassigning a role (Signed by manager and currently assigned role) ===> Assigning new role (Signed by manager and address we want to assign role to)

- Manager - Signed by current manager and address we want to assign manager role to. Other address must also be an admin

Changes not in scope of the task:
- Refactored tests that include signed task changes to avoid code duplication.
- Refactored `finalizeTask` function to fix "stack too deep" error.
- Added helper function `makeTask`.

### DIFF
--- a/contracts/Authority.sol
+++ b/contracts/Authority.sol
@@ -33,6 +33,7 @@ contract Authority is DSRoles {
     bytes4 addGlobalSkillSig = bytes4(keccak256("addGlobalSkill(uint256)"));
     bytes4 setOwnerRoleSig = bytes4(keccak256("setOwnerRole(address)"));
     bytes4 removeAdminRoleSig = bytes4(keccak256("removeAdminRole(address)"));
+    bytes4 upgradeSig = bytes4(keccak256("upgrade(uint256)"));
 
     // functions for admin + owner role
     bytes4 moveFundsBetweenPotsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,address)"));
@@ -54,6 +55,8 @@ contract Authority is DSRoles {
     setRoleCapability(ownerRole, colony, setOwnerRoleSig, true);
     // Remove admin role
     setRoleCapability(ownerRole, colony, removeAdminRoleSig, true);
+    // Upgrade colony
+    setRoleCapability(ownerRole, colony, upgradeSig, true);
 
     // Allocate funds
     setRoleCapability(adminRole, colony, moveFundsBetweenPotsSig, true);
--- a/contracts/Authority.sol
+++ b/contracts/Authority.sol
@@ -22,28 +22,28 @@ import "../lib/dappsys/roles.sol";
 
 
 contract Authority is DSRoles {
-  uint8 ownerRole = 0;
   uint8 adminRole = 1;
 
   constructor(address colony) public {
-    bytes4 makeTaskSig = bytes4(keccak256("makeTask(bytes32,uint256)"));
-    setRoleCapability(ownerRole, colony, makeTaskSig, true);
-    setRoleCapability(adminRole, colony, makeTaskSig, true);
-
-    bytes4 acceptTaskSig = bytes4(keccak256("finalizeTask(uint256)"));
-    setRoleCapability(ownerRole, colony, acceptTaskSig, true);
-    setRoleCapability(adminRole, colony, acceptTaskSig, true);
-
-    bytes4 setTaskDueDateSig = bytes4(keccak256("setTaskDueDate(uint256,uint256)"));
-    setRoleCapability(ownerRole, colony, setTaskDueDateSig, true);
-    setRoleCapability(adminRole, colony, setTaskDueDateSig, true);
-
-    bytes4 setTaskPayoutSig = bytes4(keccak256("setTaskPayout(uint256,uint256,address,uint256)"));
-    setRoleCapability(ownerRole, colony, setTaskPayoutSig, true);
-    setRoleCapability(adminRole, colony, setTaskPayoutSig, true);
-
     bytes4 moveFundsBetweenPotsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,address)"));
-    setRoleCapability(ownerRole, colony, moveFundsBetweenPotsSig, true);
+    bytes4 addDomainSig = bytes4(keccak256("addDomain(uint256)"));
+    bytes4 makeTaskSig = bytes4(keccak256("makeTask(bytes32,uint256)"));
+    bytes4 startNextRewardPayoutSig = bytes4(keccak256("startNextRewardPayout(address)"));
+    bytes4 cancelTaskSig = bytes4(keccak256("cancelTask(uint256)"));
+    bytes4 setAdminSig = bytes4(keccak256("setAdmin(address)"));
+
+    // Admin
+    // Allocate funds
     setRoleCapability(adminRole, colony, moveFundsBetweenPotsSig, true);
+    // Add domain
+    setRoleCapability(adminRole, colony, addDomainSig, true);
+    // Add task
+    setRoleCapability(adminRole, colony, makeTaskSig, true);
+    // Start next reward payout
+    setRoleCapability(adminRole, colony, startNextRewardPayoutSig, true);
+    // Cancel task
+    setRoleCapability(adminRole, colony, cancelTaskSig, true);
+    // Set admin
+    setRoleCapability(adminRole, colony, setAdminSig, true);
   }
 }
--- a/contracts/Authority.sol
+++ b/contracts/Authority.sol
@@ -22,28 +22,53 @@ import "../lib/dappsys/roles.sol";
 
 
 contract Authority is DSRoles {
+  uint8 ownerRole = 0;
   uint8 adminRole = 1;
 
   constructor(address colony) public {
+    // functions for owner role
+    bytes4 setTokenSig = bytes4(keccak256("setToken(address)"));
+    bytes4 bootstrapColonySig = bytes4(keccak256("bootstrapColony(address[],int256[])"));
+    bytes4 mintTokensSig = bytes4(keccak256("mintTokens(uint256)"));
+    bytes4 addGlobalSkillSig = bytes4(keccak256("addGlobalSkill(uint256)"));
+    bytes4 removeAdminRoleSig = bytes4(keccak256("removeAdminRole(address)"));
+
+    // functions for admin + owner role
     bytes4 moveFundsBetweenPotsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,address)"));
     bytes4 addDomainSig = bytes4(keccak256("addDomain(uint256)"));
     bytes4 makeTaskSig = bytes4(keccak256("makeTask(bytes32,uint256)"));
     bytes4 startNextRewardPayoutSig = bytes4(keccak256("startNextRewardPayout(address)"));
     bytes4 cancelTaskSig = bytes4(keccak256("cancelTask(uint256)"));
-    bytes4 setAdminSig = bytes4(keccak256("setAdmin(address)"));
+    bytes4 setAdminRoleSig = bytes4(keccak256("setAdminRole(address)"));
 
-    // Admin
+    // Set token
+    setRoleCapability(ownerRole, colony, setTokenSig, true);
+    // Bootstrap colony
+    setRoleCapability(ownerRole, colony, bootstrapColonySig, true);
+    // Mint tokens
+    setRoleCapability(ownerRole, colony, mintTokensSig, true);
+    // Add global skill
+    setRoleCapability(ownerRole, colony, addGlobalSkillSig, true);
+    // Remove admin role
+    setRoleCapability(ownerRole, colony, removeAdminRoleSig, true);
+
     // Allocate funds
     setRoleCapability(adminRole, colony, moveFundsBetweenPotsSig, true);
+    setRoleCapability(ownerRole, colony, moveFundsBetweenPotsSig, true);
     // Add domain
     setRoleCapability(adminRole, colony, addDomainSig, true);
+    setRoleCapability(ownerRole, colony, addDomainSig, true);
     // Add task
     setRoleCapability(adminRole, colony, makeTaskSig, true);
+    setRoleCapability(ownerRole, colony, makeTaskSig, true);
     // Start next reward payout
     setRoleCapability(adminRole, colony, startNextRewardPayoutSig, true);
+    setRoleCapability(ownerRole, colony, startNextRewardPayoutSig, true);
     // Cancel task
     setRoleCapability(adminRole, colony, cancelTaskSig, true);
+    setRoleCapability(ownerRole, colony, cancelTaskSig, true);
     // Set admin
-    setRoleCapability(adminRole, colony, setAdminSig, true);
+    setRoleCapability(adminRole, colony, setAdminRoleSig, true);
+    setRoleCapability(ownerRole, colony, setAdminRoleSig, true);
   }
 }
--- a/contracts/Authority.sol
+++ b/contracts/Authority.sol
@@ -31,6 +31,7 @@ contract Authority is DSRoles {
     bytes4 bootstrapColonySig = bytes4(keccak256("bootstrapColony(address[],int256[])"));
     bytes4 mintTokensSig = bytes4(keccak256("mintTokens(uint256)"));
     bytes4 addGlobalSkillSig = bytes4(keccak256("addGlobalSkill(uint256)"));
+    bytes4 setOwnerRoleSig = bytes4(keccak256("setOwnerRole(address)"));
     bytes4 removeAdminRoleSig = bytes4(keccak256("removeAdminRole(address)"));
 
     // functions for admin + owner role
@@ -49,6 +50,8 @@ contract Authority is DSRoles {
     setRoleCapability(ownerRole, colony, mintTokensSig, true);
     // Add global skill
     setRoleCapability(ownerRole, colony, addGlobalSkillSig, true);
+    // Transfer ownership
+    setRoleCapability(ownerRole, colony, setOwnerRoleSig, true);
     // Remove admin role
     setRoleCapability(ownerRole, colony, removeAdminRoleSig, true);
 
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -23,6 +23,7 @@ import "./IColonyNetwork.sol";
 import "./IColony.sol";
 import "./ColonyStorage.sol";
 import "./PatriciaTree/PatriciaTreeProofs.sol";
+import "./Authority.sol";
 
 
 contract Colony is ColonyStorage, PatriciaTreeProofs {
@@ -30,6 +31,15 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   // This function, exactly as defined, is used in build scripts. Take care when updating.
   // Version number should be upped with every change in Colony or its dependency contracts or libraries.
   function version() public pure returns (uint256) { return 1; }
+
+  function setAdmin(address _user) public auth {
+    Authority(authority).setUserRole(_user, ADMIN_ROLE, true);
+  }
+
+  // Can only be called by the owner.
+  function removeAdmin(address _user) public auth {
+    Authority(authority).setUserRole(_user, ADMIN_ROLE, false);
+  }
 
   function setToken(address _token) public
   auth
@@ -84,7 +94,6 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     token.transfer(colonyNetworkAddress, _wad);
   }
 
-  //TODO: Secure this function 'properly'
   function addGlobalSkill(uint _parentSkillId) public
   auth
   returns (uint256)
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -56,10 +56,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     colonyNetworkAddress = _address;
 
     // Initialise the task update reviewers
-    setFunctionReviewers(0xda4db249, 0, 2); // setTaskBrief => manager, worker
-    setFunctionReviewers(0xcae960fe, 0, 2); // setTaskDueDate => manager, worker
-    setFunctionReviewers(0x6fb0794f, 0, 1); // setTaskEvaluatorPayout => manager, evaluator
-    setFunctionReviewers(0x2cf62b39, 0, 2); // setTaskWorkerPayout => manager, worker
+    setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskEvaluatorPayout(uint256,address,uint256)")), MANAGER, EVALUATOR);
+    setFunctionReviewers(bytes4(keccak256("setTaskWorkerPayout(uint256,address,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("removeTaskEvaluatorRole(uint256)")), MANAGER, EVALUATOR);
+    setFunctionReviewers(bytes4(keccak256("removeTaskWorkerRole(uint256)")), MANAGER, WORKER);
+
+    setRoleAssignmentFunction(bytes4(keccak256("setTaskManagerRole(uint256,address)")));
+    setRoleAssignmentFunction(bytes4(keccak256("setTaskEvaluatorRole(uint256,address)")));
+    setRoleAssignmentFunction(bytes4(keccak256("setTaskWorkerRole(uint256,address)")));
 
     // Initialise the root domain
     IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
@@ -133,6 +140,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   {
     uint8[2] memory _reviewers = [_firstReviewer, _secondReviewer];
     reviewers[_sig] = _reviewers;
+  }
+
+  function setRoleAssignmentFunction(bytes4 _sig) private {
+    roleAssignmentSigs[_sig] = true;
   }
 
   modifier verifyKey(bytes key) {
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -34,8 +34,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
   function setOwnerRole(address _user) public auth {
     // To allow only one owner at a time we have to remove current owner from their role
-    Authority(authority).setUserRole(msg.sender, OWNER_ROLE, false);
-    Authority(authority).setUserRole(_user, OWNER_ROLE, true);
+    Authority colonyAuthority = Authority(authority);
+    colonyAuthority.setUserRole(msg.sender, OWNER_ROLE, false);
+    colonyAuthority.setUserRole(_user, OWNER_ROLE, true);
   }
 
   function setAdminRole(address _user) public auth {
@@ -65,6 +66,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), MANAGER, WORKER);
     setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), MANAGER, WORKER);
     setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskManagerPayout(uint256,address,uint256)")), MANAGER, MANAGER);
     setFunctionReviewers(bytes4(keccak256("setTaskEvaluatorPayout(uint256,address,uint256)")), MANAGER, EVALUATOR);
     setFunctionReviewers(bytes4(keccak256("setTaskWorkerPayout(uint256,address,uint256)")), MANAGER, WORKER);
     setFunctionReviewers(bytes4(keccak256("removeTaskEvaluatorRole(uint256)")), MANAGER, EVALUATOR);
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -33,6 +33,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   function version() public pure returns (uint256) { return 1; }
 
   function setOwnerRole(address _user) public auth {
+    // To allow only one owner at a time we have to remove current owner from their role
+    Authority(authority).setUserRole(msg.sender, OWNER_ROLE, false);
     Authority(authority).setUserRole(_user, OWNER_ROLE, true);
   }
 
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -24,6 +24,7 @@ import "./IColony.sol";
 import "./ColonyStorage.sol";
 import "./PatriciaTree/PatriciaTreeProofs.sol";
 import "./Authority.sol";
+import "./EtherRouter.sol";
 
 
 contract Colony is ColonyStorage, PatriciaTreeProofs {
@@ -33,7 +34,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   function version() public pure returns (uint256) { return 1; }
 
   function setOwnerRole(address _user) public auth {
-    // To allow only one owner at a time we have to remove current owner from their role
+    // To allow only one address to have owner role at a time, we have to remove current owner from their role
     Authority colonyAuthority = Authority(authority);
     colonyAuthority.setUserRole(msg.sender, OWNER_ROLE, false);
     colonyAuthority.setUserRole(_user, OWNER_ROLE, true);
@@ -43,7 +44,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     Authority(authority).setUserRole(_user, ADMIN_ROLE, true);
   }
 
-  // Can only be called by the owner.
+  // Can only be called by the owner role.
   function removeAdminRole(address _user) public auth {
     Authority(authority).setUserRole(_user, ADMIN_ROLE, false);
   }
@@ -66,6 +67,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), MANAGER, WORKER);
     setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), MANAGER, WORKER);
     setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), MANAGER, WORKER);
+    // We are setting a manager to both reviewers, but it will require just one signature from manager
     setFunctionReviewers(bytes4(keccak256("setTaskManagerPayout(uint256,address,uint256)")), MANAGER, MANAGER);
     setFunctionReviewers(bytes4(keccak256("setTaskEvaluatorPayout(uint256,address,uint256)")), MANAGER, EVALUATOR);
     setFunctionReviewers(bytes4(keccak256("setTaskWorkerPayout(uint256,address,uint256)")), MANAGER, WORKER);
@@ -180,6 +182,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     bytes32 impliedHash = getImpliedRoot(key, value, branchMask, siblings);
     require(rootHash==impliedHash, "colony-invalid-reputation-proof");
     return true;
+  }
+
+  function upgrade(uint256 _newVersion) public auth {
+    // Upgrades can only go up in version
+    uint256 currentVersion = version();
+    require(_newVersion > currentVersion);
+    // Requested version has to be registered
+    address newResolver = IColonyNetwork(colonyNetworkAddress).getColonyVersionResolver(_newVersion);
+    require(newResolver != 0x0);
+    EtherRouter e = EtherRouter(address(this));
+    e.setResolver(newResolver);
   }
 
   function initialiseDomain(uint256 _skillId) private skillExists(_skillId) {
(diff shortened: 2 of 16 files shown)


.solcover.js
```diff
@@ -4,7 +4,8 @@ module.exports = {
       'EtherRouter.sol',
       'gnosis',
       'Token.sol',
-      'PatriciaTree'
+      'PatriciaTree',
+      'SafeMath.sol'
```
elenadimitrova: Why was this ignored from coverage?

filiplazovic: The coverage went down when i added these changes, my guess is that `solidity-coverage` can't recognize what are we calling with `address.call`. To try to bring it up, i removed `SafeMath` since its a library and we dont have any tests for it. What do you think we should do?

elenadimitrova: Looking at test coverage reports functions called only via `address.call` show as executed, e.g. `setTaskSkill` shows as called 273 times https://1753-89702863-gh.circle-artifacts.com/0/home/circleci/colonyNetwork/coverage/contracts/ColonyTask.sol.html

elenadimitrova: Additionally if `.call` isn't working, `.delegatecall` would likely not be working too and that is at the basis of `EtherRouter` which is how almost everything is called. Look at the actually coverage results before making conclusions and always have a really good motivation before excluding anything from being covered in tests. Especially if it has to do with safely working with integers on chain :)

filiplazovic: I made a guess, and i was wrong. I tried to open coverage results locally, but no success. I figured it out now, thanks :)

contracts/ColonyNetwork.sol
```diff
@@ -120,9 +121,15 @@ contract ColonyNetwork is ColonyNetworkStorage {
     IColony colony = IColony(etherRouter);
     colony.setToken(_tokenAddress);
 
-    Authority authority = new Authority(colony);
+    ColonyAuthority colonyAuthority = new ColonyAuthority(colony);
+    Authority authority = new Authority(address(colonyAuthority));
```
elenadimitrova: Why are you explicitly casting `ColonyAuthority` to `Authority` here. This is needed in JS but Solidity understands inheritance so there shouldn't be a need.

contracts/ColonyTask.sol
```diff
@@ -260,24 +261,23 @@ contract ColonyTask is ColonyStorage, DSMath {
...
   taskExists(_id)
   taskNotFinalized(_id)
   domainExists(_domainId)
   {
-    require(tasks[_id].roles[MANAGER].user == msg.sender);
     tasks[_id].domainId = _domainId;
   }
 
   // TODO: Restrict function visibility to whoever submits the approved Transaction from Client
   // Maybe just the administrator is adequate for the skill?
```
elenadimitrova: You can remove this `todo`. It is solved by introducing the restrictions of `self`

contracts/ColonyTask.sol
```diff
@@ -309,13 +309,15 @@ contract ColonyTask is ColonyStorage, DSMath {
...
   function finalizeTask(uint256 _id) public
-  auth
   taskWorkRatingsAssigned(_id)
   taskNotFinalized(_id)
   {
     Task storage task = tasks[_id];
     IColonyNetwork colonyNetworkContract = IColonyNetwork(colonyNetworkAddress);
 
+    // Stack too deep, so we have to reuse task variable
+    require(task.roles[MANAGER].user == msg.sender);
```
elenadimitrova: Stack depth is becoming a real nuisance. I have some hope it maybe solved by the refactoring of this function that Daniel is doing but meanwhile can we keep the modifier and start removing the local variables, like `colonyNetworkContract` and `task`

helpers/test-data-generator.js
```diff
@@ -28,16 +28,23 @@ export async function setupAssignedTask({ colonyNetwork, colony, dueDate, domain
...
-    await colony.setTaskSkill(taskId, rootGlobalSkill);
+
+    const txData = await colony.contract.setTaskSkill.getData(taskId, rootGlobalSkill);
+    const sigs = await createSignatures(colony, [MANAGER, worker], 0, txData);
+    await colony.executeTaskChange(sigs.sigV, sigs.sigR, sigs.sigS, [0, 0], 0, txData);
   } else {
-    await colony.setTaskSkill(taskId, skill);
+    const txData = await colony.contract.setTaskSkill.getData(taskId, skill);
+    const sigs = await createSignatures(colony, [MANAGER, worker], 0, txData);
+    await colony.executeTaskChange(sigs.sigV, sigs.sigR, sigs.sigS, [0, 0], 0, txData);
```
elenadimitrova: Once you rebase on top of the work from #203 you can save some processing here by keeping the task unassigned until later and setting the task skill with just the manager signature.

helpers/test-helper.js
```diff
@@ -180,6 +180,10 @@ export async function forwardTime(seconds, test) {
   return p;
 }
 
+export function getFunctionSignature(sig) {
+  return web3Utils.sha3(sig).slice(0, 10);
```
elenadimitrova: There must be a function in one of the packages we import that already does that. Like `ethers` or https://web3js.readthedocs.io/en/1.0/web3-eth-abi.html#encodefunctionsignature

test/colony.js
```diff
@@ -733,4 +735,63 @@ contract("Colony", addresses => {
       await checkErrorRevert(colony.claimPayout(taskId, MANAGER_ROLE, token.address, { from: OTHER }));
     });
   });
+
+  describe("roles and permissions", () => {
```
elenadimitrova: Can you merge this section with `when working with permissions` above? When reading the descriptions they seem the same.

contracts/ColonyNetwork.sol
filiplazovic: So i'm creating authority for colony's authority and setting it [here](https://github.com/JoinColony/colonyNetwork/pull/224/files#diff-94738a1e114b0a8e909bb03e34628bf6R126), as you explained me. I'm just not sure how to name it, `ColonyAuthorityAuthority` sounds funny. 😄

contracts/ColonyNetwork.sol
```diff
@@ -120,9 +121,15 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+    colonyAuthority.setAuthority(authority);
+
     DSAuth dsauth = DSAuth(etherRouter);
-    dsauth.setAuthority(authority);
+    dsauth.setAuthority(colonyAuthority);
+
+    colonyAuthority.setRootUser(msg.sender, true);
+    colonyAuthority.setOwner(msg.sender);
```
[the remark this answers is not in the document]
filiplazovic: In `DSAuth`'s constructor sender is set as owner, so sender is ColonyNetwork. I'm setting the user (the one who called `createColony`) to be an owner, as it was before. I'm also setting the user to an owner for `Authority` (colony's authority authority 😄)

helpers/test-helper.js
filiplazovic: The only package we have is `web3-utils` which doesn't have `encodeFunctionSignature`. I just thought that maybe we shouldn't add whole library for one function, except if we are going to use it later

contracts/ColonyNetwork.sol
filiplazovic: This part is confusing, so i will add some comments

elenadimitrova: My note was about the cast, not about the name.

filiplazovic: The only cast i see here is `ColonyAuthority` to `address`, then passing it to `Authority`'s constructor

elenadimitrova: Sorry I'm blind. Ignore me.

contracts/ColonyNetwork.sol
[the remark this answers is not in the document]
elenadimitrova: Just thinking about permissions inheritance here, is there a problem with potentially the colony authority admins having permissions to change the authority themselves?

contracts/ColonyTask.sol
elenadimitrova: Daniel's PR was merged so you can test stack depth again here.

contracts/ColonyTask.sol
```diff
@@ -266,13 +266,12 @@ contract ColonyTask is ColonyStorage, DSMath {
     return taskWorkRatings[_id].secret[_role];
   }
 
-  // TODO: Restrict function visibility to whoever submits the approved Transaction from Client
   // Note task assignment is agreed off-chain
   function setTaskRoleUser(uint256 _id, uint8 _role, address _user) public
+  confirmTaskRoleIdentity(_id, MANAGER)
```
elenadimitrova: Can we refactor `isManager` to `confirmTaskRoleIdentity` as it is doing the same thing?

gasCosts/gasCosts.js
```diff
@@ -110,56 +110,63 @@ contract("All", accounts => {
...
     it("when working with a Task", async () => {
-      await colony.makeTask(SPECIFICATION_HASH, 1);
-      await colony.setTaskDomain(1, 1);
-      await colony.setTaskSkill(1, 7);
-      await colony.setTaskRoleUser(1, EVALUATOR_ROLE, EVALUATOR);
-      await colony.setTaskRoleUser(1, WORKER_ROLE, WORKER);
+      const { logs } = await colony.makeTask(SPECIFICATION_HASH, 1);
+      const taskId = logs[0].args.id.toNumber();
+      await colony.setTaskRoleUser(taskId, EVALUATOR_ROLE, EVALUATOR);
+      await colony.setTaskRoleUser(taskId, WORKER_ROLE, WORKER);
```
elenadimitrova: Can you delay setting the worker and execute the `setTaskDomain` and `setTaskSkill` with just the manager signature? This will reflect on the reported gas prices to show a better average inclusive of this case.

gasCosts/gasCosts.js
```diff
@@ -110,56 +110,63 @@ contract("All", accounts => {
     });
 
     it("when working with a Task", async () => {
-      await colony.makeTask(SPECIFICATION_HASH, 1);
-      await colony.setTaskDomain(1, 1);
-      await colony.setTaskSkill(1, 7);
-      await colony.setTaskRoleUser(1, EVALUATOR_ROLE, EVALUATOR);
-      await colony.setTaskRoleUser(1, WORKER_ROLE, WORKER);
+      const { logs } = await colony.makeTask(SPECIFICATION_HASH, 1);
+      const taskId = logs[0].args.id.toNumber();
```
elenadimitrova: 👍

helpers/test-data-generator.js
```diff
@@ -28,13 +28,19 @@ export async function setupAssignedTask({ colonyNetwork, colony, dueDate, domain
...
-    await colony.setTaskSkill(taskId, rootGlobalSkill);
+
+    const txData = await colony.contract.setTaskSkill.getData(taskId, rootGlobalSkill.toNumber());
+    const sigs = await createSignatures(colony, taskId, [MANAGER], 0, txData);
+    await colony.executeTaskChange(sigs.sigV, sigs.sigR, sigs.sigS, [0, 0], 0, txData);
   } else {
-    await colony.setTaskSkill(taskId, skill);
+    const txData = await colony.contract.setTaskSkill.getData(taskId, skill);
+    const sigs = await createSignatures(colony, taskId, [MANAGER], 0, txData);
+    await colony.executeTaskChange(sigs.sigV, sigs.sigR, sigs.sigS, [0, 0], 0, txData);
```
elenadimitrova: This is good but can we refactor it so we're not repeating all this code?

helpers/test-helper.js
elenadimitrova: The only other option is the `Resolver.stringToSig` function but let's leave it as it is for now.

test/meta-colony.js
```diff
@@ -389,8 +389,15 @@ contract("Meta Colony", accounts => {
 
     it("should be able to set domain on task", async () => {
       await colony.addDomain(3);
-      await colony.makeTask(SPECIFICATION_HASH, 1);
-      await colony.setTaskDomain(1, 2);
+      const { logs } = await colony.makeTask(SPECIFICATION_HASH, 1);
+      const taskId = logs[0].args.id.toNumber();
+      await colony.setTaskRoleUser(1, WORKER_ROLE, WORKER);
```
elenadimitrova: Now that we can execute with just 1 signature, let's keep test data setup simpler. You can remove this line and execute just with the Manager's signature.

test/meta-colony.js
```diff
@@ -426,8 +433,15 @@ contract("Meta Colony", accounts => {
       await metaColony.addGlobalSkill(1);
       await metaColony.addGlobalSkill(4);
 
-      await colony.makeTask(SPECIFICATION_HASH, 1);
-      await colony.setTaskSkill(1, 5);
+      const { logs } = await colony.makeTask(SPECIFICATION_HASH, 1);
+      const taskId = logs[0].args.id.toNumber();
+
+      await colony.setTaskRoleUser(1, WORKER_ROLE, WORKER);
```
elenadimitrova: Same as above

test/colony.js
```diff
@@ -123,30 +125,87 @@ contract("Colony", addresses => {
...
+        from: address1
+      });
+
+      const hasRole = await colonyAuthority.hasUserRole(address2, adminRole);
+      assert(hasRole, `Admin role not assigned to ${address2}`);
+    });
+
+    it("should allow admin to call predetermined functions", async () => {
+      const adminRole = 1;
+      const address = addresses[3];
```
elenadimitrova: Can you name this constant better? `user3` should do it

test/colony.js
```diff
@@ -123,30 +125,87 @@ contract("Colony", addresses => {
...
+      await colonyAuthority.setUserRole(OTHER, 1, false);
+      const admin = await colonyAuthority.hasUserRole.call(OTHER, 1);
       assert.isFalse(admin);
     });
+
+    it("should allow admin to assign admin role", async () => {
+      const adminRole = 1;
+
+      const address1 = addresses[1];
+      const address2 = addresses[5];
```
elenadimitrova: Since an `address` can be a contract or a user, can we call these `user1` and `user5` or something like that?

test/colony.js
```diff
@@ -123,30 +125,87 @@ contract("Colony", addresses => {
...
-      await authority.setUserRole(OTHER, 1, true);
-      await authority.setUserRole(OTHER, 1, false);
-      const admin = await authority.hasUserRole.call(OTHER, 1);
+      await colonyAuthority.setUserRole(OTHER, 1, true);
+      await colonyAuthority.setUserRole(OTHER, 1, false);
+      const admin = await colonyAuthority.hasUserRole.call(OTHER, 1);
       assert.isFalse(admin);
     });
+
+    it("should allow admin to assign admin role", async () => {
```
elenadimitrova: admin to assign admin role -> admin to assign colony admin role

contracts/ColonyNetwork.sol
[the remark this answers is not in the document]
filiplazovic: I think there is no problem, we just have to decide if we want to give them that permission.

contracts/ColonyTask.sol
filiplazovic: The problem remains, and removing local variables won't help. The only way to solve this is with internal functions.

contracts/ColonyTask.sol
```diff
@@ -266,13 +260,12 @@ contract ColonyTask is ColonyStorage, DSMath {
     return taskWorkRatings[_id].secret[_role];
   }
 
-  // TODO: Restrict function visibility to whoever submits the approved Transaction from Client
   // Note task assignment is agreed off-chain
   function setTaskRoleUser(uint256 _id, uint8 _role, address _user) public
```
elenadimitrova: Ok so we are securing this down to the manager only. That means that they can change the assigned worker at any point which is not right. Have you discussed this scenario with Pat is this what they want?

contracts/ColonyTask.sol
```diff
@@ -283,26 +276,23 @@ contract ColonyTask is ColonyStorage, DSMath {
   }
 
   function setTaskDomain(uint256 _id, uint256 _domainId) public
+  self()
```
elenadimitrova: Here we extend the authorisation to need not just the manager but the worker too. Is this something the product team have requested? Same question for `setSkill` function authorisation requirements

contracts/ColonyTask.sol
```diff
@@ -391,6 +370,21 @@ contract ColonyTask is ColonyStorage, DSMath {
     return (role.user, role.rateFail, uint8(role.rating));
   }
 
+  function updateReputation(uint8 roleId, Task storage task) internal {
```
elenadimitrova: 👍

gasCosts/gasCosts.js
```diff
@@ -110,56 +110,94 @@ contract("All", accounts => {
...
+
+      // setTaskDomain
+      await executeSignedTaskChange({
+        colony,
+        functionName: "setTaskDomain",
+        taskId,
+        signers: [MANAGER],
+        sigTypes: [0],
+        args: [taskId, 1]
+      });
```
elenadimitrova: Same as this comment https://github.com/JoinColony/colonyNetwork/pull/214#discussion_r195729747 
Applicable to the rest of the JS changes in this PR

contracts/ColonyAuthority.sol
```diff
@@ -0,0 +1,50 @@
...
+    setRoleCapability(adminRole, colony, makeTaskSig, true);
+    // Start next reward payout
+    setRoleCapability(adminRole, colony, startNextRewardPayoutSig, true);
+    // Cancel task
+    setRoleCapability(adminRole, colony, cancelTaskSig, true);
+  }
+
+  function setAdmin(address _user) public auth {
+    setUserRole(_user, adminRole, true);
+  }
```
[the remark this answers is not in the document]
filiplazovic: I will check with product team, but there was no mention of this.

contracts/ColonyAuthority.sol
```diff
@@ -0,0 +1,50 @@
...
+
+contract ColonyAuthority is DSRoles {
+  uint8 adminRole = 1;
+
+  constructor(address colony) public {
+    bytes4 moveFundsBetweenPotsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,address)"));
+    bytes4 addDomainSig = bytes4(keccak256("addDomain(uint256)"));
+    bytes4 makeTaskSig = bytes4(keccak256("makeTask(bytes32,uint256)"));
+    bytes4 startNextRewardPayoutSig = bytes4(keccak256("startNextRewardPayout(address)"));
+    bytes4 cancelTaskSig = bytes4(keccak256("cancelTask(uint256)"));
```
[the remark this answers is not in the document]
filiplazovic: These are the functions that i got from the product team. Every other function with `auth` can be called only by owner. I can double check again

contracts/ColonyTask.sol
filiplazovic: As i said in PR description, Pat decided to delegate task level permissions to us, developers. You can also see in the description who can call which function, so if you have different opinion, we can change it.

contracts/ColonyTask.sol
filiplazovic: Ok, we can restrict `setTaskDomain` and `setSkill` just to managers 👍

filiplazovic:
@elenadimitrova `executeSignedTaskChange` 
 was not here last time. You mentioned we would like to avoid code duplication https://github.com/JoinColony/colonyNetwork/pull/224#discussion_r195704976

gasCosts/gasCosts.js
filiplazovic: For this particular example, i formated it this way, it was more readable. And for https://github.com/JoinColony/colonyNetwork/pull/214#discussion_r195729747, it was `prettier` that gave me an error.

contracts/ColonyTask.sol
filiplazovic: Can you also see every other task related function. You can see just the description first, and the code later

contracts/Colony.sol
```diff
@@ -89,7 +91,6 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     token.transfer(colonyNetworkAddress, _wad);
   }
 
-  //TODO: Secure this function 'properly'
```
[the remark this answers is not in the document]
filiplazovic: I thought this was about permissions, i will revert it.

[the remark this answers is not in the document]
elenadimitrova: It is about permissions and these permissions need to be implemented and this todo removed, I just can't see it done in the Roles assignment

contracts/ColonyAuthority.sol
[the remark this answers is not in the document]
elenadimitrova: Make sure they understand there is only one owner as opposed to the "owner role" which used to contain multiple users.

contracts/ColonyTask.sol
elenadimitrova: You mean in the PR description? I can't comment against the description to extracting it here for feedback (if that's what you meant):

>setTaskSkill - Since everybody will receive the reputation in that skill, this should be probably agreed by everyone? Right now it is agreed between manager and worker.

Not everyone receives rep in the global skill, just the worker. It's probably best to leave it as-is to be agreed between the manager and worker.

>setTaskDomain - Since worker is going to receive reputation in that domain, manager and worker should agree.

Every role received domain rep actually but since the manager determines the domain with the task creation I don't see a point in restricting that down to other roles if they decided to change it.

elenadimitrova: >setTaskManagerPayout - Can only be called by manager. Should we allow manager to set his payout?

Yes, I think they should be able to determine their own payout.

elenadimitrova: > setTaskRoleUser
Ideally we would get mutual approval between the manager and whichever role is being assigned here. Not sure how possible/expensive is this in code as `executeTaskChange` relies on the reviewer roles being present to execute. In cases where one is missing, the other one executes alone. Can you find out?

helpers/test-data-generator.js
elenadimitrova: I meant this code repeating
```
// If the skill is not specified, default to the root global skill
  if (skill === 0) {
    const rootGlobalSkill = await colonyNetwork.getRootGlobalSkillId.call();
    if (rootGlobalSkill.toNumber() === 0) throw new Error("Meta Colony is not setup and therefore the root global skill does not exist");

    await executeSignedTaskChange({
      colony,
      functionName: "setTaskSkill",
      taskId,
      signers: [MANAGER],
      sigTypes: [0],
      args: [taskId, rootGlobalSkill.toNumber()]
    });
  } else {
    await executeSignedTaskChange({
      colony,
      functionName: "setTaskSkill",
      taskId,
      signers: [MANAGER],
      sigTypes: [0],
      args: [taskId, skill]
    });
  }
```
and refactoring it to something like
```
// If the skill is not specified, default to the root global skill
  if (skill === 0) {
    const rootGlobalSkill = await colonyNetwork.getRootGlobalSkillId.call();
    if (rootGlobalSkill.toNumber() === 0) throw new Error("Meta Colony is not setup and therefore the root global skill does not exist");
    skill = rootGlobalSkill.toNumber();
  }
    await executeSignedTaskChange({ colony, functionName: "setTaskSkill", taskId, signers: [MANAGER], sigTypes: [0], args: [taskId, skill]});
```

elenadimitrova: But the `executeSignedTaskChange` is good too so leave that refactoring in.

gasCosts/gasCosts.js
elenadimitrova: I personally prefer the one line formatting. But this is a topic for a different conversation that I'm happy to have outside of this PR. I don't think we should partially change the formatting here, for now just follow the one-line style which is what we've used in the remainder of the codebase.

contracts/ColonyAuthority.sol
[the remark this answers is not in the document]
filiplazovic: Yes, they understand, i talked with them about this.

contracts/Colony.sol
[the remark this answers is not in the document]
filiplazovic: So this is `addGlobalSkill`, which can be called only by `owner`

contracts/ColonyTask.sol
filiplazovic: Ok, i switched `setTaskDomain` and `setTaskSkill`, sorry: So,
- `setTaskSkill` - manager and worker.
- `setTaskDomain` - manager.

contracts/Colony.sol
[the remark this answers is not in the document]
elenadimitrova: Only one person can add skills in the meta colony? And admins can't? If this is what product team wants, ok. I thought that meta colony admins should be able to manage that too maybe. I also don't see this function listed in the PR description so I thought you haven't clarified it yet. If all is done you can remove the todo.

contracts/ColonyTask.sol
elenadimitrova: Can you have a separate commit testing how the implementation of `setTaskRoleUser` would work? Then we can decide whether to include it. The main problem with this is the `executeTaskChange` implementation which won't cover this scenario in its current shape, i.e. `setTaskWorker` would require manager and worker to agree but worker is not assigned here as the function itself does that.

elenadimitrova: ok for `setTaskDomain` and `setTaskSkill`

elenadimitrova:
ok I think we're reasonably clear on this apart from a couple of things to check with product and how we'd cover the assigning of workers in `executeTaskChange` function.

area:
Been giving this a little thought in the shower this morning; this is always going to be a special case, but what about setting roles only ever does one of two things:

* If worker is already assigned, require manager+worker agreement, and only allow setting of worker to 0x0.
* If worker is not already assigned, require manager only, and set to whatever address the second signature is from.

Needs to also accommodate the corner-case of manager setting themselves as the worker, but I think I prefer this to requiring triple-confirmation on-chain.

Note changing from one worker to another can still be displayed as a single triple-confirmation process in the client, it's just done over two transactions behind the scenes.

elenadimitrova:
I agree with @area but wanted to be clear on this one:

> If worker is not already assigned, require manager only, and set to whatever address the second signature is from

We should probably additionally confirm the second signature is the `_user` specified in the `setTaskRoleUser(uint256 _id, uint8 _role, address _user)` call. Although "stack too deep" maybe catching up on us again in `executeTaskChange`..

filiplazovic:
Yes @elenadimitrova, that's exactly what i was planing to implement.
My only concern is that if i put this functionality in `executeTaskChange` its going to be really messy, because this role change is very different from task change, so we will end up with one function that does totaly different things at different times.

elenadimitrova:
On this
>how are we able to change manager with triple-confirmation process?

I don't see it as a "triple confirmation" as the manager role can never be unassigned right? So we simply require that both the new and old manager sign the change?

filiplazovic:
Yea, that does make sense. I experimented with making admin + manager sign the changes, but that makes more sense 👍

filiplazovic:
`addGlobalSkill` still remains available only to owner
`removeAdmin` is added, can only be called by owner
Functionality for signing roles looks complex, and its a challange to write it simpler. In the meantime, this can be reviewed.

contracts/ColonyTask.sol
```diff
@@ -402,10 +517,12 @@ contract ColonyTask is ColonyStorage, DSMath {
...
   // Get the function signature and task id from the transaction bytes data
   // Note: Relies on the encoded function's first parameter to be the uint256 taskId
-  function deconstructCall(bytes _data) internal pure returns (bytes4 sig, uint256 taskId) {
+  function deconstructCall(bytes _data) internal pure returns (bytes4 sig, uint256 taskId, uint8 roleId, address userAddress) {
     assembly {
       sig := mload(add(_data, 0x20))
-      taskId := mload(add(_data, add(0x20, 4))) // same as calldataload(72)
+      taskId := mload(add(_data, 0x24)) // same as calldataload(72)
+      roleId := mload(add(_data, 0x44))
+      userAddress := mload(add(_data, 0x64))
```
elenadimitrova: `roleId` and  `userAddress ` won't work for anything but the `setTaskRoleUser` function. This feels very wrong..

test/colony-network-mining.js
```diff
@@ -1527,7 +1527,7 @@ contract("ColonyNetworkStaking", accounts => {
       const taskId = await setupRatedTask({
         colonyNetwork,
         colony: metaColony,
-        worker: MAIN_ACCOUNT,
+        worker: accounts[4],
```
area: What's going on here?

filiplazovic: `MAIN_ACCOUNT` is manager, worker can't be a manager.

area: There should be no restrictions on whether multiple roles can be the same person. We can't stop the same person having two different addresses after all.

filiplazovic: Ok, that shouldn't be hard to fix.

test/colony.js
```diff
@@ -142,30 +150,88 @@ contract("Colony", addresses => {
...
+      const hasRole = await authority.hasUserRole(user5, adminRole);
+      assert(hasRole, `Admin role not assigned to ${user5}`);
     });
 
-    it("should be able to remove a colony owner", async () => {
-      await authority.setUserRole(OTHER, 0, true);
-      await authority.setUserRole(OTHER, 0, false);
-      const owner = await authority.hasUserRole.call(OTHER, 0);
-      assert.isFalse(owner);
+    it("should allow owner to remove colony admin role", async () => {
```
area: Is there a test anywhere checking that admins can't do this?

test/colony.js
```diff
@@ -189,25 +255,30 @@ contract("Colony", addresses => {
     });
 
     it("should fail if a non-admin user tries to make a task", async () => {
-      await checkErrorRevert(colony.makeTask(SPECIFICATION_HASH, 1, { from: OTHER }));
+      await checkErrorRevert(
```
area: This appears to be a bit of a fast-and-lose-use of `checkErrorRevert`, both here and elsewhere in the PR. It seems to work, but I don't think the test would fail cleanly if the `makeTask` call succeeded; `makeTask` returns a `taskId`, which couldn't be used by `checkErrorRevert` to get a transaction receipt to check the status field...

(I actually don't think `checkErrorRevert` fails cleanly as-is if it's passed a promise for a transaction that succeeds, but this sort of usage would make it a lot harder)

contracts/ColonyFunding.sol
```diff
@@ -39,7 +39,9 @@ contract ColonyFunding is ColonyStorage, DSMath {
     return 100;
   }
 
-  function setTaskManagerPayout(uint256 _id, address _token, uint256 _amount) public isManager(_id) {
+  function setTaskManagerPayout(uint256 _id, address _token, uint256 _amount) public
```
area: Is it weird that this doesn't go through `executeTaskChange`, but the other payouts do? I understand why (we never need a second signature), but would it be easier for people to use if they were all the same?

contracts/IColony.sol
```diff
@@ -100,6 +100,21 @@ contract IColony {
   /// that control has to be transferred to the colony after this call
   function setToken(address _token) public;
 
+  /// @notice Set new colony owner.
+  /// Can be called by colony network.
```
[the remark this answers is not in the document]
filiplazovic: Yes, this was not updated.

test/colony.js
filiplazovic: Ok, reverting this part will probably fix this.

contracts/ColonyFunding.sol
filiplazovic: Yeah, it will make it consistent.

contracts/ColonyTask.sol
```diff
@@ -401,12 +454,59 @@ contract ColonyTask is ColonyStorage, DSMath {
...
+
+  function deconstructRoleChangeCall(bytes _data) internal pure returns (bytes4 sig, uint256 taskId, address userAddress) {
+    assembly {
+      sig := mload(add(_data, 0x20))
+      taskId := mload(add(_data, 0x24)) // same as calldataload(72)
+      userAddress := mload(add(_data, 0x44))
     }
   }
+
+  function setTaskRoleUser(uint256 _id, uint8 _role, address _user) private
```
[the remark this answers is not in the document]
filiplazovic: I will leave a todo, and investigate this, as this PR has enough thinks that are not in scope

test/colony.js
```diff
@@ -95,7 +103,7 @@ contract("Colony", addresses => {
       assert.equal(colonyBalance.toNumber(), 1);
     });
 
-    it("should take colony network as an owner", async () => {
+    it("should set colony network to be owner", async () => {
```
elenadimitrova: Like I mentioned in the contract logic above, I don't understand this decision. After the colony is created, its permissions should be contained to the creator and the network should no longer have any control over it.

contracts/Colony.sol
```diff
@@ -46,10 +63,18 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
 
     // Initialise the task update reviewers
-    setFunctionReviewers(0xda4db249, 0, 2); // setTaskBrief => manager, worker
-    setFunctionReviewers(0xcae960fe, 0, 2); // setTaskDueDate => manager, worker
-    setFunctionReviewers(0x6fb0794f, 0, 1); // setTaskEvaluatorPayout => manager, evaluator
-    setFunctionReviewers(0x2cf62b39, 0, 2); // setTaskWorkerPayout => manager, worker
+    setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskManagerPayout(uint256,address,uint256)")), MANAGER, MANAGER);
```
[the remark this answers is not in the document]
filiplazovic: This was change from last review. Alex thinks that we should be consistent with these task change calls. It will be easier for the front-end.

contracts/ColonyNetwork.sol
```diff
@@ -120,11 +120,14 @@ contract ColonyNetwork is ColonyNetworkStorage {
     IColony colony = IColony(etherRouter);
     colony.setToken(_tokenAddress);
 
+    // Creating new instance of colony's authority
     Authority authority = new Authority(colony);
+
     DSAuth dsauth = DSAuth(etherRouter);
     dsauth.setAuthority(authority);
-    authority.setRootUser(msg.sender, true);
```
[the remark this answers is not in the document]
filiplazovic: Because root user would be able to call `setResolver`.

contracts/Colony.sol
```diff
@@ -31,6 +32,22 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   // Version number should be upped with every change in Colony or its dependency contracts or libraries.
   function version() public pure returns (uint256) { return 1; }
 
+  function setOwnerRole(address _user) public auth {
```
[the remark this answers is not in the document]
elenadimitrova: I won't insist. If that's what they want and the discussion has been had with them that's fine.

contracts/Authority.sol
```diff
@@ -26,24 +26,52 @@ contract Authority is DSRoles {
   uint8 adminRole = 1;
```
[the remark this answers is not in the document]
elenadimitrova: I'm not sure if there's an issue at all here with `0` value roles. It's more of a suspicion

contracts/ColonyNetwork.sol
```diff
@@ -155,8 +158,8 @@ contract ColonyNetwork is ColonyNetworkStorage {
     address etherRouter = colonies[_id];
     // Check the calling user is authorised
     DSAuth auth = DSAuth(etherRouter);
-    DSAuthority authority = auth.authority();
-    require(authority.canCall(msg.sender, etherRouter, 0x0e1f20b4));
+    address authority = auth.authority();
+    require(Authority(authority).hasUserRole(msg.sender, 0));
```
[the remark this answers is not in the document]
filiplazovic: Because `DSAuthority` doesn't have `hasUserRole`, only `DSRoles`. And since we have `Authority`, i thougth i could reuse it, to avoid another import.

contracts/Authority.sol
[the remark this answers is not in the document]
area: There shouldn't be. The number used here determines which bit is used in the mask `DSRoles` uses to indicate that a user has a particular role. If we didn't use 0, we'd 'only' be able to have 255 roles rather than 256, and the LSB would never be used.

contracts/Colony.sol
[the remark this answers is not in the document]
area: Hm, this doesn't really solve the inconsistency issue if we need to double-sign from the manager.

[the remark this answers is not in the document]
filiplazovic: From the client it can be signed only by manager. There is a [check](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L170) in `executeTaskChange` that will solve this. Its a little hacky maybe, but it works.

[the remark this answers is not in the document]
area: Ah, I hadn't clocked that check would also catch this, but you're right, it does. Can we add an additional comment there? The comment currently only talks about both roles being assigned to the same user, which while it is sort-of what is happening here, is quite misleading.

area:
We got there eventually!

elenadimitrova:
Can we merge this?

PR #225, opened 2018-06-07
Prevent reputation overflows or underflows

area:
If a user fails a task, the reputation they lose could exceed the reputation they have. Their reputation is not meant to become negative, but instead become zero and no lower. This PR introduces this behaviour, where previously the client would attempt to mine a negative reputation and fail.

Similarly, if a user had a lot of reputation (or received a very large payout), their reputation could overflow. This caps their reputation at `2**256 - 1`, which is the largest value able to be represented in `uint256`.

### DIFF
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -633,9 +633,15 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
       u[U_REQUIRE_REPUTATION_CHECK] = 1;
     }
 
-    // TODO: Is this safe? I think so, because even if there's over/underflows, they should
-    // still be the same number.
-    require(int(agreeStateReputationValue)+reputationUpdateLog[reputationTransitionIdx].amount == int(disagreeStateReputationValue));
+    // We don't care about underflows for the purposes of comparison, but for the calculation we deem 'correct'.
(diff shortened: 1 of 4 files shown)


contracts/ReputationMiningCycle.sol
```diff
@@ -633,9 +633,15 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
...
-    // still be the same number.
-    require(int(agreeStateReputationValue)+reputationUpdateLog[reputationTransitionIdx].amount == int(disagreeStateReputationValue));
+    // We don't care about underflows for the purposes of comparison, but for the calculation we deem 'correct'.
+    // i.e. a reputation can't be negative.
+    if (reputationUpdateLog[reputationTransitionIdx].amount < 0 && uint(reputationUpdateLog[reputationTransitionIdx].amount * -1) > agreeStateReputationValue ) {
+      require(disagreeStateReputationValue == 0);
+    } else {
+      // TODO: Is this safe? I think so, because even if there's over/underflows, they should
+      // still be the same number.
+      require(int(agreeStateReputationValue)+reputationUpdateLog[reputationTransitionIdx].amount == int(disagreeStateReputationValue));
```
elenadimitrova: Can we use our `SafeMath` to ensure no overflows? Underflows can't be a problem.

PR #226, opened 2018-06-07
Upgrade truffle to 0.4.11

area:
As well as preparing us further for solidity 0.5.0, I believe this fixes the issues with Truffle being not very talkative when tests fail.


area:
I approve, but I made the PR so can't approve it on GH. Using superpowers to merge anyway.

PR #227, opened 2018-06-09
Add a few permissions and fix oracle server errors

area:
If there's going to be a public instance of the network anywhere, we need to make sure that people can't mess it up. This adds permissions to 

* `addGlobalSkill` (colony owner only, and only works from metaColony, so effectively only the metaColony's owner)
* `addDomain` (colony owner only)
* `setTaskRoleUser`, `setTaskDomain`, `setTaskSkill` are now restricted to the manager of the task.

I also took the opportunity to fix a couple of issues floating around in the reputation miner that I needed fixed.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -89,8 +89,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     token.transfer(colonyNetworkAddress, _wad);
   }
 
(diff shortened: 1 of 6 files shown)


area:
For our peace of mind, I would appreciate a *post hoc* review of these changes at some point, @elenadimitrova.

PR #228, opened 2018-06-11
Implement three-point ratings

### LINKED ISSUE #154: Move to three point grading scale.

kronosapiens:
Closes #154 

- Switch to 3-point scale with corresponding [-1, 1, 1.5] reputation boost.
- Evaluator receives Satisfactory rating by default.
- Scale implemented as enum, meaning `role.rated` boolean is no longer needed.
- **However, we repurpose the storage slot to `role.rateFail`**, set to `true` if the user does not rate their counterpart (set in `assignWorkRating`).
- Reputation calculation moved into a separate (internal, pure) function of `(payout, rating, rateFail)`.
- Update tests to specify ratings on the 1-3 point scale.
- Update tests to conform to usage of `TaskRatings` enum and `rateFail` boolean.
- Update tests with new rating and reputation values.

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -109,8 +109,8 @@ contract ColonyTask is ColonyStorage, DSMath {
   }
 
   modifier taskWorkRatingsAssigned(uint256 _id) {
-    require(tasks[_id].roles[WORKER].rated);
-    require(tasks[_id].roles[MANAGER].rated);
+    require(tasks[_id].roles[WORKER].rating != TaskRatings.None);
+    require(tasks[_id].roles[MANAGER].rating != TaskRatings.None);
     _;
   }
 
@@ -129,8 +129,8 @@ contract ColonyTask is ColonyStorage, DSMath {
     tasks[taskCount] = task;
     tasks[taskCount].roles[MANAGER] = Role({
       user: msg.sender,
-      rated: false,
-      rating: 0
+      rateFail: false,
+      rating: TaskRatings.None
     });
 
     pots[potCount].taskId = taskCount;
@@ -228,29 +228,29 @@ contract ColonyTask is ColonyStorage, DSMath {
     bytes32 ratingSecret = generateSecret(_salt, _rating);
     require(ratingSecret == taskWorkRatings[_id].secret[_role]);
 
-    Role storage role = tasks[_id].roles[_role];
-    role.rated = true;
-    role.rating = _rating;
+    TaskRatings rating = TaskRatings(_rating);
+    require(rating != TaskRatings.None, "Cannot rate None!");
+    tasks[_id].roles[_role].rating = rating;
   }
 
   // In the event of a user not committing or revealing within the 10 day rating window,
   // their rating of their counterpart is assumed to be the highest possible
-  // and their own rating is decreased by 5 (e.g. 0.5 points)
+  // and they will receive a reputation penalty
   function assignWorkRating(uint256 _id) public
   taskWorkRatingsClosed(_id)
   {
     Role storage managerRole = tasks[_id].roles[MANAGER];
     Role storage workerRole = tasks[_id].roles[WORKER];
+    Role storage evaluatorRole = tasks[_id].roles[EVALUATOR];
 
-    if (!workerRole.rated) {
-      workerRole.rated = true;
-      workerRole.rating = 50;
+    if (workerRole.rating == TaskRatings.None) {
+      evaluatorRole.rateFail = true;
+      workerRole.rating = TaskRatings.Excellent;
     }
 
-    if (!managerRole.rated) {
-      managerRole.rated = true;
-      managerRole.rating = 50;
-      workerRole.rating = (workerRole.rating > 5) ? (workerRole.rating - 5) : 0;
+    if (managerRole.rating == TaskRatings.None) {
+      workerRole.rateFail = true;
+      managerRole.rating = TaskRatings.Excellent;
     }
   }
 
@@ -275,8 +275,8 @@ contract ColonyTask is ColonyStorage, DSMath {
     require(tasks[_id].roles[MANAGER].user == msg.sender);
     tasks[_id].roles[_role] = Role({
       user: _user,
-      rated: false,
-      rating: 0
+      rateFail: false,
+      rating: TaskRatings.None
     });
 
     emit TaskRoleUserChanged(_id, _role, _user);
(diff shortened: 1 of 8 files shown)


test/colony-task-work-rating.js
```diff
@@ -222,63 +232,50 @@ contract("Colony Task Work Rating", () => {
...
-
-      const roleWorker = await colony.getTaskRole.call(taskId, WORKER_ROLE);
-      assert.isTrue(roleWorker[1]);
-      assert.equal(roleWorker[2].toNumber(), 0);
+      assert.equal(roleManager[2].toNumber(), 3);
 
-      const roleManager = await colony.getTaskRole.call(taskId, MANAGER_ROLE);
-      assert.isTrue(roleManager[1]);
-      assert.equal(roleManager[2].toNumber(), 50);
+      const roleEvaluator = await colony.getTaskRole.call(1, EVALUATOR_ROLE);
```
elenadimitrova: User `taskId` instead of hardcoding `1`

kronosapiens: 👍

contracts/IColony.sol
```diff
@@ -407,4 +408,9 @@ contract IColony {
   /// @param _token Address of the token, `0x0` value indicates Ether
   /// @return Total amount of tokens in pots other than the rewards pot (id 0)
   function getNonRewardPotsTotal(address _token) public view returns (uint256);
+
+  /// @notice Convert old- and new-style ratings to new-style ratings
+  /// @param rating Original rating
+  /// @return New-style rating
+  function convertRating(uint8 rating) public pure returns(uint8);
```
area: Does this function need to be public and/or on the interface?

contracts/ColonyTask.sol
```diff
@@ -228,29 +229,29 @@ contract ColonyTask is ColonyStorage, DSMath {
     bytes32 ratingSecret = generateSecret(_salt, _rating);
     require(ratingSecret == taskWorkRatings[_id].secret[_role]);
 
-    Role storage role = tasks[_id].roles[_role];
-    role.rated = true;
-    role.rating = _rating;
+    uint8 rating = convertRating(_rating);
+    require(rating != NONE, "Cannot rate None!");
```
area: With the `convertRating` function, this require can never not be true. Arguably worth keeping in to remind us when we remove the old-style ratings entirely though (and thus the convert function).

test/reputation-update.js
```diff
@@ -204,8 +167,7 @@ contract("Colony Reputation Updates", () => {
       });
       await metaColony.finalizeTask(taskId1);
       let repLogEntryWorker = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(3);
-      const result = web3Utils.toBN("1").mul(WORKER_PAYOUT);
-      assert.equal(repLogEntryWorker[1].toString(), result.toString());
+      assert.equal(repLogEntryWorker[1].toNumber(), 300 * 1e18);
```
area: Casting these to a number means it only works because we have a lot of 0s on the end. We would have a loss of precision if that were not true. I'm not going to ask you to change these back, but worth remembering that there is a reason we need to use BigNumber. The cast to string was so that assert had to primitives to compare (two different BigNumber objects are not the same, even if they represent the same value), and they are guaranteed to have whatever precision we need (i.e. infinite).

test/colony-task-work-rating.js
```diff
@@ -137,13 +148,16 @@ contract("Colony Task Work Rating", () => {
     it("should allow revealing a rating by evaluator and worker", async () => {
       await fundColonyWithTokens(colony, token, INITIAL_FUNDING);
       const taskId = await setupRatedTask({ colonyNetwork, colony, token });
+
       const roleManager = await colony.getTaskRole.call(taskId, MANAGER_ROLE);
-      assert.isTrue(roleManager[1]);
```
area: This wasn't worth an `assert.isFalse`, but the equivalent for `roleWorker` was? 😛

test/colony-task-work-rating.js
```diff
@@ -155,29 +169,34 @@ contract("Colony Task Work Rating", () => {
 
       await forwardTime(SECONDS_PER_DAY * 5 + 1, this);
       await colony.revealTaskWorkRating(1, WORKER_ROLE, WORKER_RATING, RATING_2_SALT, { from: EVALUATOR });
+
       const roleWorker = await colony.getTaskRole.call(taskId, WORKER_ROLE);
-      assert.isTrue(roleWorker[1]);
```
area: This could be changed to `assert.isFalse` too rather than just deleted. I'm giving the ones further down a pass, because we're checking extra things instead, but here we're just making the test less strict.

contracts/IColony.sol
kronosapiens: I made it public so I could test it; otherwise the code coverage build would fail.

test/colony-task-work-rating.js
kronosapiens: Oop, that should actually refer to `roleManager[3]`. The difference is that since the manager doesn't submit a rating, we don't need to check their `rateFail` boolean. We need to check the `roleWorker`'s because they submit a rating to the manager. But in any case this test is accessing the wrong field.

contracts/ColonyTask.sol
kronosapiens: Mm that's true, I can remove it in the backwards-compatibility commit (it'll come back when we roll back).

test/colony-task-work-rating.js
kronosapiens: I deleted it because `rated` is no longer part of the flow and IMO shouldn't be tested in tests outside of the ones which test for backwards compatibility; since this test is only concerned with the evaluator's evaluation (and doesn't deal with the worker's evaluation, whether submitted on time submitted late), there's no value of the `rateFail` boolean that we could meaningfully check.

I'm happy to add a test for this boolean in the backwards-compatibility test that you wrote.

test/reputation-update.js
kronosapiens: That's fair, I can make this a Big Number.

kronosapiens:
🎉🎉 🎉

PR #233, opened 2018-06-13
NetworkLoader for Colony contracts

chmanie:
This PR adds a ColonyJS compatible loader for the network contracts. In the future we expect the contracts to be compiled and then partly moved into this module automatically, whenever we deploy a new version. These artifacts will be checked into source control as they won't change for these versions anyways. This time I added the necessary artifacts manually. Later on we can automate this process.

There are two categories of contracts: static ones and versioned ones. The static ones will never be upgraded and hence can live in the static folder directly which is not versioned. Then there are the versioned contracts which lie in folders named [network]-v[version]. The version number starts at 1.

This sub-package can then be published to npm and people can install the needed ABIs conveniently.


packages/colony-contract-loader-network/src/NetworkLoader.js
```diff
@@ -0,0 +1,49 @@
...
+import config from "../config.json";
+
+const NETWORKS = ["rinkeby"];
+const LATEST_VERSION = config.LATEST_CONTRACT_VERSION;
+const VERSIONED_CONTRACTS = ["IColony", "IColonyNetwork", "Authority"];
+const STATIC_CONTRACTS = ["EtherRouter", "Token"];
+
+const CONTRACTS = {};
+NETWORKS.forEach(network => {
+  [...Array(LATEST_VERSION).keys()].forEach(version => {
```
JamesLefrere: This seems a little hard to understand. `LATEST_VERSION` is the string `1`, but `version` in this loop is `0` (and we add one to that to define a version for each network). 

Won't `LATEST_VERSION` always be just one entry? And aren't we ignoring it here?

PR #237, opened 2018-06-15
Fix README install instructions in networkloader.

gichiba:
Quick fix for a minor typo in the networkloader.


area:
The docs were right, the package name was wrong. Fixed by #241.

PR #238, opened 2018-06-15
Supply the selected network ID to the loader (`colony-network-contract-loader` package)

JamesLefrere:
This PR makes a small change to the loader so that the selected networkId (set in the loader instance) is supplied to the transform function. This makes it possible to find the address for the selected network automatically when loading the contract.

Ronseal

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,5 +5,4 @@ contracts/*
 coverage/*
 keys/*
 lib/*
-packages/reputation-miner/node_modules/*
-packages/colony-contract-loader-network/node_modules/*
+packages/**/node_modules
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,3 +6,4 @@ coverage/*
 keys/*
 lib/*
 packages/reputation-miner/node_modules/*
+packages/colony-contract-loader-network/node_modules/*
--- a/packages/colony-js-contract-loader-network/package.json
+++ b/packages/colony-js-contract-loader-network/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js-contract-loader-network",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Load the Colony contract ABIs directly from this package",
   "main": "lib/index.js",
   "scripts": {
--- a/packages/colony-js-contract-loader-network/package.json
+++ b/packages/colony-js-contract-loader-network/package.json
@@ -11,7 +11,7 @@
   "author": "Christian Maniewski <chris@colony.io>",
   "license": "GPL-3.0",
   "dependencies": {
-    "@colony/colony-js-contract-loader": "^1.3.0",
+    "@colony/colony-js-contract-loader": "^1.4.1",
     "assert": "^1.4.1",
     "babel-runtime": "^6.26.0"
   },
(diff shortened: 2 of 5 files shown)


packages/colony-contract-loader-network/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-contract-loader-network",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
JamesLefrere: The version is incremented here with the expectation that we'll publish it afterwards; if I could write straight to `develop`, I wouldn't add this here.

packages/colony-contract-loader-network/src/__tests__/NetworkLoader.test.js
```diff
@@ -18,6 +19,28 @@ describe("colony-contract-loader-network - NetworkLoader", () => {
...
+    expect(contract).toHaveProperty("abi", expect.any(Array));
+    expect(contract).toHaveProperty("address", EtherRouter.networks.rinkeby.address);
+  });
+
+  test("It should not load a versioned router contract that is not defined on the selected network", async () => {
+    const mainLoader = new NetworkLoader({ network: "main" });
+    try {
+      await mainLoader.load({
+        contractName: "IColonyNetwork",
+        routerName: "EtherRouter"
```
JamesLefrere: This is what happens when we get a new network client.

packages/colony-contract-loader-network/src/__tests__/NetworkLoader.test.js
```diff
@@ -18,6 +19,28 @@ describe("colony-contract-loader-network - NetworkLoader", () => {
     expect(contract).toHaveProperty("address", contractAddress);
   });
 
+  test("It should load a versioned router contract that is defined, without an address or version", async () => {
+    const contract = await loader.load({
+      contractName: "IColonyNetwork",
+      routerName: "EtherRouter"
```
JamesLefrere: This is what happens when we get a new network client.

.eslintignore
```diff
@@ -6,3 +6,4 @@ coverage/*
 keys/*
 lib/*
 packages/reputation-miner/node_modules/*
+packages/colony-contract-loader-network/node_modules/*
```
chmanie: It might be good to do a
packages/**/node_modules here

packages/colony-contract-loader-network/src/__tests__/NetworkLoader.test.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-env jest */
 
 import NetworkLoader from "../NetworkLoader";
+import EtherRouter from "../../contracts/static/EtherRouter.json";
```
chmanie: Does it makes sense to mock it here to be independent of the root?

.eslintignore
JamesLefrere: That is a better idea.

packages/colony-contract-loader-network/src/__tests__/NetworkLoader.test.js
JamesLefrere: I suppose so, yes 🤔

PR #253, opened 2018-06-28
Token Locking

### LINKED ISSUE #217: Token Locking

filiplazovic:
Closes #217 

I think we are familiar with locking functionality with counters. Additional changes include:
- Token lock counters will replace reward payout counters
- Waiving reward payouts will be done on token locking contract by increasing token lock count
- A colony can lock everyones tokens, except for addresses that have 0 deposited
- When address that have 0 deposited balance deposits the tokens, their token lock count will be set to total lock count

I know that tests for reward payouts need refactoring, it will be fixed in separate PR

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -59,14 +59,6 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
-  function updateTokenLocking() public auth {
-    tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
-  }
-
-  function getTokenLocking() public view returns (address) {
-    return tokenLocking;
-  }
-
   function initialiseColony(address _address) public {
     require(colonyNetworkAddress == 0x0, "colony-initialise-bad-address");
     colonyNetworkAddress = _address;
@@ -86,10 +78,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setRoleAssignmentFunction(bytes4(keccak256("setTaskEvaluatorRole(uint256,address)")));
     setRoleAssignmentFunction(bytes4(keccak256("setTaskWorkerRole(uint256,address)")));
 
-    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
-    tokenLocking = ITokenLocking(colonyNetwork.getTokenLocking());
-    uint256 rootLocalSkill = colonyNetwork.getSkillCount();
     // Initialise the root domain
+    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    uint256 rootLocalSkill = colonyNetwork.getSkillCount();
     initialiseDomain(rootLocalSkill);
   }
 
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -59,6 +59,14 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
+  function updateTokenLocking() public auth {
+    tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+  }
+
+  function getTokenLocking() public view returns (address) {
+    return tokenLocking;
+  }
+
   function initialiseColony(address _address) public {
     require(colonyNetworkAddress == 0x0, "colony-initialise-bad-address");
     colonyNetworkAddress = _address;
@@ -78,9 +86,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setRoleAssignmentFunction(bytes4(keccak256("setTaskEvaluatorRole(uint256,address)")));
     setRoleAssignmentFunction(bytes4(keccak256("setTaskWorkerRole(uint256,address)")));
 
-    // Initialise the root domain
     IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    tokenLocking = ITokenLocking(colonyNetwork.getTokenLocking());
     uint256 rootLocalSkill = colonyNetwork.getSkillCount();
+    // Initialise the root domain
     initialiseDomain(rootLocalSkill);
   }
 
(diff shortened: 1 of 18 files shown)


contracts/ColonyFunding.sol
```diff
@@ -166,44 +167,38 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
-  function getGlobalRewardPayoutCount() public view returns (uint256) {
-    return globalRewardPayoutCount;
-  }
-
-  function getUserRewardPayoutCount(address _user) public view returns (uint256) {
-    return userRewardPayoutCount[_user];
-  }
-
   function startNextRewardPayout(address _token) public auth {
+    ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
```
filiplazovic: I guess whoever controls the colony network will be able to change token locking address without colony knowing, so i will change that

contracts/Authority.sol
```diff
@@ -26,55 +26,50 @@ contract Authority is DSRoles {
...
+    setAdminRoleCapability(colony, "setAdminRole(address)");
+    setOwnerRoleCapability(colony, "setAdminRole(address)");
+  }
+
+  function setOwnerRoleCapability(address colony, bytes sig) private {
+    bytes4 functionSig = bytes4(keccak256(sig));
+    setRoleCapability(ownerRole, colony, functionSig, true);
+  }
+
+  function setAdminRoleCapability(address colony, bytes sig) private {
```
elenadimitrova: Sorry what does this have to do with token locking?

contracts/TokenLocking.sol
```diff
@@ -0,0 +1,80 @@
...
+
+  function getTotalLockCount(address _token) public view returns (uint256) {
+    return totalLockCount[_token];
+  }
+
+  function getUserLockCount(address _token, address _user) public view returns (uint256) {
+    return userLocks[_token][_user].lockCount;
+  }
+
+  function getUserDepositedBalance(address _token, address _user) public view returns (uint256) {
```
elenadimitrova: Can we merge these two functions to return the `Lock` struct?

contracts/Colony.sol
```diff
@@ -59,6 +59,14 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
+  function updateTokenLocking() public auth {
+    tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
+  }
```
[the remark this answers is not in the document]
filiplazovic: Upgrading token locking contract is one thing, but i have another concern.
Colony network owners are able to add function on `ColonyNetwork` contract that can completely change address of token locking contract

contracts/TokenLocking.sol
filiplazovic: Right now these two functions are not used anywhere together, so i would like it to stay the as it is.

contracts/ColonyFunding.sol
```diff
@@ -166,44 +167,37 @@ contract ColonyFunding is ColonyStorage, DSMath {
...
 
-    emit RewardPayoutCycleStarted(globalRewardPayoutCount);
+    emit RewardPayoutCycleStarted(totalLockCount);
   }
 
   function claimRewardPayout(uint256 _payoutId, uint256[7] _squareRoots, uint256 _userReputation, uint256 _totalReputation) public {
     RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
+    // Checking if payout is active
     require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
-    require(_payoutId - userRewardPayoutCount[msg.sender] == 1, "colony-reward-payout-bad-id");
```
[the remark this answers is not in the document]
filiplazovic: No, will add it.
Since we only allow claiming next in 'queue', we could also not pass `_payoutId` and just take current users token lock count and add `1`. But i guess its not significantly different that this

filiplazovic:
I will log an issue and replace staking implementation with this in another PR.
Authority contract changes will be moved to another branch, since we will have `stack too deep` error as soon as we add one more variable.

test/colony-funding.js
```diff
@@ -842,6 +826,158 @@ contract("Colony Funding", addresses => {
...
+      ({ logs } = await colony2.startNextRewardPayout(otherToken.address));
+      const payoutId2 = logs[0].args.id;
+
+      await colony1.claimRewardPayout(payoutId1.toString(), initialSquareRoots1, userReputation1.toString(), totalReputation.toString(), {
+        from: userAddress1
+      });
+
+      await colony2.claimRewardPayout(payoutId2.toString(), initialSquareRoots1, userReputation1.toString(), totalReputation.toString(), {
+        from: userAddress1
+      });
```
elenadimitrova: Can we check the user balance to confirm success?

test/colony-funding.js
```diff
@@ -842,6 +826,158 @@ contract("Colony Funding", addresses => {
...
+      ({ colonyAddress } = logs[0].args);
+      const colony2 = IColony.at(colonyAddress);
+
+      // Giving both colonies the capability to call `mint` function
+      const adminRole = 1;
+      const newRoles = await DSRoles.new();
+      await newRoles.setUserRole(colony1.address, adminRole, true);
+      await newRoles.setUserRole(colony2.address, adminRole, true);
+      await newRoles.setRoleCapability(adminRole, newToken.address, sha3("mint(uint256)").slice(0, 10), true);
+      await newToken.setAuthority(newRoles.address);
```
elenadimitrova: Can't we use `await newToken.setOwner(colony1 .address);` to simplify the above?

filiplazovic: To replace it with all of this above? Both `colony1` and `colony2` need to have permission to call `mint` function on the token contract

elenadimitrova: ah yes, there can only be one owner..

PR #256, opened 2018-07-01
Fix denominator constraint

filiplazovic:
Fixed a bug where user was able to pass denominator smaller than the correct value, therefore maximizing his reward and draining reward pot.

### DIFF
--- a/gasCosts/gasCosts.js
+++ b/gasCosts/gasCosts.js
@@ -357,7 +357,7 @@ contract("All", accounts => {
       });
 
       const workerReputationSqrt = bnSqrt(workerReputation);
-      const totalReputationSqrt = bnSqrt(workerReputation.add(managerReputation), true);
+      const totalReputationSqrt = bnSqrt(totalReputation, true);
       const numeratorSqrt = bnSqrt(workerReputationSqrt.mul(workerReputationSqrt));
       const denominatorSqrt = bnSqrt(totalReputationSqrt.mul(totalReputationSqrt), true);
 
--- a/gasCosts/gasCosts.js
+++ b/gasCosts/gasCosts.js
@@ -325,7 +325,7 @@ contract("All", accounts => {
     });
 
     it("when working with reward payouts", async () => {
-      const totalReputation = toBN(350 * 1e18);
+      const totalReputation = toBN(300 * 1e18);
       const workerReputation = toBN(200 * 1e18);
       const managerReputation = toBN(100 * 1e18);
       const initialFunding = toBN(360 * 1e18);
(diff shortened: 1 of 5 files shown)


gasCosts/gasCosts.js
```diff
@@ -313,7 +313,7 @@ contract("All", accounts => {
     });
 
     it("when working with reward payouts", async () => {
-      const totalReputation = toBN(350 * 1e18);
+      const totalReputation = toBN(300 * 1e18);
```
area: You've changed this to be accurate, but I don't think this variable is used anywhere in the test?

helpers/test-helper.js
```diff
@@ -253,5 +253,9 @@ export function bnSqrt(bn) {
       .add(a)
       .div(web3Utils.toBN(2));
   }
+  // Do we need while loop? Not sure how precise the formula above is
```
area: I don't believe the loop is needed (though obviously the addition is). I believe you could also replace the addition with `b = b.addn(1)`.

test/colony-funding.js
```diff
@@ -812,24 +799,28 @@ contract("Colony Funding", addresses => {
         "colony-reward-payout-invalid-parametar-user-token",
         "colony-reward-payout-invalid-parametar-total-reputation",
         "colony-reward-payout-invalid-parametar-total-tokens",
-        "colony-reward-payout-invalid-parametar-amount",
         "colony-reward-payout-invalid-parametar-numerator",
-        "colony-reward-payout-invalid-parametar-denominator"
+        "colony-reward-payout-invalid-parametar-denominator",
```
area: It's spelled `parameter` 😄

test/colony-funding.js
```diff
@@ -842,28 +833,72 @@ contract("Colony Funding", addresses => {
...
+
+    it("should return correct info about reward payout", async () => {
+      const { logs } = await colony.startNextRewardPayout(otherToken.address);
+      const payoutId = logs[0].args.id;
+
+      const balance = await colony.getPotBalance(0, otherToken.address);
+      const blockTimestamp = await currentBlockTime();
+
+      const info = await colony.getRewardPayoutInfo(payoutId);
+      assert.equal(info[1].toString(), userTokens.toString());
```
area: Is there a reason we're not checking `info[0]`? We could compare it to `colonyNetwork.getReputationRootHash()`, right? (Although without going through all the mining setup, I suppose that's going to just be `0x0`, which might be the reason you've skipped it?)

filiplazovic: Yea, i thought i can add it when we add proofs (which will require mining process), but i guess we can do it now

gasCosts/gasCosts.js
filiplazovic: We are passing it to `claimRewardPayout`

PR #265, opened 2018-07-05
Compile warnings

filiplazovic:
Changes include:
- Refactored Authority contract to avoid stack depth error
- Removed unused imports
- Fixed compile and solium warnings such as:
  - line length
  - formating of more than 5 function arguments
  - unused function arguments
  - etc.

There are solium warnings for experimental features, im not sure if its safe to remove these.
Solc warnings for external libraries will unfortunately stay: https://github.com/ethereum/solidity/issues/2675

### DIFF
--- a/contracts/PatriciaTree/PatriciaTreeProofs.sol
+++ b/contracts/PatriciaTree/PatriciaTreeProofs.sol
@@ -14,8 +14,7 @@ contract PatriciaTreeProofs {
   using Data for Data.Label;
 
   function getImpliedRoot(bytes key, bytes value, uint branchMask, bytes32[] siblings) public // solium-disable-line security/no-assign-params
-  view
-  returns (bytes32)
+  pure returns (bytes32)
   {
     Data.Label memory k = Data.Label(keccak256(key), 256);
     Data.Edge memory e;
--- a/contracts/PatriciaTree/PatriciaTreeProofs.sol
+++ b/contracts/PatriciaTree/PatriciaTreeProofs.sol
@@ -5,6 +5,7 @@ pragma experimental "ABIEncoderV2";
 import {Data} from "./Data.sol";
 import {Bits} from "./Bits.sol";
 
+
 /// @title Functions related to checking Patricia Tree proofs
 /// @notice More info at: https://github.com/chriseth/patricia-trie
 contract PatriciaTreeProofs {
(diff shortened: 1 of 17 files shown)


contracts/PatriciaTree/PatriciaTreeProofs.sol
```diff
@@ -5,14 +5,18 @@ pragma experimental "ABIEncoderV2";
...
 /// @notice More info at: https://github.com/chriseth/patricia-trie
 contract PatriciaTreeProofs {
   using Bits for uint;
   using Data for Data.Edge;
   using Data for Data.Label;
 
-  function getImpliedRoot(bytes key, bytes value, uint branchMask, bytes32[] siblings) public view returns (bytes32) { // solium-disable-line security/no-assign-params
+  function getImpliedRoot(bytes key, bytes value, uint branchMask, bytes32[] siblings) public // solium-disable-line security/no-assign-params
+  view
+  returns (bytes32)
```
elenadimitrova: I understand we reach the line limit above but can we at least keep `view returns (bytes32)` on the same line?

contracts/ReputationMiningCycle.sol
```diff
@@ -555,7 +577,11 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
     }
   }
 
-  function getExpectedSkillIdAndAddress( ReputationLogEntry storage logEntry, uint256 updateNumber ) internal returns (uint256 expectedSkillId, address expectedAddress) {
+  function getExpectedSkillIdAndAddress(ReputationLogEntry storage logEntry, uint256 updateNumber)
+  internal
+  view
```
elenadimitrova: Put `internal view` at the end of the function declaration line above. That's what we've followed everywhere else I think. The `returns..` can be on its own line to stand out

elenadimitrova:
Sorry, forgot to mention in review, to check if we can disable `no-experimental` in `solium` checks to avoid these warnings

filiplazovic:
I prefer suppressing these warnings by fixing them. In case where that is not possible i would leave them. But i guess we can supress them and be more cautious, no strong feelings

elenadimitrova:
You are not sufficiently informed when you say this, no hard feelings :)
To understand whether we need `experimental pragmas` you need to understand what they do, so read the documentation first before you decide to "fix" them. I want you to come back and tell me what each of the two experimental pragmas does and whether and why we need it.

Then check https://github.com/duaraghav8/Solium/issues/201 where we've agreed this is a bad warning anyway that will be fixed in upcoming `solium`

filiplazovic:
I agree that we should keep `experimental pragmas`, but my comment was about whether we should supress warnings (any warning, not just about `experimental pragmas`) by turning off the rule in config file.

You are right that i could do a research of what benefits of these experimental features give us. Looking from solidity releases i can see that `ABIEncoderV2` allows us to use structs and arbitrary length arrays as functions arguments and return values. And `v0.5.0` will disallow uninitialized storage pointers, old constructor sintax, `var` keywork etc.

elenadimitrova:
Yes, I was not arguing that we disable the rule to simply silence the warning. I was saying we should silence the rule because 
a) we need those experimental pragmas so we can't remove like solium is asking 
b) that solium rule is just wrong and will be removed in the next release anyway

PR #268, opened 2018-07-05
Improve adding domains

### LINKED ISSUE #266: Local skills tree not enforced to be colony specific
When adding a colony domain via a call to `function addDomain(uint256 _parentSkillId)` it is possible to add a new local skill under another colony's local skill. Therefore we should switch this to accept the `parentDomainId` instead to ensure the local skills tree is colony specific. It is also more intuitive.

elenadimitrova:
Closes #266 

Also fixes a test issue using invalid skill id https://github.com/JoinColony/colonyNetwork/blob/develop/test/meta-colony.js#L372

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -118,6 +118,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   }
 
   function addDomain(uint256 _parentDomainId) public
+  domainExists(_parentDomainId)
   auth
   {
     // Note: Remove when we want to allow more domain hierarchy levels
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -117,19 +117,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return colonyNetwork.addSkill(_parentSkillId, true);
   }
 
-  function addDomain(uint256 _parentSkillId) public
+  function addDomain(uint256 _parentDomainId) public
   auth
-  localSkill(_parentSkillId)
   {
-    // Note: remove that when we start allowing more domain hierarchy levels
-    // Instead check that the parent skill id belongs to this colony own domain
-    // Get the local skill id of the root domain
-    uint256 rootDomainSkillId = domains[1].skillId;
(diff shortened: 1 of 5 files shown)


contracts/Colony.sol
```diff
@@ -119,19 +119,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return colonyNetwork.addSkill(_parentSkillId, true);
   }
 
-  function addDomain(uint256 _parentSkillId) public
+  function addDomain(uint256 _parentDomainId) public
   auth
-  localSkill(_parentSkillId)
```
area: We don't need it now because of the `require(_parentDomainId == 1, "colony-parent-skill-not-root");`, but worth throwing in the `domainExists` modifier here now to save us forgetting to in the future?

test/meta-colony.js
```diff
@@ -343,25 +343,6 @@ contract("Meta Colony", accounts => {
       assert.equal(rootSkillChild3.toNumber(), 6);
     });
 
-    it("should NOT be able to add a child local skill more than one level from the root local skill", async () => {
```
area: Why delete these tests?

contracts/Colony.sol
elenadimitrova: Since this wasn't a breaking change for us in the future I didn't think to include it but it doesn't hurt to be future-proof.

test/meta-colony.js
elenadimitrova: - `should NOT be able to add a child local skill more than one level from the root local skill` is already covered by `should NOT be able to add a child domain more than one level away from the root domain`
- `should NOT be able to add a child local skill to a global skill parent` - is no longer relevant. Passing in a domain id in and creating the local skills within the same call ensures we cannot add a domain against a global skill

elenadimitrova: I was actually tempted to merge the `when adding domains in the meta colony` and `when adding domains in a regular colony` as there is nothing special in the meta colony setup as far as domains are concerned. What do you think?

PR #270, opened 2018-07-06
Implement recovery mode

### LINKED ISSUE #18: Recovery mode for contracts
### 2.2 Contract security

At launch, colonies will be able to be put into a ‘recovery mode’. In this state, whitelisted
addresses are able to access functions that allow the state of the contract to be directly edited — in
practise, this will correspond to access to the functions to allow setting of variables, as well as being
able to upgrade the contract. With the agreement of multiple whitelisted addresses, the contract
will then be able to be taken out of recovery mode once the contract has been returned to a rational
state. Removal from recovery mode requires the approval of multiple whitelisted addresses.

Any approvals from whitelisted addresses to leave recovery mode must be reset whenever a
variable is edited. A whitelisted address agreeing to leave recovery mode records the timestamp at
which the agreement occurred, and any change of variables also update a timestamp indicating the
last edit. When attempting to leave recovery mode, only agreements made after the last edit are
counted towards meeting the threshold.

The first whitelisted address is added at colony creation and is the creator of the colony.
Whitelisted addresses can be added or removed by a simple majority vote of existing whitelisted
addresses.

kronosapiens:
Closes #18

WIP

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -49,6 +49,22 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     Authority(authority).setUserRole(_user, ADMIN_ROLE, false);
   }
 
+  // Can only be called by the owner role.
+  function setRecoveryRole(address _user) public auth {
+    if (!Authority(authority).hasUserRole(_user, RECOVERY_ROLE)) {
+      recoveryWhitelistCount++;
+    }
+    Authority(authority).setUserRole(_user, RECOVERY_ROLE, true);
+  }
+
+  // Can only be called by the owner role.
+  function removeRecoveryRole(address _user) public auth {
+    if (Authority(authority).hasUserRole(_user, RECOVERY_ROLE)) {
+      recoveryWhitelistCount--;
+    }
+    Authority(authority).setUserRole(_user, RECOVERY_ROLE, false);
+  }
+
   function setToken(address _token) public
   auth
   {
@@ -208,5 +224,31 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
     emit DomainAdded(domainCount);
     emit PotAdded(potCount);
+  }
+
+  function enterRecoveryMode() public auth {
+    recoveryMode = true;
+  }
+
+  function setStorageSlotRecovery(uint256 _slot, bytes32 _value) public inRecoveryMode auth {
+    recoveryApprovalCount = 0;
+    recoveryEditedTimestamp = now;
+    uint x = _slot;
+    bytes32 y = _value;
+    assembly {
+      sstore(x, y)
+    }
+  }
+
+  function approveExitRecovery() public inRecoveryMode auth {
+    require(recoveryApprovalTimestamps[msg.sender] < recoveryEditedTimestamp, "recovery-approval-already-given");
+    recoveryApprovalTimestamps[msg.sender] = now;
+    recoveryApprovalCount++;
+  }
+
+  function exitRecoveryMode() public inRecoveryMode {
+    uint numRequired = (recoveryWhitelistCount + 1) / 2 + 1;
+    require(recoveryApprovalCount >= numRequired, "recovery-exit-insufficient-approvals");
+    recoveryMode = false;
   }
 }
--- a/contracts/ColonyStorage.sol
+++ b/contracts/ColonyStorage.sol
@@ -38,7 +38,7 @@ contract ColonyStorage is DSAuth {
 
   // Mapping function signature to 2 task roles whose approval is needed to execute
   mapping (bytes4 => uint8[2]) reviewers;
-  
+
   // Role assignment functions require special type of sign-off.
   // This keeps track of which functions are related to role assignment
   mapping (bytes4 => bool) roleAssignmentSigs;
@@ -84,9 +84,17 @@ contract ColonyStorage is DSAuth {
 
   uint8 constant OWNER_ROLE = 0;
   uint8 constant ADMIN_ROLE = 1;
+  uint8 constant RECOVERY_ROLE = 2;
   uint8 constant MANAGER = 0;
   uint8 constant EVALUATOR = 1;
   uint8 constant WORKER = 2;
+
+  // Variables for recovery mode
+  bool recoveryMode;
+  uint64 recoveryWhitelistCount;
+  uint64 recoveryApprovalCount;
+  uint256 recoveryEditedTimestamp;
+  mapping (address => uint256) recoveryApprovalTimestamps;
 
   // Mapping task id to current "active" nonce for executing task changes
   mapping (uint256 => uint256) taskChangeNonces;
@@ -189,6 +197,12 @@ contract ColonyStorage is DSAuth {
     require(Authority(authority).hasUserRole(_user, ADMIN_ROLE));
     _;
   }
+
+  modifier inRecoveryMode() {
+    require(recoveryMode, "not-in-recovery");
+    _;
+  }
+
 
   modifier self() {
     require(address(this) == msg.sender);
--- a/contracts/Authority.sol
+++ b/contracts/Authority.sol
@@ -24,8 +24,15 @@ import "../lib/dappsys/roles.sol";
 contract Authority is DSRoles {
   uint8 ownerRole = 0;
   uint8 adminRole = 1;
+  uint8 recoveryRole = 2;
 
   constructor(address colony) public {
+    ownerRoles(colony);
+    ownerAndAdminRoles(colony);
+    ownerAndRecoveryRoles(colony);
+  }
+
+  function ownerRoles(address colony) public {
     // functions for owner role
     bytes4 setTokenSig = bytes4(keccak256("setToken(address)"));
     bytes4 bootstrapColonySig = bytes4(keccak256("bootstrapColony(address[],int256[])"));
@@ -33,15 +40,9 @@ contract Authority is DSRoles {
     bytes4 addGlobalSkillSig = bytes4(keccak256("addGlobalSkill(uint256)"));
     bytes4 setOwnerRoleSig = bytes4(keccak256("setOwnerRole(address)"));
     bytes4 removeAdminRoleSig = bytes4(keccak256("removeAdminRole(address)"));
+    bytes4 setRecoveryRoleSig = bytes4(keccak256("setRecoveryRole(address)"));
+    bytes4 removeRecoveryRoleSig = bytes4(keccak256("removeRecoveryRole(address)"));
     bytes4 upgradeSig = bytes4(keccak256("upgrade(uint256)"));
-
-    // functions for admin + owner role
-    bytes4 moveFundsBetweenPotsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,address)"));
-    bytes4 addDomainSig = bytes4(keccak256("addDomain(uint256)"));
-    bytes4 makeTaskSig = bytes4(keccak256("makeTask(bytes32,uint256)"));
-    bytes4 startNextRewardPayoutSig = bytes4(keccak256("startNextRewardPayout(address)"));
-    bytes4 cancelTaskSig = bytes4(keccak256("cancelTask(uint256)"));
-    bytes4 setAdminRoleSig = bytes4(keccak256("setAdminRole(address)"));
 
     // Set token
     setRoleCapability(ownerRole, colony, setTokenSig, true);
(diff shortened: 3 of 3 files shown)


test/colony.js
```diff
@@ -1720,4 +1721,88 @@ contract("Colony", addresses => {
...
+      const unprotected1 = await web3GetStorageAt(colony.address, 5);
+      assert.notEqual(unprotected0.toString(), unprotected1.toString());
+    });
+
+    it("should not allow editing of protected variables", async () => {
+      const owner = addresses[0];
+      const protectedLoc = 0;
+
+      await colony.enterRecoveryMode({ from: owner });
+      await checkErrorRevert(colony.setStorageSlotRecovery(protectedLoc, fromAscii("0xdeadbeef"), { from: owner }));
```
area: Strictly, it doesn't need correcting, but it's weird you're using `fromAscii` on `0xdeadbeef`, which means it stores the bytes for the ASCII representation of the string `0xdeadbeef` rather than the bytes `0xdeadbeef`.

test/colony.js
```diff
@@ -1720,4 +1721,88 @@ contract("Colony", addresses => {
...
+    it("should allow editing of general variables", async () => {
+      const owner = addresses[0];
+
+      await colony.enterRecoveryMode({ from: owner });
+
+      const unprotected0 = await web3GetStorageAt(colony.address, 5);
+      await colony.setStorageSlotRecovery(5, fromAscii("0xdeadbeef"), { from: owner });
+
+      const unprotected1 = await web3GetStorageAt(colony.address, 5);
+      assert.notEqual(unprotected0.toString(), unprotected1.toString());
```
area: Should check that it's actually been set correctly!

test/colony.js
```diff
@@ -1720,4 +1721,88 @@ contract("Colony", addresses => {
       await checkErrorRevert(colony.claimPayout(taskId, MANAGER_ROLE, token.address, { from: OTHER }));
     });
   });
+
+  describe("when in recovery mode", () => {
+    it("should be able to add and remove recovery roles", async () => {
```
area: Allowing this feels like a bad idea...

test/colony.js
kronosapiens: That's fair. I was using `fromAscii` before to pass in `bytes32` and just kept using it. I can make it another string so it's not confusing.

test/colony.js
kronosapiens: Reasonable.

test/colony.js
kronosapiens: Mm that's a good point. I can change it so you can't add recovery roles in recovery mode.

kronosapiens:
My thought is: why disable other functionality? Given that someone in recovery mode can edit all values, they could in principle make any changes manually that could be made via the normal colony functionality. Unless the point is that we don't want non-whitelisted users to be mutating any values while the recovery is ongoing.

elenadimitrova:
>Unless the point is that we don't want non-whitelisted users to be mutating any values while the recovery is ongoing

That is exactly the point I think. If a function has been exploited (intentionally or not) we want it to stop being available until we roll out a fix and amend any bad state on chain.

To @area 's suggestion, I can't think of issues with it, pending when a fix is rolled out to a contract that gets redeployed, references in old/real Resolver need to be updated too. 

Another option can be to use `DSRoles.setPublicCapability` and `auth` every function, thus ensuring its security is both explicitly defined and also can be changed when needed. Admittedly this is better for finer granularity of recoveries and can be used _in addition to_ the above approach for removing access for example to `createColony` when a Colony Resolver is being swapped out, to prevent further colonies from creating with the potential exploit/bug.

kronosapiens:
In discussion with @area we worked through the details of the afore-mentioned proposals, but ran into some blockers:

1. The multiple-resolver approach allowed no easy way to continue to allow access to `public view` functions without necessitating keeping a mirror of all view functions in the recovery resolver (thus requiring an update to the recovery resolver for every general resolver update).

2. The `setPublicCapability` approach would involve a high gas overhead, with 20k gas per function being needed to set `publicCapability = true`.

After this discussion, we concluded that a simple modifier-based approach (`normal` vs `recovery`) would be the cheapest and least-unmaintainable way to enable/disable functionality in recovery mode.

contracts/ColonyStorage.sol
```diff
@@ -185,6 +193,16 @@ contract ColonyStorage is DSAuth, DSMath {
...
 
+  modifier recovery() {
+    require(recoveryMode, "not-in-recovery-mode");
+    _;
+  }
+
+  modifier normal() {
+    require(!recoveryMode, "in-recovery-mode");
+    _;
+  }
```
elenadimitrova: I should have probably remembered to mention this back when you commented on how you've agreed to do it but it just dawned on me this type of logic is already available in the dappsys lib we might be able to reuse https://github.com/dapphub/dappsys-monolithic/blob/002389d43cf54e8f0b919fee1fc364b20ebdf626/stop.sol#L23

contracts/IColony.sol
```diff
@@ -134,6 +134,19 @@ contract IColony {
...
+  /// @param _user User we want to give a recovery role to
+  function setRecoveryRole(address _user) public;
+
+  /// @notice Remove colony recovery role.
+  /// Can only be called by owner role.
+  /// @param _user User we want to remove recovery role from
+  function removeRecoveryRole(address _user) public;
+
+  /// @notice Return number of recovery roles.
+  function numRecoveryRoles() public view returns(uint64);
```
elenadimitrova: Add return param natspec

contracts/IColony.sol
```diff
@@ -193,6 +206,25 @@ contract IColony {
   /// is not currently exposed on the Colony's EtherRouter.
   function verifyReputationProof(bytes key, bytes value, uint256 branchMask, bytes32[] siblings) public view returns (bool isValid);
 
+  /// @notice Put colony into recovery mode.
+  /// Can only be called by user with recovery role.
+  function enterRecoveryMode() public;
+
+  /// @notice Update value of arbitrary storage variable.
+  /// Can only be called by user with recovery role.
+  /// @param _slot Uint address of storage word to be updated
```
elenadimitrova: storage word -> storage slot?

kronosapiens:
@elenadimitrova re:

1. Storage slot helper -- do you know if Solidity provides any reflection capabilities? From the assembly example in the link, you can access the slot of variable `b` as `b_slot`, but what would you do if someone passes the string `"b"`? I could imagine writing a per-variable helper which hard-codes the `x_slot` reference, but that would be a lot of boilerplate.

2. For the `stopped` modifier, the only difference in our case is that recovery mode switches off some functions and switches on others, so we would need to add a `notStopped` modifier to the recovery-specific functions. I also think the semantics would be slightly less clear than `normal` / `recovery`, which might create challenges for future maintainers.

Everything else you suggested sounds great.

contracts/IColony.sol
kronosapiens: 👍

elenadimitrova:
@kronosapiens continuing onto your points from above:
1- Storage slot helper - it appears there is an `eth_getStorageAt` method https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getstorageat which allows us to read that client side so I guess we're fine. I just didn't know about it and guessed we'd have to have something in the contract to improve our experience with storage slot updates
2-`stopped` modifier reuse - we could call the new modifier `recovery` and use that with `stoppable`? I leave that decision to you but if we stay on the current implementation can we perhaps think of a better name than `normal`?

kronosapiens:
@elenadimitrova we can definitely find a new name; since it appears a lot I was hoping to find something succinct and visually unobtrusive (as opposed to something more verbose like `notInRecovery`).

In contrast with `recovery`, we could have something like `normal`, `regular`, `healthy`, `basic`. Do any of those jump out at you?

Also, the function you describe will pull out the 256-bit word at the storage slot you give; it wont necessarily tell you what storage variable points to that slot. I was thinking, maybe it could be possible to write a script which parses `colonyStorage` and outputs a JSON of all the storage variables and their slots and offsets? We could make that available for the client to reference.

elenadimitrova:
@kronosapiens I did spend some time thinking about the right name. It has to be right because it's everywhere! Unfortunately I couldn't find anything better than `stoppable` :/ maybe `pausable` but that is terrible to pronounce.. I guess I'm still leaning towards my last proposal above which is `stoppable`\`recovery` combo. We have to indicate in meaning that the function can be temporarily stopped/paused/disabled to enable some recovery of a bad state. 

Regarding retrieval of storage slots, indeed what I offered was a way to get the value from a given slot, rather than the slot number for a given parameter. From what I know, the latter can only really be done at assembly level in the contract but the former gives us at least some way of confirming a slot contains what we expect. We won't avoid counting storage slot numbers in the `*Storage` contracts though but at least we can do some checks before we overwrite slots in storage.

kronosapiens:
Ok, I'm happy to change `normal` to `stoppable`. I don't see a benefit to using the `stop.sol` contract in particular, since we have custom enter/exit logic for recovery mode.

For `We have to indicate in meaning that the function can be temporarily stopped/paused/disabled to enable some recovery of a bad state.`, do you mean adding a natspec comment to every `stoppable` function, or something else?

elenadimitrova:
The storage contracts themselves are kind of a list themselves though. Or is there additional information we can get?

kronosapiens:
I was thinking more for a recovery mode UI, which could read this JSON file and maybe render a list of variables, plus a button to query their current values, and an update form. Users could in principle figure this out from source contracts, but it would be less error-prone to figure it out for them.

area:
At least some of that calculation is going to have to be done dynamically by the client - the storage slots of interest for a mapping that uses addresses as the key depend on the address, for example.

I am very much in favour of a script that checks that all (public) functions are labelled as `stoppable` or `onlyRecovery` (pending whatever we decides as names for them).

kronosapiens:
Ok, so we'll leave that for later.

Do u mean dis one? 👇 
https://github.com/JoinColony/colonyNetwork/blob/7234fbdf05d5444d4c360e5e9f73cb615e8ee9e7/scripts/check-recovery.js

kronosapiens:
@elenadimitrova I updated the modifier names, let me know if there's anything else you'd like me to revisit.

elenadimitrova:
I just the notes from previous review @kronosapiens If those are resolved, then mark it ready-for-review and I'll come back to it for another thorough check today or tomorrow

contracts/Colony.sol
```diff
@@ -29,23 +29,44 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
 
+  // Can only be called by the owner role.
+  function setRecoveryRole(address _user) public stoppable auth {
+    if (!Authority(authority).hasUserRole(_user, RECOVERY_ROLE)) {
+      Authority(authority).setUserRole(_user, RECOVERY_ROLE, true);
+      recoveryRolesCount = add(recoveryRolesCount, 1);
+    }
+  }
+
+  // Can only be called by the owner role. Must always be at least one user with recovery role.
```
area: This must is no longer true, right?

contracts/ColonyNetwork.sol
```diff
@@ -135,6 +135,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
 
     authority.setOwner(etherRouter);
     colony.setOwnerRole(msg.sender);
+    colony.setRecoveryRole(msg.sender);
```
area: Yeah, this is nice, but there's not *actually* any reason to do this here, is there? The owner could do this further down the road for themselves, once they've been educated as to what it means / does.

contracts/ColonyStorage.sol
```diff
@@ -85,10 +85,18 @@ contract ColonyStorage is DSAuth, DSMath {
...
   uint8 constant OWNER_ROLE = 0;
   uint8 constant ADMIN_ROLE = 1;
+  uint8 constant RECOVERY_ROLE = 2;
   uint8 constant MANAGER = 0;
   uint8 constant EVALUATOR = 1;
   uint8 constant WORKER = 2;
 
+  // Variables for recovery mode
+  bool recoveryMode;
+  uint64 recoveryRolesCount;
```
area: Why `uint64`? I think it's just more expensive than `uint256` here, no?

area: I am being outvoted on whether we should be optimising for recovery mode or the Magical Christmasland where we don't need to use it, so this can stand for now 😄

contracts/Colony.sol
kronosapiens: Correct

contracts/ColonyNetwork.sol
kronosapiens: Good point

contracts/ColonyStorage.sol
kronosapiens: ✨ 🎄 ⛰ 🚫

scripts/check-recovery.js
```diff
@@ -0,0 +1,45 @@
...
+function correctRecoveryModifier(functionDef) {
+  const isPrivate = ["private", "internal"].indexOf(functionDef.visibility) > -1;
+  const isView = ["view", "pure"].indexOf(functionDef.stateMutability) > -1;
+  const hasModifier = functionDef.modifiers.filter(mod => ["stoppable", "recovery"].indexOf(mod.name) > -1).length > 0;
+  return isPrivate || (isView || hasModifier);
+}
+
+walkSync("./contracts/").forEach(contractName => {
+  // Only these contracts are involved in recovery mode
+  if (["contracts/Colony.sol", "contracts/ColonyFunding.sol", "contracts/ColonyTask.sol"].indexOf(contractName) < 0) {
```
area: I think this script would work better in the future if we list the paths that it shouldn't run over, like I currently do in the storage variables script. If we add a new contract to Colony (which we surely will in the future), we have to remember to add it here, rather than being reminded to if it's implemented the other way and we forget.

kronosapiens: That's reasonable. I went this way because relatively few contracts need to be checked, but I see how the "opt-out" approach will make it harder to screw up down the road.

PR #273, opened 2018-07-09
Specify return variable names for functions with public interfaces

JamesLefrere:
This PR makes some alterations to the interface contracts in order to specify the names of all function returns. One benefit of this is that the ABI now receives the return variable names, which makes making sense of returned values simpler.

The difference in the truffle artifact generated looks like this:

screen shot 2018-07-09 at 12 02 18

**Other changes**

* NatSpec `@return` entries now reference the relevant variable name
* Remove a NatSpec return variable which does not exist in the underlying function for the interface contract (`getRewardPayoutInfo`)
* `startNextRewardPayout` appears to have no return value; remove this from the interface

### DIFF
--- a/contracts/IColony.sol
+++ b/contracts/IColony.sol
@@ -92,18 +92,18 @@ contract IColony {
 
   // Implemented in DSAuth.sol
   /// @notice Get the `Authority` for the colony
-  /// @return The `Authority` contract address
+  /// @return authority The `Authority` contract address
   function authority() public view returns (address authority);
 
   /// @notice Get the colony `owner` address. This should be 0x0 at all times
   /// @dev Used for testing.
-  /// @return Address of the colony owner
+  /// @return owner Address of the colony owner
   function owner() public view returns (address owner);
 
   // Implemented in Colony.sol
   /// @notice Get the Colony contract version
   /// Starts from 1 and is incremented with every deployed contract change
-  /// @return Version number
+  /// @return version Version number
   function version() public pure returns (uint256 version);
 
   /// @notice Upgrades a colony to a new Colony contract version `_newVersion`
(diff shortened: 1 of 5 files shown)


contracts/IColony.sol
```diff
@@ -446,13 +446,12 @@ contract IColony {
...
-  /// @return Remaining (unclaimed) amount of tokens
-  /// @return Token address
-  /// @return Block number at the time of creation
-  function getRewardPayoutInfo(uint256 _payoutId) public view returns (bytes32, uint256, uint256, uint256, address, uint256);
+  /// @return reputationState Reputation root hash at the time of creation
+  /// @return totalTokens Total colony tokens at the time of creation
+  /// @return amount Total amount of tokens taken aside for reward payout
+  /// @return tokenAddress Token address
+  /// @return blockTimestamp Block number at the time of creation
+  function getRewardPayoutInfo(uint256 _payoutId) public view returns (bytes32 reputationState, uint256 totalTokens, uint256 amount, address tokenAddress, uint256 blockTimestamp);
```
elenadimitrova: I understand you think this is unused but eventually when we come to returning the entire `RewardPayoutCycle` struct, that will return all 5 members regardless and reverting back to it will be a breaking change. Let's keep it as-is

JamesLefrere: Okay, that makes sense; I'll rebase and skip this commit 👍

JamesLefrere:
> I was a little twitchy about the underlying contracts not being updated to use the same return declarations but that is probably ok

Yes, I considered whether that would be necessary or useful, but perhaps it's more work than it's worth, unless we enforce it globally with linting for example.

JamesLefrere:
It seems like `getRewardPayoutInfo` was just changed in #256 - see https://github.com/JoinColony/colonyNetwork/commit/f7f0b103b972f3f7f8528ab224813ef003a27d79#diff-181fc6eef83c50f44d61b3e11baf76a9R454

I'll rebase onto this (so we won't have the

PR #275, opened 2018-07-10
Make GreenKeeper lockfile updates strictly execute first in CI

elenadimitrova:
Continuing attempts to make GK update the package lockfile in its PRs. This time we're potentially fixing the issue of GK requirement that it is the first job to be run and exiting before successful lockfile upload with `Only running on first push of a new branch` , example in  https://circleci.com/gh/JoinColony/colonyNetwork/2411

To ensure it is the first job to be run, we isolate the GreenKeeper updates using [sequential job execution](https://circleci.com/docs/2.0/workflows/#sequential-job-execution-example) to ensure the job that runs `greenkeeper-lockfile` is always executed first.

Solution inspired by https://github.com/greenkeeperio/greenkeeper-lockfile/issues/50

Resulting workflow build https://circleci.com/workflow-run/2cf9a44a-b22a-450c-a8a5-81114a67098a

Additionally we make some improvements to the circle config by using templates to isolate repeating steps and moving the `node_modules` cache to its own `~/.cache/yarn`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,22 +1,50 @@
 version: 2
 
+job_common: &job_common
+  docker:
+    - image: circleci/node:9.11.2
+  working_directory: ~/colonyNetwork
+step_save_cache: &step_save_cache
+  save_cache:
+    paths:
+      - ~/.cache/yarn
+    key: node-modules-{{ checksum "yarn.lock" }}
+step_restore_cache: &step_restore_cache
+  restore_cache:
+    keys:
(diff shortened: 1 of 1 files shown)


area:
Is there a way to stop the build from continuing if the lockfile was updated? The subsequent steps triggered by this build won't have an updated lockfile, I don't think, and so their results aren't of interest to us (but the builds triggered by GK pushing the lockfile update will be).

elenadimitrova:
We still need to run the unit tests suite sure to check if a dependency update has broken build?

PR #277, opened 2018-07-12
Moved staking logic to token locking contract

### LINKED ISSUE #264: Using token locking in `ColonyNetworkStaking`
## Summary
Do a research on how to replace current deposit/withdraw functionality with token locking implementation.

## Possible Solution
- Check if user has submited root hash in `withdraw` function on token locking contract. This will lock the withdrawal of CLNY token only.
- Rewarding user will happend by doing a transfer on token contract.
- Add function equivalent to `punishStakers` on token locking contract.

## Current Behaviour
- Tokens are deposited on colony network address, and balances are stored in dedicated mapping.
- Tokens are locked when user submits new root hash, and are locked only for them.
- Only `withdraw` is locked, meaning that user can deposit as much times as they want.
- Tokens are unlocked when new mining cycle starts.
- Rewards are given by adding amount to staked balance, and stakers are punished by setting their staked balance to `0`.

filiplazovic:
Closes #264 

Changes include:
- Check if user has submited root hash in `withdraw` function on token locking contract. This will lock the withdrawal of CLNY token only.
- Rewarding user will happend by doing a transfer on token contract.
- Add function equivalent to `punishStakers` on token locking contract.
- Passing token locking address and token address to `ReputationMiningCycle` constructor to save gas.
- Switched order of migration scripts. Deploying token locking first so we can deposit tokens necessary for first mining cycle.
- Renamed `ColonyNetworkStaking` to `ColonyNetworkMining`.


contracts/TokenLocking.sol
```diff
@@ -59,11 +66,32 @@ contract TokenLocking is TokenLockingStorage, DSMath {
...
+      (submittedHash,,,,,,,,,,) = IReputationMiningCycle(IColonyNetwork(colonyNetwork).getReputationMiningCycle(true)).reputationHashSubmissions(msg.sender);
+      require(submittedHash == 0x0);
+    }
+
     userLocks[_token][msg.sender].balance = sub(userLocks[_token][msg.sender].balance, _amount);
 
     require(ERC20Extended(_token).transfer(msg.sender, _amount), "token-locking-transfer-failed");
   }
 
+  function punishStakers(address[] _users) public onlyReputationMiningCycle {
```
elenadimitrova: It feels wrong that the `rewardStakers` is in `ColonyNetworkMining` and `punishStakers` is in the locking contract however since that logic will likely undergo refactoring I won't argue against it living here for now. Meanwhile just add a `todo` to note this context.

test/colony-network-mining.js
```diff
@@ -603,9 +613,9 @@ contract("ColonyNetworkStaking", accounts => {
       assert(nSubmittedHashes.equals(1));
       await repCycle.confirmNewHash(0);
 
-      // Check that they have had their staked balance increase
-      const balance1Updated = await colonyNetwork.getStakedBalance(MAIN_ACCOUNT);
-      assert.equal(balance1Updated.toString(), new BN("3").mul(new BN("10").pow(new BN("18"))).toString(), "Account was not rewarded properly");
+      // Check that they received the reward
+      const balance1Updated = await clny.balanceOf(MAIN_ACCOUNT);
+      assert.equal(balance1Updated.toString(), new BN("2").mul(new BN("10").pow(new BN("18"))).toString(), "Account was not rewarded properly");
```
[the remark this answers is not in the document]
filiplazovic: Before this change, we were giving a reward by appending it to staked balance in `ColonyNetworkStaking` https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyNetworkStaking.sol#L130. Which meant 1e18 (staked) + 2e18 (reward) = 3e18.

Now, we are giving a rewards by doing a transfer on token contract, and reward is just 2e18. https://github.com/JoinColony/colonyNetwork/pull/277/files#diff-655a4fbf6264d2d100633df713ab4b48R88

contracts/IReputationMiningCycle.sol
```diff
@@ -39,6 +39,23 @@ contract IReputationMiningCycle {
     uint256 providedPreviousReputationUID
     );
 
+  /// @notice The getter generated by solidity for the hashSubmissions mapping, which keeps track of submissions by user.
+  /// @param _user Address of the user
+  /// @return The elements of the Submission struct for the submission requested. See ReputationMiningCycle.sol for the full description
+  function reputationHashSubmissions(address _user) public view returns (
```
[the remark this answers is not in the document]
elenadimitrova: yes good spot I must have missed that from previous review

contracts/TokenLocking.sol
elenadimitrova: Can't find a `todo` added for this https://github.com/JoinColony/colonyNetwork/pull/277/files#diff-2143b2fd4ef32ee8dc03526a77a6a429R85

filiplazovic: I missed that, its fixed now

Issue #251, opened 2018-06-26
Allow for additional task properties to be specified during creation


ConnorChristie:
## Summary

It would be nice if there was a way to set the due date when creating the task since right now it requires human (multi)signatures to perform that action.

## Expected Behavior

Allow specifying additional properties for a task upon the `makeTask` call.

## Current Behaviour

The `makeTask` call only accepts a specificationHash and domainId right now and the dueDate can only be changed through a txn signed by a human, contracts are unable to change it.

## Possible Solution

Add optional parameters to the makeTask signature to allow for setting additional properties at the time of task creation.

## Context

I am having multiple individuals working on the same task and I want the author to be able to assign the due date instead (without multisig). I am also using an external contract to facilitate the multiple parties and as of now, the contract itself is unable to set the due date and has to ask the author to set it instead.

KevinLiLu:
@elenadimitrova @ConnorChristie 

I'd like to take this issue. Let me know if there are any conflicts.

### Possible Implementations:

**(1) Use optional parameter**

This would likely be the best solution however optional parameters do not seem to exist in Solidity yet [solidity/#232](https://github.com/ethereum/solidity/issues/232).

**(2) Overload method with additional one that takes in `uint256 _dueDate`**

We would have two methods:
- `function makeTask(bytes32 _specificationHash, uint256 _domainId) public`
- `function makeTask(bytes32 _specificationHash, uint256 _domainId, uint256 _dueDate) public`

Ideally, the second method would call the first method and then call `setTaskDueDate`.

**(3) Make `uint256 _dueDate` a required parameter with a zero value representing unset**

Function declaration would be:
- `function makeTask(bytes32 _specificationHash, uint256 _domainId, uint256 _dueDate) public`

We can do a explicit check `if (_dueDate > 0)` before setting the Task's due date so that passing in a zero value for `_dueDate` will not set the due date (how the current function operates).

<br>

Additionally, should we add a requirement that you cannot call `setTaskDueDate` with a `_dueDate` in the past (if this has not been implemented already)?

elenadimitrova:
Function overloading is cleanest however I'm not sure it's supported (yet) in truffle, details below:
https://github.com/trufflesuite/truffle/issues/737
https://github.com/trufflesuite/truffle/issues/569

Can you test this option and if it fails perhaps explore the darq-truffle solution suggested in 737 above?

KevinLiLu:
[...]

Issue #201, opened 2018-05-01
Add contract error messages


elenadimitrova:
With the introduction of return error messages for `revert` and `require` failures in `solc@0.4.23` (as upgraded in #198) we need to format and populate those message strings in contracts. There are >100 instances where this is required which in itself warrants a dedicated backlog item here.

Additionally after tooling to support these error messages in JS is introduced, tests have to be refactored to check expected error is returned as wired in `todo` here: https://github.com/JoinColony/colonyNetwork/blob/develop/helpers/test-helper.js#L85

elenadimitrova:
After #288 is merged we should be able to complete this work

KevinLiLu:
I'd like to take on this monster as it will get me familiar with more of the codebase.

elenadimitrova:
Thanks @KevinLiLu 
I have pushed in #288 updates you'll need for this, specifically the `test-helper` change to check `reason`
https://github.com/JoinColony/colonyNetwork/pull/288/files#diff-cfe0340d8acf7ab0683cf395091cd627R97

#288 still is a little rough but should allow you to start working on this reasonably smoothly.

area:
A few thoughts:

Error messages should all start with `colony-` to help users identify where errors are coming from, in the event a transaction touches many different contracts.

All tests that check that a require is triggered correctly should check the 'right' require has been triggered. This includes mining tests in `colony-network-mining.js`, which currently don't use `checkErrorRevert` for a lot of tests that just use `accommodateChallengeAndInvalidateHash`. To make things trickier,  that function just swallows the error, which is generated in different places in that function, depending on what is being tested. If that proves too tricky (there is quite a steep learning curve on those tests), do say so.

elenadimitrova:
Also in the next version of `solium@1.1.8` https://github.com/duaraghav8/Solium/releases/tag/1.1.8 they have introduced `error-reason` ruleset which ensures error messages are provided for revert() and require() statements. Maybe @filiplazovic can upgrade to that disabling the single rule that you can switch on at the end of this work?

PR #288, opened 2018-07-20
Upgrade to Truffle v5

elenadimitrova:
Choosing to do an early upgrade as that has a number of features we probably can't live without for the production ready mvp of the network including:

- support for revert reason strings - we can watch/test for error reasons 🎉 
- support for function overloads - we can have as many combinations of the `makeTask` function inputs as we like, ref #251 
- choosing the `solc` version to compile with rather that the hard fixed version that truffle gives us. This decoupling gives us a lot of freedom
- `truffle-contract` which is used in testing is upgraded to `web3v1`

Plus a number of other really nice to have updates like docker support for faster tests and events emitter interface and probably more I haven't yet discovered.

Early release notes I am following for this upgrade, kindly provided by @cgewecke here https://github.com/trufflesuite/truffle/pull/1129

image

**Notable changes**:
- `revert` error strings are returned and can be asserted for in `checkErrorRevert`. We are all eagerly awaiting @KevinLiLu to plug this in all ~190 instances as part of #201 💪  
- `await` all `.at` calls, in the truffle console too, sorry @area
- Since the sync web3 api is gone, things like `web3.eth.accounts` in `constants` is gone too. It's easiest to use the `accounts` input that truffle gives us anyway in tests. There's the `web3GetAccounts` helper if you want to get accounts (async only) anywhere
- All function returns are now `BN.js` rather than `BigNumber.js`. No more thinking of what bignumber implementation you are working on and frees up brain for more important processing 🥇 
- No more `..contract.function.getData`, use `encodeABI`, e.g. 
`etherRouter.contract.setResolver.getData` -> `etherRouter.contract.methods.setResolver("0xb3e2b6020926af4763d706b5657446b95795de57").encodeABI()`
- Parity is now initialised with network id `1998` to match the `integration` network id in truffle. Otherwise when deploying to some network ids (1,2, etc) `trufflev5` automatically triggers dry-run as a safety precaution meaning the main client is actually never used. There is wip to have a config option that disables that behaviour but for now it's the default.

**todos added**:
- Get the client via web3GetClient() which requires web3 beta34 for `getNodeInfo`. But this only means our tests will temporarily not work on `parity`. 

Bad tests:
- `respondToChallenge` called with 9 rather than 10 member array parameter
https://github.com/JoinColony/colonyNetwork/blob/develop/test/colony-network-mining.js#L1723
- swapped parameter values (i.e. in the wrong order)
https://github.com/JoinColony/colonyNetwork/pull/288/files#diff-29fe366dfa20b2486ba3b3a71852fa5eL395

In addition here we introduce `bn-chai` for big numbers testing, replacing all `isTrue(someBN.eq(otherBN))` occurrences. https://github.com/JoinColony/colonyNetwork/issues/305 is logged to deal with the remaining few hundred of them as it is too big a refactoring to be included here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,7 @@ jobs:
       - <<: *step_restore_cache
       - <<: *step_setup_global_packages
       - <<: *step_setup_greenkeeper
+      - setup_remote_docker
       - <<: *step_pull_solc_docker
       - run:
           name: "Setup parity"
@@ -108,6 +109,7 @@ jobs:
       - checkout
       - <<: *step_restore_cache
       - <<: *step_setup_global_packages
+      - setup_remote_docker
       - <<: *step_pull_solc_docker
       - run:
           name: "Running unit tests with coverage"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,10 @@ step_setup_greenkeeper: &step_setup_greenkeeper
   run:
     name: "Add greenkeeper-lockfile-update"
     command: yarn global add greenkeeper-lockfile@1
+step_pull_solc_docker: &step_pull_solc_docker
+    run:
+      name: "Pull solc 0.4.23 docker image"
+      command: docker pull ethereum/solc:0.4.23
 jobs:
   greenkeeper-updates:
     <<: *job_common
@@ -45,6 +49,7 @@ jobs:
       - <<: *step_restore_cache
       - <<: *step_setup_global_packages
       - <<: *step_setup_greenkeeper
+      - <<: *step_pull_solc_docker
       - run:
           name: "Setup parity"
           command: |
@@ -103,6 +108,7 @@ jobs:
       - checkout
       - <<: *step_restore_cache
       - <<: *step_setup_global_packages
+      - <<: *step_pull_solc_docker
       - run:
           name: "Running unit tests with coverage"
           command: yarn run test:contracts:coverage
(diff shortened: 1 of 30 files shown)


elenadimitrova:
@area maybe you can work out why is adding docker not improved test run times here :/

KevinLiLu:
@elenadimitrova 

I'm currently attacking the beast!

How should I add my #201 changes? Should I create PRs to `experiment/darq-truffle`? I don't want to drop a super massive PR, but the tests will not pass unless I add/fix all the error messages...

Are you okay with the ci tests failing so we can review the error messages in batches? Or is there a better way to do it?

elenadimitrova:
@KevinLiLu Your PR should be to merge against `develop` but still be rebased on `experiment/darq-truffle` which will be merged before you finish the revert error checks.
CI should not be failing even if you're working in batches. The `checkErrorRevert` errMsg parameter is optional so if you don't pass it it won't throw. (Although remember to change that at the very end)

KevinLiLu:
@elenadimitrova 

Ah I see. That makes sense.

The test failures I'm seeing are due to bugs where some existing error messages do not match up 😝

elenadimitrova:
Yes, brace yourself for finding actual bugs and let us know if you need help as it may not be easy working through them 💪

elenadimitrova:
You're right about both points you raise. Just on the second one, about `assert.isTrue` vs `assert.equal`. After failing to use `.toNumber()` and having to compare strings instead I did some experimentation using `assert(someBN.eq(otherBN))`, `assert.deepEqual` and `assert.strictEqual` and they all failed at comparing bn.js instances. In my desperation I just turned to `.eq` again to try and stay a BN. I haven't given up yet, so I'll try `bn-chai` next .

PR #289, opened 2018-07-20
Token issuance rate, initial supply, supply ceiling

### LINKED ISSUE #25: Token Supplier
### 5.1.1 Token Generation and Initial Supply

When a colony is created, the `TokenSupplyCeiling` and the `TokenIssuanceRate` are set. The former is the total number of colony tokens that will be created and the latter is the rate at which they become available to the colony-wide domain to assign to tasks or subdomains. The number of tokens available to the colony-wide domain can be updated at any time by a transaction from any user.
At colony creation, some tokens must also be assigned to addresses to allow users to stake tokens to create the first tasks. A one-off lump sum may also be created and made available to the colony-wide domain.

### 5.1.2 Increasing the TokenSupplyCeiling
It is crucial that new tokens cannot be generated without widespread consensus — especially if
tokens have a financial value. Consequently, such decisions require a vote with high quorum and
majority requirements involving both the token holders and reputation holders.

### 5.1.3 Changing the TokenIssuanceRate
The TokenSupplyCeiling represents the tokens that the token holders have granted to the colony
in order to conduct business: to fund tasks and domains, and to hire workers and contributors.
This is especially important during the early life of a colony when it has little-to-no revenue in other
tokens to fall back on.
The TokenIssuanceRate controls how rapidly the colony receives the new tokens. If the rate
is ‘too high’, tokens will accumulate in the pot of the top-level colony domain (or other pots lower
in the hierarchy); usually this is not a big problem. If the rate is too low, this signals that the
colony has a healthy amount of activity and that the issuance rate has become a bottleneck. In
such situations it may be desirable to increase the rate of issuance without necessarily increasing
the maximum supply.
Increasing and decreasing the TokenIssuanceRate by up to 10% can be done by the reputation
holders alone and this action can be taken no more than once every 4 weeks. Larger changes to the
issuance rate should additionally require the agreement of existing token holders.

### LINKED ISSUE #26: Increasing the TokenSupplyCeiling
### 5.1.2 Increasing the TokenSupplyCeiling
It is crucial that new tokens cannot be generated without widespread consensus — especially if
tokens have a financial value. Consequently, such decisions require a vote with high quorum and
majority requirements involving both the token holders and reputation holders.

### LINKED ISSUE #27: Changing the TokenIssuanceRate
### 5.1.3 Changing the TokenIssuanceRate
The `TokenSupplyCeiling` represents the tokens that the token holders have granted to the colony
in order to conduct business: to fund tasks and domains, and to hire workers and contributors.
This is especially important during the early life of a colony when it has little-to-no revenue in other
tokens to fall back on.
The `TokenIssuanceRate` controls how rapidly the colony receives the new tokens. If the rate
is ‘too high’, tokens will accumulate in the pot of the top-level colony domain (or other pots lower
in the hierarchy); usually this is not a big problem. If the rate is too low, this signals that the
colony has a healthy amount of activity and that the issuance rate has become a bottleneck. In
such situations it may be desirable to increase the rate of issuance without necessarily increasing
the maximum supply.
Increasing and decreasing the `TokenIssuanceRate` by up to 10% can be done by the reputation
holders alone and this action can be taken no more than once every 4 weeks. Larger changes to the
issuance rate should additionally require the agreement of existing token holders.

filiplazovic:
Closes #25 , Closes #26 , Closes #27 

## Tokens issuance rate
For calculating rate change used this formula:
a1 - current rate amount
a2 - new rate amount
```
11*a1 >= 10*a2 if a2 > a1
9*a1 <= 10*a2 if a2 < a1
```
This change is not calculated the first time the rate is set, so basically users can set any rate they want, because every token has a different value, therefore requires different rate.

## Initial supply
When colony is in bootstrap state `mintTokens` can be called without any constraints.

## Token supply ceiling
Is a limit for how many tokens there are in the circulation. Constraints that ensures that we don't exceed this limit are set in `mintTokens`.

## Other
A lot of tests were failing because of token issuance limit, so they are refactored. Because amount issued in tests are too big, we are issuing them in bootstrap mode, or by setting very high rates.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -76,6 +76,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setRoleAssignmentFunction(bytes4(keccak256("setTaskEvaluatorRole(uint256,address)")));
     setRoleAssignmentFunction(bytes4(keccak256("setTaskWorkerRole(uint256,address)")));
 
+    // Setting this so there will be no tokens available for issuing when the colony is created
+    tokenIssuanceRate.totalAmountIssuedUnderRate = now;
+
     // Initialise the root domain
     IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
     uint256 rootLocalSkill = colonyNetwork.getSkillCount();
@@ -83,6 +86,36 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
     // Set initial colony reward inverse amount to the max indicating a zero rewards to start with
     rewardInverse = 2**256 - 1;
+  }
+
+  function setTokenIssuanceRate(uint128 _amount) public auth stoppable {
+    // Minimum amount is 10 because of a rounding error
+    require(_amount >= 10, "token-issuance-rate-bad-inputs");
+    require(sub(now, tokenIssuanceRate.timestamp) >= 4 weeks, "token-issuance-rate-change-not-allowed");
+
+    // If we are setting the rate for the first time we will not go through these checks
+    if (tokenIssuanceRate.amount > 0) {
+      if (_amount > tokenIssuanceRate.amount) {
+        require(11 * tokenIssuanceRate.amount >= 10 * _amount, "token-issuance-rate-change-too-big");
+      } else {
+        require(9 * tokenIssuanceRate.amount <= 10 * _amount, "token-issuance-rate-change-too-big");
+      }
+    }
+
+    tokenIssuanceRate = TokenIssuanceRate(_amount, now, now);
+  }
+
+  function setTokenSupplyCeiling(uint256 _amount) public auth stoppable {
+    require(_amount > token.totalSupply(), "token-supply-ceiling-bad-input");
+    tokenSupplyCeiling = _amount;
+  }
+
+  function getTokenSupplyCeiling() public view returns (uint256) {
+    return tokenSupplyCeiling;
+  }
+
+  function getTokenIssuanceRate() public view returns (uint256, uint256, uint256) {
+    return (tokenIssuanceRate.amount, tokenIssuanceRate.timestamp, tokenIssuanceRate.totalAmountIssuedUnderRate);
   }
 
   function bootstrapColony(address[] _users, int[] _amounts) public
@@ -100,11 +133,35 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     }
   }
 
-  function mintTokens(uint _wad) public
+  function mintTokens(uint _amount) public
   stoppable
   auth
   {
-    return token.mint(_wad);
+    require(add(_amount, token.totalSupply()) <= tokenSupplyCeiling, "token-issuance-amount-exceeds-supply-ceiling");
+
+    if (taskCount > 0) {
+      uint256 availableAmount;
+      uint256 availableSeconds;
+      (availableAmount, availableSeconds) = getAvailableIssuanceAmountAndSeconds();
+
+      require(_amount <= availableAmount, "token-issuance-rate-exceeded");
+
+      // Extract how many seconds corresponds to `_amount` of tokens
+      uint256 time = mul(_amount, availableSeconds) / availableAmount;
+      // Register that we issued amount corresponding to `time` amount of seconds
+      tokenIssuanceRate.totalAmountIssuedUnderRate = add(tokenIssuanceRate.totalAmountIssuedUnderRate, time);
+    }
+
+    return token.mint(_amount);
+  }
+
+  function getAvailableIssuanceAmountAndSeconds() public view returns (uint256, uint256) {
+    // Get how many seconds passed from last issuance
+    uint256 secondsSinceLastIssuance = sub(now, tokenIssuanceRate.totalAmountIssuedUnderRate);
+    // How much is available for issuance
+    uint256 availableAmount = mul(secondsSinceLastIssuance, tokenIssuanceRate.amount) / 30 days;
+
+    return (availableAmount, secondsSinceLastIssuance);
   }
 
   function mintTokensForColonyNetwork(uint _wad) public stoppable {
--- a/contracts/ColonyStorage.sol
+++ b/contracts/ColonyStorage.sol
@@ -38,6 +38,22 @@ contract ColonyStorage is CommonStorage, DSMath {
   ERC20Extended token;
   uint256 rewardInverse;
 
+  struct TokenIssuanceRate {
+    // Amount describing how many tokens can be issued for one month
+    uint128 amount;
+    // Last time the token issuance rate is changed
+    // It is used to assure that rate can be changed once every 4 weeks
+    uint256 timestamp;
+    // Keeps track of how many tokens are issued in terms of seconds
+    // TODO: If we decide to mint entire available amount in `mintTokens`, there will be no need for this variable
+    uint256 totalAmountIssuedUnderRate;
+  }
+
+  // Keeps information about token issuance rate amount and last change timestamp
+  TokenIssuanceRate tokenIssuanceRate;
+  // Amount of tokens allowed in circulation
+  uint256 tokenSupplyCeiling;
+
   // Mapping function signature to 2 task roles whose approval is needed to execute
   mapping (bytes4 => uint8[2]) reviewers;
 
--- a/test/colony-network-auction.js
+++ b/test/colony-network-auction.js
@@ -2,9 +2,9 @@
 import { BN } from "bn.js";
 import chai from "chai";
 import bnChai from "bn-chai";
+import { toBN } from "web3-utils";
 
 import { getTokenArgs, web3GetTransactionReceipt, web3GetCode, checkErrorRevert, forwardTime, getBlockTime } from "../helpers/test-helper";
-import { giveUserCLNYTokens } from "../helpers/test-data-generator";
 
 const { expect } = chai;
 chai.use(bnChai(web3.utils.BN));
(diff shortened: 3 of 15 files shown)


contracts/Colony.sol
```diff
@@ -74,12 +74,54 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+      uint256 diff = (oldRate > newRate) ? sub(oldRate, newRate) : sub(newRate, oldRate);
+      require(diff > 0, "token-issuance-rate-invalid-rate");
+      // How much is new rate different (in percent) relative to the current one
+      uint256 percent = mul(diff, 100) / oldRate;
+      // Must be less than 10 percent
+      require(percent <= 10, "token-issuance-rate-change-too-big");
+    }
+
+    lastIssuenceRateChangeTimestamp = now;
+
```
elenadimitrova: Additionally here call `mintTokens` for the full remaining amount allowed to be minted as otherwise this would cause the new issuance rate to be "backdated" to the last mint time under the previous rate.

contracts/Colony.sol
```diff
@@ -74,12 +74,54 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+      uint256 newRate = mul(_amount, _precision) / _period;
+      // this will throw if `_precision` is not high enough
+      require(oldRate > 0 && newRate > 0, "token-issuance-rate-bad-precision");
+      // Get the difference between rates
+      uint256 diff = (oldRate > newRate) ? sub(oldRate, newRate) : sub(newRate, oldRate);
+      require(diff > 0, "token-issuance-rate-invalid-rate");
+      // How much is new rate different (in percent) relative to the current one
+      uint256 percent = mul(diff, 100) / oldRate;
+      // Must be less than 10 percent
+      require(percent <= 10, "token-issuance-rate-change-too-big");
```
elenadimitrova: For `a` and `p` are as defined in PR description this requirement can be simplified to the following
`11*a1*p2 >=10*a2*p1` where `a2/p2 > a1/p1` i.e. increasing issuance rate
`9*a1*p2 <= 10*a2*p1` where `a1/p1 > a2/p2` i.e. decreasing issuance rate
thus removing the need for `precision` and other calculations. 

To protect against overflows I'd suggest changing the `uint256` types of `tokenIssuanceRate.amount` to `uint128` and `tokenIssuanceRate.period` to `uint32`. 
`uint128` => upper limit = `340282366920938463463374607431768211455` token wei
`uint32` => upper limit = `((2^32-1)/2628000)/12 = 136.19 years` which is more than sufficient. 

It should provide sufficient range of values yet prevent overflows and also save 20,000 gas and an extra slot per `TokenIssuanceRate` entry.

elenadimitrova: Following a conversation with @area he suggested to fix the issuance rate period to say a month and only allow the `amount` to vary. This would limit to some negligible extent the issuance rate variations to min `30*24*3600=2592000` tokens per month which considering it is significantly less than `10^18` is deemed acceptable.
Can you think of any issues with that model? Perhaps plug this (even simpler) version in the suggestion above.

test/colony.js
```diff
@@ -1583,7 +1956,6 @@ contract("Colony", addresses => {
         sigTypes: [0, 0],
         args: [taskId, WORKER]
       });
-      await colony.mintTokens(100);
```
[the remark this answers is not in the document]
filiplazovic: We can, its just simpler this way. Is there any problem doing it this way?

contracts/Colony.sol
```diff
@@ -74,12 +74,54 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setRoleAssignmentFunction(bytes4(keccak256("setTaskEvaluatorRole(uint256,address)")));
     setRoleAssignmentFunction(bytes4(keccak256("setTaskWorkerRole(uint256,address)")));
 
+    // Setting this so there will be no tokens available for issuing when the colony is created
+    totalIssuedInSeconds = now;
```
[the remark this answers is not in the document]
filiplazovic: Sorry, i don't understand what you mean.

contracts/Colony.sol
filiplazovic: Done.

But what do you think about just setting `totalAmountIssuedUntilTimestamp = now`. This way, user will be able to choose whether they want to mint remaining tokens before they change the rate.

test/colony.js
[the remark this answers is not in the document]
elenadimitrova: I can't remember the problem I saw if any.. Let's see if I remember what I meant here before I finish the review

contracts/Colony.sol
```diff
@@ -74,12 +74,54 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+
     // Initialise the root domain
     IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
     uint256 rootLocalSkill = colonyNetwork.getSkillCount();
     initialiseDomain(rootLocalSkill);
   }
 
+  function setTokenIssuanceRate(uint256 _amount, uint256 _period, uint256 _precision) public auth {
+    require(_amount > 0 && _period > 0, "token-issuance-rate-bad-inputs");
+    require(sub(now, lastIssuenceRateChangeTimestamp) >= 4 weeks, "token-issuance-rate-change-not-allowed");
```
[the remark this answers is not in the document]
elenadimitrova: Can you just add the change I mention to the list here so it's not forgotten  https://github.com/JoinColony/Colony-Whitepaper/issues/7

contracts/Colony.sol
elenadimitrova: Good point. Sorry I saw this just now, maybe I prefer your idea to what I suggested above. Let's switch to that.

elenadimitrova: Just add a note in natspec for describing that we restart the time for token issuance rate when this is set.

contracts/ColonyStorage.sol
```diff
@@ -37,6 +37,22 @@ contract ColonyStorage is DSAuth, DSMath {
...
+    uint256 timestamp;
+  }
+
+  // Keeps information about token issuance rate amount and last change timestamp
+  TokenIssuanceRate tokenIssuanceRate;
+  // Amount of tokens allowed in circulation
+  uint256 tokenSupplyCeiling;
+  // Keeps track of how many tokens are issued in terms of seconds
+  // TODO: If we decide to mint entire available amount in `mintTokens`, there will be no need for this variable
+  uint256 totalAmountIssuedUntilTimestamp;
```
elenadimitrova: I wonder if we can make this implementation really clean by also moving this `totalAmountIssuedUntilTimestamp` to the `TokenIssuanceRate` struct params and calling it `totalAmountIssuedUnderRate`

elenadimitrova: This way there will be 2 added variables to Colony: 1 for the supply ceiling and the other for the issuance rate. Nice and clean. I suspect the logic that you've already refactored quite well will become even simpler.

test/colony-network-auction.js
```diff
@@ -205,7 +226,8 @@ contract("ColonyNetworkAuction", accounts => {
 
   describe("when bidding", async () => {
     it("can bid", async () => {
-      await giveUserCLNYTokens(colonyNetwork, BIDDER_1, "1000000000000000000");
```
elenadimitrova: I don't get why is this replaced by recreating the colony and then using `bootstrap` to give user tokens.

contracts/ColonyStorage.sol
```diff
@@ -37,6 +37,22 @@ contract ColonyStorage is DSAuth, DSMath {
   address colonyNetworkAddress;
   ERC20Extended token;
 
+  struct TokenIssuanceRate {
+    // Amount describing how many tokens can be issued for one month
+    uint256 amount;
```
elenadimitrova: `amount` should be a `uint128`

test/colony-network-auction.js
filiplazovic: Ok, recreating the colony is definitely not needed, but as we talked, we are minting a very large amount somewhere in the test (~`10**54`). Since maximum rate is ~`3.4e38` we would need to forward time 1 billion years to be able to mint `10**54`, and ganache just breaks with that amount for some reason.

elenadimitrova: How did you calculate the ~3.4e38 maximum rate? I am getting different values? Is that per second, per month?

filiplazovic: Its per month, the maximum value we can pass to `setTokenIssuanceRate`

elenadimitrova: I am getting 244894656421309 years which is still a very large number to work with considering 10^54 is probably a reasonable amount of tokens people might want minted. After all `TokenIssuanceRate.amount` might need to be a `uint256`?

elenadimitrova: The max `uint128` is `131281777361473172632474771385713` which should be the max issuance rate we pass right?

filiplazovic: For `uint128` i'm getting => https://www.google.rs/search?q=2+**+128&oq=2+**+12&aqs=chrome.0.69i59j69i57j6j69i59.2522j0j7&sourceid=chrome&ie=UTF-8

If we change `TokenIssuanceRate.amount` to `uint256` there is possible overflow [here](https://github.com/JoinColony/colonyNetwork/pull/289/files#diff-669f459dea66cccc363b550a82126c97R167) and [here](https://github.com/JoinColony/colonyNetwork/pull/289/files#diff-669f459dea66cccc363b550a82126c97R179)

elenadimitrova: Yes you're right, that's per month and mine is per second.

filiplazovic: So is `3.4e38` enough as upper limit?

elenadimitrova: Let's leave it like that for now

elenadimitrova:
Since #25 , #26 and #27 were de-scoped from the first release of the network - Colony Contribute, we'll hold off merging this until a further release.
We should consider again whether the uint128 max value is sufficient to cater for issuance rates per month as discussed here https://github.com/JoinColony/colonyNetwork/pull/289#discussion_r206482436

elenadimitrova:
Thanks for rebasing @filiplazovic !
Closing this temporarily until it's ready to be reviewed and merged in one of the next Network releases.

Issue #211, opened 2018-05-25
Make mining client suitable for real-world use


area:
Currently, the mining client works in principle, but does not have many of the features needed to ensure that people can use it for its intended purpose - reputation mining by the public CLNY holders on the deployed-on-main-net colonyNetwork.

Currently, no state persists between restarts. There is no way to get it 'up-to-speed' with the current reputation state on first boot (currently, all tests start from the reputation state of `0x0`. Currently, there are not even tests to make sure that it can persist state across update cycles (although I'm 99% sure that it can). It also needs to be made user-friendly, and based on the discussions that #55 alludes  to, have some sort of adaptive gas-pricing algorithm. It should automatically submit reputation hashes a user is entitled to, as soon as they are able to.

I'm sure there's more, which I'll add here as I think of it!

area:
Closed by #596

PR #293, opened 2018-07-29
Fix TODOs for reputation mining submission windows

### LINKED ISSUE #292: Fix withinTarget timestamp TODO
## Summary

```
  modifier withinTarget(bytes32 newHash, uint256 entryIndex) {
    require(reputationMiningWindowOpenTimestamp > 0);
    // Check the ticket is a winning one.
    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
    // require((now-reputationMiningWindowOpenTimestamp) <= 3600);
```

kronosapiens:
Closes #292 #321

### DIFF
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -69,16 +69,19 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
   /// @dev A submission will only be accepted from a reputation miner if `keccak256(address, N, hash) < target`
   /// At the beginning of the submission window, the target is set to 0 and slowly increases to 2^256 - 1 after an hour
   modifier withinTarget(bytes32 newHash, uint256 entryIndex) {
-    require(reputationMiningWindowOpenTimestamp > 0, "colony-reputation-mining-cycle-not-open");
     // Check the ticket is a winning one.
-    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
-    // require((now-reputationMiningWindowOpenTimestamp) <= 3600);
-    // x = floor(uint((2**256 - 1) / 3600)
-    if (now - reputationMiningWindowOpenTimestamp <= 3600) {
+    if (!submissionWindowClosed()) {
+      // x = floor(uint((2**256 - 1) / 3600)
       uint256 x = 32164469232587832062103051391302196625908329073789045566515995557753647122;
-      uint256 target = (now - reputationMiningWindowOpenTimestamp ) * x;
-      require(uint256(getEntryHash(msg.sender, entryIndex, newHash)) < target, "colony-reputation-mining-cycle-closed");
+      uint256 target = (now - reputationMiningWindowOpenTimestamp) * x;
+      require(uint256(getEntryHash(msg.sender, entryIndex, newHash)) < target, "colony-reputation-mining-cycle-submission-not-within-target");
     }
+    _;
+  }
+
+  modifier submissionsOpen() {
+    require(reputationMiningWindowOpenTimestamp > 0, "colony-reputation-mining-cycle-not-open");
+    require(!submissionWindowClosed() || nSubmittedHashes == 0, "colony-reputation-mining-cycle-submissions-closed");
     _;
   }
 
@@ -120,6 +123,7 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
   }
 
   function submitRootHash(bytes32 newHash, uint256 nNodes, uint256 entryIndex) public
+  submissionsOpen()
   entryQualifies(newHash, nNodes, entryIndex)
   withinTarget(newHash, entryIndex)
   {
@@ -175,7 +179,8 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
   function confirmNewHash(uint256 roundNumber) public
   finalDisputeRoundCompleted(roundNumber)
   {
-    // TODO: Require some amount of time to have passed (i.e. people have had a chance to submit other hashes)
+    require(submissionWindowClosed(), "colony-reputation-mining-submission-window-still-open");
+
     Submission storage submission = disputeRounds[roundNumber][0];
     IColonyNetwork(colonyNetworkAddress).setReputationRootHash(
       submission.proposedNewRootHash,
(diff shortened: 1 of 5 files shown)


kronosapiens:
Noting that this PR is also being extended to include this todo:

https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ReputationMiningCycle.sol#L226

test/colony-network-mining.js
```diff
@@ -668,40 +696,34 @@ contract("ColonyNetworkMining", accounts => {
...
-      await repCycle.submitRootHash("0x12345678", 10, 8);
-      await repCycle.submitRootHash("0x12345678", 10, 9);
-      await repCycle.submitRootHash("0x12345678", 10, 10);
-      await repCycle.submitRootHash("0x12345678", 10, 11);
-      await repCycle.submitRootHash("0x12345678", 10, 12);
-      await checkErrorRevert(repCycle.submitRootHash("0x12345678", 10, 13), "colony-reputation-mining-max-number-miners-reached");
+      let entryNumber = await getValidEntryNumber(colonyNetwork, MAIN_ACCOUNT, "0x12345678", 1);
+      for (let i = 1; i <= 12; i += 1) {
+        await repCycle.submitRootHash("0x12345678", 10, entryNumber); // eslint-disable-line no-await-in-loop
+        entryNumber = await getValidEntryNumber(colonyNetwork, MAIN_ACCOUNT, "0x12345678", entryNumber + 1); // eslint-disable-line no-await-in-loop
```
elenadimitrova: This iteration got me confused. I _think_ what you're doing here is starting from the first available `entryNumber` and continuing until there are `12` but if so why not iterate from `entryNumber` to `12` instead of `1` to `12`?

test/colony-network-mining.js
```diff
@@ -1506,8 +1544,12 @@ contract("ColonyNetworkMining", accounts => {
       await badClient.initialise(colonyNetwork.address);
       await goodClient.addLogContentsToReputationTree();
       await badClient.addLogContentsToReputationTree();
+
+      await forwardTime(3540, this);
       await goodClient.submitRootHash();
       await badClient.submitRootHash();
+      await forwardTime(60, this);
```
elenadimitrova: Any reason why you've squeezed in the hash submissions into smaller time here than the other tests?

contracts/ReputationMiningCycle.sol
```diff
@@ -178,7 +182,8 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
   function confirmNewHash(uint256 roundNumber) public
   finalDisputeRoundCompleted(roundNumber)
   {
-    // TODO: Require some amount of time to have passed (i.e. people have had a chance to submit other hashes)
+    require(submissionWindowClosed(), "colony-reputation-mining-submission-window-still-open");
```
elenadimitrova: This is not been tested for failure.

contracts/ReputationMiningCycle.sol
```diff
@@ -72,19 +72,22 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
   /// @dev A submission will only be accepted from a reputation miner if `keccak256(address, N, hash) < target`
   /// At the beginning of the submission window, the target is set to 0 and slowly increases to 2^256 - 1 after an hour
   modifier withinTarget(bytes32 newHash, uint256 entryIndex) {
-    require(reputationMiningWindowOpenTimestamp > 0, "colony-reputation-mining-cycle-not-open");
     // Check the ticket is a winning one.
-    // TODO Figure out how to uncomment the next line, but not break tests sporadically.
-    // require((now-reputationMiningWindowOpenTimestamp) <= 3600);
-    // x = floor(uint((2**256 - 1) / 3600)
-    if (now - reputationMiningWindowOpenTimestamp <= 3600) {
+    if (!submissionWindowClosed()) {
```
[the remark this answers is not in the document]
area: If no submission has been made to the current cycle, but the window has been open for more than an hour, submissionsOpen() does not throw. In that case, the calculation of `target` would overflow, and most submissions would be ineligible again. We want to bypass the eligibility check after the window has been open for an hour, hence this check.

Perhaps renaming `submissionWindowClosed` to `minimumSubmissionWindowElapsed` would  make this easier to understand what's happening?

packages/reputation-miner/ReputationMiner.js
```diff
@@ -489,17 +496,56 @@ class ReputationMiner {
...
-    const repCycle = new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+    return new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+  }
+
+  /**
+   * Submit what the client believes should be the next reputation state root hash to the `ReputationMiningCycle` contract
+   * @param startIndex What index to start searching at when looking for a valid submission
+   * @return {Promise}
+   */
+  async submitRootHash(startIndex = 1) {
```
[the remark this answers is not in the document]
area: The intention here is for this parameter to be passed with the index after any index they have already submitted to the current cycle, in order to avoid attempting to resubmit a hash with an index they have already used.

Upon reflection, this isn't strictly optimal behaviour for the miner - it skips any earlier entries that might now be valid that were not valid earlier. I think that behaviour change would be more appropriate in the 'autonomous mining' work I'm intending to do as part of #211.

test/colony-network-mining.js
area: `entryNumber` relates to which 'set' of `10**15` CLNY are used to submit an entry. The twelve entires might involve sets 1, 5, 13, 300, 341, 444, 510, 630, 635, 687, 731, 899. You are right that we want twelve to be submitted, but the entryNumber has (almost) no relation to the `i` being counted here (which is the number of the submission that is about to be made).

Again, perhaps you're being tripped up by my bad naming here 😢

test/colony-network-mining.js
area: No reason. I think this is just a relic from initial tinkering. Will change to 3500 which is at least consistent with a lot of the others. Even doing that though, looks like we've ended up with a split between skips of 3500/100 and 1800/1800... shall I make those consistent where possible (which should be most places)?

elenadimitrova: yes please! Considering we'll be likely changing this timeframe to more than an hour, it maybe worth extracting this logic of submissions into a helper like here https://github.com/JoinColony/colonyNetwork/pull/357/files#diff-15aaaaa81a55d904a938772d3e2a3a82R2855?

contracts/ReputationMiningCycle.sol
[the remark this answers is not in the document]
elenadimitrova: Or keep the window open until either we have 1 submission or time has elapsed, whichever is sooner?

PR #294, opened 2018-07-31
Proving reputation in reward payouts

filiplazovic:
Unexpected changes include:
- Refactored `claimRewardPayout` to avoid stack depth error
- Proving colony wide reputation inside `startNextRewardPayout` to avoid too many function arguments.

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -187,19 +187,32 @@ contract ColonyFunding is ColonyStorage {
     return nonRewardPotsTotal[_token];
   }
 
-  function startNextRewardPayout(address _token) public stoppable auth {
+  function startNextRewardPayout(address _token, bytes key, bytes value, uint256 branchMask, bytes32[] siblings) public auth stoppable {
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
     uint256 totalLockCount = tokenLocking.lockToken(address(token));
 
     require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
 
     uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
-    require(totalTokens > 0, "colony-reward-payout-invalid-total-tokens");
+    require(token.totalSupply() > 0, "colony-reward-payout-invalid-total-tokens");
+
+    bytes32 rootHash = IColonyNetwork(colonyNetworkAddress).getReputationRootHash();
+    uint256 colonyWideReputation = checkReputation(
+      rootHash,
+      domains[1].skillId,
+      0x0,
+      key,
+      value,
+      branchMask,
+      siblings
+    );
+    require(colonyWideReputation > 0, "colony-reward-payout-invalid-colony-wide-reputation");
 
     activeRewardPayouts[_token] = true;
 
     rewardPayoutCycles[totalLockCount] = RewardPayoutCycle(
-      IColonyNetwork(colonyNetworkAddress).getReputationRootHash(),
+      rootHash,
+      colonyWideReputation,
       totalTokens,
       pots[0].balance[_token],
       _token,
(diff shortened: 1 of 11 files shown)


test/colony-funding.js
```diff
@@ -634,43 +689,68 @@ contract("Colony Funding", accounts => {
...
+      const colonyWideReputationKey = makeReputationKey(colony.address, rootDomainSkill.toNumber());
+      const { key, value, branchMask, siblings } = await miningClient.getReputationProofObject(colonyWideReputationKey);
+      colonyWideReputationProof = [key, value, branchMask, siblings];
 
       await checkErrorRevert(
-        colony.claimRewardPayout(payoutId.toString(), initialSquareRoots, 0, 0, {
-          from: userAddress1
-        }),
+        newColony.startNextRewardPayout(otherToken.address, ...colonyWideReputationProof),
         "colony-reward-payout-invalid-total-reputation"
```
area: You have removed this require string from the contracts, so this test cannot possibly be testing for it. Is it meant to be `colony-reward-payout-invalid-colony-wide-reputation`? If so, this test is wrong, because according to coverage, that require is never tripped.

test/colony-funding.js
```diff
@@ -634,43 +689,68 @@ contract("Colony Funding", accounts => {
...
+
+      const result = await colony.getDomain(1);
+      const rootDomainSkill = result.skillId;
+      const colonyWideReputationKey = makeReputationKey(newColony.address, rootDomainSkill.toNumber());
+      const { key, value, branchMask, siblings } = await miningClient.getReputationProofObject(colonyWideReputationKey);
+      colonyWideReputationProof = [key, value, branchMask, siblings];
+
+      await checkErrorRevert(
+        newColony.startNextRewardPayout(otherToken.address, ...colonyWideReputationProof),
+        "colony-reward-payout-invalid-total-tokens"
```
area: This require does not have both branches run, according to coverage, so this transaction is failing in an unintended place.

test/colony-funding.js
```diff
@@ -748,19 +860,23 @@ contract("Colony Funding", accounts => {
...
+      const userReputationKey = makeReputationKey(colony.address, rootDomainSkill.toNumber(), userAddress3);
+      ({ key, value, branchMask, siblings } = await miningClient.getReputationProofObject(userReputationKey));
+      const userReputationProof3 = [key, value, branchMask, siblings];
+
       await checkErrorRevert(
-        colony.claimRewardPayout(payoutId, squareRoots, 0, totalReputation.toString(), {
+        colony.claimRewardPayout(payoutId, squareRoots, ...userReputationProof3, {
           from: userAddress3
         }),
         "colony-reward-payout-invalid-user-reputation"
```
area: According to coverage, this call doesn't fail here.

test/colony-funding.js
```diff
@@ -703,8 +797,12 @@ contract("Colony Funding", accounts => {
...
+      const userReputationKey = makeReputationKey(colony.address, rootDomainSkill.toNumber(), userAddress3);
+      ({ key, value, branchMask, siblings } = await miningClient.getReputationProofObject(userReputationKey));
+      const userReputationProof3 = [key, value, branchMask, siblings];
+
       await checkErrorRevert(
-        colony.claimRewardPayout(payoutId, squareRoots, userReputation3.toString(), totalReputation.toString(), {
+        colony.claimRewardPayout(payoutId, squareRoots, ...userReputationProof3, {
           from: userAddress3
         }),
         "colony-reward-payout-invalid-user-tokens"
```
area: Coverage says that this transaction doesn't fail here.

contracts/ColonyFunding.sol
```diff
@@ -166,19 +166,32 @@ contract ColonyFunding is ColonyStorage {
...
-  function startNextRewardPayout(address _token) public stoppable auth {
+  function startNextRewardPayout(address _token, bytes key, bytes value, uint256 branchMask, bytes32[] siblings) public auth stoppable {
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
     uint256 totalLockCount = tokenLocking.lockToken(address(token));
 
     require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
 
     uint256 totalTokens = sub(token.totalSupply(), token.balanceOf(address(this)));
-    require(totalTokens > 0, "colony-reward-payout-invalid-total-tokens");
+    require(token.totalSupply() > 0, "colony-reward-payout-invalid-total-tokens");
```
area: Why the change in this requirement? Now it's just checking that there are any tokens at all, rather than checking the colony doesn't just own all of them? Wasn't this guard partially here to stop us trying to divide by zero later?

contracts/ColonyFunding.sol
```diff
@@ -244,17 +240,88 @@ contract ColonyFunding is ColonyStorage {
...
+
+    assembly {
+      reputationValue := mload(add(value, 32))
+      keyColonyAddress := mload(add(key, 20))
+      keySkill := mload(add(key, 52))
+      keyUserAddress := mload(add(key, 72))
+    }
+
+    require(keyColonyAddress == address(this), "colony-reputation-invalid-colony-address");
+    require(keySkill == skillId, "colony-reputation-invalid-skill-id");
```
area: There are no tests trying to check this require, or the next one. I could live without this one being tested (the worst thing that happens - where this is currently used, at least - is that they supply a different skill, which *must* have either the same or less reputation than they're entitled to claim against, so they're shorting themselves.

If the next require were accidentally removed somehow, then people would be able to claim against reputation held by another user, which seems bad...

packages/reputation-miner/ReputationMiner.js
```diff
@@ -254,7 +255,10 @@ class ReputationMiner {
     // TODO: Include updates for all child skills if x.amount is negative
     // We update colonywide sums first (children, parents, skill)
     // Then the user-specifc sums in the order children, parents, skill.
-    await this.insert(colonyAddress, skillId, userAddress, score, updateNumber);
+
+    // Converting to decimal, since its going to be converted to hex inside `insert`
+    const skillIdDecimal = new BN(skillId, 16).toString();
```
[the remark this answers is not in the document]
filiplazovic: In case of skill id of 16:
I would get hex value [here](https://github.com/JoinColony/colonyNetwork/pull/294/files/8b9c8fb595fb476066b1ba198c9ad552039831e9#diff-9c228e1a1f7fe6683164300c770b62ffR254) (0000000000000000000000000000000000000000000000000000000000000010)
and then inside `insert` it will be converted to hex again and we will get (000000000000000000000000000000000000000000000000000000000000000a)
which is wrong skill id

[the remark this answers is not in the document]
area: Oh darn, yeah. I guess this stuff hasn't actually seen such high skill ids before - that's entirely on me. This feels like a bit of a hacky solution, though; I think a more robust solution would be for `breakKeyInToElements` to return `0x`-prefixed strings.

PR #295, opened 2018-08-01
Add contract and test error messages

### LINKED ISSUE #201: Add contract error messages

KevinLiLu:
Closes #201 

Error messages are prefixed with `colony-`, `colony-reputation-mining` and `colony-task` to identify better the error source. 

`solium` rules for missing error messages as well as long lines are switched to errors to force us to use revert errors going forward. It helps avoid false positives in tests and makes debugging much easier.

### DIFF
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -122,7 +122,7 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
     uint256 balance;
     (, balance) = ITokenLocking(tokenLockingAddress).getUserLock(clnyTokenAddress, msg.sender);
     require(entryIndex <= balance / 10**15, "colony-reputation-mining-stake-minimum-not-met");
-    require(entryIndex > 0);
+    require(entryIndex > 0, "colony-reputation-mining-zero-entry-index-passed");
     // If this user has submitted before during this round...
     if (reputationHashSubmissions[msg.sender].proposedNewRootHash != 0x0) {
       // ...require that they are submitting the same hash ...
@@ -648,7 +648,9 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
 
     // Check that the supplied log entry corresponds to this update number
     require(updateNumber >= logEntry.nPreviousUpdates, "colony-reputation-mining-update-number-part-of-previous-log-entry-updates");
-    require(updateNumber < logEntry.nUpdates + logEntry.nPreviousUpdates, "colony-reputation-mining-update-number-part-of-following-log-entry-updates");
+    require(
+      updateNumber < logEntry.nUpdates + logEntry.nPreviousUpdates,
+      "colony-reputation-mining-update-number-part-of-following-log-entry-updates");
     uint expectedSkillId;
     address expectedAddress;
     (expectedSkillId, expectedAddress) = getExpectedSkillIdAndAddress(logEntry, updateNumber);
@@ -901,7 +903,8 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
     // TODO: we're calling this twice during submitJRH. Should only need to call once.
     uint256 reputationRootHashNNodes = IColonyNetwork(colonyNetworkAddress).getReputationRootHashNNodes();
 
-    uint256 nUpdates = reputationUpdateLog[nLogEntries-1].nUpdates + reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes;
+    uint256 nUpdates = reputationUpdateLog[nLogEntries-1].nUpdates +
+    reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes;
     bytes memory nUpdatesBytes = new bytes(32);
     disputeRounds[round][index].jrhNnodes = nUpdates + 1;
     bytes32 submittedHash = disputeRounds[round][index].proposedNewRootHash;
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -832,7 +832,7 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
       } else {
         // TODO: Is this safe? I think so, because even if there's over/underflows, they should
         // still be the same number.
-        require(int(agreeStateReputationValue)+logEntry.amount == int(disagreeStateReputationValue));
+        require(int(agreeStateReputationValue)+logEntry.amount == int(disagreeStateReputationValue), "colony-invalid-newest-reputation-proof");
       }
     }
   }
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -90,8 +90,8 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
   /// @notice A modifier that checks that the supplied `roundNumber` is the final round
   /// @param roundNumber The `roundNumber` to check if it is the final round
   modifier finalDisputeRoundCompleted(uint256 roundNumber) {
-    require (nSubmittedHashes - nInvalidatedHashes == 1);
-    require (disputeRounds[roundNumber].length == 1); //i.e. this is the final round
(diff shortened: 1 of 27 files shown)


test/token-locking.js
```diff
@@ -68,8 +68,7 @@ contract("TokenLocking", accounts => {
       await checkErrorRevert(
         tokenLocking.deposit(token.address, usersTokens, {
           from: userAddress
-        }),
-        "token-locking-transfer-failed"
```
KevinLiLu: The error message generated from `deposit` is actually empty, as a nested error is being generated and overrides the one we specified.

The expected error: https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/TokenLocking.sol#L69

The actual error occurs at: https://github.com/dapphub/dappsys-monolithic/blob/002389d43cf54e8f0b919fee1fc364b20ebdf626/base.sol#L52

Not approved means the `_approvals[src][msg.sender]` value is 0, and the test attempts to subtract 6 from 0.

elenadimitrova: Good spot @KevinLiLu 
Note the `require` failure in `DSMath` contract and any external libraries in general is out of scope for us

KevinLiLu: I'm just going to remove the require part then? 
https://github.com/JoinColony/colonyNetwork/blob/e5de2fac169a29c90749fc438db17bd08642cba8/contracts/TokenLocking.sol#L69

`transferFrom` always returns true unless there is some error (in which it errors instead of returning false).
https://github.com/dapphub/dappsys-monolithic/blob/002389d43cf54e8f0b919fee1fc364b20ebdf626/base.sol#L47-L61

elenadimitrova: yes sure

area: Doesn't removing the require go against what we were advised in review we got done of the vesting contract?

test/colony-funding.js
```diff
@@ -804,7 +807,7 @@ contract("Colony Funding", addresses => {
         colony.claimRewardPayout(payoutId2, initialSquareRoots, userReputation.toString(), totalReputation.toString(), {
           from: userAddress1
         }),
-        "colony-reward-payout-bad-id"
+        "colony-reward-payout-invalid-parameter-amount"
```
filiplazovic: The correct error message here should be `colony-token-locking-invalid-lock-id`. It says that there is invalid amount is because we are creating reward payout with colony's native token, which is not possible. Sorry, it was my mistake.
I'm not sure if i have permission to push a code to your branch so here is the correct test:
```javascript
const tokenArgs = getTokenArgs();
const newToken = await Token.new(...tokenArgs);
await fundColonyWithTokens(colony, newToken, initialFunding.toString());

await colony.startNextRewardPayout(otherToken.address);

const { logs } = await colony.startNextRewardPayout(newToken.address);
const payoutId2 = logs[0].args.id;

await checkErrorRevert(
  colony.claimRewardPayout(payoutId2, initialSquareRoots, userReputation.toString(), totalReputation.toString(), {
    from: userAddress1
  }),
  "colony-token-locking-invalid-lock-id"
);
```

KevinLiLu: Thanks @filiplazovic. I'll change it to the correct test.

KevinLiLu:
@area  Can you assist me with a `colony-network-mining` test?

CI job has the error: 

```
1) Contract: ColonyNetworkMining
       Misbehaviour during dispute resolution
         should correctly check the proof of the previously newest reputation, if necessary.:
     AssertionError: expected undefined to equal 'colony-invalid-newest-reputation-proof'
      at _callee$ (helpers/test-helper.js:107:12)
      at tryCatch (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as throw] (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:114:21)
      at step (node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
      at node_modules/babel-runtime/helpers/asyncToGenerator.js:30:13
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:182:7)

  2) Contract: ColonyNetworkMining
       "before each" hook for "should correctly require the proof of the reputation under dispute before and after the change in question":
     Error: Transaction: 0x794b7eced61292f319cc8bbe9f293ea5777ef7adfb8e27ad24971e38ebebfc90 exited with an error (status 0). 
     Please check that the transaction:
     - satisfies all conditions set by Solidity `require` statements.
     - does not trigger a Solidity `revert` statement.

      at StatusError.ExtendableError (node_modules/truffle/build/webpack:/packages/truffle-error/index.js:10:1)
      at new StatusError (node_modules/truffle/build/webpack:/packages/truffle-contract/lib/statuserror.js:32:1)
      at Promise.receipt (node_modules/truffle/build/webpack:/packages/truffle-contract/lib/handlers.js:109:1)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:182:7)
```

It is failing at :
https://github.com/JoinColony/colonyNetwork/blob/develop/test/colony-network-mining.js#L1411.

I saw a similar test failure comparing `undefined` to a specified message when another error happened before our `require` statement happened (that does not specify an error message).

test/token-locking.js
KevinLiLu: Ok I will leave it as is then. 

The current test case will not be able to check the `token-locking-transfer-failed` error message then... Do we need to create a new test case? This is the only place where we test the error message `token-locking-transfer-failed`.

KevinLiLu:
Alright, all the unit tests are passing now!

@area @elenadimitrova can you guys look into why the test-coverage CI is failing? It looks like all the error messages are not appearing (`undefined`)?

area:
This is because `solidity-coverage` uses an outdated version of `ganache`. You won't need to do anything to fix it other than update the `solidity-coverage` version once a new one is released. I'm working on it...

area:
Once #304 is merged, once you rebase, you should have coverage passing even with the error messages.

contracts/ReputationMiningCycle.sol
```diff
@@ -832,7 +832,7 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
       } else {
         // TODO: Is this safe? I think so, because even if there's over/underflows, they should
         // still be the same number.
-        require(int(agreeStateReputationValue)+logEntry.amount == int(disagreeStateReputationValue));
+        require(int(agreeStateReputationValue)+logEntry.amount == int(disagreeStateReputationValue), "colony-invalid-newest-reputation-proof");
```
KevinLiLu: @area @elenadimitrova 

Stuck on something again... When I make this change, the migration `6_setup_meta_colony` step fails immediately.

```
6_setup_meta_colony.js
======================
Transaction: 0xbe82e5ff07add1ae1181106abc5f5824b0275da50c3dcb38e16ce2a6863f8a44 exited with an error (status 0). 
     Please check that the transaction:
     - satisfies all conditions set by Solidity `require` statements.
     - does not trigger a Solidity `revert` statement.
```

I'm not sure how adding this error message can cause step 6 to fail.

Any suggestions on how to get a more meaningful error message to debug this?

elenadimitrova: It's failing on `startNextCycle` which I'm pretty sure has run out of gas :(
Without the error message, that call costs `6,582,315` gas and after you put it in it throws consuming all `6,721,975`. Although 140,000 seems like too much to add on by an extra error message. Continuing to investigate..

elenadimitrova: Can you leave a todo there for now, summarising the above. This will likely need a separate piece of work altogether. Hopefully it won't trip you into adding most of the other `ReputationMiningCycle` error messages

PR #299, opened 2018-08-02
Updated contract loader to exact version 1.5.4

filiplazovic:
The reason why greenkeeper wasn't updating `@colony/colony-js-contract-loader-fs` is because we set a version `^1.3.0` which would cover everything to version `2.0.0`.
I set the version to be exact, so we will get PR from greenkeeper on every patch update.

Question: Is there any reason why `@colony/colony-js-contract-loader-fs` is not in `devDependencies`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "web3-utils": "^1.0.0-beta.34"
   },
   "dependencies": {
-    "@colony/colony-js-contract-loader-fs": "^1.3.0"
+    "@colony/colony-js-contract-loader-fs": "1.5.4"
   },
   "private": true,
   "workspaces": [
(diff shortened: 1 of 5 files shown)


elenadimitrova:
It can be a dev dependency it seems so you can move it

filiplazovic:
Ok, added `greenkeeper.json` which will handle monorepos. According to https://github.com/greenkeeperio/greenkeeper/issues/594, after this is merged, it should create new initial PR.

package.json
```diff
@@ -45,6 +45,7 @@
   },
   "homepage": "https://github.com/JoinColony/colonyNetwork#readme",
   "devDependencies": {
+    "@colony/colony-js-contract-loader-fs": "1.5.4",
```
elenadimitrova: Does this even need to be in the root `package.json`? I suspect it is only used in the reputation miner

filiplazovic: It's used in `test/colony-network-mining.js`, `gasCosts` and will be in `test/colony-funding.js` after #294 is merged.

Issue #57, opened 2017-09-26
Emergency Shutdown


elenadimitrova:
### 7.9 Emergency Shutdown
Section 2.2 described a transaction from a whitelisted address can put a colony into ‘recovery mode’ during which the state can be edited, the effects of bugs can be corrected and upgrades can be made.
Similarly, the reputation mining process will also have an emergency stop-and-repair mechanism (to begin with). This will allow the whitelisted addresses to revert the reputation root hash to a previous version and halt all updates to the reputation state until the issues have been resolved (which will likely involve a contract upgrade). The colonies will be able to continue operations as usual using the reputations of the last valid state, which will be temporarily frozen and not decay.

kronosapiens:
@elenadimitrova is this separate functionality from whats in the pipeline in #270 ?

I suppose the reputation mining stop-and-repair is different, but the function to put the contract into recovery mode is under review.

elenadimitrova:
Yes it is although logic can be reused. #270 deals with recovery mode for colonies only while the reputation mining functionality is managed at the colony network level, e.g. `reputationRootHash` is stored in `ColonyNetworkStorage`
@area anything else to add?

area:
Not really. Some consideration needs to be given to what happens when a cycle is stopped mid-process, but I think most things can be solved by just deploying another `ReputationUpdateCycle`, and pointing colonyNetwork to the new one(s). Log entries can be manually moved over, if necessary, with custom functions on that 'transitional' `ReputationUpdateCycle`.

filiplazovic:
@area Should i be worried about log entries being corrupted/wrong?

elenadimitrova:
Just recording a discussion on this logic I had with @area 
Ultimately, the Meta colony will have the Recovery capability. This type of decisions there would normally go through a vote, but in most cases we'd want to enter recovery as soon as possible in which case the voting process will likely prove too slow. Instead we want an `Authority` for people to be able to put it in to that state, and potentially only the owner (ultimately the Meta Colony) to be able to take it out of that state.

elenadimitrova:
Just as a comparison to the recovery mode capability for a `Colony` https://github.com/JoinColony/colonyNetwork/pull/270

PR #303, opened 2018-08-07
Emergency shutdown

filiplazovic:


.circleci/config.yml
```diff
@@ -79,6 +79,9 @@ jobs:
       - run:
           name: "Checking contract recovery modifiers"
           command: yarn run check:recoverymods
+      - run:
+          name: "Checking contract repair modifiers"
+          command: yarn run check:repairmods
```
elenadimitrova: Can we merge this with `check:recoverymods` as well?

contracts/ColonyNetworkMining.sol
```diff
@@ -31,9 +31,33 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
...
+  modifier repair() {
+    require(inRepairMode, "colony-network-mining-not-in-repair-mode");
+    _;
+  }
+
+  function activateRepairMode() public auth {
+    inRepairMode = true;
+  }
+
+  function deactivateRepairMode() public auth {
```
elenadimitrova: Regarding `auth` permissions here, we possibly need a Recovery role on a Network authority level similarly to https://github.com/JoinColony/colonyNetwork/pull/270/files#diff-bbdfafcbef1d71c176dea0989160badd

contracts/ReputationMiningCycle.sol
```diff
@@ -152,8 +167,8 @@ contract ReputationMiningCycle is PatriciaTreeProofs, DSMath {
   }
 
   /// @notice Constructor for this contract.
-  constructor(address _tokenLockingAddress, address _clnyTokenAddress) public {
-    colonyNetworkAddress = msg.sender;
+  constructor(address _colonyNetwoekAddress, address _tokenLockingAddress, address _clnyTokenAddress) public {
```
elenadimitrova: typo

contracts/IColonyNetwork.sol
```diff
@@ -146,19 +151,59 @@ contract IColonyNetwork {
   /// @return resolverAddress Address of the `Resolver` contract
   function getColonyVersionResolver(uint256 _version) public view returns (address resolverAddress);
 
+  /// @notice Activates recovery mode
+  /// @dev Can only be called by authorised address
+  function stop() public;
```
elenadimitrova: These are actually called differently in the Colony recovery implementation. Also only the trivial stop/start functionality is being refactored whereas the entire recovery review and exit is not.

contracts/ColonyNetworkStorage.sol
```diff
@@ -68,6 +68,15 @@ contract ColonyNetworkStorage is DSAuth, DSMath {
...
   // The number of nodes in the reputation state tree that was accepted at the end of the last mining cycle
   uint256 reputationRootHashNNodes;
+
+  struct ReputationRootHash {
+    bytes32 rootHash;
+    uint256 nNodes;
+  }
+
+  // History of all reputation hashes
+  ReputationRootHash[] reputationRootHashHistory;
```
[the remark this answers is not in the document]
elenadimitrova: While rebasing Alex's PR https://github.com/JoinColony/colonyNetwork/pull/311 I noticed his approach to retrieve historical  values is by looking at the events triggered when a new root hash is successfully confirmed. We can't use events here but in #311 we could use the history array for consistency. I'd prefer to wait for Alex on Monday to decide how to align these two.

filiplazovic:
@elenadimitrova After talking with Alex, we decided to simplify this implementation by making something similar to recovery mode for Colony. Since current implementation is limited, in a way that we can't change reputation state to the correct one (we can just revert to previous one) and `ColonyNetworkMining` was hitting gas limit, we will allow changes of reputation state and reputation update logs directly in the storage, and hopefully fix those issues

elenadimitrova:
@filiplazovic let us know the state of this work in a way of a handover so either myself or @area can finish it?

filiplazovic:
I dont think there is anything to add to PR description. I think there are tests left for checking if miner is ignoring bad update logs, and if challanges can be completed correctly. The best way to understand is to dig into the code

area:
This PR now adds a full 'recovery mode' to ColonyNetwork, the same as Colony has. In order to avoid code duplication, I have created a CommonStorage contract that holds the storage variables related to recovery mode that is inherited by both ColonyStorage and ColonyNetworkStorage. In the case of Colony, there is an extra protected variable; `setStorageSlotRecovery` calls `checkNotAddtionalProtectedVariable` in order to make this check, and so both contrats implement that function, even though in ColonyNetwork it does nothing.

Because ColonyNetwork has a DSAuth owner, rather than a role that we call the owner (as is the case with Colony), I have tweaked the check in `exitRecoveryMode` to explicitly check if the owner is set, and if it is, check if it has the recovery role, in order to avoid double-counting the owner and having the threshold wrong for exiting recovery mode. I *think* this correctly deals with all possible schenanigans around owners being assigned the recovery role, changing owner to another address that either does or doesn't already have the recovery role, and so forth.

There is now an Authority contract for the ColonyNetwork; this records who has the recovery role. There are currently no other roles.

The recovery mode allows us to deal with errors in the mining cycle that we find while the logs are still on-chain. The approach for doing this is via deploying new reputation mining cycles, porting over any logs, correcting any logs that are wrong, and then changing the addresses of the ReputationMiningCycle instances that ColonyNetwork claims are the active and inactive cycles. This (half-) deals with case (3) detailed in [this comment](https://github.com/JoinColony/colonyNetwork/issues/57#issuecomment-414640849).

If a bad state is committed, we are able to enter recovery mode, set a 'good' state, and then deal with whatever other changes are required - no other changes, in case 2(a), deploying new instances of ReputationMiningCycle (as above), and updating ColonyNetwork so all future instances of ReputationMiningCycle are fixed in the case of 2(b), and in addition releasing an update to the mining client in the case of 2(c).

Note that by freezing mid-dispute and redeploying new ReputationMiningCycle instances, anyone who is defending a wrong state in good faith will not be punished. Unfortunately, any compensation for good-faith defences of bad states that has already been punished will have to be paid for our of our own pocket.

Case (1) is also dealt with by being able to set the reputation state directly.

The other major piece of functionality this PR introduces deals with the other half of case (3), where the incorrect logs in the reputation state are not on-chain any longer, because the reputation cycle they were in has been completed and their effects are incorporated in to the reputation state. This is dealt with via `setReplacementReputationUpdateLogEntry`, which allows us to specify an 'overriden' value for a historical log entry. When a reputation miner syncs from scratch, they check the ColonyNetwork contract to see if they should use any replacement log entries when processing logs. The introduction of this feature means that in the event we do have to change historical logs, a client update shouldn't be required even for those people who want to sync from scratch. The mining client does not check every single log entry, just those in mining cycles where it has been noted (on-chain) at least one log entry has been replaced.

There is a dummy Contract called ContractEditing that is used for testing purposes, and hypothetically could be temporarily added to new ReputationMiningCycles being deployed in order to correctly port log entries across. This is done in tests.

I introduce another acceptable recovery mode modifier, `always`, to indicate a function that should work in both recovery mode and not. Currently, only the `upgrade` function for a colony is blessed with this power. Exiting recovery mode needed to work the same for both ColonyNetwork and Colony, and so having the `version` parameter no longer made sense. If a colony wishes to upgrade prior to exiting recovery mode, they call the `upgrade` function as normal.


### Fixes 
In the miningClient, `loadState` now correctly zeros the current state before adding the loaded log entries to the tree.

test/colony-network-recovery.js
```diff
@@ -0,0 +1,498 @@
...
+          await forwardTime(3600, this);
+          await miningClient.submitRootHash();
+
+          addr = await colonyNetwork.getReputationMiningCycle.call(true);
+          repCycle = await IReputationMiningCycle.at(addr);
+
+          const invalidEntry = await repCycle.getReputationUpdateLogEntry(5);
+          invalidEntry.amount = 0;
+
+          await repCycle.confirmNewHash(0);
```
elenadimitrova: Here the currently active reputation log is destructed, the inactive becomes active and a blank inactive cycle is created.

test/colony-network-recovery.js
```diff
@@ -0,0 +1,498 @@
...
+          await colonyNetwork.setReplacementReputationUpdateLogEntry(
+            addr,
+            5,
+            invalidEntry.user,
+            invalidEntry.amount,
+            invalidEntry.skillId,
+            invalidEntry.colony,
+            invalidEntry.nUpdates,
+            invalidEntry.nPreviousUpdates
+          );
```
elenadimitrova: This sets a what is now a destructed ReputationMiningCycle entry, for the purposes of correct future client `sync` calls

test/colony-network-recovery.js
```diff
@@ -0,0 +1,498 @@
...
+          await miningClient.loadState(startingHash);
+          // This sync call will log an error - this is because we've changed a log entry, but the root hash
+          // on-chain which .sync() does a sanity check against hasn't been updated.
+          await miningClient.sync(startingBlockNumber, true);
```
elenadimitrova: This syncs with the recovered log entry update.

test/colony-network-recovery.js
```diff
@@ -0,0 +1,498 @@
...
+          console.log("The WARNING and ERROR immediately preceeding can be ignored (they are expected as part of the test)");
+          const rootHash = await miningClient.getRootHash();
+          const nNodes = await miningClient.nReputations;
```
elenadimitrova: We get the "right" root hash and nNodes and set them on the network to correct the "wrong" ones from the previous update.

PR #309, opened 2018-08-14
Refactoring task update calls

### LINKED ISSUE #296: Refactoring task update calls
Due to the number of transactions required in a task lifecycle and the granular approach we've taken in the contract design vs product design we'd like to look for opportunities to pull some actions together into a single function call and therefore a single transaction interaction in the UI. Below is the initial set of ideas for what can be bundled but more could come of of the refactoring if opportunities are found:

- Setting the evaluator to the task creator (similarly to how the manager is assigned) as for the Colony Contribute these 2 roles will be fulfilled by the same person. We'll still allow for the role to be reassigned later
- Submitting the deliverable and rating for the task manager. If they are submitting the task, the immediate next thing they need to do is rate the manager
- Changing task role assignments currently requires `removeTaskXRole` followed by `setTaskXRole`, perhaps these can be part of the same call?
- Public functions `assignWorkRating` + `finalizeTask`?

#251 is already going to handle the `makeTask` changes to allow in addition to the mandatory `specificationHash` and `domainId` to also set the `dueDate`, `skillId` as well as payouts for all 3 roles

kronosapiens:
Closes #296

- [x] Setting the evaluator to the task creator (similarly to how the manager is assigned) as for the Colony Contribute these 2 roles will be fulfilled by the same person. We'll still allow for the role to be reassigned later
- [x] Submitting the deliverable and rating for the task manager. If they are submitting the task, the immediate next thing they need to do is rate the manager
- [x] Public functions assignWorkRating + finalizeTask?
- [ ] Changing task role assignments currently requires removeTaskXRole followed by setTaskXRole, perhaps these can be part of the same call?

SO:

- Automatically assign task creator as both manager and evaluator
- Create `submitTaskDeliverableAndRating` function which wraps `submitTaskDeliverable` and `submitTaskWorkRating`
- Make `assignWorkRating` an internal function and bundle it into a more general `finalizeTask`.

Will not create a combined `reassignTaskXRole` function now since it involves doing something fancy with signatures.

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -114,12 +114,8 @@ contract ColonyTask is ColonyStorage {
     task.domainId = _domainId;
     task.skills = new uint256[](1);
     tasks[taskCount] = task;
-    tasks[taskCount].roles[MANAGER] = Role({
-      user: msg.sender,
-      rateFail: false,
-      rating: TaskRatings.None
-    });
-
+    tasks[taskCount].roles[MANAGER].user = msg.sender;
+    tasks[taskCount].roles[EVALUATOR].user = msg.sender;
     pots[potCount].taskId = taskCount;
 
     emit PotAdded(potCount);
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -395,6 +395,7 @@ contract ColonyTask is ColonyStorage {
 
   function finalizeTask(uint256 _id) public
   stoppable
+  taskExists(_id)
   canFinalizeTask(_id)
   taskNotFinalized(_id)
   {
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -95,16 +95,8 @@ contract ColonyTask is ColonyStorage {
     _;
   }
 
-  modifier taskWorkRatingsClosed(uint256 _id) {
-    uint taskCompletionTime = tasks[_id].deliverableTimestamp != 0 ? tasks[_id].deliverableTimestamp : tasks[_id].dueDate;
-    // More than 10 days from work submission have passed
-    require(sub(now, taskCompletionTime) > add(RATING_COMMIT_TIMEOUT, RATING_REVEAL_TIMEOUT), "colony-task-rating-period-still-open");
-    _;
-  }
-
-  modifier taskWorkRatingsAssigned(uint256 _id) {
-    require(tasks[_id].roles[WORKER].rating != TaskRatings.None, "colony-task-worker-rating-missing");
-    require(tasks[_id].roles[MANAGER].rating != TaskRatings.None, "colony-task-manager-rating-missing");
+  modifier canFinalizeTask(uint256 _id) {
+    require(taskWorkRatingsAssigned(_id) || taskWorkRatingsClosed(_id), "colony-task-cannot-finalize");
     _;
   }
 
@@ -297,28 +289,6 @@ contract ColonyTask is ColonyStorage {
     emit TaskWorkRatingRevealed(_id, _role, _rating);
   }
 
-  // In the event of a user not committing or revealing within the 10 day rating window,
-  // their rating of their counterpart is assumed to be the highest possible
-  // and they will receive a reputation penalty
-  function assignWorkRating(uint256 _id) public
-  stoppable
-  taskWorkRatingsClosed(_id)
-  {
-    Role storage managerRole = tasks[_id].roles[MANAGER];
-    Role storage workerRole = tasks[_id].roles[WORKER];
-    Role storage evaluatorRole = tasks[_id].roles[EVALUATOR];
-
-    if (workerRole.rating == TaskRatings.None) {
-      evaluatorRole.rateFail = true;
-      workerRole.rating = TaskRatings.Excellent;
-    }
-
-    if (managerRole.rating == TaskRatings.None) {
-      workerRole.rateFail = true;
-      managerRole.rating = TaskRatings.Excellent;
-    }
-  }
-
   function generateSecret(bytes32 _salt, uint256 _value) public pure returns (bytes32) {
     return keccak256(abi.encodePacked(_salt, _value));
   }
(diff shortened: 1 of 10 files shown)


kronosapiens:
@elenadimitrova On the topic of merging `assignWorkRating` and `finalizeTask`, I'm inclined to merge them together entirely and have a more involved function which assigns ratings if they've not been assigned and then does the rest of the finalize task logic. Is there a need for a dedicated `assignWorkRating` function?

elenadimitrova:
@kronosapiens yeah exactly my thinking and why I suggested it here in the first place 👍

test/colony.js
```diff
@@ -614,15 +657,6 @@ contract("Colony", accounts => {
...
       const taskId = await makeTask({ colony });
 
-      await executeSignedRoleAssignment({
-        colony,
-        taskId,
-        functionName: "setTaskEvaluatorRole",
-        signers: [MANAGER],
-        sigTypes: [0],
-        args: [taskId, MANAGER]
-      });
```
elenadimitrova: We can remove the unnecessary assert that checks this too further down in test, i.e. 
```
const evaluatorInfo = await colony.getTaskRole(taskId, EVALUATOR_ROLE);
assert.equal(evaluatorInfo[0], MANAGER);
```

test/colony.js
```diff
@@ -664,34 +687,22 @@ contract("Colony", accounts => {
...
-        taskId,
-        functionName: "setTaskEvaluatorRole",
-        signers: [MANAGER, EVALUATOR],
-        sigTypes: [0, 0],
-        args: [taskId, EVALUATOR]
-      });
-
       let evaluator = await colony.getTaskRole(taskId, EVALUATOR_ROLE);
-      assert.equal(evaluator[0], EVALUATOR);
+      assert.equal(evaluator[0], MANAGER);
```
elenadimitrova: We should really leave `EVALUATOR` here. Any logic that checks that the evaluator role is in fact assigned to the manager may lead to misunderstandings

test/colony.js
```diff
@@ -1558,9 +1552,9 @@ contract("Colony", accounts => {
 
       // Run through the task flow
       await colony.submitTaskDeliverable(taskId, DELIVERABLE_HASH, { from: WORKER });
-      await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: EVALUATOR });
+      await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: MANAGER });
       await colony.submitTaskWorkRating(taskId, MANAGER_ROLE, RATING_1_SECRET, { from: WORKER });
-      await colony.revealTaskWorkRating(taskId, WORKER_ROLE, WORKER_RATING, RATING_2_SALT, { from: EVALUATOR });
+      await colony.revealTaskWorkRating(taskId, WORKER_ROLE, WORKER_RATING, RATING_2_SALT, { from: MANAGER });
```
elenadimitrova: As above, let's keep the `EVALUATOR` constant reference here. It is whoever is assigned to the evaluator role that rates even if they happen to be the same person as the manager.

helpers/test-data-generator.js
```diff
@@ -54,7 +54,7 @@ export async function executeSignedRoleAssignment({ colony, functionName, taskId
 export async function setupAssignedTask({ colonyNetwork, colony, dueDate, domain = 1, skill = 0, evaluator, worker }) {
   const accounts = await web3GetAccounts();
   const manager = accounts[0];
-  evaluator = evaluator || accounts[1]; // eslint-disable-line no-param-reassign
```
elenadimitrova: We default the evaluator to `accounts[1]` in a few more places in the test-data-generator which need changing

gasCosts/gasCosts.js
```diff
@@ -211,11 +202,11 @@ contract("All", accounts => {
       await colony.submitTaskDeliverable(taskId, DELIVERABLE_HASH, { from: WORKER, gasPrice });
 
       // submitTaskWorkRating
-      await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: EVALUATOR, gasPrice });
+      await colony.submitTaskWorkRating(taskId, WORKER_ROLE, RATING_2_SECRET, { from: MANAGER, gasPrice });
```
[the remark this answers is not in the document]
kronosapiens: I was just thinking about that, actually.

helpers/test-data-generator.js
```diff
@@ -82,25 +82,33 @@ export async function setupAssignedTask({ colonyNetwork, colony, dueDate, domain
...
-  signers = manager === worker ? [manager] : [manager, worker];
-  sigTypes = Array.from({ length: signers.length }, () => 0);
+  const signers = manager === worker ? [manager] : [manager, worker];
+  const sigTypes = Array.from({ length: signers.length }, () => 0);
 
   await executeSignedRoleAssignment({
     colony,
-    taskId,
     functionName: "setTaskWorkerRole",
+    taskId,
```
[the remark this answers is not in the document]
kronosapiens: Ok.

gasCosts/gasCosts.js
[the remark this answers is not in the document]
kronosapiens: Yup

helpers/test-data-generator.js
kronosapiens: Updated

test/colony.js
kronosapiens: Yup

test/colony.js
kronosapiens: Ok, but I'd like to add a note then that only the manager's signature is required on the unassignment because `manager === evaluator`

test/colony.js
kronosapiens: Yup

helpers/test-data-generator.js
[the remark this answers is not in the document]
kronosapiens: Done

contracts/ColonyTask.sol
```diff
@@ -420,12 +386,23 @@ contract ColonyTask is ColonyStorage {
...
+  {
+    submitTaskDeliverable(_id, _deliverableHash);
+    submitTaskWorkRating(_id, 0, _ratingSecret);
+  }
+
   function finalizeTask(uint256 _id) public
   stoppable
   taskExists(_id)
-  taskWorkRatingsAssigned(_id)
+  canFinalizeTask(_id)
```
elenadimitrova: Rename `canFinalizeTask` to `taskWorkRatingsComplete`? 
canFinalize implies potentially the other modifier rules we have in there, e.g. task exists and task not already finalized

elenadimitrova: The associated error message `colony-task-cannot-finalize` can also be improved to `colony-task-ratings-incomplete`

PR #312, opened 2018-08-15
Upgrade truffle to next.9

cgewecke:
`next.9` upgrades web3 to `beta.35`. They changed the way the `status` field on the receipt is formatted from hex number to boolean. Just couple places in one of the test helpers where this value is checked directly - have updated those.

And `getNodeInfo` is available so have re-enabled the client check for `forwardTime`.


area:
:heart:

PR #320, opened 2018-08-22
Move ReputationMiningCycle to an EtherRouter pattern

area:
This has been on the horizon for a while, to massively reduce the cost of calling `confirmNewHash` (from 4472809 gas to 690445, according to the gas tests... which I think could still come down, but I'll take this for now!), which had finally caused us to come a cropper (in Fillip's work on #303, and we were also expecting it to rear its head in #280 when Elena rebased) when adding new functionality to ReputationMiningCycle.

The only thing I've not given any consideration to is the permissions on the resolver. I think only ColonyNetwork should be able to make changes to the resolver, and only when in recovery mode. This removes the need (in most instances) to 'port' log entries between old and new ReputationMiningCycles if we are sufficiently clever enough with our new ReputationMiningCycle implementation.

EDIT: Alternatively, it would be cheaper to use libraries, but we wouldn't be able to do in-place upgrades, and we would have to port log entries, which would cost us an arbitrarily large amount...

It does mean that, in theory, recovery mode lets us cause miners to lose their tokens (by changing the functionality so that we can directly call `punishStaker` for whoever), but I think that was going to be true so long as recovery mode for reputation mining exists and is under our control.


contracts/IReputationMiningCycle.sol
```diff
@@ -196,5 +196,22 @@ contract IReputationMiningCycle {
   /// same address backed the same hash multiple times with different entries.
   function rewardStakersWithReputation(address[] stakers, address commonColonyAddress, uint reward, uint miningSkillId) public;
 
-  function reputationMiningWindowOpenTimestamp() public view returns (uint timestamp);
+  /// @notice Get the timestamp that the current reputation mining window opened
+  function getReputationMiningWindowOpenTimestamp() public view returns (uint timestamp);
```
elenadimitrova: Return `uint` -> `uint256`

contracts/IReputationMiningCycle.sol
```diff
@@ -196,5 +196,22 @@ contract IReputationMiningCycle {
...
+
+  /// @notice Initialise this reputation mining cycle.
+  /// @dev This will only be called once, by ColonyNetwork, in the same transaction that deploys this contract
+  function initialise(address tokenLocking, address clnyToken) public;
+
+  /// @notice Get the number of hashes that have been submitted this mining cycle
+  function getNSubmittedHashes() public view returns (uint nSubmittedHashes);
```
elenadimitrova: Return `uint256`

contracts/IReputationMiningCycle.sol
```diff
@@ -196,5 +196,22 @@ contract IReputationMiningCycle {
...
+
+  /// @notice Get the number of hashes that have been invalidated this mining cycle
+  function getNInvalidatedHashes() public view returns (uint nInvalidatedHashes);
```
elenadimitrova: Return `uint256`

contracts/ReputationMiningCycleRespond.sol
```diff
@@ -0,0 +1,387 @@
...
+  uint constant U_AGREE_STATE_NNODES = 3;
+  uint constant U_AGREE_STATE_BRANCH_MASK = 4;
+  uint constant U_DISAGREE_STATE_NNODES = 5;
+  uint constant U_DISAGREE_STATE_BRANCH_MASK = 6;
+  uint constant U_PREVIOUS_NEW_REPUTATION_BRANCH_MASK = 7;
+  uint constant U_REQUIRE_REPUTATION_CHECK = 8;
+  uint constant U_LOG_ENTRY_NUMBER = 9;
+  uint constant U_DECAY_TRANSITION = 10;
+
+  function respondToChallenge(
```
elenadimitrova: `solium` complains about a blank line rule broken here

test/colony-network-mining.js
```diff
@@ -793,7 +793,7 @@ contract("ColonyNetworkMining", accounts => {
       );
     });
 
-    it('should not allow "startNextCycle" to be called if a cycle is in progress', async () => {
+    it('should not allow "repCycle.nSubmittedHashesCycle" to be called if a cycle is in progress', async () => {
```
elenadimitrova: Is this a mass rename gone wrong? The original test title was correct.

test/colony.js
```diff
@@ -87,6 +89,11 @@ contract("Colony", accounts => {
     const clnyToken = await Token.new("Colony Network Token", "CLNY", 18);
     await colonyNetwork.createMetaColony(clnyToken.address);
 
+    const reputationMiningCycle = await ReputationMiningCycle.new();
+    const reputationMiningCycleRespond = await ReputationMiningCycleRespond.new();
+    const reputationMiningCycleResolver = await Resolver.new();
+    await setupReputationVersionResolver(reputationMiningCycle, reputationMiningCycleRespond, reputationMiningCycleResolver, colonyNetwork);
```
elenadimitrova: There's no need to wire this up again, we've done it already in migrations and can use the existing `miningCycleResolver`.

test/meta-colony.js
```diff
@@ -45,6 +47,11 @@ contract("Meta Colony", accounts => {
     const metaColonyAddress = await colonyNetwork.getMetaColony();
     metaColony = await IColony.at(metaColonyAddress);
 
+    const reputationMiningCycle = await ReputationMiningCycle.new();
+    const reputationMiningCycleRespond = await ReputationMiningCycleRespond.new();
+    const reputationMiningCycleResolver = await Resolver.new();
+    await setupReputationVersionResolver(reputationMiningCycle, reputationMiningCycleRespond, reputationMiningCycleResolver, colonyNetwork);
```
elenadimitrova: Same comment as for the `colony` test setup, this is already done in migrations.

test/reputation-update.js
```diff
@@ -59,10 +61,16 @@ contract("Colony Reputation Updates", accounts => {
       .mul(new BN(1000))
       .toString();
     await fundColonyWithTokens(metaColony, colonyToken, amount);
+
+    const reputationMiningCycle = await ReputationMiningCycle.new();
+    const reputationMiningCycleRespond = await ReputationMiningCycleRespond.new();
+    const reputationMiningCycleResolver = await Resolver.new();
+    await setupReputationVersionResolver(reputationMiningCycle, reputationMiningCycleRespond, reputationMiningCycleResolver, colonyNetwork);
```
elenadimitrova: Same again

contracts/IColonyNetwork.sol
```diff
@@ -189,4 +189,6 @@ contract IColonyNetwork {
   /// @notice Register a "colony.joincolony.eth" label. Can only be called by a Colony.
   /// @param subnode The keccak256 hash of the label to register.
   function registerColonyLabel(bytes32 subnode) public;
+
+  function setMiningResolver(address miningResolverAddress) public;
```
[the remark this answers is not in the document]
area: Given the contracts are relatively short lived, if we wanted to upgrade the mining cycle contracts while not in recovery mode, I would in the first instance try to just change `miningCycleResolver` on colonyNetwork and have all future instances of the contract be the upgraded ones, rather than upgrading the active ones in-place.

If we do need to upgrade them in place though, then yeah, I'm on board with only being able to do that in recovery mode.

helpers/upgradable-contracts.js
```diff
@@ -113,3 +113,13 @@ export async function setupUpgradableTokenLocking(etherRouter, resolver, tokenLo
   const registeredResolver = await etherRouter.resolver();
   assert.equal(registeredResolver, resolver.address);
 }
+
+export async function setupReputationVersionResolver(reputationMiningCycle, reputationMiningCycleRespond, resolver, colonyNetwork) {
```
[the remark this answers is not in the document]
area: I have gone for `setupReputationMiningCycleResolver`. `setupUpgradableReputationMiningCycle` made it sound like we were setting up an instance of an upgradeable ReputationMiningCycle, which isn't really what it's doing.

contracts/ReputationMiningCycleRespond.sol
area: Fixed, but this was only a warning, which is why linting passed on the precommit hook. Should we make it an error?

test/colony-network-mining.js
area: I have absolutely no clue what happened here 😕

contracts/ReputationMiningCycleRespond.sol
elenadimitrova: I have been going for selectively switching _on_ `solium` warning rules to errors, like line length for example but I find we need to go for maximum strictness on contracts linting so I like the suggestion of just making all warnings errors and selectively switch _off_ rules if we feel the need.

PR #325, opened 2018-08-29
Give miner ability to persist reputation states

### LINKED ISSUE #302: Allow reputation mining state to persist
Currently, when the reputation mining client is closed, it forgets the current state. Instead, the current state should persist between reboots. Ideally, it should also 'catch-up' if the state has changed between when it was closed and reopened, likely using functionality introduced in #301.

area:
Closes #302 

## Implementation details

I have introduced an [SQLite](https://www.sqlite.org/) database to the miner for persisting state. This is an internal SQL instance (i.e. not a separate server) that persists to disk invisibly (i.e. no need to explicitly call a 'save' function on the database instance created. The schema of the database I have used is defined by

```
CREATE TABLE users ( address text NOT NULL UNIQUE );
CREATE TABLE reputation_states ( root_hash text NOT NULL UNIQUE, n_nodes INTEGER NOT NULL );
CREATE TABLE colonies ( address text NOT NULL UNIQUE );
CREATE TABLE skills ( skill_id INTEGER PRIMARY KEY );
CREATE TABLE reputations ( root_hash text NOT NULL, colony_address INTEGER NOT NULL, skill_id INTEGER NOT NULL, user_address INTEGER NOT NULL, value text NOT NULL);
```
Note that each table (apart from `skills`, which is defined with `PRIMARY KEY`) also has a [`rowid`](https://www.sqlite.org/lang_createtable.html#rowid) column, which acts as the primary key for each table. Note that while `address` would be a good natural key for `users`, if it appeared repeatedly as a foreign key in `reputations`, there would be a great deal of wasted disk space. By using surrogate keys, the drive requirements for a reputation miner are vastly reduced.

Individual reputations are stored in the `reputations` table. By storing these, we are able to recreate any past reputation state that the miner has stored, and in turn any reputation proof for a past state. This will be a requirement in the future when miners are acting as oracles, and must provide users proofs for their historical reputations, not only the current one (e.g. when a reward payout is occurring). I believe this is a reasonable tradeoff between storage space required, ease of development, and computation required to respond. For example, the same would be possible if we just stored all logs (which would require much less space), but having to replay all history to recreate an old state would be interminably slow going forward.

Going forward, old states that are stored that no longer have a use (e.g. no reward payouts or disputes relying on a particular state) can be purged to recover disk space.

It has been many years since I had to think about optimising SQL, and I'm sure what I have here could be a lot better. In particular, I have completely ignored having to deal with race conditions by having each query implement sequentially (leading to a lot of `//eslint-disable-line no-await-in-loop`). I don't particularly want to waste a lot of time improving those now, but if someone more knowledgeable than me could spot something easy, I'd cheerfully do it.

## Other changes

I have added the ability to call `sync` with a new parameter; if set to true, then all (valid) states that are encountered are saved to the database. This will be useful for a new miner coming online that wishes to act as a reputation oracle.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -775,9 +775,7 @@ class ReputationMiner {
     const db = await sqlite.open(this.dbPath, { Promise });
 
     const currentRootHash = await this.getRootHash();
-    let res = await db.all(`SELECT rowid, root_hash FROM reputation_states WHERE root_hash='${currentRootHash}' AND n_nodes='${this.nReputations}'`);
-
-    res = await db.run(`INSERT OR IGNORE INTO reputation_states (root_hash, n_nodes) VALUES ('${currentRootHash}', ${this.nReputations})`);
+    let res = await db.run(`INSERT OR IGNORE INTO reputation_states (root_hash, n_nodes) VALUES ('${currentRootHash}', ${this.nReputations})`);
 
     for (let i = 0; i < Object.keys(this.reputations).length; i += 1) {
       const key = Object.keys(this.reputations)[i];
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -777,18 +777,9 @@ class ReputationMiner {
     const currentRootHash = await this.getRootHash();
     let res = await db.all(`SELECT rowid, root_hash FROM reputation_states WHERE root_hash='${currentRootHash}' AND n_nodes='${this.nReputations}'`);
 
-    let rootHashRowId;
-    if (res.length === 0) {
-      res = await db.run(`INSERT INTO reputation_states (root_hash, n_nodes) VALUES ('${currentRootHash}', ${this.nReputations})`);
-      rootHashRowId = res.lastID;
-    } else {
-      rootHashRowId = res[0].rowid;
-    }
+    res = await db.run(`INSERT OR IGNORE INTO reputation_states (root_hash, n_nodes) VALUES ('${currentRootHash}', ${this.nReputations})`);
 
     for (let i = 0; i < Object.keys(this.reputations).length; i += 1) {
-      let colonyRowId;
-      let userRowId;
-      let skillRowId;
       const key = Object.keys(this.reputations)[i];
       const value = this.reputations[key];
       let [colonyAddress, skillId, userAddress] = await ReputationMiner.breakKeyInToElements(key); // eslint-disable-line no-await-in-loop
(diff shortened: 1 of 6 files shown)


packages/reputation-miner/ReputationMiner.js
```diff
@@ -758,12 +761,131 @@ class ReputationMiner {
...
+    await db.run("CREATE TABLE users ( address text NOT NULL UNIQUE )");
+    await db.run("CREATE TABLE reputation_states ( root_hash text NOT NULL UNIQUE, n_nodes INTEGER NOT NULL )");
+    await db.run("CREATE TABLE colonies ( address text NOT NULL UNIQUE )");
+    await db.run("CREATE TABLE skills ( skill_id INTEGER PRIMARY KEY )");
+    await db.run(
+      `CREATE TABLE reputations (
+        root_hash text NOT NULL,
+        colony_address INTEGER NOT NULL,
+        skill_id INTEGER NOT NULL,
+        user_address INTEGER NOT NULL,
```
elenadimitrova: The column namings confused me at first when I saw them for `root_hash`, `colony_address`, `skill_id` and `user_address` as those are not holding respectively a root hash, address, skill id or an address but contain sql table identifiers. We should name them perhaps postfix them with `dbId` and remove `address` from the names? e.g. `root_hash_dbId`, `colony_dbId`, `skill_dbId` and `user_dbId?

packages/reputation-miner/ReputationMiner.js
```diff
@@ -758,12 +761,131 @@ class ReputationMiner {
...
+      }
+
+      // eslint-disable-next-line no-await-in-loop
+      res = await db.get(
+        `SELECT COUNT ( * ) AS "n"
+        FROM reputations
+        WHERE root_hash=${rootHashDBId}
+        AND colony_address=${colonyDBId}
+        AND skill_id=${skillDBId}
+        AND user_address=${userDBId}`
```
elenadimitrova: The right way to do this is with inner joins instead of inferring the rowIds from the previous exciting or just inserted records. All four fields are unique keys in the other tables.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -758,12 +761,131 @@ class ReputationMiner {
...
+      if (res.length === 0) {
+        res = await db.run(`INSERT INTO users (address) VALUES ('${userAddress}')`); // eslint-disable-line no-await-in-loop
+        userDBId = res.lastID;
+      } else {
+        userDBId = res[0].rowid;
+      }
+
+      res = await db.all(`SELECT skill_id FROM skills WHERE skill_id='${skillId}'`); // eslint-disable-line no-await-in-loop
+      if (res.length === 0) {
+        res = await db.run(`INSERT INTO skills (skill_id) VALUES ('${skillId}')`); // eslint-disable-line no-await-in-loop
```
elenadimitrova: All of the above `SELECT` \ `INSERT` statements can be instead bundled with `WHERE NOT EXISTS` statements e.g.
```
INSERT INTO skills (skill_id) VALUES ('${skillId}')
WHERE NOT EXISTS (SELECT skill_id FROM skills WHERE skill_id='${skillId}')
```

packages/reputation-miner/ReputationMiner.js
```diff
@@ -758,12 +761,131 @@ class ReputationMiner {
...
+        AND colony_address=${colonyDBId}
+        AND skill_id=${skillDBId}
+        AND user_address=${userDBId}`
+      );
+
+      if (res.n === 0) {
+        // eslint-disable-next-line no-await-in-loop
+        await db.run(
+          `INSERT INTO reputations (root_hash, colony_address, skill_id, user_address, value)
+          VALUES (${rootHashDBId}, ${colonyDBId}, ${skillDBId}, ${userDBId}, '${value}')`
```
elenadimitrova: We should really bundle up all `INSERT`s into a single transaction to ensure atomicity and data consistency. Documentation is missing on how to achieve that with `sqlite` but there is a test for it in https://github.com/mapbox/node-sqlite3/blob/113be5a8a6ec0e4f7c96045ab69119ea4e9a1427/benchmark/insert.js#L14

elenadimitrova:
Also git ignore `reputationStates.sqlite`

packages/reputation-miner/ReputationMiner.js
area: Postfixed with `rowid`, as that is the name of the column that holds these keys in the tables where they are the primary keys.

packages/reputation-miner/ReputationMiner.js
area: I need the `rowid` after the fact though. So with your approach, I still have to do a `SELECT`, as far as I can tell, resulting in a guaranteed two database transactions, rather than maybe one with the approach I took here (i.e. if it already exists, only one database transaction occurs).

packages/reputation-miner/ReputationMiner.js
area: What race condition are you trying to guard against here? We're not doing any queries simultaneously.

packages/reputation-miner/ReputationMiner.js
elenadimitrova: You can use a sub statement to `SELECT` the rowIds or if syntax allows temp vars in the transaction when bundling all the inserts

packages/reputation-miner/ReputationMiner.js
elenadimitrova: It wasn't a race condition but if any of the `INSERT`s fails in `saveCurrentState` fails

Issue #319, opened 2018-08-22
Nightly integration build


elenadimitrova:
In addition to thorough unit testing, with coverage ~100%, the Colony Network should be tested end-to-end under larger volumes of interactions as an integration load and system test.

Additionally, unit testing is predominantly done using `ganache-cli` whereas both `parity` and `geth` should be considered where possible, i.e. not having to rely on forwarding time functionality.

The above checks can be done as part of a nightly or even weekly (depending on complexity and duration) Circle build.

dguido:
Hey Elena, I think we integrated all of the MAIAN checks into Slither. MAIAN's toolchain is not really keeping up with development of Solidity, but Slither is. SMTChecker is similarly deprecated.

elenadimitrova:
Hey @dguido , isn't Solidity's `SMTChecker` actively worked on? I believe that's what @leonardoalt is doing mostly as part of this project https://github.com/ethereum/solidity/projects/8

leonardoalt:
@dguido not sure where you got that info but it's just false, the SMTChecker was never deprecated and is under continuous development.
Also, why shouldn't they run MAIAN even though Slither already does everything? Different implementations might yield different results.
I always wonder why people try to diminish other tools to promote their own.

ggrieco-tob:
I told @dguido that the SMTChecker was deprecated but I was wrong. I was confused with [the why3 output for formal verification that was deprecated in 0.5.0](https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html#command-line-and-json-interfaces). My apologies to @leonardoalt.

leonardoalt:
No worries @ggrieco-tob ! I thought that this was indeed the confusion

dguido:
Sorry for the confusion Leonardo.

> Also, why shouldn't they run MAIAN even though Slither already does everything? Different implementations might yield different results. I always wonder why people try to diminish other tools to promote their own.

It's a practical matter. MAIAN was a one-and-done academic project that has seen no development since it was released last year. People have continued to file issues against the repository for it that have gone unanswered, and this will continue. Further, the researchers created their own dynamic analysis framework to discover 3 bugs. It is complex, slow, and prone to breaking. On the other hand, the issues they discuss in their research are amenable to static discovery. As implemented in Slither, they return results in less than 1 second and gain the benefit of a maintained framework that keeps up with Solidity development.

PR #327, opened 2018-08-30
Switch parity to newer version download from working mirror

elenadimitrova:
Parity setup on [CI started failing on the evening (GMT) of 29th August](https://circleci.com/gh/JoinColony/colonyNetwork/3596) due to the [parity downloads mirror](https://parity-downloads-mirror.parity.io/) being unavailable.
It appears parity have switched their downloads mirror to a different location for quite some releases now and I am guessing are phasing out the old mirror. Old versions are not available in the new working mirror so our hand's been twisted into upgrading. ~I don't particularly mind as it's a non breaking version for us so just moving build to it to get CI working again.~
New version https://github.com/paritytech/parity-ethereum/releases/tag/v1.11.8

After a conversation with the parity team in Gitter Tbaut there says:
>each new release discontinues the support for older versions. There might be in the future some LTS versions if there is the need. Things are moving too fast now for us to be able to do this

Which essentially means we have to be investing much more time and effort into parity upgrade support and maintenance and perhaps we should be switching the two test suites that use it to `geth`?  Food for thought for when we do #319 perhaps.


area:
> non breaking

Yeah right

PR #330, opened 2018-09-04
Make the due date default to 90 days in future

### LINKED ISSUE #329: Deal with funds being able to be stuck in a task
After merging #324, there is the ability for a worker acting in bad faith to prevent a colony from removing funds from a task:

* A manager assigns a worker to a task with no due date. They assign a payout to the worker.
* Funds are moved to fund the task.

At this point, if the worker disappears, the funds are stuck in the task forever, because there is no due date.

There are two possible solutions:

1. Have an on-chain mechanism where the manager says they want to set a due date (to be able to finish off the task, give the worker a poor rating and reclaim the funds), and if the worker doesn't object, then the change can be enacted.
2. Have a 'default' timestamp some large-ish time in the future (3 months), and do not allow the timestamp to be unset.

The second option is much quicker to implement, and requires far fewer design decisions to be made that might trip us up down the line.

EDIT: After discussion, we are going to implement option 2.

area:
Closes #329 

I am not anticipating this to be a permanent solution; I feel like once disputes are fully fleshed out and begin to be implemented, we can apply them here as one of the first testbeds for them. This is why I have left the 'unset due date' test in, but `skip`ed it. I don't really have much to say here; I was pretty surprised at how straightforward this change ultimately turned out to be.

### DIFF
--- a/upgrade-test/colony-upgrade.js
+++ b/upgrade-test/colony-upgrade.js
@@ -1,5 +1,5 @@
 /* globals artifacts */
-import { getTokenArgs } from "../helpers/test-helper";
+import { currentBlockTime, getTokenArgs } from "../helpers/test-helper";
 import { setupColonyVersionResolver } from "../helpers/upgradable-contracts";
 import { SPECIFICATION_HASH, SPECIFICATION_HASH_UPDATED } from "../helpers/constants";
 import { makeTask } from "../helpers/test-data-generator";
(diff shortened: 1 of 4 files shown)


upgrade-test/colony-upgrade.js
```diff
@@ -27,6 +27,8 @@ contract("Colony contract upgrade", accounts => {
   let updatedColony;
   let updatedColonyVersion;
 
+  const dueDate = 112233445566;
```
elenadimitrova: Is this a sticky key or something? What's wrong which `currentBlockTime`?

area: Yeah, don't know what I was thinking here.

area:
The test that was meant to be testing the missing bit of `afterDueDate` modifier was skipped accidentally. On the `beforeDueDate` front... I don't think anything is tied to being before the due date now, so I have removed this - I think it should have come out in colonyNetwork#324, which tied being able to submit the deliverable to the task being completed, rather than being before the due date, but we missed it.

PR #337, opened 2018-09-11
Deal with base issues in mining client cleanly

### LINKED ISSUE #318: Cleanly deal with base issues in mining client
Filip introduced a hacky (but correct) solution in #294 to deal with [accidental confusion between bases](https://github.com/JoinColony/colonyNetwork/pull/294/files#r211222092). I would like a less hacky solution - ideally, the one I suggested in the comment, making it clear that the strings returned from `breakKeyInToElements` prefixed with `0x` to make it clear that they are hexadecimal. Filip tried this, and it breaks some other stuff - to be expected, but more than we wanted to fix in #294 directly.

area:
Closes #318 

Everything returned by `breakKeyInToElements` is now prefixed by `0x` to make it clear that they are base16, and getKey can now deal with a base 16 string passed to it.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -257,9 +257,7 @@ class ReputationMiner {
     // We update colonywide sums first (children, parents, skill)
     // Then the user-specifc sums in the order children, parents, skill.
 
-    // Converting to decimal, since its going to be converted to hex inside `insert`
-    const skillIdDecimal = new BN(skillId, 16).toString();
-    await this.insert(colonyAddress, skillIdDecimal, userAddress, score, updateNumber);
+    await this.insert(colonyAddress, skillId, userAddress, score, updateNumber);
   }
 
   /**
@@ -288,6 +286,14 @@ class ReputationMiner {
     let colonyAddress = _colonyAddress;
     let userAddress = _userAddress;
 
+    let base = 10;
+    let skillId = _skillId.toString();
+    if (skillId.slice(0, 2) === "0x") {
+      // We've been passed a hex string
+      skillId = skillId.slice(2);
+      base = 16;
+    }
+
     let isAddress = web3Utils.isAddress(colonyAddress);
     // TODO should we return errors here?
     if (!isAddress) {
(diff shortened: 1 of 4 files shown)


packages/reputation-miner/ReputationMiner.js
```diff
@@ -375,7 +381,7 @@ class ReputationMiner {
     const colonyAddress = key.slice(2, 42);
     const skillId = key.slice(42, 106);
     const userAddress = key.slice(106);
-    return [colonyAddress, skillId, userAddress];
+    return [`0x${colonyAddress}`, `0x${skillId}`, `0x${userAddress}`];
```
elenadimitrova: How did this pass build if `saveCurrentState` is still converting these https://github.com/JoinColony/colonyNetwork/blob/develop/packages/reputation-miner/ReputationMiner.js#L783

packages/reputation-miner/ReputationMiner.js
```diff
@@ -257,9 +257,7 @@ class ReputationMiner {
     // We update colonywide sums first (children, parents, skill)
     // Then the user-specifc sums in the order children, parents, skill.
 
-    // Converting to decimal, since its going to be converted to hex inside `insert`
-    const skillIdDecimal = new BN(skillId, 16).toString();
-    await this.insert(colonyAddress, skillIdDecimal, userAddress, score, updateNumber);
+    await this.insert(colonyAddress, skillId, userAddress, score, updateNumber);
```
elenadimitrova: Do we need to `breakKeyIntoElements` only in the `insert` to compose it again?

packages/reputation-miner/ReputationMiner.js
area: You've done a rebase here, right? I think it was written before `saveCurrentState` was merged in.

packages/reputation-miner/ReputationMiner.js
area: Heh, probably not. It's been like that for a while, but yeah, I can't see a reason.

packages/reputation-miner/ReputationMiner.js
elenadimitrova: Ah I've broken it! Do you want me to fix?

elenadimitrova:
Also [this](https://github.com/JoinColony/colonyNetwork/pull/325#issuecomment-418762756) wasn't added in #325 and I seem to have missed it in my last review. Let's add it here.

Issue #317, opened 2018-08-20
Test suite refactoring


elenadimitrova:
This is a living issue to keep track of any ideas around making our test suite better

- [x] Simplify the test data setup to use the new `makeTask` overloads and not set the `dueDate` via a signed transaction, e.g. in https://github.com/JoinColony/colonyNetwork/pull/315/files#diff-c973fa0b369a707dd3dcba1d13c542bfL60
- [x] @area pointed out the `colony-reputation-mining-failed-to-respond-in-time` error message shows when the submission in question has not yet failed to respond whereas the message definitely makes it sound like it has. Maybe `colony-reputation-mining-challenge-not-timed-out` is more appropriate?
- [x] Check there are no simple `assert()`s and everything is `assert.isTrue` 
- [x] For the mining tests, explore other ways to `giveUserCLNYTokensAndStake` rather than completing a task. Possibly if we create the token authority for the CLNY token and allow the coinbase account to call `mint`, we can achieve that. Currently it complicates our reputation mining tests as these task completions incur reputation changes.
Note that as part of #387 we are already introducing the CLNY Authority as defined in the colonyToken repo. 
- [x] In mining tests, when a bad client fails to complete a challenge, we should check the error returned is what we expect, e.g. in `respondToChallenge` function the bad client should get back an error it's failing at the point we are testing. Also outlined by @area [here](https://github.com/JoinColony/colonyNetwork/issues/201#issuecomment-408474449)
- [x] General refactoring of the entire main test suite to get rid of unnecessary setup and checks in tests and focus on _unit_ testing as some tests feel more like integration testing. e.g. Decide if we really need the `colony.js`, `metacolony.js`, `reputation-change.js` tests to redeploy an entire network at the start of every test.
- [x] Look for ways to speed up coverage testing, e.g. exclude tests that are not testing contract functionality such as the reputation mining client test set 
- [x] At the very least investigate and maybe complete the work for bringing test coverage up to as close to 100% as possible. Log any new issues for distinct pieces of logic that need covering if it proves too much for the scope of this work.
- [x] Add test coverage and checks for the reputation mining client (logged as https://github.com/JoinColony/colonyNetwork/issues/523)
- [x] In mining tests, work on restoring good state even after a test fails. Currently if a test fails in `colony-network-mining.js` test fails, the test run exits. (logged separately as https://github.com/JoinColony/colonyNetwork/issues/524)

elenadimitrova:
[...]

PR #357, opened 2018-09-14
Allow reputation oracle to return historical proofs

area:
Now that the history of all reputation states can be stored in the database, we needed to extend the reputation oracle to be able to return those historical states - showing a user their current reputation is great from a not-actually-doing-anything perspective, but if a reward payout is occurring (or, in the future, a dispute or a vote), the user needs to be able to prove their *historical* reputation to the contract.

This adds an extra parameter `rootHash` to the oracle route. If the root hash is the current root hash, then the reputation and proof is returned easily. If it is not the current root hash, the oracle attempts to  reconstruct that reputation state based on what it has in the database in a new (JS) patricia tree. Upon doing so, it calculates the proof for the requested reputation and returns it.

I have no idea how the response time scales with very large historical reputation states, but that's not really my concern right now.

I have also introduced an `auto` parameter to the `ReputationMiningClient`, which allows the autonomous submission to be toggled on and off (it is toggled off for the new tests to avoid race conditions).

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -376,7 +376,7 @@ class ReputationMiner {
     return key;
   }
 
-  static async breakKeyInToElements(key) {
+  static breakKeyInToElements(key) {
     const colonyAddress = key.slice(2, 42);
     const skillId = key.slice(42, 106);
     const userAddress = key.slice(106);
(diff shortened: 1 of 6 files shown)


test/colony-network-mining.js
```diff
@@ -3012,5 +3015,121 @@ contract("ColonyNetworkMining", accounts => {
...
+  });
+
+  describe("Reputation Mining Client", () => {
+    let client;
+    beforeEach(async () => {
+      await giveUserCLNYTokensAndStake(colonyNetwork, MAIN_ACCOUNT, "1000000000000000000");
+      await advanceTimeSubmitAndConfirmHash();
+      await goodClient.saveCurrentState();
+
+      const loader = new TruffleLoader({
```
elenadimitrova: We have a global constant for that called `contractLoader`, can we reuse it?

test/colony-network-mining.js
```diff
@@ -3012,5 +3015,121 @@ contract("ColonyNetworkMining", accounts => {
...
+        assert.equal(siblings[i], retrievedSiblings[i]);
+        assert.equal(siblings[i], retrievedSiblings[i]);
+      }
+
+      const clientHash3 = await goodClient.reputationTree.getRootHash();
+      assert.equal(clientHash2, clientHash3);
+    });
+  });
+
+  describe("Reputation Mining Client", () => {
```
elenadimitrova: This test file is becoming ridiculous. Can we extract mining client only functionality out or at least this new `describe` and perhaps the one above it for miner syncing?
Note that besides this test file, all `describe`s are camelcase like, e.g. they start with lowercase letter.

test/colony-network-mining.js
```diff
@@ -2849,19 +2852,19 @@ contract("ColonyNetworkMining", accounts => {
     it.skip("should abort if a deposit did not complete correctly");
   });
 
+  async function advanceTimeSubmitAndConfirmHash(test) {
+    await forwardTime(3600, test);
+    await goodClient.addLogContentsToReputationTree();
+    await goodClient.submitRootHash();
```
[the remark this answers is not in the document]
area: Correct.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -520,6 +520,60 @@ class ReputationMiner {
...
+      INNER JOIN reputation_states ON reputation_states.rowid=reputations.reputation_rowid
+      WHERE reputation_states.root_hash="${rootHash}"
+      AND users.address="0x${userAddress}"
+      AND reputations.skill_id="${skillId}"
+      AND colonies.address="0x${colonyAddress}"`
+    );
+
+    if (res.length === 0) {
+      return new Error("No such reputation");
+    }
```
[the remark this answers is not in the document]
area: I guess we could. It should never be more than one though, because of the check done [when saving the state](https://github.com/JoinColony/colonyNetwork/blob/develop/packages/reputation-miner/ReputationMiner.js#L804).

test/colony-network-mining.js
area: I was hoping I was going to get away with this for now given #317 is on the cards...

PR #359, opened 2018-09-19
Add ENS bits required by DDB team

area:
The previous ENS implementation only really went half way - allowed people to claim ownership of ENS addresses, but it didn't actually set them up to resolve to anything. This PR claims to do so, as well as adding reverse lookup for registered user/colony names and allowing the querying of an ENS hash to return an orbitDB string. The colonyNetwork contract is supplied as the resolver to the ENS Registry for registered subdomains, and is ENS compliant (so names registered in colony should 'just work' everywhere that implements ENS support).

Because we want to force people to allow reverse lookups, the contract is now passed the actual string that is their username, rather than the hash of the string that is their username. Note that the contract does not enforce the required [normalization](http://docs.ens.domains/en/latest/implementers.html#normalising-and-validating-names), which should be enforced clientside, and clients should check that any usernames that are reverse-looked-up and normalised do not have their the representation changed by the normalization in order to avoid impersonation.


contracts/IColonyNetwork.sol
```diff
@@ -183,12 +183,34 @@ contract IColonyNetwork {
...
+  /// @notice Reverse lookup a username from an address.
+  /// @param addr The address we wish to find the corresponding ENS domain for (if any)
+  /// @return domain A string containing the colony-based ENS name corresponding to addr
+  function lookupUsername(address addr) public view returns(string domain);
+
+  /// @notice Query if a contract implements an interface
+  /// @param interfaceID The interface identifier, as specified in ERC-165
+  /// @dev Interface identification is specified in ERC-165.
+  /// @return `true` if the contract implements `interfaceID`
+  function supportsInterface(bytes4 interfaceID) public view returns (bool);
```
elenadimitrova: We can make this an `external` function as specified in the ERC but also because it'll be cheaper. (Note to self we should review our function visibility modifiers across everything)

contracts/ColonyNetworkENS.sol
```diff
@@ -0,0 +1,112 @@
...
+
+/// @title First-In-First-Served ENS Registrar
+/// @notice A registrar that allocates subdomains to the first person to claim them.
+/// @notice Source modified from https://github.com/ensdomains/ens/blob/master/contracts/FIFSRegistrar.sol
+contract ColonyNetworkENS is ColonyNetworkStorage {
+
+  bytes32 constant USER_HASH = keccak256("user");
+  bytes32 constant COLONY_HASH = keccak256("colony");
+
+  modifier unowned(bytes32 node, string username) {
```
elenadimitrova: username -> domainName

contracts/ColonyNetworkENS.sol
```diff
@@ -0,0 +1,112 @@
...
+
+  function getProfileDBAddress(bytes32 node) public view returns (string) {
+    return records[node].orbitdb;
+  }
+
+  function lookupUsername(address addr) public view returns(string) {
+    if (bytes(userLabels[addr]).length != 0) {
+      return string(abi.encodePacked(userLabels[addr], ".user.joincolony.eth"));
+    } else if (bytes(colonyLabels[addr]).length != 0) {
+      return string(abi.encodePacked(colonyLabels[addr], ".colony.joincolony.eth"));
```
elenadimitrova: BUG: The register functions allows me to register the same name under both users and colonies so this reverse lookup would only ever return the user address.

contracts/ColonyNetworkENS.sol
```diff
@@ -0,0 +1,112 @@
...
+    ENS(ens).setSubnodeOwner(rootNode, COLONY_HASH, this);
+  }
+
+  function registerUserLabel(string username, string orbitdb)
+  public
+  unowned(userNode, username)
+  {
+    require(bytes(username).length > 0, "colony-user-label-invalid");
+    require(bytes(userLabels[msg.sender]).length == 0, "colony-user-label-already-owned");
+    bytes32 subnode = keccak256(username);
```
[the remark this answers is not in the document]
area: The packed encoding of `username` is just `username`, right?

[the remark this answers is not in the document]
elenadimitrova: yes, but there is a warning thrown atm because `keccak` expects `bytes` only and we're giving it a string. I think this is by design which seems will carry over in 0.5.0 https://github.com/ethereum/solidity/issues/3955

To eliminate the warning explicitly convert via `bytes(userName)`

test/colony-network.js
```diff
@@ -385,11 +386,45 @@ contract("ColonyNetwork", accounts => {
...
+    });
+
+    it("should return a blank address if looking up an address with no Colony-based ENS name", async () => {
+      const lookedUpENSDomain = await colonyNetwork.lookupUsername(accounts[2]);
+      assert.equal(lookedUpENSDomain, "");
+    });
+
+    it("should respond correctly to queries regarding ENS interfaces it supports", async () => {
+      let response = await colonyNetwork.supportsInterface("0x01ffc9a7"); // supports 'supportsInterface(bytes4)'
+      assert(response);
```
elenadimitrova: Can we use `assert.isTrue` as this if error simply produces a non-specific error `AssertionError: Unspecified AssertionError`

PR #361, opened 2018-09-24
Amend 'ratings' description in docs

gichiba:
As a result of answering a question on build.colony.io, I realized that the description in the docs leave it ambiguous whether a task's payout is determined by the ratings, or just the reputation gains/losses. This fixes the issue.


elenadimitrova:
Putting [skip ci] in the PR title does nothing @gichiba This is only applicable to commits. Either that or you just couldn't bother changing the PR description when opening it from the branch with 1 commit of the same title. I can't tell which is worse..

gichiba:
😶

gichiba:
I don't know which is worse, but now I know that! I just wanted to be sure because at least I know that docs changes don't need to run through any tests... :)

PR #381, opened 2018-10-09
Review task lifecycle functions and guards

### LINKED ISSUE #338: End to end task lifecycle review
Go over the task lifecycle functions, perhaps even document them here for posterity https://docs.google.com/spreadsheets/d/1HiL-jJd22BzX4aAMVaOv0ptYkBfnkfY4yhJkLAHVOGI/edit#gid=0 and fix any gaps that may break the cycle.

Solving the todo here https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L589
is also part of this work.

```
// TODO: Check if we are changing a role before due date and before work has been submitted
function setTaskRoleUser(uint256 _id, uint8 _role, address _user)
```

Also the following:
- cancelling a task after a deliverable is submitted should not be possible

elenadimitrova:
Closes #338 

Includes refactoring of the task tests to a separate test file and away from `colony-test`, firstly as those don't require a new colony to be created before each and secondly because it'll be nice to test with anything than the first or second colony task.

Additionally switch all `taskNotFinalized` to `taskNotComplete` except for `finalize` and `claimPayout`

Task lifecycle methods are documented for posterity [here](https://docs.google.com/spreadsheets/d/1HiL-jJd22BzX4aAMVaOv0ptYkBfnkfY4yhJkLAHVOGI/edit#gid=0 )


elenadimitrova:
@area there is this one test here, which I couldn't fix. Can you please have a look as well? I spent a couple of hours just testing colony and task funding worked, task is finalized correctly and the payout is claimed. I even thought it maybe a ganache bug, similarly to how we had the balance issues before. If worse comes to worst I might test the event amount instead of getting the total balance?  https://github.com/JoinColony/colonyNetwork/pull/381/files#diff-124616d3d79cacf55da9363cc4b3b015R1461

elenadimitrova:
@area when @itspatch was going through the spreadsheet he noticed the `cancelTask` was allowed after a user is assigned. The whitepaper agrees that this shouldn't be allowed so I will bring that under the `executeSignedTaskChange` umbrella.

area:
I can't see anything major here that would prevent me from accepting this as 'good enough', but a few stray ideas:

1. Is it odd that `setTaskSkill` goes through `executeTaskChange`, but `setTaskDomain` does not? They both relate to reputation earned by the worker...
2. `taskComplete` modifier implicitly checks that `taskExists`, so anywhere we have both, we only really need the former.
3. The `beforeDueDate` modifier is not used anywhere, but we still have it defined (this one's my fault from #324 - it should have been removed there.

elenadimitrova:
Note on your points above Alex
1. Indeed the domain affects reputation for every role on task. This was a product decision more than a technical one back when this was implemented, @itspatch and @collinvine I believe agreed with Filip the requirements around `setTaskDomain` here: https://github.com/JoinColony/colonyNetwork/pull/224#issue-192827824

The alternative is to have to get everyone affected agree to the change (all 3 roles) or at a minimum the worker in addition to manager.

collinvine:
Copying in the conversation from Slack:

Collin Vine [11:05 AM]
I'm okay with that. To be honest, I thought domain was a multisig tx.

However, we're working off of this pattern:

* Before assignment: any task attribute can be changed by the task owner (the manager) as all of this is stored either locally or on the ddb. Not on chain (yet).
* After assignment: we're using (want to use) the assignment as the trigger to call `makeTask` and its associated functions. Any change—to the skill, domain, funding, due date—will require approval from the worker.

PR #386, opened 2018-10-14
Delegate Network property setting to the Meta Colony

### LINKED ISSUE #20: Secure given Colony Network parameters to Meta Colony
### 3.2 Role of CLNY Holders and the Meta Colony
There will be permissioned functions on the Network Contract to allow fundamental parameters of the network to be set, which can only be called by the Meta Colony.
For these permissioned functions to be called by the Meta Colony, a vote open to all CLNY
and reputation holders must be conducted (see Section 11.4).

Examples of such network parameters are the Network fee, registering new Colony versions.

Note **Section 3.3 Handling off decision-making power to the Meta Colony**
Common Colony Token holders are responsible for Reputation Mining from the start, but decisions
about the underlying properties of the network will initially be made by a multisignature contract
controlled by the Colony team. As the network develops and is proved to be effective, control over
these decisions will cede to the Common Colony.

### LINKED ISSUE #379: Consider behaviour of fees when payout is less than 100 tokens
I think we just need to round up when calculating the fee, but worth checking.

elenadimitrova:
Closes #20 Closes #379 

The Colony Network fee inverse is here abstracted to a parameter settable by the Meta Colony.

Because of the additional functionality available to the MetaColony in comparison to a Colony, I've abstracted the MetaColony interface which makes it clear now how are the two different. I thought about separating the actual implementation e.g. Colony -> MetaColony and Authority -> MetaAuthority but that might mean maintaining distinct versions for the MetaColony so I'd probably prefer to keep Meta colony distinction separation up to just the interface level.

The rewards parameter is now settable on a colony by its owner with an initial value of `2**256-1` indicating a zero rewards in colony by default.

Registering new Colony versions is also delegated to the Meta Colony. It's a catch 22 though as to create a colony (including a meta colony) we need to have registered a colony version but to register a colony version we need make that function call from the Meta Colony. We solve this by introducing an explicit `IColonyNetwork.initialise` function that sets the resolver for the very first colony version.

Additionally we deal with the problem presented in #379 where we round up network fees in cases where the payout is smaller than the fee inverse. e.g. if a payout is 98 tokens and the fee inverse is 100, the final network fee will be 1. Same for any payouts from 1 to 100. 
I wasn't sure how to deal with payouts which are 1 token worth, and decided that 1 token should be collected as fee leaving the payout effectively as 0. Seems like the saner approach than leave 0 to the network and 1 to the user, opening us to exploits and free network use.


kronosapiens:
Part of me wants to say that a payout of 1 token would go to the network. It means no payout for the user but it avoid that exploit, and they shouldn't be setting payouts that small anyway.

contracts/ColonyAuthority.sol
```diff
@@ -46,6 +46,12 @@ contract ColonyAuthority is CommonAuthority {
     setOwnerRoleCapability(colony, "upgrade(uint256)");
     // Claim colony ENS label
     setOwnerRoleCapability(colony, "registerColonyLabel(string)");
+    // Set Network fee inverse
+    setOwnerRoleCapability(colony, "setNetworkFeeInverse(uint256)");
+    // Set Network fee inverse
```
kronosapiens:
```suggestion
    // Set reward inverse
```

contracts/Colony.sol
```diff
@@ -78,6 +78,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
     uint256 rootLocalSkill = colonyNetwork.getSkillCount();
     initialiseDomain(rootLocalSkill);
+
+    // Set initial colony reward inverse amount
```
kronosapiens: Shouldn't there be a require over this? Or a call to the `setRewardInverse` function?

kronosapiens: One alternative would be to set a default value in the `initialiseColony ` function and allow people to update it later. I like that as well because it keeps the `initialiseColony` interface small and so makes me happy from an aesthetic perspective.

contracts/ColonyNetworkStorage.sol
```diff
@@ -114,4 +117,9 @@ contract ColonyNetworkStorage is CommonStorage, DSMath {
     require(!_isColony[msg.sender], "colony-caller-must-not-be-colony");
     _;
   }
+
+  modifier calledByMetaColony() {
```
kronosapiens: Interesting that this didn't exist before

contracts/IColony.sol
```diff
@@ -420,14 +409,13 @@ contract IColony is IRecovery {
   /// @return rating Rating the user received
   function getTaskRole(uint256 _id, uint8 _role) public view returns (address user, bool rateFail, uint8 rating);
 
-  // Implemented in ColonyFunding.sol
-  /// @notice Return 1 / the fee to pay to the network. e.g. if the fee is 1% (or 0.01), return 100
-  /// @return feeInverse The inverse of the network fee
-  function getFeeInverse() public pure returns (uint256 feeInverse);
+  /// @notice Set the reward inverse to pay out from revenue. e.g. if the fee is 1% (or 0.01), set 100
+  /// @return _rewardInverse The inverse of the reward
```
kronosapiens:
```suggestion
```

test/colony-network-recovery.js
```diff
@@ -211,8 +211,8 @@ contract("Colony Network Recovery", accounts => {
 
       await colonyNetwork.enterRecoveryMode();
 
-      await colonyNetwork.setStorageSlotRecovery(19, "0x02");
-      await colonyNetwork.setStorageSlotRecovery(20, `0x${new BN(7).toString(16, 64)}`);
+      await colonyNetwork.setStorageSlotRecovery(20, "0x02");
```
kronosapiens: Fixing these failures can't have been fun... 😭

test/colony-network.js
```diff
@@ -90,129 +104,126 @@ contract("ColonyNetwork", accounts => {
   describe("when creating new colonies", () => {
     it("should allow users to create new colonies", async () => {
       const token = await Token.new(...TOKEN_ARGS);
-      const { logs } = await colonyNetwork.createColony(token.address);
+      const { logs } = await colonyNetwork.createColony(token.address, 100);
       const { colonyAddress } = logs[0].args;
       const colonyCount = await colonyNetwork.getColonyCount();
       assert.notEqual(colonyAddress, 0x0);
-      assert.equal(colonyCount.toNumber(), 1);
+      expect(colonyCount).to.eq.BN(2);
```
kronosapiens: I would mildly prefer the change in testing style to be a separate PR (perhaps part of a more comprehensive test suite refactor) but I'll survive.

truffle.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     }
   },
   mocha: {
-    reporter: "mocha-circleci-reporter",
+    reporter: "eth-gas-reporter",
```
kronosapiens: Would be nice if there was an easy way to exclude this change from the PR (since it's an artifact of our test setup)

test/colony-recovery.js
```diff
@@ -32,10 +34,15 @@ contract("Colony", accounts => {
...
-    await setupColonyVersionResolver(colonyTemplate, colonyTask, colonyFunding, contractRecovery, resolver, colonyNetwork);
+    await setupColonyVersionResolver(colonyTemplate, colonyTask, colonyFunding, contractRecovery, resolver);
+    await colonyNetwork.initialise(resolver.address);
 
     clnyToken = await Token.new("Colony Network Token", "CLNY", 18);
-    await colonyNetwork.createMetaColony(clnyToken.address);
+    await colonyNetwork.createMetaColony(clnyToken.address, 100);
+    const metaColonyAddress = await colonyNetwork.getMetaColony();
+    metaColony = await IMetaColony.at(metaColonyAddress);
+    await metaColony.setNetworkFeeInverse(100);
```
kronosapiens: Why not give the network a default fee inverse? It would let us avoid some of this testing boilerplate.

contracts/ColonyNetworkStorage.sol
elenadimitrova: We only use this check in `allowedToAddSkill` modifier on the network as only the Meta Colony is allowed to add global skills (stored on the network)

contracts/IColony.sol
elenadimitrova: This should actually be a `@param` natspec not `@return`

test/colony-network-recovery.js
elenadimitrova: Alex did the hard work on finding out the exact storage slots in he first place, here I just added 1 to the slots positioned _after_ the one new parameter I introduced.

test/colony-network.js
elenadimitrova: I would normally agree, but I just couldn't help it! Comparing numbers drives me up a wall. I'll try to keep my hands off `bn-chai` for now 👐

truffle.js
elenadimitrova: That's my mistake for committing it. Will undo.

contracts/IColony.sol
kronosapiens: 👍

contracts/Colony.sol
kronosapiens: Bump. Someone could set a 0 here and leave it unchanged.

contracts/ColonyNetwork.sol
```diff
@@ -157,22 +162,33 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+    require(currentColonyVersion > 0, "colony-network-not-intialised-cannot-add-colony-version");
+    
     colonyVersionResolver[_version] = _resolver;
     if (_version > currentColonyVersion) {
       currentColonyVersion = _version;
     }
   }
 
+  function initialise(address _resolver) public 
+  auth
```
area: Is this required? It can only be called once, has to be before anything productive can be done, and will be before the address is public...

contracts/ColonyFunding.sol
```diff
@@ -388,4 +390,15 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
 
     updateTaskPayoutsWeCannotMakeAfterBudgetChange(_id, _token, currentTotalAmount);
   }
+
+  function calculateNetworkFeeForPayout(uint256 _payout) private view returns (uint256 fee) {
+    IColonyNetwork colonyNetworkContract = IColonyNetwork(colonyNetworkAddress);
+    uint256 feeInverse = colonyNetworkContract.getFeeInverse();
+
```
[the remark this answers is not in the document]
elenadimitrova: With this ☝️ if they are both 2 we get a fee of 2 which means effective fee of 100% rather than the intended 50%, that error margin decreases as the values increase, i.e. if they are both 100 we get an effective fee of 2% rather than intended 1% etc. We are slightly overcharging using this calculation as opposed to slightly undercharging using mine above.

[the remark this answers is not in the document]
elenadimitrova: With this ☝️ if they are both 2 we get a fee of 2 which means effective fee of 100% rather than the intended 50%, that error margin decreases as the values increase, i.e. if they are both 100 we get an effective fee of 2% rather than intended 1% etc. We are slightly overcharging using this calculation as opposed to slightly undercharging using mine above.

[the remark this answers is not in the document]
elenadimitrova: With this ☝️ if they are both 2 we get a fee of 2 which means effective fee of 100% rather than the intended 50%, that error margin decreases as the values increase, i.e. if they are both 100 we get an effective fee of 2% rather than intended 1% etc. We are slightly overcharging using this calculation as opposed to slightly undercharging using mine above.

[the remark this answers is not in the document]
elenadimitrova: With this ☝️ if they are both 2 we get a fee of 2 which means effective fee of 100% rather than the intended 50%, that error margin decreases as the values increase, i.e. if they are both 100 we get an effective fee of 2% rather than intended 1% etc. We are slightly overcharging using this calculation as opposed to slightly undercharging using mine above.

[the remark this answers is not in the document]
elenadimitrova: With this ☝️ if they are both 2 we get a fee of 2 which means effective fee of 100% rather than the intended 50%, that error margin decreases as the values increase, i.e. if they are both 100 we get an effective fee of 2% rather than intended 1% etc. We are slightly overcharging using this calculation as opposed to slightly undercharging using mine above.

kronosapiens:
I'm happy with the changes, the last thing I see is the discussion with Alex: https://github.com/JoinColony/colonyNetwork/pull/386#discussion_r226950999

contracts/ColonyFunding.sol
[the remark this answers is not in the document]
elenadimitrova: We agreed to take Alex's calculation as that slightly favours the network in terms of error margins in fee calculations as opposed to mine which is slightly more in favour of the client. This has been implemented.

PR #395, opened 2018-10-24
Economic tweaks

### LINKED ISSUE #353: Burn CLNY used in auctions
CLNY spent in auctions to purchase other native tokens should be burned, rather than returned to the MetaColony. 

Related changes after conversation with Jack:
Fees on CLNY should be burned rather than auctioned for CLNY (~and possibly the MC should not pay a fee on CLNY payouts, but probably not~)
Fee is the same for all tokens (even native tokens)
Network fee is applied to both task payouts and colony rewards.

area:
Closes #353

* CLNY used in auctions is burned, not sent to MetaColony
* If an auction of CLNY is attempted to be started, all CLNY held by ColonyNetwork is burned
* Rewards have the network fee applied to them, same as task payouts

My notes are unclear as to whether 'fees being the same on all tokens, including native ones' applies to `claimColonyFunds`, so I have left that unchanged for now but should check with Jack.

### DIFF
--- a/contracts/ColonyNetworkAuction.sol
+++ b/contracts/ColonyNetworkAuction.sol
@@ -25,11 +25,24 @@ contract ColonyNetworkAuction is ColonyNetworkStorage {
   event AuctionCreated(address auction, address token, uint256 quantity);
 
   function startTokenAuction(address _token) public stoppable {
+    require(_token != 0x0, "colony-auction-invalid-token");
+
     uint lastAuctionTimestamp = recentAuctions[_token];
     require(lastAuctionTimestamp == 0 || now - lastAuctionTimestamp >= 30 days, "colony-auction-start-too-soon");
+
     address clny = IColony(metaColony).getToken();
-    DutchAuction auction = new DutchAuction(clny, _token, metaColony);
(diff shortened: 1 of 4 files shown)


contracts/ColonyNetworkAuction.sol
```diff
@@ -25,11 +25,24 @@ contract ColonyNetworkAuction is ColonyNetworkStorage {
...
     address clny = IColony(metaColony).getToken();
-    DutchAuction auction = new DutchAuction(clny, _token, metaColony);
+    require(clny != 0x0, "colony-auction-invalid-token");
+
     uint availableTokens = ERC20Extended(_token).balanceOf(this);
+
+    if (_token==clny) {
+      // We don't auction CLNY. We just burn it instead.
+      // Note we can do this more often than every 30 days.
+      ERC20Extended(clny).burn(availableTokens);
```
elenadimitrova: ColonyNetwork will receive CLNY from:
- MetaColony task payout fees
- MetaColony rewards payout fees
(no longer auctions since this PR)

The above will be burned here so wanted to make sure we're okay with that.

PR #397, opened 2018-10-26
Feature/miner stake

### LINKED ISSUE #344: Definitively establish minimum stake amount
A minimum amount of CLNY is required to stake. We need to ensure that people lose more CLNY than it costs for good actors to defend their submission. Note that because the value of CLNY will vary compared to ETH, we should build in a comfortable buffer (e.g. an order of magnitude).

[TODO](https://github.com/JoinColony/colonyNetwork/blob/8a94d4dbf1a790ff68325bc29f3561fce54aae3b/contracts/ReputationMiningCycle.sol#L54)

### LINKED ISSUE #352: Penalising stakers backing the wrong reputation root hash
After child reputation updates #280 is done complete the remaining `todo`s in `punishStakers` function [ref](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/TokenLocking.sol#L94)

kronosapiens:
Closes #344
Closes #352

### DIFF
--- a/contracts/TokenLocking.sol
+++ b/contracts/TokenLocking.sol
@@ -91,16 +91,15 @@ contract TokenLocking is TokenLockingStorage, DSMath {
   // This function is only used in context of reputation mining
   // TODO: After we add formula to calculate user's loss, refactor accordingly and/or
   // move some of the functionality to `ColonyNetworkMining` if needed
-  function punishStakers(address[] _users) public onlyReputationMiningCycle {
+  function punishStakers(address[] _stakers, address _beneficiary) public onlyReputationMiningCycle {
     address clnyToken = IColony(IColonyNetwork(colonyNetwork).getMetaColony()).getToken();
+    uint256 lostStake;
     // Passing an array so that we don't incur the EtherRouter overhead for each staker if we looped over
     // it in ReputationMiningCycle.invalidateHash;
-    for (uint256 i = 0; i < _users.length; i++) {
-      // This is pretty harsh! Are we happy with this?
-      // Alternative: lose more than they would have gained for backing the right hash.
-      userLocks[clnyToken][_users[i]].balance = 0;
-      // TODO: Where do these staked tokens go? Maybe split between the person who did the 'invalidate' transaction
-      // and the colony network?
+    for (uint256 i = 0; i < _stakers.length; i++) {
+      lostStake = userLocks[clnyToken][_stakers[i]].balance;
+      userLocks[clnyToken][_stakers[i]].balance = 0;
+      userLocks[clnyToken][_beneficiary].balance = add(userLocks[clnyToken][_beneficiary].balance, lostStake);
       // TODO: Lose rep?
     }
   }
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -310,6 +310,7 @@ export function makeReputationKey(colonyAddress, skill, accountAddress = undefin
   return key;
 }
 
+// Note: value can be anything with a `.toString()` method -- a string, number, or BN.
 export function makeReputationValue(value, repuationId) {
   return `0x${(new BN(value.toString())).toString(16, 64)}${(new BN(repuationId)).toString(16, 64)}`; // eslint-disable-line
 }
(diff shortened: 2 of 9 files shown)


test/colony-network-mining.js
```diff
@@ -41,6 +41,10 @@ const contractLoader = new TruffleLoader({
 
 const useJsTree = true;
 
+const WAD = new BN(10).pow(new BN(18));
+const MIN_STAKE = WAD.muln(2000);
+const DEFAULT_STAKE = MIN_STAKE.muln(12);
```
area: Can we increase this to say, 1000? Currently, tests that involve stakers that use `submitAndForwardTimeToDispute` advance 1800 seconds, then get all the clients to submit. With a stake of `MIN_STAKE.muln(12)`, there's a 1-in-4096 chance that any client passed in to that function will fail (because none of its entries are eligible yet after only 1800 seconds). We call that function 46 times in this file, so there's somewhere between a 1-in-45 to 1-in-90 (depending on how many clients are passed in to that function) that a test will randomly fail. By increasing this to `.muln(1000)`, which is effectively what it was before, those numbers are on the order of `10**300`, which I am much more comfortable with.

kronosapiens: 100%

kronosapiens: And I still can't find the minimum stake hardcoded in `ReputationMiner.js`

kronosapiens: @area bump

area: https://github.com/JoinColony/colonyNetwork/blob/84b8106ab8f124230e1c5fb6c758f683af946883/packages/reputation-miner/ReputationMiner.js#L535

helpers/test-helper.js
```diff
@@ -329,7 +330,7 @@ export async function getValidEntryNumber(colonyNetwork, account, hash, starting
   const userBalance = userLockInformation.amount;
 
   // What's the largest entry they can submit?
-  const nIter = userBalance.div(new BN(10).pow(new BN(15)));
+  const nIter = userBalance.div(new BN(10).pow(new BN(18)).muln(2000));
```
elenadimitrova: Sorry just having a last look through this.
Isn't the divisor here `MIN_STAKE`?

migrations/8_setup_meta_colony.js
```diff
@@ -9,6 +9,8 @@ const ITokenLocking = artifacts.require("./ITokenLocking");
 const EtherRouter = artifacts.require("./EtherRouter");
 const Token = artifacts.require("./Token");
 
+const DEFAULT_STAKE = "24000000000000000000000";
```
elenadimitrova: Why is this default stake different to the default for tests?

contracts/TokenLocking.sol
```diff
@@ -91,16 +91,15 @@ contract TokenLocking is TokenLockingStorage, DSMath {
...
-    for (uint256 i = 0; i < _users.length; i++) {
-      // This is pretty harsh! Are we happy with this?
-      // Alternative: lose more than they would have gained for backing the right hash.
-      userLocks[clnyToken][_users[i]].balance = 0;
-      // TODO: Where do these staked tokens go? Maybe split between the person who did the 'invalidate' transaction
-      // and the colony network?
+    for (uint256 i = 0; i < _stakers.length; i++) {
+      lostStake = userLocks[clnyToken][_stakers[i]].balance;
+      userLocks[clnyToken][_stakers[i]].balance = 0;
+      userLocks[clnyToken][_beneficiary].balance = add(userLocks[clnyToken][_beneficiary].balance, lostStake);
```
elenadimitrova: There is no test for this functionality

helpers/test-helper.js
kronosapiens: Yes, but because `constants.js` imports from `test-helper.js`, you can't import `MIN_STAKE` because of circularity.

migrations/8_setup_meta_colony.js
kronosapiens: No real reason, the build was running when I noticed this discrepancy and didn't want to restart. I can update.

contracts/TokenLocking.sol
kronosapiens: Here: https://github.com/JoinColony/colonyNetwork/pull/397/files#diff-15aaaaa81a55d904a938772d3e2a3a82R727

Issue #393, opened 2018-10-23
In reputation mining, use bytes32 arrays for values rather than bytes.


area:
Values we pass in as bytes are really two bytes32 variables concatenated, and we do some assembly to tease them apart. We could just pass them in pre-separated and concatenate them where we need them. This also has the benefit of ensuring the variables passed are the right length; currently, if they're too short, the assembly will copy from the start of the next variable along. I can't see how that would be abused yet, especially because we only read from these variables and don't save to them, but it's a pretty classic start of something bad being able to happen.

We could also do the same with keys, making them bytes32[3], but would have to be careful concatenating as one of the elements is an address which is obviously shorter.

At which point, we could in theory collapse all the bytes arguments down in to a bytes32[18], freeing up a lot of stack space, which I know we've found ourselves bumping up against...

PR #400, opened 2018-10-27
Refactor reputation value inputs as members of uint256 array

elenadimitrova:
As part for #393 , refactors the reputation value as additional `uint256` members of the `respondToChallenge` `uint256[] u` array


contracts/ReputationMiningCycleRespond.sol
```diff
@@ -313,95 +300,69 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
...
     if (_agreeStateReputationUID != 0) {
       // i.e. if this was an existing reputation, then require that the ID hasn't changed.
       require(_agreeStateReputationUID == _disagreeStateReputationUID, "colony-reputation-mining-uid-changed-for-existing-reputation");
       emit ProveUIDSuccess(_agreeStateReputationUID, _disagreeStateReputationUID, true);
     } else {
-      uint256 previousNewReputationUID;
-      assembly {
-        previousNewReputationUID := mload(add(_previousNewReputationValue, 64))
-      }
+      uint256 previousNewReputationUID = u[U_PREVIOUS_NEW_REPUTATION_UID];
```
elenadimitrova: No need for more local variables

contracts/ReputationMiningCycleRespond.sol
```diff
@@ -519,24 +477,18 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
...
     require(impliedRoot == disputeRounds[u[U_ROUND]][u[U_IDX]].jrh, "colony-reputation-mining-origin-skill-state-disagreement");
   }
 
-  function saveProvedReputation(uint256[11] u, bytes previousNewReputationValue) internal {
-    uint256 previousReputationUID;
-    assembly {
-      previousReputationUID := mload(add(previousNewReputationValue,0x40))
-    }
+  function saveProvedReputation(uint256[19] u) internal {
+    uint256 previousReputationUID = u[U_PREVIOUS_NEW_REPUTATION_UID];
```
elenadimitrova: Again no need for another local variable

contracts/ReputationMiningCycleRespond.sol
```diff
@@ -546,4 +498,14 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
     // Save the index for tiebreak scenarios later.
     disputeRounds[u[U_ROUND]][u[U_IDX]].provedPreviousReputationUID = previousReputationUID;
   }
+
+  function concatenateToBytes64(uint256 a, uint256 b) internal pure returns (bytes) {
+    bytes memory retValue = new bytes(64);
+    assembly {
+      // Seems as if you access an external variable by name, you get the value, straight up...?
```
elenadimitrova: What do you mean "you get the value" I think you get the variable location in memory slot

contracts/ReputationMiningCycleRespond.sol
```diff
@@ -244,34 +237,29 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
...
     bytes32 reputationRootHash = getImpliedRoot(_reputationKey, agreeStateReputationValue, u[U_REPUTATION_BRANCH_MASK], reputationSiblings);
-    bytes memory jhLeafValue = new bytes(64);
+    bytes memory jhLeafValue = concatenateToBytes64(uint256(reputationRootHash), u[U_AGREE_STATE_NNODES]);
     bytes memory lastAgreeIdxBytes = new bytes(32);
+
     assembly {
-      mstore(add(jhLeafValue, 0x20), reputationRootHash)
-      let x := mload(add(u, mul(32,3))) // 3 = U_AGREE_STATE_NNODES. Constants not supported by inline solidity
-      mstore(add(jhLeafValue, 0x40), x)
       mstore(add(lastAgreeIdxBytes, 0x20), lastAgreeIdx)
```
elenadimitrova: We can abstract the 3 instances where we calculate `lastAgreeIdxBytes`

elenadimitrova: There is another instance of `firstDisagreeIdxBytes` but it is only a single one so probably not worth abstracting. This will take assembly level code in this contract down to 4 which makes me happy

PR #412, opened 2018-11-05
Enable contract error messages on mining client

elenadimitrova:
Here we move back to the vanilla `ethers` release which [now provides support for making contract calls using the `blockTag` parameter](https://github.com/ethers-io/ethers.js/issues/226), https://github.com/ethers-io/ethers.js/issues/329
This is required functionality for the client syncing logic that was introduced in a custom version of ours of both `ethers` and `ganache-core` back in https://github.com/JoinColony/colonyNetwork/pull/311

Since the `ethers` release has jumped up a major version since then, we include a set of breaking fixes for v3 to v4.

Introducing `checkErrorRevertEthers` - an `ethers` specific helper function to check expected error message is present in the response. This is useful for the revert error checks coming back from the reputation mining client which currently uses `ethers.contract` (as opposed to `truffle-contract` as our tests do).

### DIFF
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -144,7 +144,6 @@ export async function checkErrorRevert(promise, errorMessage) {
     // If the promise is from Truffle, then we have the receipt already.
     // If this tx has come from the mining client, the promise has just resolved to a tx hash and we need to do the following
     if (!receipt) {
-      console.log("no receipt returned");
       const txid = await promise;
       receipt = await web3GetTransactionReceipt(txid);
     }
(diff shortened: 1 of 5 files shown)


helpers/test-helper.js
```diff
@@ -92,6 +92,43 @@ export function web3GetAccounts() {
...
+
+  return new Promise((resolve, reject) => {
+    web3.currentProvider.send(packet, (err, res) => {
+      if (err !== null) return reject(err);
+      return resolve(res);
+    });
+  });
+}
+
+export function extractReasonString(res) {
```
area: Should we credit these functions to truffle-contract?

PR #418, opened 2018-11-07
Chinese translation for bug bounty plan

bobjiang:
This is a Chinese version for bug bounty plan. I will translate following document in the docs/ dir one by one into docs-cn/ dir.

BoB Jiang
Thanks a lot!

Closes #[GH issue]
Implements [X] part for #[GH issue]


gichiba:
Hi Bob, we're going to put this on hold for a while because a) nobody can verify the translation and b) since the rest of the program is in English, bugs submitted must also be submitted in English. 

But thanks very much for this and for your effort!

PR #430, opened 2018-11-12
Working with low auction price

### LINKED ISSUE #377: Destructing a Token auction can be stalled
If someone sends CLNY to an auction contract (outside of taking part in the auction, after the auction has ended), then [this `assert`](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyNetworkAuction.sol#L214) will never pass meaning the auction will never be able to `destruct`.

### LINKED ISSUE #416: Token lock in Auction when the lot price is decreased
## Summary
`ColonyNetworkAuction.sol` implements a dutch auction with dynamically decreasing price. When user makes a bid, the function first checks how many CLNY tokens it is required to bid to finish the auction. The calculation is made via `DSMath.sub`, which throws if resulting amount is smaller than 0. In other words, the function will throw if `_totalToEndAuction - receivedTotal < 0`. If the price will decrease to the extent that it will become smaller than `receivedTotal`, the bid will throw forever, regardless of the passed `_amount` parameter.
Note that calling `bid` function is the only way to close the auction, and closing auction is the only way to finalize it, and finalizing it is the only way to claim lot for bidder. Therefore, locking bid function is basically locking the auction lot.
This lock will happen regardless of the number of bidders, bids, and bid amounts. The only requirement is that the price will become low enough without additional bids.

## Steps to Reproduce (for bugs)
1. Start an auction.
2. Make bids.
3. Wait until `_totalToEndAuction - receivedTotal < 0`.

## Expected Behavior
Contract should allow to stop the auction.

## Current Behaviour
Contract will not allow to make bids or closing auction in any other way.

## Possible Solution
Allow to bid with dust amounts even if the `_totalToEndAuction - receivedTotal < 0`. Alternatively, make a function `close` which will close the auction if the amount of received tokens is high enough given the current price.

elenadimitrova:
Here we restrict auctions functionality as they will not be used in Glider release. Also all possible improvements to test coverage are done.

Additionally the following issues are fixed:
Closes #416 
Closes #377 

- causing the `price` to go below the minimum when the auction remained open after 36 days
- causing the `totalToEndAuction` to be `0` due to rounding down minimum price for `quantity < 1e18`.
```
totalToEndAuction =(quantity * price()) / 1e18
```
For `quantity < 1e18` minimum price = `1e18 / quantity`, therefore substituting in the above
```
totalToEndAuction =(quantity * (1e18 / quantity)) / 1e18
totalToEndAuction = 1
```

For `quantity >= 1e18` minimum price = `1`, therefore substituting in the above
```
totalToEndAuction =(quantity * 1) / 1e18
totalToEndAuction = quantity / 1e18 which is always >1
```
Meaning `totalToEndAuction` is always `1` after a low quantity auction falls to the minimum price. 

- causing the `finalPrice` to be incorrect for low quantity auctions. After some deliberation, using the `finalPrice` in calculations introduces further inaccuracies in `claim`, so we are selectively calculating a claim based on whether the minimum price was reached and is this a low quantity auction, details below:
```
claim = amount * TOKEN_MULTIPLIER / finalPrice
```
where `finalPrice = mul(receivedTotal, TOKEN_MULTIPLIER) / quantity`, substituting in the above:
```
claim = amount * TOKEN_MULTIPLIER / ((receivedTotal * TOKEN_MULTIPLIER) / quantity)
           = amount * quantity / receivedTotal
```
In cases where `amount * quantity < receivedTotal` leading to 0 result, we can still use the original calculation including the `finalPrice`


.eslintrc
```diff
@@ -16,7 +16,8 @@
     "flowtype/require-valid-file-annotation": "off",
     "import/named": "off",
     "no-console": "off",
-    "no-unused-expressions": "off"
+    "no-unused-expressions": "off",
+    "no-await-in-loop": "off"
```
kronosapiens: 🥇

contracts/ColonyNetworkAuction.sol
```diff
@@ -137,7 +137,7 @@ contract DutchAuction is DSMath {
   auctionStartedAndOpen
   returns (uint)
   {
-    return mul(quantity, price()) / TOKEN_MULTIPLIER;
+    return sub((mul(quantity, price()) / TOKEN_MULTIPLIER), 1);
```
kronosapiens: Why? It needs to be at least 1 or it throws?

test/colony-network-auction.js
```diff
@@ -297,6 +298,64 @@ contract("ColonyNetworkAuction", accounts => {
...
+      await clny.approve(tokenAuction.address, quantity, { from: BIDDER_1 });
+
+      let endTime = await tokenAuction.endTime();
+      const amount = new BN(10).pow(new BN(20));
+
+      while (endTime.isZero()) {
+        await forwardTime(SECONDS_PER_DAY * 3, this);
+
+        await tokenAuction.bid(amount, { from: BIDDER_1 });
+        // const price = await tokenAuction.price();
```
kronosapiens: What are these **COMMENTED OUT LINES** doing here.

test/colony-network-auction.js
```diff
@@ -297,6 +298,64 @@ contract("ColonyNetworkAuction", accounts => {
...
+        // const receivedTotal = await tokenAuction.receivedTotal();
+        endTime = await tokenAuction.endTime();
+        // console.log("price", price.toString());
+        // console.log("totalToEndAuction", totalToEndAuction.toString());
+        // console.log("receivedTotal", receivedTotal.toString());
+      }
+
+      await tokenAuction.finalize();
+      const receivedTotal = await tokenAuction.receivedTotal();
+      const endPrice = new BN(10)
```
kronosapiens: Could import `WAD` from constants.

test/colony-network-auction.js
```diff
@@ -297,6 +298,64 @@ contract("ColonyNetworkAuction", accounts => {
...
+      await clny.approve(tokenAuction.address, quantity, { from: BIDDER_1 });
+
+      await forwardTime(SECONDS_PER_DAY * 34, this);
+      let endTime = await tokenAuction.endTime();
+      const amount = new BN(10).pow(new BN(17));
+
+      while (endTime.isZero()) {
+        await forwardTime(SECONDS_PER_DAY, this);
+        await tokenAuction.bid(amount, { from: BIDDER_1 });
+        // const price = await tokenAuction.price();
```
kronosapiens: https://gph.is/1LnCqDw

test/colony-network-auction.js
kronosapiens: I mean, calling it something else is an option, but I don't 100% see the benefit since other codebases speak in WADs and we need some word to describe `10 ** 18`. Mostly I think the semantics are clearer when we share the same constant, since the intention is not `10 ** 18` as a number but rather as an order of magnitude for other numbers. We can call it something different if we feel strongly.

test/colony-network-auction.js
kronosapiens: I mean, I'll live it just seems not ideal.

elenadimitrova:
Putting this on hold as I'm still unhappy with the various patches to make this work sufficiently accurately with the very wide range of price and quantity.

helpers/test-helper.js
```diff
@@ -114,6 +114,36 @@ export function web3GetRawCall(params) {
   });
 }
 
+export function stopMining() {
```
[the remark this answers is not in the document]
elenadimitrova: Not used anywhere in the final version but indeed I tried to use them to eliminate the proximity calculations in finalPrice estimates. That didn't work because we have to forward time as well in between opening the auction, bidding and closing it :(

contracts/ColonyNetworkAuction.sol
```diff
@@ -187,29 +214,38 @@ contract DutchAuction is DSMath {
...
   auctionFinalized
   returns (bool)
   {
-    uint amount = bids[msg.sender];
+    uint amount = bids[recipient];
     require(amount > 0, "colony-auction-zero-bid-total");
 
-    uint tokens = mul(amount, TOKEN_MULTIPLIER) / finalPrice;
+    uint tokens;
+    if (mul(amount, quantity) < receivedTotal) {
```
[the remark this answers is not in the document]
elenadimitrova: Yes I actually wrote the justification in the PR description as it's non-trivial but definitely need to add notes here too.

test/colony-network-auction.js
```diff
@@ -304,6 +307,450 @@ contract("Colony Network Auction", accounts => {
...
+        await tokenAuction.bid(amount, { from: BIDDER_1 });
+        endTime = await tokenAuction.endTime();
+      }
+
+      await tokenAuction.finalize();
+      const finalPrice = await tokenAuction.finalPrice();
+      expect(finalPrice).to.eq.BN(new BN("40000000000000000000000"));
+    });
+
+    it("functions correctly even when price has reached the minimum", async () => {
```
area: This isn't the scenario tested in this test, which has led to a race condition and the flickery test. If 10**16 tokens are minted, then the minimum price is 100. The price starts at 10**36, and goes down by an order of magnitude each day, so I'd expect the minimum price to only be reached at 34 days (and then only one person to be able to bid 1 CLNY wei before the auction closed).

elenadimitrova: I've named the test more appropriately now and changed the calculations to be dynamic so to avoid marginal inaccuracies.

Issue #305, opened 2018-08-08
Use bn-chai for big numbers testing


elenadimitrova:
Testing big number results using `chai` is just awful and you have two (terrible) ways to do that: either use `assert.isTrue(someBN.eq(otherBN))` or `assert.equal(someBN.toString(), otherBN.toString())`. So far in the network we're favoured the latter as on test failures that gives the actual different big number values back as strings, the former just says that the assertion failed with no additional information on the differing values. The advantage of the latter however is that it uses the native `bn.js` comparison thus avoiding conversions and comparing strings, rather than numbers. 

`assert.deepEqual` and `assert.strictEqual` were also tested comparing `bn.js` instances and they both failed.

Here comes [`bn-chai`](https://github.com/EthWorks/bn-chai) which gives us the best of both worlds via natural language syntax e.g. `expect(someBN).to.eq.BN(otherBN);` with errors giving you all the information you need on a failure, e.g. 
`AssertionError: expected 57896044618658097711785492504343953926634992332820282019728792003956564819965 to equal 57896044618658097711785492504343953926634992332820282019728792003956564819966`

Refactoring of the `assert.isTrue` tests using the above syntax was done as part of #288 in commit 
https://github.com/JoinColony/colonyNetwork/pull/288/commits/4db2ee7887a1931252e08bb49a8126e7039cde2c 
This issue targets the remaining `assert.equal` instance comparing big numbers and also additional conversions we do. The aim is to stay in native `bn.js` world for as much as possible.

Issue #283, opened 2018-07-17
Return structs in read functions


elenadimitrova:
With the experimental ABIv2 we can return Structs out of function calls (as well as pass those in). It it worth exploring as a design in our contracts potentially with the following benefits:

- Enforces return data structures to comply with those declared in storage. This was the reason behind the misalignment highlighted in `getSkill` return value issue here  https://github.com/JoinColony/colonyNetwork/issues/249
- Simplifies upgrades as extending a struct (e.g. adding a new member) won't necessitate upgrade of the function return parameter type - subject to testing

Candidates for this refactoring are essentially all getter functions where a struct is returned: `getSkill`, `getTask`, `getDomain`, `getTaskWorkRatings`, `getTaskRole`, `getRewardPayoutInfo`, `getUserLock` etc

filiplazovic:
I encountered some strange error when i tried to used `ABIEncoderV2` in `ReputationMiningCycle`. After a little research i found out that creating a new instance of a contract with `pragma experimental "ABIEncoderV2"` in it costs ~300k gas more. I will investigate on this a little bit more and update you if i find more info

PR #435, opened 2018-11-14
Upgrades truffle to beta.2 version

elenadimitrova:
Update all packages to latest including notably:

- `truffle` to @beta 2 release which reintroduces support for `watch` flag that I hope to improve dev quality of life. Additionally this adds support for `ABIEncoderV2` including returning `struct` s from functions which we need for #283 

**Notes**: 
      `0x0`  is no longer considered a valid address, use `ZERO_ADDRESS` from `constants` instead.
       When calling a function with integer inputs, pass in `BN.js` instances rather than strings and JS numbers, e.g. there's no need for `toString()` and `toNumber()`. The remaining instances of those are used in `assert` compares and those will be eliminated when #305 is completed.  

- `web3` to beta.36 which is part of above but also unfortunately [introduces an issue](https://github.com/ethereum/web3.js/issues/1920) with encoding ABI using `BN.js` input parameters 

- [Related `eth-gas-reporter` issue](https://github.com/cgewecke/eth-gas-reporter/issues/104) relevant here as it affected this set of changes


elenadimitrova:
Note: There appears to be an issue with `truffle` getting deployed transactions to `parity` so putting this on hold for now.

PR #436, opened 2018-11-15
Rename Colony owner role to founder

### LINKED ISSUE #369: Rename the 'owner' role
DSAuth has it's own concept of an `owner`, which is the one person that can call any function. For Colonies, there is a role in DSRoles we're calling owner, which isn't a superuser (indeed, we make sure [there is no such user](https://github.com/JoinColony/colonyNetwork/blob/58c3ea1d46f9492807bf9bc041da6aec2145e748/contracts/ColonyNetwork.sol#L150). This is going to trip us up at some point if we don't rename the role in DSRoles...

elenadimitrova:
Closes #369 

Unfortunately `ENS` and `MultiSigWallet` contracts also declare their separate `owner`s, however since those are built in their specifications and they are probably sufficiently detached from the Colony permissions model we've chosen not to refactor it.

### DIFF
--- a/contracts/ContractRecovery.sol
+++ b/contracts/ContractRecovery.sol
@@ -72,7 +72,7 @@ contract ContractRecovery is CommonStorage {
 
   function exitRecoveryMode() public recovery auth {
     uint totalAuthorized = recoveryRolesCount;
-    // Don't double count the owner (if set);
+    // Don't double count the founder (if set);
     if (owner != 0x0 && !CommonAuthority(authority).hasUserRole(owner, RECOVERY_ROLE)) { 
       totalAuthorized += 1; 
     }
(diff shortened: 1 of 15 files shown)


contracts/ContractRecovery.sol
```diff
@@ -72,7 +72,7 @@ contract ContractRecovery is CommonStorage {
 
   function exitRecoveryMode() public recovery auth {
     uint totalAuthorized = recoveryRolesCount;
-    // Don't double count the owner (if set);
+    // Don't double count the founder (if set);
```
area: The comment has been changed, but the code has not...

elenadimitrova: I think the code doesn't need to change. The comment needs to be reverted.

elenadimitrova: This is the owner which is derived down from `DSAuth`

PR #439, opened 2018-11-16
Return structs from function calls

### LINKED ISSUE #283: Return structs in read functions

elenadimitrova:
Closes #283

Depends on the truffle upgrade in #435 which introduces support for [`ABIEncoderV2`](https://solidity.readthedocs.io/en/v0.5.0/layout-of-source-files.html?highlight=ABIEncoderV2#abiencoderv2) including returning structs from functions. 

Here we demonstrate how `getDomain` can return the `Domain` struct. In tests this is backward compatible as it currently returns the two sets for the same values, once as plain data types and secondly as named parameters. For instance `getDomain` returns: 
```
[ '4', '1', skillId: '4', potId: '1' ]
```
so we can do both `domain[0]` and `domain.skillId`.

Overall to proceed we should consider whether using `ABIEncoderV2` is production ready for us. There are varying opinions of this out there, [`ethers` for example](https://blog.ricmoo.com/solidity-abiv2-a-foray-into-the-experimental-a6afd3d47185) is (as always) on the cautious side, while `0x` are sufficiently confident to [roll it out in production back in February](https://blog.0xproject.com/development-update-6-february-2018-8d6f6018eabf).

At present remix IDE is not yet working for the new 0.5.0 compiler so cannot try this with the new red hot `solc` release. However there's nothing in the release notes mentioning support for this and I'm more certain than not that returning structs is still only in the experimental `ABIEncoderV2`.

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -17,6 +17,7 @@
 
 pragma solidity ^0.4.22;
 pragma experimental "v0.5.0";
+pragma experimental "ABIEncoderV2";
 
 import "./ColonyStorage.sol";
 import "./ITokenLocking.sol";
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -322,7 +322,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
 
     uint256 userTokens;
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
-    (, userTokens,) = tokenLocking.getUserLock(address(token), msg.sender);
+    userTokens = tokenLocking.getUserLock(address(token), msg.sender).balance;
 
     require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
     require(userReputation > 0, "colony-reward-payout-invalid-user-reputation");
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -15,7 +15,7 @@
   along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
 */
 
-pragma solidity ^0.4.23;
+pragma solidity ^0.4.22;
 pragma experimental "v0.5.0";
 
 import "./ColonyStorage.sol";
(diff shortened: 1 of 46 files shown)


area:
Given we're only proposing we use it for reading from a contract for now, rather than passing things in (though I can see it being useful for packing the inputs to `respondToChallenge` down even further), I can't see an issue with it still being considered experimental?

contracts/ColonyFunding.sol
```diff
@@ -15,8 +15,9 @@
   along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
 */
 
-pragma solidity ^0.4.23;
+pragma solidity ^0.4.22;
```
elenadimitrova: ? 🤔

area: No clue what happened here :confused:

contracts/ColonyNetwork.sol
```diff
@@ -70,9 +71,16 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+  function getSkill(uint256 _skillId) public view returns (Skill skill) {
+    skill = skills[_skillId];
+  }
+
+  function isGlobalSkill(uint256 _skillId) public view returns (bool globalSkill) {
+    globalSkill = skills[_skillId].globalSkill;
+  }
+
+  function getSkillNParents(uint256 _skillId) public view returns (uint256 nParents) {
+    nParents = skills[_skillId].nParents;
```
elenadimitrova: I can't understand why this is needed?

PR #440, opened 2018-11-20
Submit JRH with submission

### LINKED ISSUE #404: JRH should be submitted alongside proposed new hashes
A bad actor wishes to cause the submitted (correct) new reputation state to not be selected. They can do this by submitting another reputation state of their choice, and then once `submitJRH` becomes available for the good state, submitting a bad JRH for the good state (which should be the root hash for the tree generated by taking the correct justification tree and replacing all leaves except the first and last with garbage). This means the good submission cannot be defended correctly, and the bad submission will be accepted almost by default.

The solution is to accept the JRH in `submitNewHash`, validating the first and last leaves at the time, and treating submissions with the same `nNodes` and same `proposedNewRootHash` but different `jrh` as different submissions, which should be played off against each other as is currently done for submissions with the same `proposedNewRootHash` but different `nNodes`.

### LINKED ISSUE #446: confirmBinarySearchResult guard can be worked around
Currently, the guard for `confirmBinarySearchResult` looks like

```
    require(disputeRounds[round][idx].lowerBound == disputeRounds[round][idx].upperBound, "colony-reputation-binary-search-incomplete");
    require(
      2**(disputeRounds[round][idx].challengeStepCompleted - 2) <= disputeRounds[round][idx].jrhNnodes,
      "colony-reputation-binary-search-result-already-confirmed"
    );
```
However, consider a malicious submission that doesn't confirm their JRH, but their opponent does. The opponent cannot confirm the binary search result, because it hasn't occurred. However, by not confirming their JRH, the malicious submission has lowerBound and upperBound set to `0`, and so will pass the first `require`. They also pass the second `require`, as `disputeRounds[round][idx].challengeStepCompleted - 2` will be -1, interpreted as `UINTMAX` due to an underflow, and then `2**UINTMAX` overflows prodigiously to be 0, which is equal to `disputeRounds[round][idx].jrhNnodes` for a submission with no confirmed JRH.

Both submissions have the same `challengeStepCompleted`, and so both will be invalidated even if one was a correct submission.

### LINKED ISSUE #447: A new node can be claimed to be added to 'time out' legitimate submission
If a bad miner silently increases the number of nodes they claim a state has (without actually adding a node, thus keeping the hash the same), the dispute resolution process will fail against even a well-defended, good submission.

We should check that (DISAGREE_STATE_NNODES - AGREE_STATE_NNODES) is 1 (a new node has been added) or 0 (an existing node has been updated).

area:
# New type of Patricia tree
I introduce a new type of Patricia tree, where the key is not hashed before it is inserted in to the tree. As a result, the `insert` command now takes a `bytes32` rather than a `bytes`. `PatriciaTreeProofs` now has two `getImpliedRoot` functions, called `getImpliedRootHashKey` and `getImpliedRootNoHashKey`, for clarity.

`PatriciaTree.sol` is the original, `HashKey` contract, and `PatriciaTreeNoHash.sol` is the new one. Each of these only has a `getImpliedRoot` function, which behaves in the way required to verify proofs from the tree in question.

# JRH now submitted alongside proposedNewRootHash

The justification for this can be found in #404. Instead of being submitted as a separate transction, it is confirmed as a separate transaction, which involves proving that the key at `0x00..000` corresponds to the currently accepted state, and the key at `0x00...nNodes` corresponds to the state they submitted.

# Changes to the Justification Tree 

The justification tree is now one of these new types of Patricia Trees. This is so that we can be more specific about the proofs we allow during the confirmation of the JRH (described above) and the binary search for the first discrepancy. By not hashing the keys, and the keys (of a correctly built JRH) being of the form `0x00...000`, `0x000...001`  up to `0x000...nNodes`, we can check that the proofs supplied are of the correct length (so there aren't any extra keys added afterwards), and we can also check that there aren't extra nodes in the middle deeper than we expected:

```
          R
        /   \
       o     L
      / \
     L   o
        / \
       L   E   

```

We only expected three leaves (`L`) in this tree, but the tree has four (`E` has been inserted). If we didn't verify the lengths of the proofs, we wouldn't be able to detect this subterfuge - they would be able to proof that all of the correct nodes were in the tree during dispute resolution, and the process would get stuck. By not hashing the leaves, an attempt to prove the second correct leaf is in the tree will fail, because we can tell the submitted proof is too long.

## Calculation of the correct number of siblings in proof

I recognise this calculation is quite obtuse. It will be easier to verify that the trees in the patricia tree implementation we are using are 'built out' from the root node. So for example, if we add nodes with keys 0x00, 0x01, 0x02, 0x03 and 0x04

```
   A         C               D                   F                       I
            / \             / \                 /  \                    / \  
           A   B           C   E               C    G                  F   J
                          / \                 / \  / \                /  \
                         A   B               A   B E  H              C    G
                                                                    / \  / \
                                                                   A   B E  H
```

You can also think about this in terms of binary - Node `J`, in the last tree, with key `0x04`, is `0b100`. Once the tree has gone down the `1` branch, there are no other keys it has to be distinguished from, and so it is placed only one level down from `I`.  Conversely, `0b000`, `0b001`, `0b010` and `0b011` require more branches to distinguish between them.

The power-of-two-bit-twiddling is taken from [this incredible page](https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2), extending to 256 bit numbers.

# Changes to Patricia Trees

I believe that the change from `>1` to `>=1` [here](https://github.com/JoinColony/colonyNetwork/compare/bug/404-submit-jrh-with-hash?expand=1#diff-1c81aa01cf67d1d93b7a053584a0feafR224) is fixing a mistake in the original implementation that was just never discovered (the odds of finding two hashes that differ by only the final bit are astronomical). 

The loop for Patricia tree proofs has been changed from being based on the branchmask to being based on the number of siblings. For full proofs, the number of 1s in the branchmask is the same as the number of siblings, but the same is not true for the partial proofs that we need for the new binary search algorithm. (We need one more bit in the branchmask than we do siblings for a partial proof; by simply looping over the number of siblings, we get to skip a lot of fiddly-bit-setting that could easily go wrong).

# Changes to binary search algorithm

The previous binary search algorithm was wrong, and should serve as a cautionary tale for diverging from the whitepaper. The new binary search algorithm follows that laid out in the whitepaper (§7.5.1, Part 2, 'Searching for a discrepancy'). I will not go in to detail here - see the whitepaper for the description. I will note that while more correct, this algorithm has worse guarantees about efficiency than the last one (which always bisected the search space at every step), hence why some tests have required additional binary search steps. The changing of algorithm necessitated the addition of `getFinalPairAndImpliedRootNoHash` to `PatriciaTreeProofs.sol`

# Test for nNodes disagreement
Elena requested tests for disagreement over Nnodes; two types are included. In one, the client just lies about nNodes it's submitting, and so fairly straightforwardly is unable to confirm the JRH. In the second type, the number of nodes is incremented by an extra amount during a transition (there are two tests - one when this transition is meant to add a new node, and one when it's an update and is not meant to add a new node). This wasn't able to be sorted by the existing code (which was a bug, see #447) without extra requires being introduced to check that the number of nodes increases (or doesn't) as expected. I believe these new requires obsolete an existing one, but I'm leaving the one I believe to be obsoleted (and the corresponding test) in for now in case I am mistaken.

# ConfirmBinarySearchResult guard
Issue detailed in #446 - I have solved it by adding an additional guard that essentially requires that the justification root hash has been confirmed.

Closes #404 
Closes #446 
Closes #447


contracts/ReputationMiningCycle.sol
```diff
@@ -296,22 +304,26 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
...
-    bytes32 jrh = disputeRounds[round][idx].jrh;
-
-    bytes memory targetNodeBytes = new bytes(32);
-    assembly {
-      mstore(add(targetNodeBytes, 0x20), targetNode)
-    }
-
-    bytes32 impliedRoot = getImpliedRoot(targetNodeBytes, jhIntermediateValue, branchMask, siblings);
-    require(impliedRoot==jrh, "colony-reputation-mining-invalid-binary-search-response");
+    require(impliedRoot==targetHashDuringSearch, "colony-reputation-mining-invalid-binary-search-response");
```
kronosapiens: Spaces around `==`?

contracts/ReputationMiningCycle.sol
```diff
@@ -545,22 +582,56 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
...
+    while (nextPowerOfTwo > 1) {
+      layers += 1;
+      nextPowerOfTwo >>= 1;
+    }
+    return layers;
+  }
+
   function processBinaryChallengeSearchStep(uint256 round, uint256 idx, uint256 targetNode) internal {
     uint256 opponentIdx = (idx % 2 == 1 ? idx-1 : idx + 1);
+    uint256 searchWidth = disputeRounds[round][idx].upperBound - disputeRounds[round][idx].lowerBound + 1;
```
kronosapiens: Adding parens here would clarify the intended order-of-operations.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -642,12 +654,60 @@ class ReputationMiner {
...
-    const [branchMask, siblings] = await this.justificationTree.getProof(targetNodeKey);
-    return repCycle.respondToBinarySearchForChallenge(round, index, intermediateReputationHash, branchMask, siblings, {
+    const proof = await this.justificationTree.getProof(targetNodeKey);
+    const [branchMask] = proof;
+    let [, siblings] = proof;
+
+    // Trim the proof if needed
+    // How long should the proof be?
+
+    // function pow2ceil(v){
```
kronosapiens: Is this *COMMENTED OUT CODE* supposed to be here?

packages/reputation-miner/ReputationMiner.js
```diff
@@ -642,12 +654,60 @@ class ReputationMiner {
...
-    const targetNode = submission[8].add(submission[9]).div(2);
+    const searchInfo = await repCycle.getDisputeRounds2(round, index);
+    const targetNode = submission[8];
     const targetNodeKey = ReputationMiner.getHexString(targetNode, 64);
 
     const intermediateReputationHash = this.justificationHashes[targetNodeKey].jhLeafValue;
-    const [branchMask, siblings] = await this.justificationTree.getProof(targetNodeKey);
-    return repCycle.respondToBinarySearchForChallenge(round, index, intermediateReputationHash, branchMask, siblings, {
+    const proof = await this.justificationTree.getProof(targetNodeKey);
+    const [branchMask] = proof;
```
kronosapiens: TIL a little more about JS variable assignment; would've thought it was `[branchMask,]`

packages/reputation-miner/patricia-test.js
```diff
@@ -129,5 +129,39 @@ contract("Javascript Patricia Tree", accounts => {
         assert.equal(jsSiblings[i], realSiblings[i]);
       }
     });
+
+    it("should recover identical root hashes from proofs", async () => {
+      const dog = web3Utils.fromAscii("dog");
```
kronosapiens: Good theme. 🐶 🐒

packages/reputation-miner/patricia-test.js
```diff
@@ -129,5 +129,39 @@ contract("Javascript Patricia Tree", accounts => {
...
+      await jsClient.reputationTree.insert(dog, rover);
+      await solClient.reputationTree.insert(dog, rover, { gasLimit: 4000000 });
+      await realPatriciaTree.insert(dog, rover);
+
+      const [jsMask, jsSiblings] = await jsClient.reputationTree.getProof(dog);
+      const [solMask, solSiblings] = await solClient.reputationTree.getProof(dog);
+      // This is the difference between what an ethers contract returns (above) and what a
+      // truffle contract returns (below)
+      const res = await realPatriciaTree.getProof(dog);
+      const realMask = res["0"];
```
kronosapiens: Keys are strings? Why not an array?

packages/reputation-miner/test/MaliciousReputationMinerWrongJRH.js
```diff
@@ -0,0 +1,67 @@
...
+    // Get how much we've staked, and thefore how many entries we have
+    let entryIndex;
+    const [, balance] = await this.tokenLocking.getUserLock(this.clnyAddress, this.minerAddress);
+    const reputationMiningWindowOpenTimestamp = await repCycle.getReputationMiningWindowOpenTimestamp();
+    const minStake = ethers.utils.bigNumberify(10).pow(18).mul(2000); // eslint-disable-line prettier/prettier
+    for (let i = ethers.utils.bigNumberify(startIndex); i.lte(balance.div(minStake)); i = i.add(1)) {
+      // Iterate over entries until we find one that passes
+      const entryHash = await repCycle.getEntryHash(this.minerAddress, i, hash); // eslint-disable-line no-await-in-loop
+
+      const miningCycleDuration = 60 * 60 * 24;
```
kronosapiens: Can we import this from `constants.js`?

kronosapiens:
Also, do you intend to bubble up the `<=` bug you identified to the original implementors?

packages/reputation-miner/patricia-test.js
area: They're actually both, but treating it as an array is more expected.

packages/reputation-miner/test/MaliciousReputationMinerWrongJRH.js
area: I think I'd rather not. We have the duration in `ReputationMiner.js` (i.e. the 'real' miner) too, and I wouldn't want to change it there either - I don't want this package repository (which will probably eventually be published as its own npm package) to be reliant on the whole of `colonyNetwork`.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -511,10 +520,12 @@ class ReputationMiner {
...
     }
-    //
+    // Get the JRH
+    const jrh = await this.justificationTree.getRootHash();
     // Submit that entry
-    const gas = await repCycle.estimate.submitRootHash(hash, this.nReputations, entryIndex);
-    return repCycle.submitRootHash(hash, this.nReputations, entryIndex, { gasLimit: `0x${gas.mul(2).toString()}` });
+    const gas = await repCycle.estimate.submitRootHash(hash, this.nReputations, jrh, entryIndex);
+
+    return repCycle.submitRootHash(hash, this.nReputations, jrh, entryIndex, { gasLimit: `0x${gas.toString(16)}` });
```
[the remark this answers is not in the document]
area: That'd be nice! If that all gets sorted before this is merged, I'll try it out.

packages/reputation-miner/test/MaliciousReputationMinerWrongJRH.js
kronosapiens: Fair enough

packages/reputation-miner/ReputationMiner.js
```diff
@@ -642,12 +656,35 @@ class ReputationMiner {
     const [round, index] = await this.getMySubmissionRoundAndIndex();
     const repCycle = await this.getActiveRepCycle();
     const submission = await repCycle.getDisputeRounds(round, index);
-    const targetNode = submission[8].add(submission[9]).div(2);
+    const searchInfo = await repCycle.getDisputeRounds2(round, index);
```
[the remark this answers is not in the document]
area: It has too many elements to return in a single function, so for now, it's not possible. I agree that once we are able to return the struct, we should as a single function.

contracts/ReputationMiningCycle.sol
```diff
@@ -352,35 +360,41 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
...
 
+
     // Check the proofs for the JRH
-    checkJRHProof1(jrh, branchMask1, siblings1, reputationRootHashNNodes);
+    checkJRHProof1(disputeRounds[round][index].jrh, branchMask1, siblings1, reputationRootHashNNodes);
     checkJRHProof2(
       round,
       index,
-      jrh,
+      disputeRounds[round][index].jrh,
```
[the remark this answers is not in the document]
area: Done it for `checkJRHProof2`, but not the other - pretty sure it'd cost more gas.

test/colony-network-mining.js
```diff
@@ -847,26 +872,331 @@ contract("ColonyNetworkMining", accounts => {
...
+
+      let addr = await colonyNetwork.getReputationMiningCycle(true);
+      let repCycle = await IReputationMiningCycle.at(addr);
+      await forwardTime(MINING_CYCLE_DURATION, this);
+      await repCycle.submitRootHash("0x12345678", 10, "0x00", 10, { from: MAIN_ACCOUNT });
+      await repCycle.confirmNewHash(0);
+
+      await giveUserCLNYTokens(colonyNetwork, MAIN_ACCOUNT, DEFAULT_STAKE);
+      await giveUserCLNYTokens(colonyNetwork, OTHER_ACCOUNT, DEFAULT_STAKE);
+      await giveUserCLNYTokens(colonyNetwork, accounts[2], DEFAULT_STAKE);
```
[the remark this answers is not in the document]
area: I'll change it for consistency.

packages/reputation-miner/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@colony/colony-js-contract-loader-fs": "1.8.1",
     "bn.js": "^4.11.8",
-    "ethers": "^4.0.13",
+    "ethers": "^4.0.17",
```
elenadimitrova: hmm

PR #441, opened 2018-11-20
Big Old Test Refactor 1

### LINKED ISSUE #317: Test suite refactoring

kronosapiens:
Partial soln to #317 

- Remove circular dependency between `constants.js` and `test-helper.js`
- Utilize bulk assignment of task properties when generating tasks (10-15% speedup).
- Replace `assert` with `assert.isTrue`.
- Clarify a reputation mining error message re: timeouts.
- Introduce `setupFinalizedTask` to remove recurring boilerplate.
- Introduce `advanceMiningCycleNoContest` and `getActiveRepCycle` to remove recurring boilerplate.

### DIFF
--- a/helpers/constants.js
+++ b/helpers/constants.js
@@ -1,5 +1,5 @@
 import { toBN, soliditySha3 } from "web3-utils";
-import { getRandomString } from "./test-helper";
+import shortid from "shortid";
 
 const MANAGER_ROLE = 0;
 const EVALUATOR_ROLE = 1;
@@ -19,14 +19,16 @@ const MANAGER_RATING = 2;
 const WORKER_RATING = 3;
 const RATING_MULTIPLIER = { 1: -1, 2: 1, 3: 1.5 };
 const SECONDS_PER_DAY = 86400;
-const RATING_1_SALT = soliditySha3(getRandomString(10));
-const RATING_2_SALT = soliditySha3(getRandomString(10));
+const RATING_1_SALT = soliditySha3(shortid.generate());
+const RATING_2_SALT = soliditySha3(shortid.generate());
 const RATING_1_SECRET = soliditySha3(RATING_1_SALT, MANAGER_RATING);
 const RATING_2_SECRET = soliditySha3(RATING_2_SALT, WORKER_RATING);
 
 const ACTIVE_TASK_STATE = 0;
 const CANCELLED_TASK_STATE = 1;
 const FINALIZED_TASK_STATE = 2;
+
+const UINT256_MAX = toBN(0).notn(256);
 
 const WAD = toBN(10).pow(toBN(18));
 const MIN_STAKE = WAD.muln(2000);
@@ -60,6 +62,7 @@ module.exports = {
   ACTIVE_TASK_STATE,
   CANCELLED_TASK_STATE,
   FINALIZED_TASK_STATE,
+  UINT256_MAX,
   WAD,
   MIN_STAKE,
   DEFAULT_STAKE,
--- a/helpers/constants.js
+++ b/helpers/constants.js
@@ -29,6 +29,7 @@ const CANCELLED_TASK_STATE = 1;
 const FINALIZED_TASK_STATE = 2;
 
 const UINT256_MAX = toBN(0).notn(256);
+const INT256_MAX = toBN(0).notn(255);
 
 const WAD = toBN(10).pow(toBN(18));
 const MIN_STAKE = WAD.muln(2000);
@@ -63,6 +64,7 @@ module.exports = {
   CANCELLED_TASK_STATE,
   FINALIZED_TASK_STATE,
   UINT256_MAX,
+  INT256_MAX,
   WAD,
   MIN_STAKE,
   DEFAULT_STAKE,
--- a/test/colony-network-recovery.js
+++ b/test/colony-network-recovery.js
@@ -12,7 +12,9 @@ import {
   currentBlockTime,
   checkErrorRevert,
   submitAndForwardTimeToDispute,
-  web3GetStorageAt
+  web3GetStorageAt,
+  getActiveRepCycle,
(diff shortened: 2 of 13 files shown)


test/reputation-update.js
```diff
@@ -201,51 +197,39 @@ contract("Colony Reputation Updates", accounts => {
...
-      const maxUIntNumber = new BN(2).pow(new BN(255)).sub(new BN(1));
-      await fundColonyWithTokens(metaColony, colonyToken, maxUIntNumber);
-      // Split the tokens as payouts between the manager and worker
+      await fundColonyWithTokens(metaColony, colonyToken, INT256_MAX);
+
+      // Split the tokens as payouts between the manager, evaluator, and worker
       const managerPayout = new BN("2");
       const evaluatorPayout = new BN("1");
-      const workerPayout = maxUIntNumber.sub(managerPayout).sub(evaluatorPayout);
+      const workerPayout = INT256_MAX.sub(managerPayout).sub(evaluatorPayout);
```
elenadimitrova: Why an int max rather than uint max?

kronosapiens: Because reputation is based on `ints`, so using `UINT256_MAX` would be interpreted as a negative value.

elenadimitrova: Okay yes, it's just a badly named old variable, i.e. `maxUIntNumber`

test/colony-network-recovery.js
```diff
@@ -250,28 +239,16 @@ contract("Colony Network Recovery", accounts => {
...
           await miningClient.addLogContentsToReputationTree();
-          await forwardTime(MINING_CYCLE_DURATION, this);
-          await miningClient.submitRootHash();
-
-          addr = await colonyNetwork.getReputationMiningCycle.call(true);
-          repCycle = await IReputationMiningCycle.at(addr);
+          await advanceMiningCycleNoContest(colonyNetwork, this, miningClient);
 
+          const addr = await colonyNetwork.getReputationMiningCycle(true);
+          const repCycle = await IReputationMiningCycle.at(addr);
```
elenadimitrova: Use the new `getActiveMineCycle` helper?

kronosapiens: Ok updated

area: There are a still a few places where a line or two could be saved with this pattern (e.g. https://github.com/JoinColony/colonyNetwork/pull/441/files#diff-15aaaaa81a55d904a938772d3e2a3a82R387) - basically, most (though not all) places where `getReputationMiningCycle(true)` still exists.

kronosapiens: The only kicker is that we occasionally use `addr` in comparisons, but we could always drop in `repCycle.address`.

kronosapiens: Alright, try this on for size

helpers/constants.js
```diff
@@ -18,15 +18,18 @@ const MANAGER_RATING = 2;
 const WORKER_RATING = 3;
 const RATING_MULTIPLIER = { 1: -1, 2: 1, 3: 1.5 };
 const SECONDS_PER_DAY = 86400;
-const RATING_1_SALT = web3Utils.soliditySha3(getRandomString(10));
-const RATING_2_SALT = web3Utils.soliditySha3(getRandomString(10));
+const RATING_1_SALT = web3Utils.soliditySha3(shortid.generate());
+const RATING_2_SALT = web3Utils.soliditySha3(shortid.generate());
```
elenadimitrova: When you rebase this pls use the `web3Utils` imports at the top rather than explicit references.

kronosapiens: Kk

PR #463, opened 2018-12-14
Fixed wrong install order in docs

wmitsuda:
The install instructions doesn't work from scratch; fixed wrong order.


elenadimitrova:
Thanks for the correction @wmitsuda !

PR #464, opened 2018-12-17
Add expected errors to accommodateChallengeAndInvalidateHash

area:
This add another `errors` parameter to `accommodateChallengeAndInvalidateHash`, which contains two properties - `client1` and `client2`. Each of these, if it exists, is expected to contain an object that contains one of the properties:

* `confirmJustificationRootHash`
* `respondToBinarySearchForChallenge`
* `confirmBinarySearchResult`
* `respondToChallenge`

If defined, these properties should contain the string of the expected error to be returned. The exception is `respondToBinarySearchForChallenge`, which should contain an array, with each element containing the expected error (if any) for the `nth` call to `respondToBinarySearchForChallenge` for that client - for example, if `error-string` was expected on the third call to `respondToBinarySearchForChallenge` for client 2, a valid object to pass would be

```
{ client2: { respondToBinarySearchForChallenge: [undefined, undefined, "error-string"] } } 
```

If an expected error is thrown, none of the other calls are fired aside from the same call on `client2` if `client1` was the one that errored.

Part of #317

### DIFF
--- a/test/colony-network-mining.js
+++ b/test/colony-network-mining.js
@@ -9,6 +9,7 @@ import {
   forwardTime,
   checkErrorRevert,
   checkErrorRevertEthers,
+  checkErrorIfRevertEthers,
   makeReputationKey,
   makeReputationValue,
   currentBlock,
@@ -134,7 +135,7 @@ contract("ColonyNetworkMining", accounts => {
     await clny.burn(userBalance, { from: MAIN_ACCOUNT });
   });
 
-  async function accommodateChallengeAndInvalidateHash(test, client1, client2) {
+  async function accommodateChallengeAndInvalidateHash(test, client1, client2, message) {
     let round2;
     let idx2;
     let toInvalidateIdx;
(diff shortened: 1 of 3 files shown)


kronosapiens:
How do you feel about appending the error string attribute with `Error`, as in `confirmJustificationRootHashError`? Having two attributes with the same name seems risky.

Also, can we move `navigateChallenge` to `test-helpers` with everything else? It seems like it belongs there more so than nestled amidst the `before` and `beforeEach` blocks.

test/colony-network-mining.js
```diff
@@ -91,6 +92,8 @@ contract("ColonyNetworkMining", accounts => {
     metaColony = await IMetaColony.at(metaColonyAddress);
     const clnyAddress = await metaColony.getToken();
     clny = await Token.at(clnyAddress);
+    goodClient = new ReputationMiner({ loader: contractLoader, minerAddress: MAIN_ACCOUNT, realProviderPort: REAL_PROVIDER_PORT, useJsTree });
```
elenadimitrova: The instantiation or the `goodClient` is in the `beforeEach` no?

elenadimitrova:
@kronosapiens you're right, I've moved the helper functions into the `test-helper` where they should be

test/colony-network-mining.js
kronosapiens: Yup! ~Arguably~ we could instantiate clients in the tests only when they are needed (since we occasionally overwrite the `badClients`, but one thing at a time.

PR #466, opened 2018-12-18
Require that miners stake before the mining cycle begins

### LINKED ISSUE #47: Submission of a new root hash - eligibility calculation

kronosapiens:
Closes #47

Tests in `colony-network-mining.js` now *begin* with `MAIN_ACCOUNT` having `DEFAULT_STAKE` staked. This is to allows `MAIN_ACCOUNT` to mine new mining cycles. For newly-staked accounts to be able to mine, we must advance one mining cycle after staking.

### DIFF
--- a/test/colony-network-mining.js
+++ b/test/colony-network-mining.js
@@ -265,7 +265,7 @@ contract("ColonyNetworkMining", accounts => {
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT, DEFAULT_STAKE);
       await forwardTime(MINING_CYCLE_DURATION, this);
 
-      const repCycle = await getActiveRepCycle(colonyNetwork);
+      let repCycle = await getActiveRepCycle(colonyNetwork);
       await checkErrorRevert(
         repCycle.submitRootHash("0x12345678", 10, "0x00", 10, { from: OTHER_ACCOUNT }),
         "colony-reputation-mining-stake-too-recent"
@@ -273,7 +273,7 @@ contract("ColonyNetworkMining", accounts => {
 
       await advanceMiningCycleNoContest({ colonyNetwork, test: this });
       await forwardTime(MINING_CYCLE_DURATION, this);
-
+      repCycle = await getActiveRepCycle(colonyNetwork);
       await repCycle.submitRootHash("0x12345678", 10, "0x00", 10, { from: OTHER_ACCOUNT });
     });
 
--- a/test/colony-network-mining.js
+++ b/test/colony-network-mining.js
@@ -229,7 +229,7 @@ contract("ColonyNetworkMining", accounts => {
 
     it("should allow a new reputation hash to be submitted", async () => {
       const repCycle = await getActiveRepCycle(colonyNetwork);
-      await forwardTime(MINING_CYCLE_DURATION);
+      await forwardTime(MINING_CYCLE_DURATION, this);
       await repCycle.submitRootHash("0x12345678", 10, "0x00", 10, { from: MAIN_ACCOUNT });
 
       const submitterAddress = await repCycle.getSubmittedHashes("0x12345678", 10, "0x00", 0);
@@ -238,7 +238,7 @@ contract("ColonyNetworkMining", accounts => {
 
     it("should only allow the first submission after the window closes to be accepted", async () => {
       const repCycle = await getActiveRepCycle(colonyNetwork);
-      await forwardTime(MINING_CYCLE_DURATION + 400); // Well after the window has closed
+      await forwardTime(MINING_CYCLE_DURATION + 400, this); // Well after the window has closed
       await repCycle.submitRootHash("0x12345678", 10, "0x00", 10, { from: MAIN_ACCOUNT });
 
       await checkErrorRevert(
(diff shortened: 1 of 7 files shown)


test/colony-network-mining.js
```diff
@@ -314,9 +316,24 @@ contract("ColonyNetworkMining", accounts => {
       assert.isTrue(nSubmittedHashes.isZero());
     });
 
-    it("should not allow someone to withdraw their stake if they have submitted a hash this round", async () => {
-      await giveUserCLNYTokensAndStake(colonyNetwork, MAIN_ACCOUNT, DEFAULT_STAKE);
+    it("should not allow someone to submit a new reputation hash if they stake after the cycle beings", async () => {
+      await forwardTime(1);
```
elenadimitrova: Why this?

test/colony-network-mining.js
```diff
@@ -288,8 +290,8 @@ contract("ColonyNetworkMining", accounts => {
     });
 
     it("should only allow the first submission after the window closes to be accepted", async () => {
-      await giveUserCLNYTokensAndStake(colonyNetwork, MAIN_ACCOUNT, DEFAULT_STAKE);
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT, DEFAULT_STAKE);
+      await advanceMiningCycleNoContest({ colonyNetwork, test: this });
```
elenadimitrova: If this is added here shouldn't we remove the next 3 lines?

test/colony-network-mining.js
```diff
@@ -751,9 +740,9 @@ contract("ColonyNetworkMining", accounts => {
 
   describe("Rewards and punishments of good and bad submissions", () => {
     it("should punish all stakers if they misbehave (and report a bad hash)", async () => {
-      await giveUserCLNYTokensAndStake(colonyNetwork, MAIN_ACCOUNT, DEFAULT_STAKE);
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT, DEFAULT_STAKE);
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT2, DEFAULT_STAKE);
+      await advanceMiningCycleNoContest({ colonyNetwork, test: this });
```
elenadimitrova: I can't work out why sometimes a new mining cycle is started in tests now and sometimes not?

test/colony-network-mining.js
```diff
@@ -314,9 +316,24 @@ contract("ColonyNetworkMining", accounts => {
       assert.isTrue(nSubmittedHashes.isZero());
     });
 
-    it("should not allow someone to withdraw their stake if they have submitted a hash this round", async () => {
-      await giveUserCLNYTokensAndStake(colonyNetwork, MAIN_ACCOUNT, DEFAULT_STAKE);
+    it("should not allow someone to submit a new reputation hash if they stake after the cycle beings", async () => {
```
elenadimitrova: typo -> beings

test/colony-network-mining.js
kronosapiens: Hmm, looking at this test more closely we may be able to do away with this setup entirely, as the modifier which checks if the submission window is open comes before the one which checks the stake and timestamp.

test/colony-network-mining.js
kronosapiens: Tk u 👍

test/colony-network-mining.js
kronosapiens: I added a comment, basically the conditional is `windowOpen >= stakeTimestamp` (done so the staking in migration 8 works), which made the tests fail randomly since sometimes the cycle was advanced in the same block as the staking occurred and so the timestamps were equal. This ensures that the staking happens strictly after the window opens.

test/colony-network-mining.js
kronosapiens: Basically anytime there is a `giveUserCLNYTokensAndStake` in the beginning of a test I added a `advanceMiningCycleNoContest` to "activate" that account's stake.

PR #467, opened 2018-12-18
Upgrade contracts to solc 0.5.0 syntax

### LINKED ISSUE #459: Upgrade contracts to solc 0.5.0
Now dappHub library and truffle are updated too there's nothing stopping us from moving to the next major solc version.

elenadimitrova:
Closes #459 

Implements fixes for the [solc 0.5.0 breaking changes](https://solidity.readthedocs.io/en/latest/050-breaking-changes.html). The only exception not yet implemented is the `address payable` modifier and this is due [to `solidity-coverage`](https://github.com/sc-forks/solidity-coverage/issues/258) and additionally [`solium`] (https://github.com/duaraghav8/Ethlint/issues/246) not yet supporting it. 

Since the CLNY token and MultiSig contracts were compiled and deployed with earlier version of `solc` here we start importing the already compiled contracts from `colonyToken` repo, as opposed to recompiling them from source internally here.


contracts/Colony.sol
```diff
@@ -24,6 +24,8 @@ import "./EtherRouter.sol";
 
 contract Colony is ColonyStorage, PatriciaTreeProofs {
 
+  function() external payable {}
```
kronosapiens: Can we add a comment explaining what this function is for?

test/colony-network-mining.js
```diff
@@ -204,7 +204,7 @@ contract("ColonyNetworkMining", accounts => {
       await giveUserCLNYTokens(colonyNetwork, OTHER_ACCOUNT, 9000);
       await clny.approve(tokenLocking.address, 10000, { from: OTHER_ACCOUNT });
 
-      await checkErrorRevert(tokenLocking.deposit(clny.address, 10000, { from: OTHER_ACCOUNT }), "ds-token-insufficient-balance");
+      await checkErrorRevert(tokenLocking.deposit(clny.address, 10000, { from: OTHER_ACCOUNT }));
```
kronosapiens: Oh well :/

contracts/Colony.sol
```diff
@@ -15,7 +15,7 @@
   along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
 */
 
-pragma solidity >0.5.0;
+pragma solidity >=0.4.23;
```
kronosapiens: Why, if we are committed to Sol 5 moving forward?

kronosapiens:
How come in the last few commits it looks like some of the changes around `address payable` and the version pragma were rolled back?

contracts/ReputationMiningCycleStorage.sol
```diff
@@ -15,8 +15,7 @@
   along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
 */
 
-pragma solidity ^0.4.23;
-pragma experimental "v0.5.0";
+pragma solidity >=0.4.23 <0.5.0;
```
kronosapiens: Should this also omit the `<0.5.0` ?

contracts/Colony.sol
```diff
@@ -275,4 +276,6 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit DomainAdded(domainCount);
     emit PotAdded(potCount);
   }
+
+  function() external payable {}
```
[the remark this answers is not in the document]
elenadimitrova: You are on an outdated commit @kronosapiens . This was removed a day before I finished the PR. Not sure how you got it?

contracts/ColonyNetworkMining.sol
```diff
@@ -85,8 +85,9 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     address clnyToken = IMetaColony(metaColony).getToken();
     require(clnyToken != address(0x0), "colony-reputation-mining-clny-token-invalid-address");
 
-    inactiveReputationMiningCycle = new EtherRouter();
-    EtherRouter(inactiveReputationMiningCycle).setResolver(miningCycleResolver);
+    EtherRouter e = new EtherRouter();
```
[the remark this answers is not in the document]
elenadimitrova: Because implicit conversion from a contract to address type is not supported in solc 0.5.0. Additionally the cast from non-payable address to a payable contract type (such as EtherRouter) is also not supported. Additionally the new single line replacement is much cleaner in its purpose.

contracts/ContractRecovery.sol
```diff
@@ -72,7 +72,7 @@ contract ContractRecovery is CommonStorage {
   function exitRecoveryMode() public recovery auth {
     uint totalAuthorized = recoveryRolesCount;
     // Don't double count the owner (if set);
-    if (owner != address(0x0) && !CommonAuthority(authority).hasUserRole(owner, RECOVERY_ROLE)) { 
+    if (owner != address(0x0) && !CommonAuthority(address(authority)).hasUserRole(owner, RECOVERY_ROLE)) {
```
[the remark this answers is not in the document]
elenadimitrova: If you check the actual declarations of these vars, they are:
```
DSAuthority public authority;
address public owner;
```

contracts/Colony.sol
elenadimitrova: This is no longer there.

elenadimitrova:
>How come in the last few commits it looks like some of the changes around address payable and the version pragma were rolled back?

This is because due to `solidity-coverage` not yet supporting solc 0.5.0 we had to stay on 0.4 compiler but still with an upgraded syntax to support the incoming solc 0.5.0. With the exception of `address payable` with is only supported from 0.5.0 onwards and so cannot be used yet here.

contracts/Colony.sol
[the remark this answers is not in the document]
kronosapiens: I was trying out a commit-by-commit review this time, to better see how the PR evolved over time. But it does make commenting a bit trickier.

contracts/ColonyNetworkMining.sol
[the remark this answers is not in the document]
kronosapiens: 👍

contracts/ContractRecovery.sol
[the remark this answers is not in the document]
kronosapiens: 👍

contracts/Colony.sol
kronosapiens: 👍

contracts/Colony.sol
kronosapiens: 👍

contracts/ReputationMiningCycleStorage.sol
kronosapiens: 👍

contracts/Colony.sol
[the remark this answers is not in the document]
elenadimitrova: Yeah I had to go back on part of the implementation. Normally I would just drop or amend commits but the change was tricker to extract this time. I'll try to keep a cleaner commit history going forward.

PR #469, opened 2018-12-18
Add MAX_PAYOUT limit to setTaskPayout

### LINKED ISSUE #394: Potential overflow in reputation calculation on task payout
When calculating reputation earned on a task in `ColonyTask.getReputation`, there is a potential overflow as that uses a multiplier of the task payout which is a `uint256` and the rating multipliers which can go up to `1.5` meaning the following calculation there will throw
```
reputation = SafeMath.mulInt(payout, ratingMultipliers[rating - 1]);
```
https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyTask.sol#L484

kronosapiens:
Closes #394 

Impose a `MAX_PAYOUT` limit of `UINT254_MAX` on task payments. This is 1/4th of the `UINT256_MAX` range (since we omit two bits). We do this to ensure we can calculate the reputation update (which involves scaling and adding to the payment) without overflowing.


contracts/ColonyTask.sol
```diff
@@ -453,25 +453,27 @@ contract ColonyTask is ColonyStorage {
...
-    int8[3] memory ratingMultipliers = [-2, 2, 3];
-    int8 ratingDivisor = 2;
+    bool negative = (rating == TaskRatings.Unsatisfactory);
+    uint256 reputation = mul(payout, (rating == TaskRatings.Excellent) ? 3 : 2);
 
-    reputation = SafeMath.mulInt(payout, ratingMultipliers[rating - 1]);
-    reputation = SafeMath.subInt(reputation, rateFail ? payout : 0); // Deduct penalty for not rating
-    reputation /= ratingDivisor; // We may lose one atom of reputation here :sad:
+    if (rateFail) {
+      reputation = negative ? add(reputation, payout) : sub(reputation, payout);
```
elenadimitrova: That's something fishy here.. I'm going 🎣

elenadimitrova:
Just to confirm with @area and @kronosapiens the current behaviour.

Say we have even payouts of 100 tokens per role. Considering the ratings below, confirm the expected reputation amount is correct:

| Task role  | TaskRatings.Unsatisfactory  | TaskRatings.Satisfactory | TaskRatings.Excellent     | Failed to rate |
| --------- | --------------------------- |------------------------ |------------------------ |------------- |
| Manager  | -100                                              | +100                                     | +150                                      | NA                    |
| Evaluator | NA (not rated explicitly)          | +100 (Implicit rating)      |  NA (not rated explicitly)  |  -150               |
| Worker(domain rep)| -100                           | +100                                     | +150                                      | +50 (assuming rated a 2) by evaluator  |
| Worker (skill rep)    | -100                             | +100                                     | +150                                      | +50               |

contracts/ColonyTask.sol
kronosapiens: I know it's not the clearest but the idea was to work with `uint`s to take advantage of the larger range, and then scale it down and add the sign at the very end.

PR #477, opened 2018-12-28
Remove the ability to change token on Colony once initialised

### LINKED ISSUE #474: Remove the ability to change the Colony Token
Currently the `IColony.setToken` function allows colony founders to switch the native token of their colony. Following discussions in dev and [product](https://joincolony.slack.com/archives/C0YNG3M8B/p1544207179079000) we think we should remove that capability.

elenadimitrova:
Closes #474

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -54,7 +54,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return ColonyAuthority(address(authority)).hasUserRole(_user, _role);
   }
 
-  function getColonyNetworkAddress() public view returns (address) {
+  function getColonyNetwork() public view returns (address) {
     return colonyNetworkAddress;
   }
 
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -58,22 +58,15 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return colonyNetworkAddress;
   }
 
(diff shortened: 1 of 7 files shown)


contracts/Colony.sol
```diff
@@ -54,26 +54,19 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
-
   function getToken() public view returns (address) {
     return address(token);
   }
 
-  function initialiseColony(address _colonyNetworkAddress) public stoppable {
-    require(colonyNetworkAddress == address(0x0), "colony-initialise-bad-address");
+  function initialiseColony(address _colonyNetworkAddress, address _token) public stoppable {
+    require(colonyNetworkAddress == address(0x0), "colony-already-initialised-network");
+    require(address(token) == address(0x0), "colony-already-initialised-token");
```
kronosapiens: Q: why cast `address(token)` if `token` is already an `address` type?

elenadimitrova: A: But it's not an address type: https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyStorage.sol#L35

kronosapiens: Hmm. Mixing address and contract types for variables seems like a recipe for confusion. Is there a reason we don't store everything as `address` types and cast them to their specific contract types when we need to use them? Some sort of convention around this seems appropriate.

PR #479, opened 2018-12-29
Integrate OpenZeppelin token

### LINKED ISSUE #405: Impossible to use OpenZeppelin v1.10 ERC20 Mintable token for Colony
# Prerequisites
- [X] I checked to make sure that this issue has not already been filed

## Summary
ColonyNetwork is incompatible with OpenZeppelin v1.10 Mintable ERC20 token due to specifiaction and implementation difference of ownership and minting logic. Specifically, Colony method `mintTokens` is calling Token `mint` method according to dappsys mintable token specification with amount of tokens to mint passed as an argunment. OpenZeppelin mintable token specification `mint` method is receiving address and amount as arguments.

Here are the links for specifications:
https://github.com/dapphub/dappsys-monolithic/blob/master/token.sol#L71
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/v1.10.0/contracts/token/ERC20/MintableToken.sol#L36

I think OpenZeppelin smart contracts library is widely used and actively developed, and should not be ignored by Colony Network.

elenadimitrova:
Closes #405 

Here we add overloads for `mint` and `burn` to the `ERC20Extended` token interface. These functions are already implemented in the DappHub token but they are also necessary for the OpenZeppelin `ERC20Mintable`  token.
Additionally we remove the `ERC20ExtendedToken` implementation and simply use the vanilla `DSToken` in tests and `ERC20Mintable` (open zeppelin token) in integration tests. 

We update the `Colony.mintTokens` implementation to use the extended `mint` overload.


contracts/ERC20ExtendedToken.sol
```diff
@@ -34,21 +34,30 @@ contract ERC20ExtendedToken is DSTokenBase(0), DSAuth, ERC20Extended {
...
+  function mint(address guy, uint wad) public auth {
+    _balances[guy] = add(_balances[guy], wad);
+    _supply = add(_supply, wad);
+    emit Mint(guy, wad);
+    emit Transfer(address(0x0), guy, wad);
   }
-}
 
+  function burn(address guy, uint wad) public {
+    if (guy != msg.sender && _approvals[guy][msg.sender] != uint(-1)) {
```
kronosapiens: What is the purpose of the second condition here? `_approvals[guy][msg.sender] != uint(-1)`?

test/colony-token-integrations.js
```diff
@@ -0,0 +1,81 @@
...
+      expect(colonyPotBalance).to.be.eq.BN(99);
+      expect(colonyTokenBalance).to.be.eq.BN(100);
+    });
+  });
+
+  describe("when working with openzeppelin-solidity/ERC20Mintable token owned by the colony", () => {
+    beforeEach(async () => {
+      await erc20Mintable.addMinter(colony.address);
+      await erc20Mintable.renounceMinter();
+      // At the point the only permitted
```
kronosapiens: Incomplete comment?

test/colony-token-integrations.js
```diff
@@ -0,0 +1,81 @@
...
+      expect(colonyRewardPotBalance).to.be.eq.BN(1);
+      expect(colonyPotBalance).to.be.eq.BN(99);
+      expect(colonyTokenBalance).to.be.eq.BN(100);
+    });
+  });
+
+  describe("when working with openzeppelin-solidity/ERC20Mintable token owned by the colony", () => {
+    beforeEach(async () => {
+      await erc20Mintable.addMinter(colony.address);
+      await erc20Mintable.renounceMinter();
```
kronosapiens: Looks like these two lines are a no-op?

contracts/ERC20ExtendedToken.sol
kronosapiens: `uint(-1)` is equivalent to `UINT256_MAX` unless I'm mistaken, so we don't subtract the approval if they are maximum?

test/colony-token-integrations.js
elenadimitrova: Why? https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/access/roles/MinterRole.sol#L30

contracts/ERC20ExtendedToken.sol
elenadimitrova: This is related to the so-called binary approval in dappsys token implementation where you approve an address to manage _all_ your tokens via approving them to the max limit, see https://github.com/dapphub/ds-token/blob/master/src/token.sol#L37
So this line is saying that if you have such approval, you cannot use the `burn` function.

test/colony-token-integrations.js
kronosapiens: Ah, it's adding the colony.address and renouncing the coinbase account. Nevermind.

kronosapiens: It's one shade more implicit that I'd like but not sure a simple alternative so #yolo.

contracts/ERC20ExtendedToken.sol
kronosapiens: Mm. That's definitely an undocumented dependency between these contracts then... can we introduce something like:

```
uint256 constant BINARY_APPROVAL = uint256(-1);
```

Which makes the semantics more clear?

test/colony-token-integrations.js
kronosapiens: Ok fixed

contracts/ERC20ExtendedToken.sol
kronosapiens: This is my last outstanding comment ^^

elenadimitrova: I've completely removed our custom ERC20 implementation in the face of `ERC20ExtendedToken` and now using vanilla `DSToken` instead. Colony creators should be able to make their own informed choice as to which token contract to use, there's plenty on the market including `DSToken` which the DAI is based on and OpenZeppelin's `ERC20Mintable` token.

elenadimitrova:
Even though neither @area nor @kronosapiens have yet pushed the green button here, I am merging because I believe all concerns they've both raised are now solved.

PR #480, opened 2018-12-29
Optimise contract storage layout

### LINKED ISSUE #460: Optimise storage layout in contracts
Considering our recovery mode and in general the `EtherRouter` inner workings rely extensively on the storage layout, we should optimise that to make our dev / dev ops life easier. e.g. `reputationRootHash` property can be moved to more towards the top of the NetworkStorage declarations instead of it being behind 4 mappings down at storage slot 20.

elenadimitrova:
Closes #460

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -31,27 +31,27 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   function setFounderRole(address _user) public stoppable auth {
     // To allow only one address to have founder role at a time, we have to remove current founder from their role
     ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
-    colonyAuthority.setUserRole(msg.sender, FOUNDER_ROLE, false);
-    colonyAuthority.setUserRole(_user, FOUNDER_ROLE, true);
+    colonyAuthority.setUserRole(msg.sender, uint8(ColonyRole.Founder), false);
+    colonyAuthority.setUserRole(_user, uint8(ColonyRole.Founder), true);
 
     emit ColonyFounderRoleSet(msg.sender, _user);
   }
 
   function setAdminRole(address _user) public stoppable auth {
-    ColonyAuthority(address(authority)).setUserRole(_user, ADMIN_ROLE, true);
+    ColonyAuthority(address(authority)).setUserRole(_user, uint8(ColonyRole.Admin), true);
 
     emit ColonyAdminRoleSet(_user);
   }
 
   // Can only be called by the founder role.
   function removeAdminRole(address _user) public stoppable auth {
-    ColonyAuthority(address(authority)).setUserRole(_user, ADMIN_ROLE, false);
+    ColonyAuthority(address(authority)).setUserRole(_user, uint8(ColonyRole.Admin), false);
 
     emit ColonyAdminRoleRemoved(_user);
   }
 
-  function hasUserRole(address _user, uint8 _role) public view returns (bool) {
-    return ColonyAuthority(address(authority)).hasUserRole(_user, _role);
+  function hasUserRole(address _user, ColonyRole _role) public view returns (bool) {
+    return ColonyAuthority(address(authority)).hasUserRole(_user, uint8(_role));
   }
 
   function getColonyNetwork() public view returns (address) {
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -69,17 +69,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     token = ERC20Extended(_token);
 
     // Initialise the task update reviewers
-    setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), MANAGER, WORKER);
-    setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), MANAGER, WORKER);
-    setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), MANAGER, WORKER);
-    setFunctionReviewers(bytes4(keccak256("setTaskDomain(uint256,uint256)")), MANAGER, WORKER);
+    setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), TaskRole.Manager, TaskRole.Worker);
+    setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), TaskRole.Manager, TaskRole.Worker);
+    setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), TaskRole.Manager, TaskRole.Worker);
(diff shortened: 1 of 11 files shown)


test/colony-network-recovery.js
```diff
@@ -390,11 +390,10 @@ contract("Colony Network Recovery", accounts => {
             await newActiveCycleAsRecovery.setStorageSlot(logEntryStartingSlot.addn(5), `0x${new BN(logEntry.nPreviousUpdates).toString(16, 64)}`);
 
             const portedLogEntry = await newActiveCycle.getReputationUpdateLogEntry(i);
-
-            assert.strictEqual(portedLogEntry.user, logEntry.user);
+            assert.equal(portedLogEntry.user, logEntry.user);
             assert.strictEqual(portedLogEntry.amount, logEntry.amount);
             assert.strictEqual(portedLogEntry.skillId, logEntry.skillId);
-            assert.strictEqual(portedLogEntry.colony, logEntry.colony);
+            assert.equal(portedLogEntry.colony, logEntry.colony);
```
kronosapiens: Why the relaxation?

contracts/ColonyStorage.sol
```diff
@@ -31,59 +31,59 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
   // this one will have the getters. Make custom getters in the contract that seems most appropriate,
   // and add it to IColony.sol
 
-  address colonyNetworkAddress;
-  ERC20Extended token;
-  uint256 rewardInverse;
+  // Colony-wide roles
+  uint8 constant FOUNDER_ROLE = 0;
```
kronosapiens: We c o u l d be using enums here instead... #semantecs

But I'll survive :)

contracts/ColonyNetworkStorage.sol
```diff
@@ -26,59 +26,62 @@ import "./IMetaColony.sol";
 
 contract ColonyNetworkStorage is CommonStorage, ColonyNetworkDataTypes, DSMath {
   // Number of colonies in the network
-  uint256 colonyCount;
+  uint256 colonyCount; // Storage slot 6
```
kronosapiens: I almost want to say that if it's a simple as counting down, these comments are redundant (after the first), but I also see the value of being thorough (although increases the chance of them being out of date if we move things around in the future).

contracts/ColonyNetworkStorage.sol
```diff
@@ -26,59 +26,62 @@ import "./IMetaColony.sol";
...
+
+  // Contains the address of the resolver for ReputationMiningCycle
+  address miningCycleResolver; // Storage slot 15
+  // Address of the currently active reputation mining cycle contract
+  address activeReputationMiningCycle; // Storage slot 16
+  // Address of the next active reputation mining cycle contract, which is where new reputation updates are put.
+  address inactiveReputationMiningCycle; // Storage slot 17
+
+  // Maps index to colony address
+  mapping (uint256 => address) colonies; // Storage slot 18
```
kronosapiens: Interesting, I vaguely recall learning that for mappings and arrays, the storage slot is elsewhere? Some function of the hash of the mapping and key names?

contracts/ColonyNetworkStorage.sol
elenadimitrova: I imagine sitting in the midst of a crisis (aka recovery) mode trying to count from 1 to 20..

contracts/ColonyNetworkStorage.sol
area: The hash is based on the key and the storage slot that is 'used' (see [the docs](https://solidity.readthedocs.io/en/v0.4.24/miscellaneous.html?highlight=slot#layout-of-state-variables-in-storage)).

contracts/CommonStorage.sol
```diff
@@ -28,14 +28,14 @@ contract CommonStorage is DSAuth {
...
 
   // Recovery variables
-  bool recoveryMode;
+  bool recoveryMode; // Storage slot 3
   uint64 recoveryRolesCount;
   uint64 recoveryApprovalCount;
-  uint256 recoveryEditedTimestamp;
-  mapping (address => uint256) recoveryApprovalTimestamps;
+  uint256 recoveryEditedTimestamp; // Storage slot 4
+  mapping (address => uint256) recoveryApprovalTimestamps; // Size in storage slot 5
```
area: RE the comment, mappings don't use the slot to store the length (because they don't really have one), only arrays do.

contracts/ColonyStorage.sol
```diff
@@ -31,59 +31,59 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+  mapping (address => bool) activeRewardPayouts; // Storage slot 17
 
   // This keeps track of how much of the colony's funds that it owns have been moved into pots other than pot 0,
   // which (by definition) have also had the reward amount siphoned off and put in to pot 0.
   // This is decremented whenever a payout occurs and the colony loses control of the funds.
-  mapping (address => uint256) nonRewardPotsTotal;
+  mapping (address => uint256) nonRewardPotsTotal; // Storage slot 18
 
-  mapping (uint256 => RatingSecrets) public taskWorkRatings;
+  mapping (uint256 => RatingSecrets) public taskWorkRatings; // Storage slot 18
```
area: Should be 19, with all subsequent ones changed too.

contracts/CommonStorage.sol
elenadimitrova: Yes, forgotten to delete this.

contracts/ColonyStorage.sol
elenadimitrova: Tbh I need to give `enum`s another chance. It's a brand new year after all.. I've been biased against them for a while and I can no longer remember why.

elenadimitrova: There we go, enums all around. See how you like that @kronosapiens

contracts/Colony.sol
```diff
@@ -31,27 +31,27 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
-    colonyAuthority.setUserRole(_user, FOUNDER_ROLE, true);
+    colonyAuthority.setUserRole(msg.sender, uint8(ColonyRole.Founder), false);
+    colonyAuthority.setUserRole(_user, uint8(ColonyRole.Founder), true);
 
     emit ColonyFounderRoleSet(msg.sender, _user);
   }
 
   function setAdminRole(address _user) public stoppable auth {
-    ColonyAuthority(address(authority)).setUserRole(_user, ADMIN_ROLE, true);
+    ColonyAuthority(address(authority)).setUserRole(_user, uint8(ColonyRole.Admin), true);
```
kronosapiens: Can these functions take `ColonyRole` enums directly? Do we have to cast?

elenadimitrova: Yes unfortunately `enum`s don't convert implicitly

kronosapiens: Right but can `setUserRole` be updated to accept the `ColonyRole` type?

elenadimitrova: `setUserRole` is part of `DSRoles` which works generically with `uint8`s and so we can't change that.

PR #481, opened 2019-01-02
Clarify ownership of Network contracts

### LINKED ISSUE #260: Agree ownership of upgradable contracts we are responsible for
We have to agree the ownership of the following contracts and their resolvers:

MetaColony
ColonyNetwork
TokenLocking
(ReputationMiningCycle)

elenadimitrova:
Relevant to #260 

For  `ReputationMiningCycle` contracts the owner is the ColonyNetwork and due to the short lifecycle (24h) of these, upgrades will not be performed but instead a seamless upgrade will happen once an old cycle completes, new reputation root hash is set and a new mining contract instantiated. 
Tests were added for demonstrating these seamless upgrades. 
The above changes uncovered issues when passing `mocha` test context to `forwardTime` function which entailed changes to stop using arrow functions in tests as [explained here](https://mochajs.org/#arrow-functions). Note that this is relevant to the entire test suite and perhaps need to be refactored everywhere potentially as part of #319 when we attempt running tests under `parity`. 

Additionally here we add a test for confirming appending struct members will be supported during upgrades.


scripts/generate-test-contracts.sh
```diff
@@ -9,12 +9,34 @@ cp ./contracts/Resolver.sol ./contracts/UpdatedResolver.sol
...
+sed -i.bak "s/ColonyStorage/UpdatedColonyStorage/g" ./contracts/UpdatedColonyStorage.sol
+sed -i.bak "s/ColonyDataTypes/UpdatedColonyDataTypes/g" ./contracts/UpdatedColonyStorage.sol
+# Modify IUpdatedColony contract
 sed -i.bak "s/contract IColony/contract IUpdatedColony/g" ./contracts/IUpdatedColony.sol
-sed -i.bak "s/contract IUpdatedColony is ColonyDataTypes, IRecovery {/contract IUpdatedColony is ColonyDataTypes, IRecovery {function isUpdated() public pure returns(bool);/g" ./contracts/IUpdatedColony.sol
+sed -i.bak "s/ColonyDataTypes/UpdatedColonyDataTypes/g" ./contracts/IUpdatedColony.sol
+sed -i.bak "s/contract IUpdatedColony is UpdatedColonyDataTypes, IRecovery {/contract IUpdatedColony is UpdatedColonyDataTypes, IRecovery {function isUpdated() public pure returns(bool);/g" ./contracts/IUpdatedColony.sol
+# Modify UpdatedReputationMiningCycle contract
+sed -i.bak "s/contract ReputationMiningCycle/contract UpdatedReputationMiningCycle/g" ./contracts/UpdatedReputationMiningCycle.sol
+sed -i.bak "s/WAD;/WAD;function isUpdated() public pure returns(uint256) {return 100;}/g" ./contracts/UpdatedReputationMiningCycle.sol
```
area: Why does this `isUpdated` return `uint256` and the others are booleans? The test still seems to work, which is a little offputting...

test-upgrade/reputation-mining-cycle-upgrade.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  describe("when upgrading ReputationMiningCycle contract", function() {
+    it("should change the ReputationMiningCycle resolver on the Network", async function() {
+      const reputationMiningResolverAfter = await colonyNetwork.getMiningResolver();
+      assert.notEqual(reputationMiningResolverBefore, reputationMiningResolverAfter);
+    });
+
+    it("should pick up the new ReputationMiningCycle contract", async function() {
+      // Cycle through the mining process twice to get the newly created inactive cycles on the updated mining contract
```
area: Once?

elenadimitrova: Indeed. I started off with going through it twice but there's no point.

scripts/generate-test-contracts.sh
elenadimitrova: Yes, indeed. The twist is that the upgrade contracts here start using `forwardTime` meaning the test checking this particular function is skipped.. I will add 2 test runs to solve this, one as it is currently, running under `parity` and another under `ganache-cli` which will include the run of this verification check.

Issue #468, opened 2018-12-18
Fix brittle reputation mining test


elenadimitrova:
The test `should cope with many hashes being submitted and eliminated before a winner is assigned` has been frequently failing in CI both as part of the [main test suite run](https://circleci.com/gh/JoinColony/colonyNetwork/5382) as well as the [coverage testing](https://circleci.com/gh/JoinColony/colonyNetwork/5404?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

It's slowing down our development pipeline and should be investigated and resolved

area:
Closed by #482

PR #482, opened 2019-01-02
(Maybe) fix flickery test

area:
I was surprised when my changes in #464 to this test passed first time when I did them, and thought I was just having a good day down in the coding mines. It turns out, when CLNY funding of addresses was changed to not use tasks in #453, this test silently broke and was (mostly) passing without issue.

What was happening was all clients were submitting the same hash, because they were told to get updates 7-15 wrong, but there were now only four updates in the log. Then, what I *think* was happening was the `confirmNewHash` call was being called before all the transactions generated by `submitAndForwardTimeToDispute` were mined. This could happen because the promises generated there resolve once the transaction is submitted, not mined. If that happened, then the first `submitRootHash` call afterwards would fail.

This antipattern is a result of us (mostly me) overlooking the subtle differences between truffle and ethers, so there's probably a lot more floating around out there 😟 

So #468 *might* be fixed by this. But equally, that test was so far from where it should have been, it might not, and might just be papering over a crack I've not spotted.

### DIFF
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -419,7 +419,8 @@ export async function submitAndForwardTimeToDispute(clients, test) {
   await forwardTime(MINING_CYCLE_DURATION / 2, test);
   for (let i = 0; i < clients.length; i += 1) {
     await clients[i].addLogContentsToReputationTree(); // eslint-disable-line no-await-in-loop
-    await clients[i].submitRootHash(); // eslint-disable-line no-await-in-loop
+    const tx = await clients[i].submitRootHash(); // eslint-disable-line no-await-in-loop
+    await tx.wait(); // eslint-disable-line no-await-in-loop
   }
   await forwardTime(MINING_CYCLE_DURATION / 2, test);
 }
--- a/test/colony-network-mining.js
+++ b/test/colony-network-mining.js
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
         // These have to be done sequentially because this function uses the total number of tasks as a proxy for getting the
         // right taskId, so if they're all created at once it messes up.
       }
+      await fundColonyWithTokens(metaColony, clny, INITIAL_FUNDING.muln(30));
 
-      // We need to complete the current reputation cycle so that all the required log entries are present
       await advanceMiningCycleNoContest({ colonyNetwork, test: this });
+
+      for (let i = 0; i < 10; i += 1) {
+        await setupFinalizedTask( // eslint-disable-line
+          {
+            colonyNetwork,
+            colony: metaColony,
+            token: clny,
+            workerRating: 2,
+            managerPayout: 10000000,
+            evaluatorPayout: 10000000,
+            workerPayout: 10000000
+          }
+        );
+      }
 
       const clients = await Promise.all(
         accounts.slice(3, 11).map(async (addr, index) => {
           const client = new MaliciousReputationMinerExtraRep(
             { loader: contractLoader, minerAddress: addr, realProviderPort: REAL_PROVIDER_PORT, useJsTree },
-            accounts.length - index,
+            accounts.length - index - 5,
             index
           );
           // Each client will get a different reputation update entry wrong by a different amount, apart from the first one which
@@ -2462,8 +2476,18 @@ contract("ColonyNetworkMining", accounts => {
         })
       );
 
-      const repCycle = await getActiveRepCycle(colonyNetwork);
+      // We need to complete the current reputation cycle so that all the required log entries are present
+      await advanceMiningCycleNoContest({ colonyNetwork, test: this, client: clients[0] });
 
+      await clients[0].saveCurrentState();
+      const savedHash = await clients[0].reputationTree.getRootHash();
+      await Promise.all(
+        clients.map(async client => {
+          client.loadState(savedHash);
+        })
+      );
+
+      const repCycle = await getActiveRepCycle(colonyNetwork);
       await submitAndForwardTimeToDispute(clients, this);
 
       const nSubmittedHashes = await repCycle.getNSubmittedHashes();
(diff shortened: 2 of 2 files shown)


test/colony-network-mining.js
```diff
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
         // These have to be done sequentially because this function uses the total number of tasks as a proxy for getting the
         // right taskId, so if they're all created at once it messes up.
       }
+      await fundColonyWithTokens(metaColony, clny, INITIAL_FUNDING.muln(30));
 
-      // We need to complete the current reputation cycle so that all the required log entries are present
       await advanceMiningCycleNoContest({ colonyNetwork, test: this });
 
+      for (let i = 0; i < 10; i += 1) {
```
kronosapiens: Unless the explicit task parameters are being used (doesn't look like they are?), you could replace the following line with:

```
await setupFinalizedTask({ colonyNetwork, colony: metaColony });
```

test/colony-network-mining.js
```diff
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
...
+          }
+        );
+      }
+
       const clients = await Promise.all(
         accounts.slice(3, 11).map(async (addr, index) => {
           const client = new MaliciousReputationMinerExtraRep(
             { loader: contractLoader, minerAddress: addr, realProviderPort: REAL_PROVIDER_PORT, useJsTree },
-            accounts.length - index,
+            accounts.length - index - 5,
```
kronosapiens: Can you add a parenthesis here to indicate the intended order of operations?

test/colony-network-mining.js
```diff
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
...
+            managerPayout: 10000000,
+            evaluatorPayout: 10000000,
+            workerPayout: 10000000
+          }
+        );
+      }
+
       const clients = await Promise.all(
         accounts.slice(3, 11).map(async (addr, index) => {
           const client = new MaliciousReputationMinerExtraRep(
```
elenadimitrova: Are all clients here "bad"?

test/colony-network-mining.js
```diff
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
...
+            colonyNetwork,
+            colony: metaColony,
+            token: clny,
+            workerRating: 2,
+            managerPayout: 10000000,
+            evaluatorPayout: 10000000,
+            workerPayout: 10000000
+          }
+        );
+      }
```
elenadimitrova: This creates 87 updates in the log. Is it completely unnecessary if we're only using bad entries at index 0 to 7

test/colony-network-mining.js
elenadimitrova: For me this is the actual line that fixes the test. We should extract that variable to be clear but it is basically the `entryToFalsify` in the bad client.

test/colony-network-mining.js
```diff
@@ -2444,15 +2444,29 @@ contract("ColonyNetworkMining", accounts => {
...
+
       const clients = await Promise.all(
         accounts.slice(3, 11).map(async (addr, index) => {
           const client = new MaliciousReputationMinerExtraRep(
             { loader: contractLoader, minerAddress: addr, realProviderPort: REAL_PROVIDER_PORT, useJsTree },
-            accounts.length - index,
+            accounts.length - index - 5,
             index
```
elenadimitrova: Can we extract this to be clear that `const incrementAmountWith = index`

test/colony-network-mining.js
elenadimitrova: Just noticed that even through the first client _is_ instantiated as bad, the amount to falsify the entry with is `0` making it in fact behaving like a good client. Perhaps add a note as it's non-obvious?

helpers/test-helper.js
```diff
@@ -419,7 +419,8 @@ export async function submitAndForwardTimeToDispute(clients, test) {
   await forwardTime(MINING_CYCLE_DURATION / 2, test);
   for (let i = 0; i < clients.length; i += 1) {
     await clients[i].addLogContentsToReputationTree(); // eslint-disable-line no-await-in-loop
-    await clients[i].submitRootHash(); // eslint-disable-line no-await-in-loop
+    const tx = await clients[i].submitRootHash(); // eslint-disable-line no-await-in-loop
+    await tx.wait(); // eslint-disable-line no-await-in-loop
```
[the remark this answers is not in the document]
elenadimitrova: I agree that potentially this fixes #468

test/colony-network-mining.js
```diff
@@ -2484,15 +2508,15 @@ contract("ColonyNetworkMining", accounts => {
...
+          if (client2idx >= clients.length - 4) {
             error = "colony-reputation-mining-decay-incorrect";
           } else {
             error = "colony-reputation-mining-invalid-newest-reputation-proof";
           }
           // eslint-disable-next-line no-await-in-loop
           await accommodateChallengeAndInvalidateHash(colonyNetwork, this, clients[i], clients[client2idx], {
             client2: { respondToChallenge: error }
-          }); // eslint-disable-line no-await-in-loop
+          });
```
elenadimitrova: Don't we have a helper function for this sort of convoluted logic?

test/colony-network-mining.js
area: By chance, this task creation is actually not needed at all.

test/colony-network-mining.js
```diff
@@ -2462,46 +2464,48 @@ contract("ColonyNetworkMining", accounts => {
...
+        client2: { respondToChallenge: "colony-reputation-mining-invalid-newest-reputation-proof" }
+      });
+      await accommodateChallengeAndInvalidateHash(colonyNetwork, this, clients[4], clients[6], {
+        client2: { respondToChallenge: "colony-reputation-mining-decay-incorrect" }
+      });
+
+      // Round 3
+      await accommodateChallengeAndInvalidateHash(colonyNetwork, this, clients[0], clients[4], {
+        client2: { respondToChallenge: "colony-reputation-mining-decay-incorrect" }
+      });
```
elenadimitrova: Lovely

PR #486, opened 2019-01-04
Update "Get Started" and reputation miner

ryanchristo:
## Description

This pull request updates the "Get Started" instructions in the colonyNetwork documentation and the miner durations. It also requires babel in the reputation miner script so that the miner can be run with node using es6 syntax.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,5 @@ lib/*
 ganache-accounts.json
 .vscode/
 yarn-error.log
+reputations.json
 reputationStates.sqlite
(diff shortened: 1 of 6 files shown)


elenadimitrova:
@ryanchristo this is broken but you've requested review?

ryanchristo:
@elenadimitrova That is correct. I should have pinged @area instead. The only changes I made were to the docs, so the failing test is a result of [cf558a8](https://github.com/JoinColony/colonyNetwork/pull/486/commits/cf558a83ba1b7cd0dc7473e3512602be816f89e3). I can take a look as well.

PR #490, opened 2019-01-07
Do not append zero-valued reputation updates to the log

### LINKED ISSUE #354: 0 value reputation update entries may be obsolete
In the reputation update log can we _not_ log `0` value updates?

kronosapiens:
Closes #354 

Implemented at the `ColonyNetwork` level rather than the `ColonyTask` level for generality.

### DIFF
--- a/contracts/ColonyNetwork.sol
+++ b/contracts/ColonyNetwork.sol
@@ -177,7 +177,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
   calledByMetaColony
   {
     require(currentColonyVersion > 0, "colony-network-not-intialised-cannot-add-colony-version");
-    
+
     colonyVersionResolver[_version] = _resolver;
     if (_version > currentColonyVersion) {
       currentColonyVersion = _version;
@@ -186,8 +186,8 @@ contract ColonyNetwork is ColonyNetworkStorage {
     emit ColonyVersionAdded(_version, _resolver);
   }
 
-  function initialise(address _resolver) public 
-  auth 
+  function initialise(address _resolver) public
+  auth
   stoppable
   {
     require(currentColonyVersion == 0, "colony-network-already-initialised");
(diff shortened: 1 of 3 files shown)


contracts/ColonyNetwork.sol
```diff
@@ -267,6 +267,10 @@ contract ColonyNetwork is ColonyNetworkStorage {
   calledByColony
   skillExists(_skillId)
   {
+    if (_amount == 0) {
+      return;
+    }
```
elenadimitrova: That helps with consistency however I think we're still paying a rather large gas price for the external contract call to here. Which was one of the main benefits of this work. Could you at least test this implementation in comparison with logic sitting in `ColonyTask` contract for `finalizeTask` call? That should help us outweigh cost vs complexity.

kronosapiens: Sure!

elenadimitrova: We've merged a fix for the gas reporter so if you rebase you should have the gasCost tests reporting all calls properly now.

elenadimitrova: Okay, let's proceed with keeping the logic in the Network contract

area: > Noting that putting the skip in ColonyFunding means that other sources of reputation updates (for example, miner rewards) currently will not be skipped if zero-valued.

As-is, mining rewards are not skipped, even if they're zero. However, I think that's a good thing, as I have definitely assumed in reputation disputes that there is always at least one entry in the update log...

kronosapiens: Hrm, but all the tests passed... (the skip is currently at the network level, i.e. affects all updates).

elenadimitrova: Perhaps we don't have a test that completes a ReputationMiningCycle and confirms a hash with an empty log?

PR #495, opened 2019-01-10
Remove redundant mining client initializations

kronosapiens:
Implements part of #317 

I noticed we initialized quite a few redundant mining clients. Why waste the cycles? Also, I reworked some variable names to take advantage of javascript destructuring in function calls and make the miner invocations more succinct.

### DIFF
--- a/test/colony-network-mining.js
+++ b/test/colony-network-mining.js
@@ -62,7 +62,7 @@ const ITokenLocking = artifacts.require("ITokenLocking");
 const Token = artifacts.require("Token");
 const IReputationMiningCycle = artifacts.require("IReputationMiningCycle");
 
-const contractLoader = new TruffleLoader({
+const loader = new TruffleLoader({
   contractDir: path.resolve(__dirname, "..", "build", "contracts")
 });
 
@@ -87,7 +87,7 @@ contract("ColonyNetworkMining", accounts => {
   let goodClient;
   let badClient;
   let badClient2;
-  const REAL_PROVIDER_PORT = process.env.SOLIDITY_COVERAGE ? 8555 : 8545;
+  const realProviderPort = process.env.SOLIDITY_COVERAGE ? 8555 : 8545;
 
   before(async () => {
     const etherRouter = await EtherRouter.deployed();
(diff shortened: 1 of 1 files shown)


test/colony-network-mining.js
```diff
@@ -1758,6 +1681,7 @@ contract("ColonyNetworkMining", accounts => {
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT2, DEFAULT_STAKE);
       await giveUserCLNYTokensAndStake(colonyNetwork, OTHER_ACCOUNT3, DEFAULT_STAKE);
 
+      await fundColonyWithTokens(metaColony, clny, INITIAL_FUNDING.muln(4));
```
elenadimitrova: Why was this additional funding necessary?

test/colony-network-mining.js
```diff
@@ -4290,11 +4165,8 @@ contract("ColonyNetworkMining", accounts => {
       const nInactiveLogEntries = await repCycle.getReputationUpdateLogLength();
       assert.equal(nInactiveLogEntries.toNumber(), 13);
 
-      badClient = new MaliciousReputationMinerExtraRep(
-        { loader: contractLoader, minerAddress: OTHER_ACCOUNT, realProviderPort: REAL_PROVIDER_PORT, useJsTree },
-        40, // Skill 4
-        "0xfffffffff"
-      );
+      // Skill 4
```
elenadimitrova: I added the comments alongside some of the bad clients `entityToFalsify` to try and deal with some of the uncertainty around verifying which entry we are falsifying so at least when we refactor we can use that bit of information without having to log the good client updates first to understand the intention. Unless we find a better way to verify these (as described [here](https://github.com/JoinColony/colonyNetwork/issues/317#issuecomment-416025328) ) I suggest we add this information to all bad clients instantiations. It probably shouldn't be done as part of this PR but still, worth mentioning it here.

test/colony-network-mining.js
kronosapiens: Not sure, it showed up when I rebased onto an earlier PR.

test/colony-network-mining.js
kronosapiens: 👍

test/colony-network-mining.js
kronosapiens: Yup, I remember now -- there was a funding line added to the `beforeEach` block for this suite of tests, so this isn't necessary. It's a tad opaque for my taste but so be it.

elenadimitrova: It might look better once I split them off and they all get their individual `beforeEach`s

kronosapiens: :D :D :D

PR #498, opened 2019-01-12
Changed data-type of element in struct of nUpdates & nPreviousUpdates from uint256 to uint128

### LINKED ISSUE #496: Reduce size of ReputationLogEntry
This:

```
  struct ReputationLogEntry {
    address user;
    int amount;
    uint256 skillId;
    address colony;
    uint256 nUpdates;
    uint256 nPreviousUpdates;
  }
```

Could be this:

```
  struct ReputationLogEntry {
    address user;
    int amount;
    uint256 skillId;
    address colony;
    uint128 nUpdates;
    uint128 nPreviousUpdates;
  }
```

Which slims the struct down by one word and thus reduces the cost of every update by 20k gas. Assuming there are less than UINT128_MAX (340282366920938463463374607431768211455) updates per cycle we will be ok.

dev1644:
Closes #496
Implements required a part for #496

Changed data-type of elements in struct ReputationLogEntry. nUpdates & nPreviousUpdates from uint256 to uint128 thus effectively reducing gas costs by 20k.

### DIFF
--- a/contracts/ColonyNetwork.sol
+++ b/contracts/ColonyNetwork.sol
@@ -78,7 +78,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
     globalSkill = skills[_skillId].globalSkill;
   }
 
-  function getSkillNParents(uint256 _skillId) public view returns (uint256 nParents) {
+  function getSkillNParents(uint256 _skillId) public view returns (uint128 nParents) {
     nParents = skills[_skillId].nParents;
   }
 
@@ -272,9 +272,9 @@ contract ColonyNetwork is ColonyNetworkStorage {
       return;
     }
 
-    uint nParents = skills[_skillId].nParents;
+    uint128 nParents = skills[_skillId].nParents;
     // We only update child skill reputation if the update is negative, otherwise just set nChildren to 0 to save gas
-    uint nChildren = _amount < 0 ? skills[_skillId].nChildren : 0;
+    uint128 nChildren = _amount < 0 ? skills[_skillId].nChildren : 0;
     IReputationMiningCycle(inactiveReputationMiningCycle).appendReputationUpdateLog(
       _user,
       _amount,
--- a/contracts/ColonyNetworkDataTypes.sol
+++ b/contracts/ColonyNetworkDataTypes.sol
@@ -97,9 +97,9 @@ contract ColonyNetworkDataTypes {
 
   struct Skill {
     // total number of parent skills
-    uint256 nParents;
+    uint128 nParents;
     // total number of child skills
-    uint256 nChildren;
+    uint128 nChildren;
     // array of `skill_id`s of parent skills starting from the 1st to `n`th, where `n` is an integer power of two larger than or equal to 1
     uint256[] parents;
     // array of `skill_id`s of all child skills
@@ -118,7 +118,7 @@ contract ColonyNetworkDataTypes {
     int amount;
     uint256 skillId;
     address colony;
-    uint256 nUpdates;
-    uint256 nPreviousUpdates;
+    uint128 nUpdates;
+    uint128 nPreviousUpdates;
   }
 }
--- a/contracts/ColonyNetworkMining.sol
+++ b/contracts/ColonyNetworkMining.sol
@@ -40,8 +40,8 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     int _amount,
     uint256 _skillId,
     address _colony,
-    uint256 _nUpdates,
-    uint256 _nPreviousUpdates)
+    uint128 _nUpdates,
+    uint128 _nPreviousUpdates)
     public recovery auth
     {
     replacementReputationUpdateLogsExist[_reputationMiningCycle] = true;
--- a/contracts/IColonyNetwork.sol
+++ b/contracts/IColonyNetwork.sol
@@ -51,8 +51,8 @@ contract IColonyNetwork is ColonyNetworkDataTypes, IRecovery {
     int _amount,
     uint256 _skillId,
     address _colony,
-    uint256 _nUpdates,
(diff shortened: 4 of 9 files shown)


dev1644:
Seems like, I must have missed some commits, Let me fix this.

dev1644:
Well, I have no idea why tests are failing now, I could use some help here.

dev1644:
In contract ReputationMiningCycle at L403 change the data-type from uint128 to uint256 of nParents and nChildren & at L418 & L420 type-casted the result to uint128.
But I think it can potentially lead to the unforeseeable assignment of values to `nUpdate`.

kronosapiens:
Hi @dev1644 , thanks for your work on this. I went over your commits and have some feedback:

- The failing test is here: https://github.com/JoinColony/colonyNetwork/blob/develop/test/colony-network-recovery.js#L299 The test is skipped by the coverage tests, which is why the failure only occurs during the unit testing.

- The failure is due to this test checking our ability to manually set values in storage. Since this PR has combined two storage slots into one (by using uint128 instead of uint256), the test currently expects the now-unnecessary storage slot to have a value. The solution is (I am 90% sure) to update the test to combine two values into a single storage slot when performing the manual setting. Specifically, these lines need to be combined into a single update at position `(logEntryStartingSlot.addn(4)`, with each of the two current values occupying half of a uint256 slot:

```
await newActiveCycleAsRecovery.setStorageSlot(logEntryStartingSlot.addn(4), `0x${new BN(logEntry.nUpdates).toString(16, 64)}`);
await newActiveCycleAsRecovery.setStorageSlot(logEntryStartingSlot.addn(5), `0x${new BN(logEntry.nPreviousUpdates).toString(16, 64)}`);
```

- We would rather you rebase your branch onto `develop` instead of merging, which creates unnecessary merge commits. Are you familiar with interactive rebasing? Running `git rebase -i HEAD~5` should let you remove the merge commits, and you can then rebase your changes onto `develop`. Overall, the changes are small so ideally could be rolled up into a single commit.

- I don't love the casting here: https://github.com/JoinColony/colonyNetwork/pull/498/files#diff-21ee0c070df2d15abfdf2c1a0e8c8737R418 As you say, this could lead to unexpected behavior. The "proper" solution would involve updating the upstream `Skill` structs to also use `uint128`s. Would you be willing to take a crack at that as well?

dev1644:
Hi @kronosapiens, Yes, I am up for the task, I will change the skill struct and will also adjust the tests.

dev1644:
- As we can see, Updating all elements inside of Skill Struct will require much more changes.

- We also need to update `skillId` of Struct ReputationLogEntry to uin128 too, because we are using `skillCount`  from  https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyNetworkStorage.sol#L39 to uint128 or either type the result to uint128 because `skillCount` is being used to assign a value of `nParent`, https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyNetwork.sol#L230

kronosapiens:
Hi @dev1644 , I took another look at your changes and I found the issue. This line needs to be updated  to advance the index the correct  distance for the new size of the struct:

```
const logEntryStartingSlot = new BN(arrayStartingSlot.slice(2), 16).add(new BN(i * 6));
```

must become

```
const logEntryStartingSlot = new BN(arrayStartingSlot.slice(2), 16).add(new BN(i * 5));
```

I made the change locally and the test passed.

kronosapiens:
Also, can you omit this change? The `i` here is not related to the struct changes we are discussing: https://github.com/JoinColony/colonyNetwork/pull/498/files#diff-21ee0c070df2d15abfdf2c1a0e8c8737R467

dev1644:
Got it, I also had my doubts about that line, was gonna try that today.
Ok, I will fix the line and will push it.

kronosapiens:
Nice catch, I didn't notice that.

kronosapiens:
Nicely done. Before we accept this, I'd like you to do an interactive git rebase to clean up the commit log, and then `rebase` onto develop.

Here is some more info: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

dev1644:
Hi @kronosapiens, I have done the rebasing and created two separate commits, one for changing the data-type of variable and other for fixing up the tests.

dev1644:
After Rebasing onto `develop`, I lost my most of my commits (It's my fault that I don't know how to use git properly) while doing the things again, I notice at https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ReputationMiningCycle.sol#L601 the tests failing because of `invalid type-conversion`, I didn't encounter `checkJRHProof2` before, so I think it must have come from the latest `develop` branch.
The test is failing because we are trying to add `uint128` to reputationRootHashNNodes`uint256`.
For right now, I am type-casting it (though I think it is dangerous) to see if other issues regarding tests are resolved or not, and I want to know more about `reputationRootHashNNodes`.

Can we convert `reputationRootHashNNodes` to uint128? If not then I would like to research different workarounds for this.

kronosapiens:
Hi @dev1644 , I took a look and it's ok. The tricky thing was that the `nUpdates` you are setting to here refers to something different than the `nUpdates` we've been working with -- it's really more like an `nTotalUpdates`, which should be a `uint256`. Update that part to this and you should be good to go:

```
    uint256 nUpdates = reputationRootHashNNodes +
      uint256(reputationUpdateLog[nLogEntries-1].nUpdates) +
      uint256(reputationUpdateLog[nLogEntries-1].nPreviousUpdates);
```

dev1644:
Edited and Pushed.

kronosapiens:
@dev1644 Thanks! Would you mind squashing that commit into your previous? Then we should be good to go :)

dev1644:
Done.

PR #500, opened 2019-01-15
Disable no-await-in-loop rule

kronosapiens:
Implements part of #317 

Disable `no-await-in-loop` rule and remove all one-off disabling comments.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,8 @@
     "flowtype/require-valid-file-annotation": "off",
     "import/named": "off",
     "no-console": "off",
-    "no-unused-expressions": "off"
+    "no-unused-expressions": "off",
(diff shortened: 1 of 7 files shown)


.eslintrc
```diff
@@ -16,7 +16,8 @@
     "flowtype/require-valid-file-annotation": "off",
     "import/named": "off",
     "no-console": "off",
-    "no-unused-expressions": "off"
+    "no-unused-expressions": "off",
+    "no-await-in-loop": "off"
```
elenadimitrova: Just remove the rule from the lower level `packages/reputation-miner/.eslintrc` now we have it project wide here.

kronosapiens: 👍

PR #508, opened 2019-01-20
Throw error for insufficient miner stake

kronosapiens:
Implements part of #317 

Currently, an insufficient stake leads the miner to `return` an Error in `submitRootHash`, due to the loop not finding a valid index. This leads to an odd behavior where the actual error which appears to the developer is a `missing method for tx.wait()`; and further, the return value for `miner.submitRootHash` is ambiguous (an Error or a Tx). We should throw the error so that the miner knows what's gone wrong, and so the method always returns a Tx.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -11,6 +11,7 @@ const sqlite = require("sqlite");
 // do in the future.
 // const accountAddress = "0xbb46703786c2049d4d6dd43f5b4edf52a20fefe4";
 const secretKey = "0xe5c050bb6bfdd9c29397b8fe6ed59ad2f7df83d6fd213b473f84b489205d9fc7";
+const miningCycleDuration = 60 * 60 * 24; // 24 hours
 
 class ReputationMiner {
   /**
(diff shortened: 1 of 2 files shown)


packages/reputation-miner/ReputationMiner.js
```diff
@@ -603,7 +603,7 @@ class ReputationMiner {
       }
     }
     if (!entryIndex) {
-      return new Error("No valid entry for submission found");
+      throw new Error("No valid entry for submission found.");
```
elenadimitrova: If I'm reading this right, `entryIndex` will always have a value if [this](https://github.com/JoinColony/colonyNetwork/pull/508/files#diff-9c228e1a1f7fe6683164300c770b62ffR593) condition passes even once. Therefore the only time we throw an error is when the above-mentioned condition  is never true?

area: Hm, entryIndex should actually only be set after the estimate for `submitRootHash` succeeds.

elenadimitrova: 👆 I agree

kronosapiens: Mkay I can update. And yeah - if there's no stake, then we never even loop and so this is `undefined`. If there is only a small stake, then we may not find a valid index, and so `undefined`.

kronosapiens: Ok I fixed.

PR #510, opened 2019-01-22
Upgrade to babel 7 and resolve source map support failure in tests

elenadimitrova:
In an attempt to coerce `mocha` to report back correct code line numbers on test failures ([sample failing build reporting bad lines](https://circleci.com/gh/JoinColony/colonyNetwork/5944)) we upgrad to `babel@7` here. Although this hasn't solved the "line number" issue it is worth upgrading as it's not a straight forward change. Because `.babelrc` is no longer recognised as a global configuration, this had to be switched to `babel.config.js`  see https://babeljs.io/docs/en/config-files#root-babelconfigjs-files and https://github.com/facebook/jest/issues/6053#issuecomment-383632515 for explanation.

Ultimately after much digging in historical builds and merged changes, we managed to pinpoint the [bad commit](https://github.com/JoinColony/colonyNetwork/commit/23e89d310bcbdcdeeae86244711d52eec6c80153) that started causing the incorrect line number reports in test failures. To solve this issue we pinned the `ganache-core` version down to a working one. Logged https://github.com/trufflesuite/ganache-core/issues/287 as a result.

Upgrading to `node@10.12` here as well to match the dApp.


test-gas-costs/gasCosts.js
```diff
@@ -214,7 +214,12 @@ contract("All", function(accounts) {
...
 
-      const goodClient = new ReputationMinerTestWrapper({ loader: contractLoader, minerAddress: STAKER1, realProviderPort: REAL_PROVIDER_PORT });
+      const goodClient = new ReputationMinerTestWrapper({
+        loader: contractLoader,
+        minerAddress: STAKER1,
+        realProviderPort: REAL_PROVIDER_PORT,
+        useJsTree: true
+      });
       const badClient = new MaliciousReputationMinerExtraRep(
         { loader: contractLoader, minerAddress: STAKER2, realProviderPort: REAL_PROVIDER_PORT },
```
area: Shouldn't the bad client get to use it too? 😄

elenadimitrova: Yes of course.. I'll do that

PR #515, opened 2019-01-28
Upgrade colony-funding.js to bn-chai

### LINKED ISSUE #305: Use bn-chai for big numbers testing

kronosapiens:
Closes #305 

Pretty straightforward. Did the thing, overall seems like a good change.

### DIFF
--- a/test/colony-funding.js
+++ b/test/colony-funding.js
@@ -62,16 +62,16 @@ contract("Colony Funding", accounts => {
       let colonyRewardPotBalance = await colony.getPotBalance(0, otherToken.address);
       let colonyPotBalance = await colony.getPotBalance(1, otherToken.address);
       let colonyTokenBalance = await otherToken.balanceOf(colony.address);
-      assert.equal(colonyTokenBalance.toNumber(), 100);
-      assert.equal(colonyPotBalance.toNumber(), 0);
-      assert.equal(colonyRewardPotBalance.toNumber(), 0);
+      expect(colonyTokenBalance).to.eq.BN(100);
+      expect(colonyPotBalance).to.eq.BN(0);
+      expect(colonyRewardPotBalance).to.eq.BN(0);
       await colony.claimColonyFunds(otherToken.address);
       colonyRewardPotBalance = await colony.getPotBalance(0, otherToken.address);
       colonyPotBalance = await colony.getPotBalance(1, otherToken.address);
       colonyTokenBalance = await otherToken.balanceOf(colony.address);
-      assert.equal(colonyTokenBalance.toNumber(), 100);
-      assert.equal(colonyPotBalance.toNumber(), 99);
-      assert.equal(colonyRewardPotBalance.toNumber(), 1);
+      expect(colonyTokenBalance).to.eq.BN(100);
+      expect(colonyPotBalance).to.eq.BN(99);
+      expect(colonyRewardPotBalance).to.eq.BN(1);
     });
 
     it("should syphon off own tokens in to the reward pot", async () => {
@@ -79,9 +79,9 @@ contract("Colony Funding", accounts => {
       const colonyRewardPotBalance = await colony.getPotBalance(0, token.address);
       const colonyPotBalance = await colony.getPotBalance(1, token.address);
       const colonyTokenBalance = await token.balanceOf(colony.address);
-      assert.equal(colonyTokenBalance.toNumber(), 100);
-      assert.equal(colonyPotBalance.toNumber(), 99);
-      assert.equal(colonyRewardPotBalance.toNumber(), 1);
+      expect(colonyTokenBalance).to.eq.BN(100);
+      expect(colonyPotBalance).to.eq.BN(99);
+      expect(colonyRewardPotBalance).to.eq.BN(1);
     });
 
     it("should let tokens be moved between pots", async () => {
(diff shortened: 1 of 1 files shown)


test/colony-funding.js
```diff
@@ -62,26 +62,26 @@ contract("Colony Funding", accounts => {
       let colonyRewardPotBalance = await colony.getPotBalance(0, otherToken.address);
       let colonyPotBalance = await colony.getPotBalance(1, otherToken.address);
       let colonyTokenBalance = await otherToken.balanceOf(colony.address);
-      assert.equal(colonyTokenBalance.toNumber(), 100);
-      assert.equal(colonyPotBalance.toNumber(), 0);
-      assert.equal(colonyRewardPotBalance.toNumber(), 0);
+      expect(colonyTokenBalance).to.eq.BN(100);
+      expect(colonyPotBalance).to.eq.BN(0);
+      expect(colonyRewardPotBalance).to.eq.BN(0);
```
elenadimitrova: Can we use the full potential of `bn-chai` and do `to.be.zero`?

test/colony-network-auction.js
```diff
@@ -313,30 +316,28 @@ contract("Colony Network Auction", accounts => {
...
-      assert.isTrue(finalized);
+      expect(finalized).to.be.true;
     });
 
     it("all CLNY sent to the auction in bids is burned", async () => {
       const balanceBefore = await clnyToken.balanceOf(tokenAuction.address);
       const supplyBefore = await clnyToken.totalSupply();
       const receivedTotal = await tokenAuction.receivedTotal();
-      assert.isFalse(receivedTotal.isZero());
+      expect(receivedTotal.isZero()).to.be.false;
```
elenadimitrova: Better be `expect(receivedTotal).to.be.zero`

test/colony-funding.js
kronosapiens: Yes! I was just circling back to make those updates.

test/colony-network-auction.js
kronosapiens: Yup!

test/colony-network-auction.js
```diff
@@ -313,30 +316,28 @@ contract("Colony Network Auction", accounts => {
...
       const balanceBefore = await clnyToken.balanceOf(tokenAuction.address);
       const supplyBefore = await clnyToken.totalSupply();
       const receivedTotal = await tokenAuction.receivedTotal();
-      assert.isFalse(receivedTotal.isZero());
+      expect(receivedTotal).to.not.be.zero;
       await tokenAuction.finalize();
 
       const balanceAfter = await clnyToken.balanceOf(tokenAuction.address);
-      assert.equal(balanceAfter.toString(), "0");
+      expect(balanceAfter).to.eq.BN(0);
```
elenadimitrova: Missed this.

test/colony-network-auction.js
```diff
@@ -413,7 +414,7 @@ contract("Colony Network Auction", accounts => {
       await tokenAuction.finalize();
       await tokenAuction.claim({ from: BIDDER_1 });
       const bid = await tokenAuction.bids(BIDDER_1);
-      assert.equal(bid.toNumber(), 0);
+      expect(bid).to.eq.BN(0);
```
elenadimitrova: And this

test/colony-network-recovery.js
```diff
@@ -84,26 +89,26 @@ contract("Colony Network Recovery", accounts => {
       let numRecoveryRoles;
 
       numRecoveryRoles = await colonyNetwork.numRecoveryRoles();
-      assert.equal(numRecoveryRoles.toNumber(), 0);
+      expect(numRecoveryRoles).to.eq.BN(0);
```
elenadimitrova: ...

test/colony-network-recovery.js
```diff
@@ -285,12 +290,12 @@ contract("Colony Network Recovery", accounts => {
...
           const newHashNNodes = await colonyNetwork.getReputationRootHashNNodes();
-          assert.equal(newHash, rootHash);
-          assert.equal(newHashNNodes.toNumber(), nNodes.toNumber());
+          expect(newHash).to.equal(rootHash);
+          expect(newHashNNodes).to.eq.BN(nNodes.toString()); // nNodes is a BigNumber :sob:
 
           await newClient.sync(startingBlockNumber);
           const newValue = newClient.reputations[reputationKey].slice(2, 66);
-          assert.equal(new BN(newValue, 16).toNumber(), 0);
+          expect(new BN(newValue, 16)).to.eq.BN(0);
```
elenadimitrova: ...

test/colony-recovery.js
```diff
@@ -25,28 +31,28 @@ contract("Colony Recovery", accounts => {
       let numRecoveryRoles;
 
       numRecoveryRoles = await colony.numRecoveryRoles();
-      assert.equal(numRecoveryRoles.toNumber(), 0);
+      expect(numRecoveryRoles).to.eq.BN(0);
```
elenadimitrova: ...

test/colony-recovery.js
```diff
@@ -133,7 +139,7 @@ contract("Colony Recovery", accounts => {
       await colony.setStorageSlotRecovery(5, "0xdeadbeef");
 
       const unprotected = await web3GetStorageAt(colony.address, 5);
-      assert.equal(unprotected.toString(), `0xdeadbeef${"0".repeat(56)}`);
+      expect(unprotected).to.eq.BN(`0xdeadbeef${"0".repeat(56)}`);
```
elenadimitrova: We ought to find a better way to pad variables to correct length.. But I guess it's not the purpose of this PR

test/colony-task-work-rating.js
```diff
@@ -69,11 +74,12 @@ contract("Colony Task Work Rating", accounts => {
       const taskId = await setupAssignedTask({ colonyNetwork, colony, dueDate });
 
       await colony.submitTaskDeliverableAndRating(taskId, DELIVERABLE_HASH, RATING_1_SECRET, { from: WORKER });
+      const currentTime2 = await currentBlockTime();
       const ratings = await colony.getTaskWorkRatings(taskId);
-      assert.equal(ratings[0].toNumber(), 1);
-      assert.closeTo(ratings[1].toNumber(), currentTime, 2);
+      expect(ratings.nSecrets).to.eq.BN(1);
+      expect(ratings.lastSubmittedAt).to.eq.BN(currentTime2);
```
elenadimitrova: Can we keep the `closeTo` here, which `chai` supports. This can be brittle with blocktime variations.

test/colony-recovery.js
kronosapiens: Fair fair. I was able to get rid of `parseInt` though :)

PR #517, opened 2019-01-28
Improve test coverage and refactoring

### LINKED ISSUE #317: Test suite refactoring

elenadimitrova:
Closes #317 
Code coverage thresholds are now at 94% for branches and 99% for statements, functions and lines. The contracts that might still be improved upon are the `ReputationMiningCycle` and `ColonyNetworkAuction` both of which have pending refactoring work that might affect these. The remaining uncovered lines in other contracts are impossible to cover due to us doing multiple checks at different levels which secure certain paths more than once.

Overall build time is down with ~11 min to 1h including the added tests which brought branches coverage up from 90% to 94%.

- Refactor `reputation-update` tests to not recreate the entire network but just clear the reputation state
- Skip reputation miner client tests in coverage testing
- Improve coverage for `ENSRegistry`, `ContractRecovery`, `Colony`
- Fixes an issue with `IColony.verifyReputationProof` which was always returning `true` probably due to the following `solc 0.5.0` compiler changes
>Pure and view functions are now called using the opcode STATICCALL instead of CALL if the EVM version is Byzantium or later. This disallows state changes on the EVM level. 

- Fixes an issue with `IColonyNetwork.createColony` where an empty token was accepted. Since we can no longer change the colony token, this should be mandatory in the create call.

Because of issues with the custom `ganache-cli` under node 10 which only surfaced here due to the yarn cache hiding the problem in CI, we upgrade to the latest vanilla `ganache-cli` but with @area 's fixed version of `ethers`


test/reputation-mining/happy-paths.js
```diff
@@ -725,8 +725,22 @@ contract("Reputation Mining - happy paths", accounts => {
       const key = makeReputationKey(metaColony.address, new BN("2"), MINER1);
       const value = goodClient.reputations[key];
       const [branchMask, siblings] = await goodClient.getProof(key);
-      const isValid = await metaColony.verifyReputationProof(key, value, branchMask, siblings, { from: MINER1 });
+      // Checking all good parameters confirms a good proof
+      let isValid = await metaColony.verifyReputationProof(key, value, branchMask, siblings, { from: MINER1 });
       assert.isTrue(isValid);
+
+      const badKey = makeReputationKey("0xdeadbeef", new BN("2"), MINER1);
+      // Check using a bad key confirms an invalid proof
```
area: You're checking bad key and a bad branchmask in this test, but not bad value or bad siblings?

.solcover.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
     skipFiles: [
       'Migrations.sol',
       'EtherRouter.sol',
-      'PatriciaTree'
+      'PatriciaTree',
+      'ContractEditing' // Covered implicitly via the colony-network-recovery.js tests
```
area: I agree it doesn't need to factor in to our coverage, but I don't understand the comment? I'd say it was because it's only used in setting up tests, never in production.

test/reputation-update.js
```diff
@@ -150,64 +178,64 @@ contract("Reputation Updates", accounts => {
...
       expect(repLogEntryWorker1.amount).to.eq.BN(WORKER_PAYOUT);
 
-      const repLogEntryWorker2 = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(3);
+      const repLogEntryWorker2 = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(4);
       expect(repLogEntryWorker2.user).to.equal(WORKER);
       expect(repLogEntryWorker2.amount).to.eq.BN(WORKER_PAYOUT);
     });
 
-    it("should set the correct reputation change amount in log when evaluator has failed to rate", async () => {
+    it("should set the correct reputation change amount in log when evaluator has failed to rate", async function() {
```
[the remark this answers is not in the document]
kronosapiens: Oh rad. 👍

PR #518, opened 2019-01-28
Fix issue around origin reputation

area:
We had, at some point, implicitly assumed that `originReputationValue` could be either the colony-wide origin reputation, or the user-specific origin reputation, and tried to accommodate that in the `checkOriginReputation` function. However, we ultimately only use the origin reputation around [this line](https://github.com/JoinColony/colonyNetwork/blob/f8d7721a29bdd6519a3449be2fdd0efb9aab5f48/contracts/ReputationMiningCycleRespond.sol#L595), where it must be the user origin reputation.

The reputation miner correctly [only ever supplies the user's origin reputation](https://github.com/JoinColony/colonyNetwork/blob/f8d7721a29bdd6519a3449be2fdd0efb9aab5f48/packages/reputation-miner/ReputationMiner.js#L266). However, a malicious miner could instead use the colony-wide origin reputation in its calculations, and pass it here during a dispute over the colony-wide total. This would cause such a submission to be able to tie the correct new reputation state, allowing any other to be set 'by default' if judiciously submitted alongside.

The first commit introduces a test that demonstrates the issue around being able to unexpectedly tie the good submission, and the second commit fixes it by making the checks around origin reputation stricter, now we have clarity of mind to realise we're only expecting the user's origin reputation to be passed in. I have also renamed variables to make it more obvious that we are only expecting the user's origin reputation.

We could, of course, now just build the originKey we're expecting from the log, and not have it passed in at all... worth doing that refactoring here, too?

### DIFF
--- a/contracts/ReputationMiningCycleRespond.sol
+++ b/contracts/ReputationMiningCycleRespond.sol
@@ -82,15 +82,14 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
 
   uint constant B_REPUTATION_KEY = 0;
   uint constant B_PREVIOUS_NEW_REPUTATION_KEY = 1;
-  uint constant B_CHILD_REPUTATION_KEY = 2;
-  uint constant B_ADJACENT_REPUTATION_KEY = 3;
+  uint constant B_ADJACENT_REPUTATION_KEY = 2;
 
   uint constant DECAY_NUMERATOR =    992327946262944; // 24-hr mining cycles
   uint constant DECAY_DENOMINATOR = 1000000000000000;
 
   function respondToChallenge(
     uint256[27] memory u, //An array of 27 UINT Params, ordered as given above.
-    bytes[4] memory b, // An array of 5 bytes params, ordered as given above
+    bytes[3] memory b, // An array of 3 bytes params, ordered as given above
     bytes32[] memory reputationSiblings,
     bytes32[] memory agreeStateSiblings,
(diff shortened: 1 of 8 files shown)


contracts/ReputationMiningCycleRespond.sol
```diff
@@ -265,59 +265,45 @@ contract ReputationMiningCycleRespond is ReputationMiningCycleStorage, PatriciaT
...
-      }
-
-      require(colonyAddressOriginRep == colonyAddressChildRep, "colony-reputation-mining-origin-colony-incorrect");
+      require(colonyAddressOriginRep == logEntry.colony, "colony-reputation-mining-origin-colony-incorrect");
       require(skillIdOriginRep == logEntry.skillId, "colony-reputation-mining-origin-skill-incorrect");
-
-      // For colony wide updates the reputation key userAddress is 0x0, otherwise ensure it the user address of the origin and child skills match
-      if (relativeUpdateNumber < nChildUpdates) {
-        // I don't think this require is possible to hit - I can't see how to get checkKeyLogEntry to pass and this to fail
-        require(userAddressChildRep == address(0x0), "colony-reputation-mining-colony-wide-update-user-nonzero");
```
elenadimitrova: I am guessing I added this before you did the edge case of disputes over colony wide child reputations and so this bug was regressed :(

area: I've definitely got some blood on my hands here! I remember futzing around with this code at the time (the comment about not being able to hit the require is definitely mine) and it didn't click then.

PR #522, opened 2019-02-01
Transparently add the Payment struct

kronosapiens:
For real this time.

- Separate the existing `Task` struct into a `Payment` and `Task` struct, with the same identifier.
- Allow the Payment manager to make changes to the struct unilaterally.
- Allow a one-way "upgrade" from Payment to Task by assigning an evaluator.
- Tasks continue to require multisig for changes.


kronosapiens:
Oh well.

PR #526, opened 2019-02-04
Take bootstrapped tokens from claimed funds only

### LINKED ISSUE #502: Incorrect pot balance after using bootstrapColony
## Summary

If colony funds are claimed before calling `bootstrapColony`, the pot balance of the root domain will be inaccurate. The `bootstrapColony` method should deduct funds from the pot balance of the root domain if the difference between the total supply of the token and the total value of claimed funds is less than the total value of funds being distributed via the `bootstrapColony` method.

## Steps to Reproduce

1. mint tokens
2. claim colony funds
3. bootstrap colony

## Possible Solution

Maybe require the funds being distributed to have been claimed by the colony. The pot balance of the root domain would then need to be greater than or equal to the total value of funds being distributed. The total value of funds being distributed would then need to be deducted from the pot balance of the root domain when calling the `bootstrapColony` method.

area:
Closes #502 

Implemented the fix suggested, which was that tokens distributed during bootstrapping have to come from the claimed funds pot. Strictly, we don't need the `not-enough-tokens` require and could just let `sub` error with `ds-math-sub-underflow`, but I thought it better to have a more explicit error.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -105,8 +105,11 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
     for (uint i = 0; i < _users.length; i++) {
       require(_amounts[i] >= 0, "colony-bootstrap-bad-amount-input");
-      
-      ERC20Extended(token).transfer(_users[i], uint(_amounts[i]));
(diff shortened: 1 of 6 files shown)


kronosapiens:
Q: Shouldn't we take a fee off of the bootstrap token transfers? Seems like otherwise it provides a backdoor to pay out native tokens w/o a fee.

Issue #527, opened 2019-02-05
Add support for 3tx Payments


kronosapiens:
There is a desire for support at the contract level for making payments in as few transactions as possible. Following January's research summit on modularity (see [writeup](https://docs.google.com/document/d/1zWVQ8cuDYaX1dsiSlnII39QV6ykmGH5_66CX6ABfecI/edit)), we would like to support a "3 transaction" payment flow, breaking down as follows:

- Permissions - Only Colony admins are able to create and subsequently amend a payment. Any admin should be able to amend any payment in their colony. There is appetite from product team to limit those powers to the narrower domain admins group when horizontal modularity is implemented.
- Payment will contain payment to a single user in a single token or eth denomination.
- Creating the payment sets recipient, token, amount, source domain (all aforementioned are mandatory props) and optionally skill. 
- Funding the payment (from the source domain).
- Claiming the payment (anyone can claim on behalf of the recipient).
- Reputation is earned only by the recipient (not the payment admin), in the payment domain and optionally in the skill (if this is assigned). The reputation amount will be equal to that earned by completing a task and getting a "meets expectations" rating of `2` i.e. reputation = token (or eth) amount
- Collect reward and network fees on payouts

Further, a "1 transaction" flow can be implemented on top of this through a separate contract (a simple "module" which defines, funds, and claims a payment on behalf of a recipient).

kronosapiens:
What that means is that I can call the `claimPayment` function on your behalf to have money sent to your account. This solves the "first ETH" problem since someone with no ether couldn't even claim their payment, but I can have the "first ETH" payment sent to their account on their behalf. It doesn't mean that I can have someone else's payment sent to _my_ account.

elenadimitrova:
Further questions:
- Domain is mandatory on task, will it be mandated on payments? Same for skills.
- We plan to support multiple skills per task, do we want to allow for multiple skills per payment as well?
- Are we collecting network and reward fees on payouts the same way we charge tasks?
- Will there be any additional rules needed around the funding and claiming of a payment, i.e. can we withdraw a payment back into the parent domain at any point? In tasks we have due dates and completion status to govern that. 
- Can the payment amount, domain, recipient, skill or any other property change after it's created? What permissions if any are needed?
- Can a Payment be cancelled?

kronosapiens:
[...]

PR #530, opened 2019-02-05
Allow anyone to claim payout on Task Role's behalf

kronosapiens:
Implements part of #527

- Remove `require` which allows anyone to call `claimPayout` on behalf of the user (funds go to the user still!).
- Update tests to de-specify the address when calling `claimPayout`. 
- Add guard against sending funds to the null address.

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -89,13 +89,12 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
   taskFinalized(_id)
   {
     Task storage task = tasks[_id];
-    require(task.roles[_role].user == msg.sender, "colony-claim-payout-access-denied");
+    uint payout = task.payouts[_role][_token];
 
-    if (task.roles[_role].rating == TaskRatings.Unsatisfactory) {
+    if (task.roles[_role].rating == TaskRatings.Unsatisfactory || payout == 0) {
       return;
     }
 
-    uint payout = task.payouts[_role][_token];
     task.payouts[_role][_token] = 0;
 
     fundingPots[task.fundingPotId].balance[_token] = sub(fundingPots[task.fundingPotId].balance[_token], payout);
@@ -106,7 +105,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
 
     if (_token == address(0x0)) {
       // Payout ether
-      msg.sender.transfer(remainder);
+      task.roles[_role].user.transfer(remainder);
       // Fee goes directly to Meta Colony
       IColonyNetwork colonyNetworkContract = IColonyNetwork(colonyNetworkAddress);
       address metaColonyAddress = colonyNetworkContract.getMetaColony();
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -89,6 +89,8 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
   taskFinalized(_id)
   {
     Task storage task = tasks[_id];
+    assert(task.roles[_role].user != address(0x0));
+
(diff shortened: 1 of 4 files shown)


contracts/ColonyFunding.sol
```diff
@@ -106,7 +105,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
 
     if (_token == address(0x0)) {
       // Payout ether
-      msg.sender.transfer(remainder);
+      task.roles[_role].user.transfer(remainder);
```
elenadimitrova: Just to be safe against sending funds into the abyss of 0x0, can we `assert` `task.roles[_role].user != address(0x0)`.

area:
> Should we similarly modify `claimRewardPayout` to match this behaviour? I'd argue in favour for the sake of consistency? @area ?

I can't think of a reason why we shouldn't.

kronosapiens:
> > Should we similarly modify `claimRewardPayout` to match this behaviour? I'd argue in favour for the sake of consistency? @area ?
> 
> I can't think of a reason why we shouldn't.

The difference is that `claimPayout` takes a `roleId` which denotes which address gets paid out; `msg.sender` is redundant there. `claimRewardPayout`, on the other hand, uses `msg.sender` to figure out to whom to send the funds. So they're different in that regard, opening up `claimRewardPayout` would necessitate adding an additional argument, which is the recipient. I'm not sure that would represent an improvement.

contracts/ColonyFunding.sol
kronosapiens: Mm I'm not 100% on this for three reasons:

1) I'm not sure this is a state we can actually get into via the current Task flow. Why? Because for any user to receive a payout, the have to take some positive action. Managers setup the task; evaluators evaluate, workers submit work. So even _if_ there was a null account, it could not receive a positive payout.

2. There are many accounts other than `address(0x0)` which should not receive tokens. Adding this check gives a partial sense of security, in that it doesn't account for any of the countless possible addresses which shouldn't hold funds.

3. Because this branch is hard to test with the current Task flow (see point 1), it brings our coverage down below the 94% threshold and makes the build fail. 😭

elenadimitrova: I disagree with all 3, to answer more specifically
1) The probability of funds being sent to an empty (vs non-empty) address combined with the fact that is an address we're guaranteed not to be able to recover funds from (as opposed to accidentally sending to another user for instance) increase the risk we fall into this scenario. Combined with the fact the ease of of this check makes it a recommended option. This is not misleading security but actually improved security of our contracts.
2) You've just described the very definition of when `assert` should be used, see the Solidity docs.
3) We have other invariants which are untestable, hence branch coverage is at 94% and not 100%. We should prefer to lower the `istanbul` coverage check instead of forsaking asserting the invariant.

kronosapiens: Otay

elenadimitrova:
But your coverage is at over 94%? https://circleci.com/gh/JoinColony/colonyNetwork/6339?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

contracts/ColonyFunding.sol
```diff
@@ -89,6 +89,8 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
   taskFinalized(_id)
   {
     Task storage task = tasks[_id];
+    require(task.roles[_role].user != address(0x0), "colony-funding-payout-null-recipient");
```
elenadimitrova: This should be `assert` not `require` as per the discussion we had around it.

kronosapiens: Fair fair will update. Also hmm the coverage is above 94 this time... 😕

PR #531, opened 2019-02-06
Rename Pot to FundingPot

elenadimitrova:
This terminology has been bothering me for a year and the time has come to make it better. `FundingPot` better demonstrates the concept intention behind the fund pools these manage and is additionally aligned with white-paper terminology.

I've tried to not be too invasive with the rename so there are still test variables called `pot` but these can be refactored along in time if we want to be perfectly aligned. For now the contracts and natspec comments suffices I think.

### DIFF
--- a/test/colony-funding.js
+++ b/test/colony-funding.js
@@ -417,13 +417,13 @@ contract("Colony Funding", accounts => {
     it("should return correct number of funding funding pots", async () => {
       const taskCountBefore = await colony.getTaskCount();
       expect(taskCountBefore).to.be.zero;
-      const potCountBefore = await colony.getPotCount();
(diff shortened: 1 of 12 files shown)


test/colony-funding.js
```diff
@@ -414,16 +414,16 @@ contract("Colony Funding", accounts => {
       expect(colonyPotBalance).to.eq.BN(297);
     });
 
-    it("should return correct number of funding pots", async () => {
+    it("should return correct number of funding funding pots", async () => {
```
kronosapiens: This one reads weird.

elenadimitrova: Yes, an overzealous rename. I'll fix

PR #533, opened 2019-02-07
Prevent bad actors responding on behalf of good

### LINKED ISSUE #410: RespondToChallenge optional branch can be skipped by bad actor
There is an 'optional' branch in `respondToChallenge`, which is executed if the submitter is proving that a reputation existed in `lastAgreeState`, and increments `challengeStepCompleted` an extra `1`. In certain circumstances (the reputation exists, and is going to become zero after the reputation update under dispute is applied), a bad actor can respond on behalf of a good actor, skipping this branch (even though the good actor could prove it), but otherwise successfully completing `respondToChallenge`, causing the good submission to be eliminated incorrectly.

area:
Closes #410... but not in a very satisfying way?

Firstly, some background. The reason this issue existed was fundamentally because there were two states that could be submitted that had a plausible numerical calculation done - a bad actor could assume a reputation didn't exist, and use a value of zero when doing the calculation, adding a new node. If a good actor proved the reputation existed, and then did the calculation correctly, we would have two submissions that successfully passed, and so had to use a tiebreak to decide between them, which was done with the additional increment of `challengeStepsCompleted` in the case of the submission that proved the reputation already existed. By responding on behalf of the good actor and skipping this proof (and thus the additional increment), the proofs were tied even with the tiebreak, and both would be eliminated.

The issue arose where the reputation existed, but the same answer was arrived at if you assumed a value of 0 for the existing reputation. In that case (which only applied for a negative reputation change), a bad actor could call `respondToChallenge` on behalf of the good actor, successfully pass all the requires, but skip the proof of the fact the reputation already existed (and thus the additional increment of `challengeStepsCompleted`. In that case, both submissions would have the same number of `challengeStepsCompleted`, and both would end up being eliminated.

This is now impossible for two reasons:

1. The introduction of `checkAdjacentReputationHash` in #506 means that, while there are two states that can have a successful calculation, one of them will never reach the calculation and fail earlier in `respondToChallenge`, as claiming a reputation doesn't exist when it actually does is now detectable based on the branchMask of the adjacent hash.

2. The other is the introduction of stricter checks for the values of `agreeStateNNodes` and `disagreeStateNNodes` in this PR. Currently, so long as they differ by either 0 or 1, regardless of the rest of the submission, the `respondToChallenge` call succeeds; this was exploited by the bad actor replying on behalf of the good. Part of this PR makes sure that the choice of 0 or 1 is self-consistent throughout the response.

It was not immediately clear to me why the same attack was not possible on the remaining 'skippable' increments. But it is my belief is that it is now not possible to (productively) respond on behalf of the good submission and bypass the optional proofs. 

Firstly, the remaining skippable increments only apply when considering a negative update. All negative updates check `userOriginReputationValue`. If the update is of a colony-wide total for a child skill, then `userChildReputationValue` is checked as well.

* Any response on behalf of the good submission that successfully calls `respondToChallenge` and skips a proof can only occur when the calculation done with the correct value for `userOriginReputationValue` or `userChildReputationValue` is the same as if the calculation is done with 0 for one of those values (otherwise the value the bad actor proves to be in `disagreeState` won't be consistent with the calculation submitted by the good actor)
* This can only occur if at exactly one of `userOriginReputation` or `userChildReputation` exists, or both exist and at least one has a value of zero, or both exist and logEntry.amount is zero (which can only occur for miners, currently).
* A bad actor can respond for the good response in this scenario and prevent the submission getting some increment(s) that it deserves, but it still gets at least one for a successful call to `respondToChallenge`.
* In this scenario, I believe that the competing state is will not be able to make a successful `respondToChallenge` at all, and will still therefore be eliminated, as I don't think the bad client can have submitted a different state that will be able to be show a plausible calculation. If it skips the proofs for the right `userOriginReputationValue` or `userChildReputationValue`, and thus uses 0 instead of their correct values, it gets the same answer as the good submission, and so would have submitted the same state. To have submitted a different state, it must therefore have submitted an updated value for the reputation under dispute that will fail the calculation check, and so the good client's submission will not be eliminated (because it got one increment for a successful `respondToChallenge`, just not as many as it could have done).

Note that we cannot get rid of the increments in the optional proof paths, in case a bad submission is made that assumes a value of 0 for the calculation, but the reputation actually does exist with some other value - we have to be able to find in favour of the submission that proved an existing reputation value. In this case, `respondToChallenge` requires the proofs to be made to complete successfully, otherwise the calculation for the good client will fail. The tests `if one person claims an origin skill doesn't exist but the other does (and proves such)` and `if one person claims a user's child skill doesn't exist but the other does (and proves such)` demonstrate this, and fail if these increments are removed.

Note that in this scenario, the bad client is _not_ able to respond on behalf of the good client successfully - if it tries to do so by skipping one of the optional reputations, it won't pass the calculation check (which used a nonzero value when the good client submitted it).

### DIFF
--- a/packages/reputation-miner/test/MaliciousReputationMinerClaimNoOriginReputation.js
+++ b/packages/reputation-miner/test/MaliciousReputationMinerClaimNoOriginReputation.js
@@ -1,25 +1,23 @@
-import BN from "bn.js";
 import ReputationMinerTestWrapper from "./ReputationMinerTestWrapper";
 
-class MaliciousReputationMiningNoOriginReputation extends ReputationMinerTestWrapper {
-  // This client will claim there is no originReputationUID, whether there is one or not
-  //
+class MaliciousReputationMinerClaimNoOriginReputation extends ReputationMinerTestWrapper {
+  // This client will claim there is no origin reputation, whether there is one or not, if told to falsify a child update.
+  // Not sure what will happen otherwise!
   constructor(opts, entryToFalsify) {
     super(opts);
     this.entryToFalsify = entryToFalsify;
   }
 
   async addSingleReputationUpdate(updateNumber, repCycle, blockNumber, checkForReplacement) {
-    if (updateNumber.toNumber() === this.entryToFalsify){
-      this.alterThisEntry = true;
-      const reputationKey = await this.getKeyForUpdateNumber(updateNumber);
-      const reputationValue = new BN(this.reputations[reputationKey].slice(2, 66), 16);
-      this.replacementAmount = reputationValue.mul(new BN(-1));
-    }
     await super.addSingleReputationUpdate(updateNumber, repCycle, blockNumber, checkForReplacement)
     if (updateNumber.toNumber() === this.entryToFalsify){
 
       // Set the origin skill key
+      // Because the amount is zero (due to our custom getAmount function below), the origin skill proof object and the user child proof object
+      // will have default (zero) values.
+      // We set the keys here, (because the keys are checked in respondToChallenge) and leave everything else the same.
+      // The amount variable represents the change in the reputation being updated, which for a child update is always zero when there is no origin reputation.
+      // The calculation is therefore self-consistent and will be able to pass respondToChallenge.
       const logEntryNumber = await this.getLogEntryNumberForLogUpdateNumber(updateNumber.sub(this.nReputationsBeforeLatestLog));
       const logEntry = await repCycle.getReputationUpdateLogEntry(logEntryNumber);
       const originSkillUpdateNumber = logEntry.nUpdates.add(logEntry.nPreviousUpdates).add(this.nReputationsBeforeLatestLog).sub(1);
@@ -37,10 +35,8 @@ class MaliciousReputationMiningNoOriginReputation extends ReputationMinerTestWra
       } else {
         childKey = await this.getKeyForUpdateNumber(updateNumber);
       }
-      this.justificationHashes[ReputationMinerTestWrapper.getHexString(updateNumber, 64)].childReputationProof = 
+      this.justificationHashes[ReputationMinerTestWrapper.getHexString(updateNumber, 64)].childReputationProof =
         await this.getReputationProofObject(childKey);
-
-      this.alterThisEntry = false;
     }
   }
 
@@ -54,4 +50,4 @@ class MaliciousReputationMiningNoOriginReputation extends ReputationMinerTestWra
   }
 }
 
-export default MaliciousReputationMiningNoOriginReputation;
+export default MaliciousReputationMinerClaimNoOriginReputation;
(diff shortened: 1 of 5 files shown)


packages/reputation-miner/test/MaliciousReputationMinerClaimNoOriginReputation.js
```diff
@@ -1,53 +1,47 @@
...
+      const nChildUpdates = nUpdates.div(2).sub(1).sub(nParents);
+      let childKey;
+      if (relativeUpdateNumber.lt(nChildUpdates)) {
+        const childSkillUpdateNumber = updateNumber.add(nUpdates.div(2));
+        childKey = await this.getKeyForUpdateNumber(childSkillUpdateNumber);
+      } else {
+        childKey = await this.getKeyForUpdateNumber(updateNumber);
+      }
+      this.justificationHashes[ReputationMinerTestWrapper.getHexString(updateNumber, 64)].childReputationProof = 
+        await this.getReputationProofObject(childKey);
```
elenadimitrova: It's not immediately obvious how this client mades an existing origin reputation disappear. Maybe add some comments? When debugging to find out more I got this out for the bad client's final set of 6 updates (that are relevant for this), see below. 
```
updateNumber 34
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000050000000000000000000000000000000000000000
amount 0
originReputationKey 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000049f485401a3c22529ab6ea15e2ebd5a8ca54a5430
childKey 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000059f485401a3c22529ab6ea15e2ebd5a8ca54a5430
updateNumber 35
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000
amount -500000000000
updateNumber 36
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000
amount -500000000000
updateNumber 37
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000059f485401a3c22529ab6ea15e2ebd5a8ca54a5430
amount -500000000000
updateNumber 38
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000019f485401a3c22529ab6ea15e2ebd5a8ca54a5430
amount -500000000000
updateNumber 39
key 0xd7474abde95761bc61742a8bb2269332d9dc81e700000000000000000000000000000000000000000000000000000000000000049f485401a3c22529ab6ea15e2ebd5a8ca54a5430
amount -500000000000
```

packages/reputation-miner/test/MaliciousReputationMinerClaimNoUserChildReputation.js
```diff
@@ -0,0 +1,58 @@
+import ReputationMinerTestWrapper from "./ReputationMinerTestWrapper";
+
+const ethers = require("ethers");
+
+class MaliciousReputationMiningNoUserChildReputation extends ReputationMinerTestWrapper {
```
elenadimitrova: Same for making the class and file names match.

PR #534, opened 2019-02-08
Remove colony-js-contract-loader-network

ryanchristo:
## Description

This pull request moves `colony-js-contract-loader-network` from colonyNetwork to colonyJS.

## Checklist

- [X] Merge JoinColony/colonyJS#356

## Related PR

JoinColony/colonyJS#356


ryanchristo:
Hmm... test coverage failed after rebasing:

```
ERROR: Coverage for branches (93.93%) does not meet global threshold (94%)
```

@kronosapiens Any ideas what might be causing this?

kronosapiens:
Hmm, nothing you did. Ideally we would keep branch coverage above 94% but since your PR didn't  change any actual code, what this is suggesting to me is that there is some non-determinism in the branch testing. I've run into something similar before, but in my case re-running Circle got me back above 94%. The most straightforward option would be to bump the coverage threshold down to 93% for the time being. Let me rope in @elenadimitrova and @area to see if they have any thoughts.

elenadimitrova:
I wouldn't want to drop down coverage thresholds with a whole percent, can we first try giving `istanbul` `93.9`, its documentation doesn't say whether it provides decimal support for thresholds.

PR #553, opened 2019-02-20
Domain-level permissions

### LINKED ISSUE #551: Implement support for domain-level permissioning (Horizontal Modularity)
As per Jan 2019's research summit (see [writeup](https://docs.google.com/document/d/1zWVQ8cuDYaX1dsiSlnII39QV6ykmGH5_66CX6ABfecI/edit?usp=sharing)), we would like to support "Horizontal Modularity" via an implementation of "domain-level permissions".

Specifically, rather than have two permission classes (`founder` and `admin`) which authorize functions across an entire Colony, we would like to have five permission classes (`funding`, `administration`, `arbitration`, `architecture`, and `root`) which (with the exception of `root`) authorize functions on a per-domain basis.

The idea is that, as these permissions give access to a wide variety of behaviors in a Colony, and that these permissions can be granted to smart contracts as well as accounts, this permission system can be used both to give humans fine-grained control over the Colony operations, as well as to allow Colonies to plug in external contracts ("modules") which implement additional behavior (such as reputation-weighted funding queues, multi-sig authorization, etc).

There are three open questions: one philosophical, two technical.

**The philosophical question: should granting a permission in a parent domain guarantee that permission in child domains, or should the permissions of every domain be wholly independent of all others?**

- The argument for the former is that it greatly simplifies the possible configurations of permissions, makes permissions easier to reason about, minimizes the likelihood of "bad" permission states, and enforces the idea that the scope of a subdomain is strictly a subset of the scope of the parent domain.

- The argument for the latter is that it provides the maximum possible flexibility for a Colony.

**The first technical question: how to represent domain-level permissions?**

Currently, our authorization contract uses the function signature to verify permissions. But with domain-level permissions, the space of verification is the cartesian product of **functions X domains**. This means we need to figure out a way to represent a domain-level permission (something like `sha256(functionsig.domain)` could suffice), as well as inferring the domain when the function is called. The latter problem may prove the trickiest, as we will need to be able to infer from the function arguments what domains are involved. For manipulating tasks and payments, a simple lookup of the domain should suffice. For transferring funds, we can access the domains of the pots. For architecture, domains are arguments to their functions. For arbitration, a mechanism of inferring domain from data being disputed is mentioned in the WP and will likely be used here.

**The second technical question: how to record and inherit domain-level permissions in a gas-efficient (or gas-bound) way?**

- One approach (in line with philosophical solution 1) is to record a permission in the parent domain and _infer_ it for child domains, an approach which minimizes storage requirements. However, given that domain trees can be of arbitrary depth, linearly traversing up or down a domain tree to assess permissions is not viable. @area has noted that, since every parent domain contains an array of children domains, users can provide a "proof" of permission by providing two integers: the domain of the permission, and the location of the target domain in the `domain.children` array.

- Another approach (in line with philosophical solution 2) is to "copy" permissions from a parent to a child domain at the time of creation, and let the permissions be independent afterwards. An even simpler approach is to initialize each new domain with no permissions and require manually updated the permissions in each domain. Unlike the first approach, this approach requires substantially more gas since permissions cannot be "inferred" from parent domains.

The first approach has the lowest gas requirements, since permissions can be inferred in most cases in constant time using a domain proof. The second approach has higher gas requirements since permissions must be explicitly set for every domain, but does not require submitting a domain proof. In both cases, verifying permission at execution time can be done in constant gas.

kronosapiens:
Closes #551 

Remove `Founder` and `Admin` roles entirely, and add support for the following permissions (and functions):

### Root (colony-wide)
- `setRootRole` (arbitrary # of holders, vs current one-only)
- `setArchitectureRole`
- `setFundingRole`
- `setAdministrationRole`
- `setRecoveryRole`
- `removeRecoveryRole`
- `bootstrapColony`
- `registerColonyLabel`
- `setRewardInverse`
- `upgrade`
- `mintTokens`

Meta Colony specific:
- `addNetworkColonyVersion`
- `setNetworkFeeInverse`
- `addGlobalSkill`

As well as the following roles which can be held at the *domain level*:

### Arbitration
- TODO

### Architecture
- `addDomain`
- `setFundingRole` (in subdomain)
- `setAdministrationRole` (in subdomain)
- `setArchitectureRole` (in subdomain)

### Funding
- `transferFundsBetweenPots`

### Administration
- `makeTask`

The domain-level permissions are verified via a new auth modifier `authDomain`. This modifier accepts three arguments:

```
  modifier authDomain(uint256 permissionDomainId,  uint256 childSkillIndex, uint256 childDomainId)
```

It performs three checks:

1. Whether the user has the permission in `permissionDomainId`,
2. Whether `childDomainId` is actually a child of the parent, using `childSkillIndex`,
3. And for the architecture permission, checks to see that the domain is strictly a child of the permission domain (i.e. not the permission domain itself).

Currently, the existing `auth` modifier is preserved and checks for permissions in the root domain. This is similar in flavor to Aragon's `auth` and `authP` which allows for additional predicate logic (such as authority up until a certain block number, for instance).

In order to provide the necessary arguments to `authDomain`, we have added two arguments to every permissioned function, which by convention are the first two arguments. The first is the permission domain, and the second is an index telling us where in the child array of the permission domain we can find the "domain of action", which we infer from the other arguments.

The domain-level permissions are implemented via an override of `roles.sol` (in `DomainRoles.sol`) which adds an additional level of mapping:

```
mapping(address=>mapping(uint256=>bytes32)) _user_domain_roles;
```

All of the accessor functions are also overloaded to accept an additional argument: the domain. DSRoles functions for managing permissions are overridden with implementation that checks permissions in the root colony domain.

This is the new inheritance graph for the Authority contracts:

```
DSAuth => DSRoles => DomainRoles => CommonAuthority => Colony{,Network}Authority
```

Note that this sits parallel with:

```
DSAuth => CommonStorage => {ColonyStorage, ContractRecovery}
```

As we support the "old interface" (without a domain) which checks for the permission in the root domain, we can support both "colony-wide" permissions such as root and recovery and "domain-specific" permissions (the others) on top of the same contracts. For colony-wide roles, we simply omit the domain when updating the role.


scripts/check-auth.js
```diff
@@ -0,0 +1,137 @@
...
+          contractName,
+          "doesn't appear to have the right auth declaration for function ",
+          functionDef.name,
+          ". Errors: ",
+          res.errors
+        );
+        process.exit(1);
+      }
+    });
+});
```
elenadimitrova: Just as a side note here, we can consider decorating _**all**_ Colony functions with `auth` and explicitly marking open functions as ["public capability"](https://github.com/dapphub/ds-roles#setpubliccapability)

contracts/ColonyRoles.sol
```diff
@@ -0,0 +1,85 @@
...
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.4.23;
+
+import "./CommonAuthority.sol";
+
+
+contract ColonyRoles is CommonAuthority {
```
elenadimitrova: I don't understand why this implementation is needed. The `DSRoles` vanilla contract allows for up to 256 roles to be mapped to essentially 63 uniquely permissioned domains (with 4 unique roles each), which I imagine is sufficiently high limit considering we inherit permissions by default.

elenadimitrova:
@kronosapiens can you check why `startNextRewardPayout` is under both Root and Administration permissions? Probably just Root or even a public function?

scripts/check-auth.js
```diff
@@ -0,0 +1,137 @@
...
+    valid = false;
+    errors.push("Second parameter to auth is not _domainId or a lookup thereof");
+  }
+  return { valid, errors };
+}
+
+walkSync("./contracts/").forEach(contractName => {
+  // Only contracts using domain-level permissions need to be checked, i.e. those that implement
+  // functions in IColony.sol
+  // Basically only Colony.sol, ColonyFunding.sol, and ColonyTask.sol (?)
```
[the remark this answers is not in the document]
kronosapiens: This has been brought up for other scripts -- the argument I heard was that it's less error-prone to have to manually exclude new contracts. I.e. say we add a new contract which needs to be checked by the script, theres' a good chance we'll forget to add them to all the scripts which might need to be checked. The way it is currently, you'll get an error during the build. cc @area

contracts/ColonyTask.sol
```diff
@@ -89,13 +89,20 @@ contract ColonyTask is ColonyStorage {
...
+    uint256 _parentDomainId,
+    uint256 _domainProofIndex,
+    bytes32 _specificationHash,
+    uint256 _domainId,
+    uint256 _skillId,
+    uint256 _dueDate
+  ) public
   stoppable
-  auth
+  auth2(_parentDomainId, _domainId, _domainProofIndex)
```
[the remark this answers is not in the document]
kronosapiens: Not necessarily, since the permission might be held in a parent domain. If I have Administration permission in domain A, and then I try to create a task in child domain B, we would need to provide the proof that B is a child domain of A.

scripts/check-recovery.js
```diff
@@ -39,6 +39,7 @@ walkSync("./contracts/").forEach(contractName => {
       "contracts/ColonyAuthority.sol",
       "contracts/ColonyNetworkAuthority.sol",
       "contracts/ColonyNetworkStorage.sol",
+      "contracts/ColonyRoles.sol",
```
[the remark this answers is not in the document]
kronosapiens: Good point!

contracts/ColonyNetwork.sol
```diff
@@ -261,6 +266,16 @@ contract ColonyNetwork is ColonyNetworkStorage {
     return skill.children[_childSkillIndex];
   }
 
+  function getChildSkillIndex(uint _skillId, uint _childSkillId) public view returns (uint256) {
```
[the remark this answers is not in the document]
kronosapiens: It's a convenience function to make it easier for clients to figure out the reputation proofs. Since it's `view` gas isn't an issue. Also it gets used in this helper function: https://github.com/JoinColony/colonyNetwork/pull/553/files#diff-cfe0340d8acf7ab0683cf395091cd627R703

It's not 100% necessary but I expect it'll be convenient to have around, especially if we want to write more involved domain tests. Having to manually keep track of all the domains and their child skill arrays seems not fun.

scripts/check-auth.js
kronosapiens: We definitely could, but is it necessary? Seems like it would be more code for the same result.

contracts/ColonyRoles.sol
kronosapiens: Mm I think it's quite risky to have an implicit cap on domains -- I can imagine many colonies who will experiment with domain structures and then "run out" of domains. Imagine a large colony who wants to grow but then realizes it's "stuck". That seems like a bad outcome. And we can have unlimited domains with this implementation, and it's not like it's a huge additional cost (esp. since there's no cost for a domain with no explicit permissions).

kronosapiens:
@elenadimitrova I can make `startNextRewardPayout` public, I was just following the existing permissions pattern.

contracts/ColonyAuthority.sol
```diff
@@ -17,64 +17,81 @@
...
+    setRootCapability(colony, "upgrade(uint256)");
+    setRootCapability(colony, "mintTokens(uint256)");
+    setRootCapability(colony, "addGlobalSkill(uint256)");
+    setRootCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
   }
 
-  function setFounderRoleCapability(address colony, bytes memory sig) private {
-    bytes4 functionSig = bytes4(keccak256(sig));
-    setRoleCapability(founderRole, colony, functionSig, true);
+  function setAdministrationCapability(address colony, bytes memory sig) private {
```
elenadimitrova: Seeing that the `colony` address is passed in all these functions yet we have it in the constructor, and considering the `ColonyAuthority` contract only ever works with a single colony, we can set the colony address as a property of the authority and reuse it in permissions management functions. It'll simplify the interface somewhat.

kronosapiens: That's fair, this contract definitely has some boilerplate we can remove.

scripts/check-recovery.js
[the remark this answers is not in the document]
kronosapiens: Except wait, these functions are not part of the IColony interface, they are called internally by other Colony functions. So perhaps they don't?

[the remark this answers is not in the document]
elenadimitrova: oh if that's the case they don't need to be but that script should skip all non public functions too right?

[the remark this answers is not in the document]
kronosapiens: Hmm that seems right to me.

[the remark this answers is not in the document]
kronosapiens: We're already doing that though: https://github.com/JoinColony/colonyNetwork/blob/develop/scripts/check-recovery.js#L15

elenadimitrova:
Just a few more clarifications I need. The Root permission is not a Domain permission right? Looking at the implementation it is authorised via `auth` rather than `authD`, can we make that clear in the PR description as currently it is listed alongside the other domain permissions. Also to your first open question on this, my personal preference is that Roles are ordered in somewhat decreasing order of permissions power: 0-Recovery, 1-Root, 2-Arbitration, 3-Architecture, 4-Administration, 5-Funding.

kronosapiens:
Will update the PR description regarding Root, and I like the reorganization you're proposing, although I would put `Administration` in last place since it can't control resources. And I'll take a look at `auth` in ReputationMiningCycleStorage.

Regarding the `Root` role and functionality, I was following the current usage of the `Founder` role. The current `Founder` role can manage network-wide state via the Meta Colony, as well as managing Colony-specific state. So this doesn't really change anything, it just depends whether `Root` is held in the Meta Colony or another Colony.

contracts/DomainRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+    return setUserRole(who, 1, role, enabled);
+  }
+
+  function hasUserRole(address who, uint8 role) public view returns (bool) {
+    return hasUserRole(who, 1, role);
+  }
+
+  function canCall(address caller, address code, bytes4 sig) public view returns (bool) {
+    return canCall(caller, 1, code, sig);
+  }
```
elenadimitrova: For the above 3 functions can we not override the `DSAuth` implementations but instead not have these functions at all and call directly the implementation` e.g `hasUserRole(who, 1, role)`

contracts/IColony.sol
```diff
@@ -233,10 +247,13 @@ contract IColony is ColonyDataTypes, IRecovery {
   /// @notice Assigning manager role
   /// Current manager and user we want to assign role to both need to agree
   /// User we want to set here also needs to be an admin
+  /// Note that the domain proof data comes at the end here to not interfere with the assembly argument unpacking
   /// @dev This function can only be called through `executeTaskRoleAssignment`
   /// @param _id Id of the task
   /// @param _user Address of the user we want to give a manager role to
-  function setTaskManagerRole(uint256 _id, address payable _user) public;
+  /// @param _parentDomainId The domain ID in which _user has the Administration permission
```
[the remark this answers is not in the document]
kronosapiens: Agree

contracts/ColonyRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+
+import "../lib/dappsys/roles.sol";
+
+
+contract ColonyRoles is DSRoles {
+  mapping(address=>mapping(uint256=>bytes32)) _user_roles;
+
+  // New function signatures taking arbitrary domains
+
+  function getUserRoles(address who, uint256 where) public view returns (bytes32) {
```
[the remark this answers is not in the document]
kronosapiens: I guess I liked the thematic consistency with the `who` that DSRoles uses. I'll swap `domainId` to `where` for the other functions so its consistent with the other functions.

contracts/ColonyRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.4.23;
+
+import "../lib/dappsys/roles.sol";
+
+
+contract ColonyRoles is DSRoles {
```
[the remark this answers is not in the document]
kronosapiens: Done!

contracts/ColonyRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+  The Colony Network is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.4.23;
```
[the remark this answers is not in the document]
kronosapiens: Good to know, I'll double check.

contracts/ColonyStorage.sol
```diff
@@ -125,13 +125,50 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+    if (canCallBecauseArchitect(msg.sender, _parentDomainId, msg.sig)) {
+      require(_parentDomainId != _childDomainId, "ds-auth-only-authorized-in-child-domain");
+    }
+    _;
+  }
+
+  modifier auth() {
+    require(isAuthorized(msg.sender, msg.sig), "ds-auth-unauthorized");
+    _;
+  }
```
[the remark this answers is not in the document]
kronosapiens: Hmm, I assumed we had to copy `auth` here since we override `isAuthorized`, but maybe that's not the case.

I think keeping `auth` is good, since it's more clear in the case of globally-`auth`ed functions like the `root` and `recovery` functions. For those we'd have to do something like `authDomain(1, 0, 0)` and not actually pass in any arguments, which to my mind is more confusing than just `auth`ing those functions. I think the distinction that `auth` refers to global permissions vs `authDomain` which refers to domains helps communicate whats going on.

contracts/DomainRoles.sol
kronosapiens: I'll double-check. The reason we did the override was otherwise if the functions were called, they would fail since they would be accessing the mapping incorrectly. But if we only call them internally then I suppose I can clear them out.

contracts/ColonyRoles.sol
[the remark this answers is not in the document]
elenadimitrova: Sure that's fine, I guess the contract is called `DomainRoles` so the _where_ should be deducible from there.

contracts/ColonyStorage.sol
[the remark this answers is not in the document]
elenadimitrova: But in the end `auth` here is in fact referring to domain permissions where the domain is the root domain id 1 right?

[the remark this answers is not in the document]
kronosapiens: Yup, but transparently, and only in ColonyStorage do you realize that. From the perspective of the other contracts that doesn't matter, which I think is valuable.

contracts/IColony.sol
```diff
@@ -55,28 +55,32 @@ contract IColony is ColonyDataTypes, IRecovery {
...
+  function setFundingRole(uint256 _parentDomainId, uint256 _domainProofIndex, address _user, uint256 _domainId, bool _setTo) public;
 
-  /// @notice Remove colony admin.
-  /// Can only be called by founder role.
-  /// @param _user User we want to remove admin role from
-  function removeAdminRole(address _user) public;
+  /// @notice Set new colony architecture role.
+  /// Can be called by root role or architecture role.
+  /// @param _user User we want to give an architecture role to
+  function setArchitectureRole(uint256 _parentDomainId, uint256 _domainProofIndex, address _user, uint256 _domainId, bool _setTo) public;
```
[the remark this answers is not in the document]
kronosapiens: Fixed

contracts/IColony.sol
```diff
@@ -55,28 +55,32 @@ contract IColony is ColonyDataTypes, IRecovery {
...
-  /// Can be called by founder role or admin role.
+  /// Can be called by root role or architecture role.
   /// @param _user User we want to give an admin role to
-  function setAdminRole(address _user) public;
+  function setAdministrationRole(uint256 _parentDomainId, uint256 _domainProofIndex, address _user, uint256 _domainId, bool _setTo) public;
+
+  /// @notice Set new colony funding role.
+  /// Can be called by root role or architecture role.
+  /// @param _user User we want to give an funding role to
+  function setFundingRole(uint256 _parentDomainId, uint256 _domainProofIndex, address _user, uint256 _domainId, bool _setTo) public;
```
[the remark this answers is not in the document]
kronosapiens: Fixed

contracts/IColony.sol
```diff
@@ -55,28 +55,32 @@ contract IColony is ColonyDataTypes, IRecovery {
...
-  /// @dev There can only be one address assigned to founder role at a time.
-  /// Whoever calls this function will lose their founder role
-  /// Can be called by founder role.
-  /// @param _user User we want to give an founder role to
-  function setFounderRole(address _user) public;
-
   /// @notice Set new colony admin role.
-  /// Can be called by founder role or admin role.
+  /// Can be called by root role or architecture role.
   /// @param _user User we want to give an admin role to
```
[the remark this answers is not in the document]
kronosapiens: Fixed

contracts/IColony.sol
```diff
@@ -55,28 +55,32 @@ contract IColony is ColonyDataTypes, IRecovery {
...
-  /// @param _user User we want to remove admin role from
-  function removeAdminRole(address _user) public;
+  /// @notice Set new colony architecture role.
+  /// Can be called by root role or architecture role.
+  /// @param _user User we want to give an architecture role to
+  function setArchitectureRole(uint256 _parentDomainId, uint256 _domainProofIndex, address _user, uint256 _domainId, bool _setTo) public;
+
+  /// @notice Set new colony root role.
+  /// Can be called by root role only.
+  /// @param _user User we want to give an root role to
```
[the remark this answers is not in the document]
kronosapiens: Fixed

contracts/ColonyFunding.sol
```diff
@@ -137,9 +137,19 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
+    uint256 _fromPot,
+    uint256 _toPot,
+    uint256 _amount,
+    address _token
+  )
+  public
   stoppable
-  auth
+  authDomain(_parentDomainId, getDomainFromFundingPot(_fromPot), _fromDomainProofIndex)
+  authDomain(_parentDomainId, getDomainFromFundingPot(_toPot), _toDomainProofIndex)
```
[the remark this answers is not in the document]
kronosapiens: Agreed

contracts/ColonyStorage.sol
```diff
@@ -125,13 +125,50 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+
+  function isValidDomainProof(uint256 parentDomainId, uint256 childDomainId, uint256 childIndex) internal view returns (bool) {
+    return (parentDomainId == childDomainId) || validateDomainProof(parentDomainId, childDomainId, childIndex);
+  }
+
+  function validateDomainProof(uint256 parentDomainId, uint256 childDomainId, uint256 childIndex) internal view returns (bool) {
```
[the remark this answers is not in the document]
kronosapiens: Yup merged, I figured out a cleaner way to approach that logic. And added a comment, and reordered the arguments (I agree they're better that way). I think that `validateDomainProof` is fine though, it's a little more abstract but I don't think it's a huge problem.

contracts/ColonyRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+
+import "../lib/dappsys/roles.sol";
+
+
+contract ColonyRoles is DSRoles {
+  mapping(address=>mapping(uint256=>bytes32)) _user_roles;
```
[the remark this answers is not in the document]
kronosapiens: The only thing is that I'd like to keep it consistent with the rest of `DSRoles` which uses underscores, without changing anything in `roles.auth`.

contracts/DomainRoles.sol
kronosapiens: So I updated everything else but this one is giving me trouble. If I remove the `canCall` override, then the contract recovery tests break. I'm not 100% sure of how the inheritance is working there since the contract recovery code inherits from DSAuth but not DomainRoles so I think the `auth` in those contracts is perhaps calling the original `canCall` which doesn't know about domains and so is breaking. I'm not quite sure yet though, still looking into it.

contracts/ColonyStorage.sol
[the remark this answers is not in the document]
kronosapiens: However I was able to remove the override of `isAuthorized` which I noticed was unnecessary.

contracts/DomainRoles.sol
kronosapiens: Ok, I think I figured it out. `ContractRecovery` does not have the `auth` and `isAuthorized` overrides in `ColonyStorage` (they are siblings in the inheritance tree). So when we call a function like `enterRecoveryMode`, we call the original `auth` in `DSAuth`, which calls into the original `isAuthorized`, which calls `authority.canCall(src, address(this), sig);`, the version without the `domainId`.

The `authority` here is `ColonyAuthority`, which knows about domains, but if we omit the explicit override of `auth` or `isAuthorized` in `DomainRoles`, then we wind up calling the version in `DSRoles` which doesn't know about domains, leading to the auth check failing.

To get around this, we'd either have to override `DSAuth` itself which seems like a bummer, or override the `isAuthorized` function in `ContractRecovery`, which seems like poor separation of concerns (since there's no reason why recovery code which only deals with Colony-wide roles should be aware of domains). So between these three options (override `DSAuth`, override `isAuthorized` in `ContractRecovery`, or put an override in `DomainRoles`), the third options seems like the best since it keeps the domains logic together.

elenadimitrova: We had a conversation with @area and @jakzilla earlier in the week around recovery roles (which is the only reason we inherit off `DSRoles` atm). I think there was a high level agreement we can decommission those in favour of the Root permission. What do you think?

elenadimitrova: I am trying to understand the different "colony-wide" roles implementation, where for the Root role we do:
```
 setUserRole(who, 1, ROOT_ROLE, true);
```
while for the Recovery role we do 
```
setUserRole(who, RECOVERY_ROLE, true);
```
meaning that the Root role is held within the `DomainRoles` while the Recovery is help within DSRoles. If we are going to define colony-wide permissions as those Root permissions held in domain 1 then we should do the same for Recovery role. I stand to be corrected if I misunderstood through.

kronosapiens: Hmm, insinctively I feel like having the `recovery` roles in addition to `root` is a good idea -- we may want some people to be able to help fix recovery state without giving them control over day-to-day operations of the colony. Users with `root` permissions can act unilaterally, whereas users with `recovery` permissions, although powerful in that they can change arbitrary state, also cannot finalize those changes without support from other recovery roles. So in that sense they're different and I don't think it would be great to combine them.

And for setting the `root` role, in this line we don't reference the domain at all: https://github.com/JoinColony/colonyNetwork/pull/553/files#diff-669f459dea66cccc363b550a82126c97R32

What example are you thinking of?

elenadimitrova: I meant that the _desing_ of how the colony-wide roles is implemented is different and should be made the same, not that the roles themselves should be merged

kronosapiens: Mm I am not sure that is correct. In `ContractRecovery.sol` we have:

```
CommonAuthority(address(authority)).setUserRole(_user, RECOVERY_ROLE, true);
```

While in `Colony.sol` we have:

```
ColonyAuthority(address(authority)).setUserRole(_user, uint8(ColonyRole.Root), _setTo);
```

Both `CommonAuthority` and `ColonyAuthority` inherit from `DomainRoles` so both of those calls reference the same code, and by extension both make use (implicitly) of the domain mapping.

contracts/CommonAuthority.sol
```diff
@@ -17,11 +17,12 @@
 
 pragma solidity >=0.5.3;
 
-import "../lib/dappsys/roles.sol";
+import "./ColonyDataTypes.sol";
```
elenadimitrova: Just noticed this import isn't used. We can remove.

contracts/DomainRoles.sol
elenadimitrova: Ah yes, you're right, so basically colony-wide permissions are implemented as domain permissions for the root domain?

kronosapiens: Exactly. As every `authority` contract inherits from `CommonAuthority`, then every `authority` has the following 2-level mapping:

```
  mapping(address=>mapping(uint256=>bytes32)) _user_roles;
```

So "colony-wide" permissions are implemented by transparently passing a 1 as the domain.

elenadimitrova: You mean `auth` as in your override of the `DSAuth` modifier [here](https://github.com/JoinColony/colonyNetwork/pull/553/files?file-filters%5B%5D=.js&file-filters%5B%5D=.sol#diff-f4e9c4137ebcc35fead90e9b07c10f4dR139)?
I do still worry about these overrides and I'm on the fence about implementing "colony-wide" permissions this way vs just using the existing `DSRoles` and `DSAuth` functionality. Did you have any particular arguments that leaned you towards one more than the other?

kronosapiens: Yup, that's the one. I'm curious as to your reservation. It seems to me that maintaining one `_user_roles` mapping is better than maintaining two, especially when the gas costs are equivalent (i.e. we don't pay any gas for having the additional layer in the mapping for colony-wide roles).

I guess if the choice is more functions, fewer storage locations and fewer functions, more storage locations I would choose the former since functions are easier to change than storage locations. But once I finish this rebase I can try removing the overrides and see if everything breaks.

contracts/ColonyStorage.sol
```diff
@@ -125,13 +125,46 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+  function validateDomainProof(uint256 permissionDomainId, uint256 childSkillIndex, uint256 childDomainId) internal view returns (bool) {
+    return (permissionDomainId == childDomainId) || (getChildSkillId(permissionDomainId, childSkillIndex) == domains[childDomainId].skillId);
+  }
+
+  function getChildSkillId(uint256 permissionDomainId, uint256 childSkillIndex) internal view returns (uint256) {
+    return IColonyNetwork(colonyNetworkAddress).getChildSkillId(domains[permissionDomainId].skillId, childSkillIndex);
+  }
+
+  function canCallBecauseArchitect(address src, uint256 domainId, bytes4 sig) internal view returns (bool) {
+    return DomainRoles(address(authority)).canCallBecause(src, domainId, uint8(ColonyRole.ArchitectureSubdomain), address(this), sig);
```
elenadimitrova: I haven't been able to explain why this returns `false` for the founder? The founder belonging to the `ArchitectureSubdomain` role is explicitly tested [here](https://github.com/JoinColony/colonyNetwork/pull/553/files#diff-c255652ac9460b92a219642618a4887cR65) yet if you try 
```
const canCallMakeTaskBecauseArchitect = await colonyAuthority.canCallBecause(FOUNDER, 1, ARCHITECTURE_ROLE, colony.address, getFunctionSignature("makeTask(uint256,uint256,bytes32,uint256,uint256,uint256)"));
``` 
in that test it returns `false` 🤔

kronosapiens: Mm, so this function checks to see if it is _only_ the architecture permission that gives this role. I.e. if you have root and architecture funding then it returns `false`, but if you have only architecture permission it returns `true`. That was how we implemented the restriction on the architecture role where you can only call certain functions in a _subdomain_ of your permission,  i.e. if you can call the role because you're an architect (i.e. not root, or even an architect in a higher domain) then we check that you are calling the function in a child domain.

elenadimitrova: Ah okay I see the different bitwise operation going on in `canCallBecause`. Maybe better to rename that function or even better, merge it into the `canCallBecauseArchitect` and rename _that_ one to something like `canCallBecauseOnlyArchitect` or something better?

kronosapiens: I think the name is ok -- `canCallBecause` tells you if someone can call a function _because_ they have a particular role (i.e. they aren't getting the permission elsewhere). I also think its cleaner to keep the `canCallBecause` in `DomainRoles` since it's accessing the underlying mappings and doing similar bit operations as the other functions there, and to keep it general since there's no reason to hardcode a particular role into `DomainRoles`. But an expository comment would be appropriate since it's not clear why we need to be doing this check at all for architect.

contracts/DomainRoles.sol
elenadimitrova: I made a [branch](https://github.com/JoinColony/colonyNetwork/compare/feature/domain-permissions...experiment/domain-permissions) with the other option we discuss here for clarity.

contracts/ColonyStorage.sol
elenadimitrova: I feel pretty strongly about this one actually and would want something renamed. I don't feel adding a comment will suffice. `canCallBecause` only half means your definition 

>someone can call a function because they have a particular role

The last part about the role is not obvious in that name. 
Plus it is only ever called once. I feel there are way too many functions in `ColonyStorage` serving `authDomain` and I understand some maybe justifiable because of the stack depth. But the following are not called anywhere else but that modifier and I'd argue for merging some of that logic:
- `isAuthorized`
- `validateDomainProof`
- `getChildSkillId` - that especially feel that should go
- `canCallBecauseArchitect` and `canCallBecause`

kronosapiens: Mm, those are reasonable points. I split them up that way largely to keep each function clean -- putting them all together will entail a big `if/else` statement and wont be as nice. I do think it's important to avoid creating local variables as we don't want to create problems for people writing functions which use these modifiers. But I also recognize that having functions which aren't used more than once creates its own confusion. I'll see if I can roll some of them together. I do think that `canCallBecause` belongs in `DomainRoles` though.

elenadimitrova: I don't disagree with keeping `canCallBecause` in `DomainRoles`. I just disagree with its name, maybe `canCallBecauseHasSingleRoleOf` or something nicer you can come up with.

contracts/DomainRoles.sol
elenadimitrova: Just discussed this with @area as I was honestly on the fence about this one. His comment below:
>I think I prefer `DomainRoles` for everything. The reason being if we straddle, it seems very arbitrary that if you wanted to give someone e.g. funding permission globally, you need to give it to them in domain 1 in `DomainRoles` rather than giving them that role in `DSRoles`
(I understand why - because funding can be given not-globally - but I'm thinking of the perspective of someone new approaching the codebase)

Which leans into the current implementation. So what I'll just try to do to give myself a peace of mind is experiment with exploits I suspect exist in the current solution and if I find none, we'll just go with this.

contracts/ColonyStorage.sol
kronosapiens: Fair

kronosapiens: So I'll update the name to `canCallOnlyBecauseArchitect`, hopefully that's more helpful. With regards to the consolidating functions, I'll point out that this will involve _more_ lines of code. Consider:

```
  function validateDomainProof(uint256 permissionDomainId, uint256 childSkillIndex, uint256 childDomainId) internal view returns (bool) {
    if (permissionDomainId == childDomainId) {
      return true;
    } else {
      uint256 childSkillId = IColonyNetwork(colonyNetworkAddress).getChildSkillId(domains[permissionDomainId].skillId, childSkillIndex);
      return childSkillId == domains[childDomainId].skillId;
    }
  }
```

vs. the original

```
   function validateDomainProof(uint256 permissionDomainId, uint256 childSkillIndex, uint256 childDomainId) internal view returns (bool) {
    return (permissionDomainId == childDomainId) || (getChildSkillId(permissionDomainId, childSkillIndex) == domains[childDomainId].skillId);
  }

  function getChildSkillId(uint256 permissionDomainId, uint256 childSkillIndex) internal view returns (uint256) {
    return IColonyNetwork(colonyNetworkAddress).getChildSkillId(domains[permissionDomainId].skillId, childSkillIndex);
   }
```

8 lines vs 6, so two extra lines for the privilege of combining functions. I'll do it if you _really_ want but I don't think having the extra functions are a big deal and IMO they make this section easier to read.

elenadimitrova: On the point of merging `getChildSkillId` in `validateDomainProof`, I was thinking of this end result:
In `authDomain`:
```
if (permissionDomainId != childDomainId) {
      require(validateDomainProof(_permissionDomainId, _childSkillIndex, _childDomainId), "ds-auth-invalid-domain-proof");
    }
```

as really the `validateDomainProof` check is only applicable to when we need to prove the child-parent relationship. 

```
function validateDomainProof(uint256 permissionDomainId, uint256 childSkillIndex, uint256 childDomainId) private view returns (bool) {
    uint256 childSkillId = IColonyNetwork(colonyNetworkAddress).getChildSkillId(domains[permissionDomainId].skillId, childSkillIndex);
    return  (childSkillId == domains[childDomainId].skillId);
  }
```
Which is 6 lines of code.

Also a bit on the naming (sorry I think I left a note on that somewhere but maybe didn't? As in that "proof" is used in Merkle tree proofs in reputation mining and shouldn't be used here) `validateDomainProof` -> `validateDomainInheritance` (Open to better names here)

`isAuthorized` I'm okay to keep separate, as it's reused in 2 places and as you say matches the `DSAuth` design.

For `canCallBecauseArchitect`, I prefer your suggestion of `canCallOnlyBecauseArchitect`.

contracts/ColonyFunding.sol
```diff
@@ -210,7 +213,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
   }
 
   function startNextRewardPayout(address _token, bytes memory key, bytes memory value, uint256 branchMask, bytes32[] memory siblings)
-  public auth stoppable
+  public stoppable auth
```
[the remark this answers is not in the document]
kronosapiens: Little cleanups as you go along, that's how you do it.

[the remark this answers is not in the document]
kronosapiens: Just reduce the entropy... bit by bit. 😂

contracts/ColonyStorage.sol
kronosapiens: Mm that's how I did the domain proof the first time around, but I combined it into a single function on the grounds that those two pieces of logic always go together, plus I liked the idea of `authDomain` being mostly just one-line requires. But I can change it back.

And fair, I can reserve the term `proof` for Merkle proofs (although this is a ~kind of proof).

contracts/ColonyNetworkENS.sol
```diff
@@ -43,7 +43,7 @@ contract ColonyNetworkENS is ColonyNetworkStorage {
       interfaceID == ADDR_INTERFACE_ID );
   }
 
-  function setupRegistrar(address _ens, bytes32 _rootNode) public auth stoppable {
+  function setupRegistrar(address _ens, bytes32 _rootNode) public stoppable auth {
```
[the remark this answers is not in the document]
kronosapiens: No, I just noticed that we usually put the recovery modifier before the auth modifier so I figured why not be consistent.

contracts/ColonyFunding.sol
```diff
@@ -381,18 +384,16 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
+    uint currentBalance = fundingPots[task.fundingPotId].balance[_token];
+    uint newBalance = add(currentBalance, _amount);
 
-    if (_prev >= totalTokenPayout) {                  // If the old amount in the pot was enough to pay for the budget
-      if (tokenPot < totalTokenPayout) {              // And the new amount in the pot is not enough to pay for the budget...
-        task.payoutsWeCannotMake += 1;                // Then this is a set of payouts we cannot make that we could before.
-      }
-    } else {                                          // If this 'else' is running, then the old amount in the pot could not pay for the budget
-      if (tokenPot >= totalTokenPayout) {             // And the new amount in the pot can pay for the budget
+    // It is not possible to defund a pot below the budget, so we don't check for that here.
```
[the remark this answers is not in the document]
kronosapiens: Mm the main argument I'd make for removing is that it reduces code coverage.

kronosapiens:
@area To your point about admin/founder permission, I would suggest we re-introduce the add/remove admin/founder functions and then implement them in terms of these new permissions. We can remove them in a future release. I didn't want to do that in this PR but I can make a second one after this one is done.

contracts/extensions/OneTxPayment.sol
```diff
@@ -24,32 +24,45 @@ import "./../ColonyAuthority.sol";
...
     uint256 _amount,
     uint256 _domainId,
     uint256 _skillId) public
   {
-    IColony colony = IColony(_colony);
-
     // Check caller is able to call makePayment on the colony
     require(
-      ColonyAuthority(colony.authority()).canCall(msg.sender, _domainId, _colony, ADD_PAYMENT_SIG),
+      ColonyAuthority(colony.authority()).canCall(msg.sender, _permissionDomainId, address(colony), ADD_PAYMENT_SIG),
```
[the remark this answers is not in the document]
kronosapiens: Hmm good point, let me think about how to do this better.

contracts/Colony.sol
```diff
@@ -28,30 +28,56 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   // Version number should be upped with every change in Colony or its dependency contracts or libraries.
   function version() public pure returns (uint256 colonyVersion) { return 1; }
 
-  function setFounderRole(address _user) public stoppable auth {
-    // To allow only one address to have founder role at a time, we have to remove current founder from their role
+  function setRootRole(address _user, bool _setTo) public stoppable auth {
```
elenadimitrova: Do we worry about `DomainRoles` Root role confused with `DSRoles` Root role?

kronosapiens: Mm I suppose slightly, we could re-use the `Founder` terminology if we prefer.

area:
> To your point about admin/founder permission, I would suggest we re-introduce the add/remove admin/founder functions and then implement them in terms of these new permissions. We can remove them in a future release. I didn't want to do that in this PR but I can make a second one after this one is done.

I'm not very enthusiastic about this, because I think there's a very good chance they would never be removed.

contracts/ColonyFunding.sol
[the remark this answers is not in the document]
area: Code coverage is a tool!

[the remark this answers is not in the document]
kronosapiens: I mean yes and no, when your build fails because of code coverage then it's a spec.

kronosapiens:
Hmm, that's a valid concern I think, but using an external contract would require being able to invoke arbitrary transactions via that contract (i.e. passing `callData` to the extension which then passes it to the Colony contracts), which I expect would create a lot of work for the Dapp team.

contracts/ColonyFunding.sol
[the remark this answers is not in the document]
kronosapiens: Although I'm realizing that after merging in Payments that invariant no longer holds since we _can_ defund payments below the payout. So much ado about nothing.

helpers/test-data-generator.js
```diff
@@ -178,7 +177,9 @@ export async function setupFundedTask({
   const evaluatorPayoutBN = new BN(evaluatorPayout);
   const workerPayoutBN = new BN(workerPayout);
   const totalPayouts = managerPayoutBN.add(workerPayoutBN).add(evaluatorPayoutBN);
-  await colony.moveFundsBetweenPots(1, fundingPotId, totalPayouts, tokenAddress);
+  await colony.setFundingRole(1, 0, manager, 1, true);
+  // TODO: extend this to allow funding pots beyond the first child of root domain... :(
```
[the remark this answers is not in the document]
kronosapiens: Saw. Although this line is more to do with passing in the domain proof to the helper function.

contracts/ColonyStorage.sol
```diff
@@ -163,13 +158,54 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+    require(domainExists(_permissionDomainId), "ds-auth-permission-domain-does-not-exist");
+    require(domainExists(_childDomainId), "ds-auth-child-domain-does-not-exist");
+    require(isAuthorized(msg.sender, _permissionDomainId, msg.sig), "ds-auth-unauthorized");
+    if (_permissionDomainId != _childDomainId) {
+      require(validateDomainInheritance(_permissionDomainId, _childSkillIndex, _childDomainId), "ds-auth-invalid-domain-inheritence");
+    }
+    if (canCallOnlyBecauseArchitect(msg.sender, _permissionDomainId, msg.sig)) {
+      require(_permissionDomainId != _childDomainId, "ds-auth-only-authorized-in-child-domain");
+    }
+    _;
```
elenadimitrova: Lovely

contracts/ColonyStorage.sol
```diff
@@ -163,13 +158,54 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
...
+  }
+
+  // Evaluates a "domain proof" which checks that childDomainId is part of the subtree starting at permissionDomainId
+  function validateDomainInheritance(uint256 permissionDomainId, uint256 childSkillIndex, uint256 childDomainId) internal view returns (bool) {
+    uint256 childSkillId = IColonyNetwork(colonyNetworkAddress).getChildSkillId(domains[permissionDomainId].skillId, childSkillIndex);
+    return childSkillId == domains[childDomainId].skillId;
+  }
```
elenadimitrova: Very lovely

contracts/ColonyTask.sol
```diff
@@ -298,10 +304,10 @@ contract ColonyTask is ColonyStorage {
     return taskWorkRatings[_id].secret[_role];
   }
 
-  function setTaskManagerRole(uint256 _id, address payable _user) public
+  function setTaskManagerRole(uint256 _id, address payable _user, uint256 _permissionDomainId, uint256 _childSkillIndex) public
```
elenadimitrova: Just on ordering of parameters. In most places you start off with the domain permission input params but here they are at the end. I don't mind which way it's done, as long as it's consistent.

contracts/ColonyTask.sol
```diff
@@ -332,10 +338,11 @@ contract ColonyTask is ColonyStorage {
   function setTaskDomain(uint256 _id, uint256 _domainId) public
   stoppable
   taskExists(_id)
-  domainExists(_domainId)
   taskNotComplete(_id)
   self()
   {
+    require(domainExists(_domainId), "colony-domain-does-not-exist");
```
elenadimitrova: 🤔 and you removed the modifier because?

contracts/DomainRoles.sol
```diff
@@ -0,0 +1,76 @@
...
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.5.3;
+
+import "../lib/dappsys/roles.sol";
+
+
+contract DomainRoles is DSRoles {
+  mapping(address=>mapping(uint256=>bytes32)) _user_domain_roles;
```
elenadimitrova: Most lovely

contracts/IRecovery.sol
```diff
@@ -38,17 +38,17 @@ contract IRecovery {
   function isInRecoveryMode() public view returns (bool inRecoveryMode);
 
   /// @notice Set new colony recovery role.
-  /// Can be called by founder.
```
elenadimitrova: As a bit of a cleanup and terminology consistency, can you replace all "founder" instances in tests with "root". My search found 36 instances which all looked straight forward search/replace.

contracts/ColonyAuthority.sol
```diff
@@ -17,85 +17,67 @@
...
+  uint8 constant ROOT_ROLE = uint8(ColonyDataTypes.ColonyRole.Root);
+
+  address colony;
+
+  constructor(address _colony) public CommonAuthority(_colony) {
+    colony = _colony;
+
+    // Add permissions for the Administration role
+    addRoleCapability(ADMINISTRATION_ROLE, "makeTask(uint256,uint256,bytes32,uint256,uint256,uint256)");
+    // TODO: add domain auth
```
elenadimitrova: Is this `todo` still valid?

package.json
```diff
@@ -7,8 +7,10 @@
     "eslint-staged": "bash ./scripts/eslint.sh",
     "eslint": "eslint .",
     "solium": "solium --dir .",
+    "notify:osx": "osascript -e 'display notification \"Ding!\"'",
     "check:storagevars": "./node_modules/@babel/node/bin/babel-node.js ./scripts/check-storage.js",
     "check:recoverymods": "./node_modules/@babel/node/bin/babel-node.js ./scripts/check-recovery.js",
+    "check:auth": "./node_modules/@babel/node/bin/babel-node.js ./scripts/check-auth.js",
```
elenadimitrova: This should be run in CI

test-upgrade/colony-upgrade.js
```diff
@@ -104,7 +104,8 @@ contract("Colony contract upgrade", accounts => {
     });
 
     it("should return correct permissions", async function() {
-      const founder = await colony.hasUserRole(ACCOUNT_ONE, 0);
+      const ROOT_ROLE = 6;
```
elenadimitrova: Can we refactor out the roles in `constants.js` for clarity

test/colony-permissions.js
```diff
@@ -0,0 +1,263 @@
...
+  const USER1 = accounts[1];
+  const USER2 = accounts[2];
+
+  // Role 0 is the recovery role
+  const ROOT_ROLE = 1;
+  // const ARBITRATION_ROLE = 2;
+  const ARCHITECTURE_ROLE = 3;
+  const ARCHITECTURE_SUBDOMAIN_ROLE = 4;
+  const FUNDING_ROLE = 5;
+  const ADMINISTRATION_ROLE = 6;
```
elenadimitrova: Can we refactor out the roles in `constants.js` for clarity

contracts/ColonyTask.sol
kronosapiens: It was redundant in all the other cases because `authDomain` checks if the domain exists. Since there was only one existing `domainExists` check which wasn't preceded by an `authDomain` check (in this function), I figured make `domainExists` into a function, use it in `authDomain`, and then throw it into a `require` this one time.

contracts/ColonyAuthority.sol
kronosapiens: Nope! Nice catch.

contracts/IRecovery.sol
kronosapiens: Yes, but the one distinction I'd make between a "founder" as the account which creates the colony vs. the "founder" role. I think it's fine to have an account called "founder", which may have multiple roles (including `root`). But I've updated the references to a `founder` _role_ to refer to the `root` role.

elenadimitrova: Do we have a "founder" role still?

kronosapiens: Nope, which is why I've removed the references to a `founder` _role_.

kronosapiens: However, I have _re-implemented_ them in terms of the other roles for the Glider release, for the reason that the Dapp has not finished designing for the new roles. Those functions are deprecated though and we should strive to remove them as quickly as possible.

kronosapiens:
Two notes:

1. I've re-introduced the old `function setFounderRole(address _user)` and `function setAdminRole(address _user, bool _setTo)`, as the Dapp will not be ready to support the new permissions at launch. These "roles" are implemented in terms of the new roles, however, and so we can remove them easily. We should make sure to consistently message that these roles are temporary.

[commit](https://github.com/JoinColony/colonyNetwork/pull/553/commits/22fd998cbd5e26fc7146cf56d1e94add9c017c36)

2. I've update the OneTxPayment contract to allow the contract and caller to hold their permissions in different domains (so two proofs: one for contract and one for caller). I assume that each entity holds both `administration` and `funding` permission in the same domain, though. It's not the most elegant solution on earth but I think it'll work just fine, especially if we're going to have only root-domain `admin` and `founder` at launch anyway. IMO the better long-term solution is to build some sort of `authority` into the extension contracts themselves, rather than pass through to the Colony authority.

[commit(https://github.com/JoinColony/colonyNetwork/pull/553/commits/89d642b54f6c899cf370d53bdba79487140b68b4)

contracts/extensions/OneTxPayment.sol
```diff
@@ -18,41 +18,82 @@
...
+
+  IColony colony;
+  IColonyNetwork colonyNetwork;
+
+  constructor(address _colony) public {
+    colony = IColony(_colony);
+    colonyNetwork = IColonyNetwork(colony.getColonyNetwork());
+  }
+
+  // Note: assumes that each entity holds administration and funding roles are held in the same domain,
```
elenadimitrova: Repetition of "holds" in description

contracts/ColonyStorage.sol
```diff
@@ -150,11 +145,6 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, DSMath {
 
     // Prevent people moving funds from the pot designated to paying out token holders
     require(_fromPot > 0, "colony-funding-cannot-move-funds-from-rewards-pot");
-
-    // Preventing sending from non-existent funding pots is not strictly necessary (if a pot doesn't exist, it can't have any funds if we
-    // prevent sending to nonexistent funding pots) but doing this check explicitly gives us the error message for clients.
-    require(_fromPot <= fundingPotCount, "colony-funding-from-nonexistent-pot"); // Only allow sending from created pots
-    require(_toPot <= fundingPotCount, "colony-funding-nonexistent-pot"); // Only allow sending to created funding pots
```
elenadimitrova: Why were these removed?

kronosapiens: That logic was moved to https://github.com/JoinColony/colonyNetwork/pull/553/files#diff-8920cdf7ab56c438c907df9bd37d2931R471

PR #555, opened 2019-02-21
Payments implementation

### LINKED ISSUE #527: Add support for 3tx Payments

elenadimitrova:
Closes #527 

The `Payment` flow (aka 4tx flow) is:
1. Create new payment: 
`addPayment(address recipient,  address token,  uint256 amount, uint256 domainId,  uint256 skillId)` 
2. Fund the payment (same as for Tasks using `moveFundsBetweenPots`).
3. Finalize the payment. When the payment becomes fully funded admins can call `finalizePayment` which marks it finalised and awards reputation.
3. Claim the payment via `claimPayment` (can be called by anyone on behalf of the recipient)

Amending the different elements of a payment can be done via: `setPaymentRecipient`, `setPaymentDomain`, `setPaymentSkill` or `setPayout`. These functions are restricted to the colony admins as is creating a payment. Note that once created any admin can manage any payment in the colony. Payments cannot be amended once finalized.

Note that even though by default a newly created Payment is 1 token amount for the recipient, there is a possibility to add more tokens in the payment for the same recipient, again similarly to tasks.

For the single transaction flow, I've reused the `OneTxPayment` contract and tests from @area 's [branch for eth-denver](https://github.com/JoinColony/colonyNetwork/tree/temp/eth-denver)

In addition to implementing the requirements for `Payment`s we refactor the following common logic between Tasks and Payments:

- Logic around tracking whether a `Task` or `Payment` is sufficiently funded via the `payoutsWeCannotMake` property. This used to be stored in the `Task` itself however it make more sense to abstract as part of the `FundingPot` model. This way `Payment`s can reuse it. A nice side effect from this is that the iterative function `getTotalTaskPayout` is no longer needed as this is kept as part of the `FundingPot.payouts` information.

- Disbursing the payment logic including deducting network fee, updating the funding pot balances and paying the recipient is consolidated in the `processPayout` private function which is reused across both `Task` and `Payment`

### DIFF
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -66,6 +66,9 @@ contract ColonyAuthority is CommonAuthority {
     setAdminRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
     setFounderRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
     // Update payments
+    setAdminRoleCapability(colony, "finalizePayment(uint256)");
+    setFounderRoleCapability(colony, "finalizePayment(uint256)");
+
     setAdminRoleCapability(colony, "setPaymentRecipient(uint256,address)");
     setFounderRoleCapability(colony, "setPaymentRecipient(uint256,address)");
     
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -78,8 +78,8 @@ contract ColonyAuthority is CommonAuthority {
     setAdminRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
     setFounderRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
 
-    setAdminRoleCapability(colony, "setPayout(uint256,address,uint256)");
-    setFounderRoleCapability(colony, "setPayout(uint256,address,uint256)");
+    setAdminRoleCapability(colony, "setPaymentPayout(uint256,address,uint256)");
+    setFounderRoleCapability(colony, "setPaymentPayout(uint256,address,uint256)");
 
     // Start next reward payout
     setAdminRoleCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -63,8 +63,8 @@ contract ColonyAuthority is CommonAuthority {
     setFounderRoleCapability(colony, "makeTask(bytes32,uint256,uint256,uint256)");
 
     // Add payment
-    setAdminRoleCapability(colony, "addPayment(address,uint256,uint256,address,uint256)");
-    setFounderRoleCapability(colony, "addPayment(address,uint256,uint256,address,uint256)");
+    setAdminRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
+    setFounderRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
 
     // Start next reward payout
     setAdminRoleCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -65,6 +65,18 @@ contract ColonyAuthority is CommonAuthority {
     // Add payment
     setAdminRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
     setFounderRoleCapability(colony, "addPayment(address,address,uint256,uint256,uint256)");
+    // Update payments
+    setAdminRoleCapability(colony, "setPaymentRecipient(uint256,address)");
+    setFounderRoleCapability(colony, "setPaymentRecipient(uint256,address)");
+    
+    setAdminRoleCapability(colony, "setPaymentDomain(uint256,uint256)");
+    setFounderRoleCapability(colony, "setPaymentDomain(uint256,uint256)");
+
+    setAdminRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
+    setFounderRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
+
+    setAdminRoleCapability(colony, "setPayout(uint256,address,uint256)");
+    setFounderRoleCapability(colony, "setPayout(uint256,address,uint256)");
 
     // Start next reward payout
     setAdminRoleCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -57,9 +57,15 @@ contract ColonyAuthority is CommonAuthority {
     // Add domain
     setAdminRoleCapability(colony, "addDomain(uint256)");
     setFounderRoleCapability(colony, "addDomain(uint256)");
+    
     // Add task
     setAdminRoleCapability(colony, "makeTask(bytes32,uint256,uint256,uint256)");
     setFounderRoleCapability(colony, "makeTask(bytes32,uint256,uint256,uint256)");
+
+    // Add payment
+    setAdminRoleCapability(colony, "addPayment(address,uint256,uint256,address,uint256)");
+    setFounderRoleCapability(colony, "addPayment(address,uint256,uint256,address,uint256)");
+
     // Start next reward payout
     setAdminRoleCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
     setFounderRoleCapability(colony, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
--- a/contracts/ColonyDataTypes.sol
+++ b/contracts/ColonyDataTypes.sol
@@ -177,7 +177,7 @@ contract ColonyDataTypes {
 
   struct Payment {
     address recipient;
-    bool claimed;
+    bool finalized;
     uint256 fundingPotId;
     uint256 domainId;
     uint256[] skills;
--- a/contracts/ColonyDataTypes.sol
+++ b/contracts/ColonyDataTypes.sol
@@ -176,7 +176,7 @@ contract ColonyDataTypes {
   }
 
   struct Payment {
-    address recipient;
+    address payable recipient;
     bool finalized;
     uint256 fundingPotId;
     uint256 domainId;
--- a/contracts/ColonyDataTypes.sol
+++ b/contracts/ColonyDataTypes.sol
@@ -240,4 +240,6 @@ contract ColonyDataTypes {
     uint256 skillId;
     uint256 fundingPotId;
   }
+
+  uint256 constant MAX_PAYOUT = 2**254 - 1; // Up to 254 bits to account for sign and payout modifiers.
 }
--- a/contracts/ColonyDataTypes.sol
+++ b/contracts/ColonyDataTypes.sol
@@ -174,7 +174,7 @@ contract ColonyDataTypes {
 
   struct Payment {
     address recipient;
-    uint256 potId;
+    uint256 fundingPotId;
     uint256 domainId;
     uint256[] skills;
     mapping (address => uint256) payouts;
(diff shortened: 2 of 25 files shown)


contracts/ColonyFunding.sol
```diff
@@ -381,54 +403,94 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
 
-    uint currentTotalAmount = getTotalTaskPayout(_id, _token);
-    tasks[_id].payouts[uint8(_role)][_token] = _amount;
+    Task storage task = tasks[_id];
+    FundingPot storage fundingPot = fundingPots[task.fundingPotId];
+    uint currentTotalAmount = fundingPot.payouts[_token];
+    uint currentTaskRolePayout = task.payouts[uint8(_role)][_token];
+    task.payouts[uint8(_role)][_token] = _amount;
+    
+    fundingPot.payouts[_token] = currentTotalAmount - currentTaskRolePayout + _amount;
```
elenadimitrova: Note to self: use safemath

contracts/ColonyFunding.sol
```diff
@@ -132,15 +135,18 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
-  stoppable
-  auth
+  function getFundingPot(uint256 _potId) public view returns
+  (FundingPotAssociatedType associatedType, uint256 associatedTypeId, uint256 payoutsWeCannotMake)
   {
+    FundingPot storage fundingPot = fundingPots[_potId];
+    return (fundingPot.associatedType, fundingPot.associatedTypeId, fundingPot.payoutsWeCannotMake);
+  }
+
+  modifier test(uint256 _fromPot, uint256 _toPot) {
```
elenadimitrova: Note to self: Name this properly and move to where modifiers live in the `DataTypes` contract

test/colony-payment.js
```diff
@@ -0,0 +1,302 @@
...
+      expect(payment.finalized).to.be.true;
+
+      const recipientBalanceBefore = await token.balanceOf(RECIPIENT);
+      const networkBalanceBefore = await token.balanceOf(colonyNetwork.address);
+      await colony.claimPayment(paymentId, token.address);
+
+      const recipientBalanceAfter = await token.balanceOf(RECIPIENT);
+      const networkBalanceAfter = await token.balanceOf(colonyNetwork.address);
+      expect(recipientBalanceAfter.sub(recipientBalanceBefore)).to.eq.BN(new BN("197"));
+      expect(networkBalanceAfter.sub(networkBalanceBefore)).to.eq.BN(new BN("2"));
```
kronosapiens: Should we also have an `otherToken` claim here?

contracts/ColonyFunding.sol
```diff
@@ -76,52 +76,55 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
+    fundingPot.payouts[_token] -= payout;
+    task.payouts[_role][_token] = 0;
 
-    if (task.roles[_role].rating == TaskRatings.Unsatisfactory || payout == 0) {
-      return;
+    bool unsatisfactory = task.roles[_role].rating == TaskRatings.Unsatisfactory;
+    if (!unsatisfactory) {
+      processPayout(task.fundingPotId, _token, payout, task.roles[_role].user);
     }
+  }
```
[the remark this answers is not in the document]
elenadimitrova: Wouldn't the dispute happen before they claim here?

test/colony-payment.js
elenadimitrova: Yes indeed this is missing. Added now.

contracts/ColonyDataTypes.sol
```diff
@@ -172,12 +175,19 @@ contract ColonyDataTypes {
     uint256 blockTimestamp;
   }
 
+  struct Payment {
+    address recipient;
+    bool finalized;
```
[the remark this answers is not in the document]
elenadimitrova: If we'll only have the 1tx flow then this doesn't apply. But otherwise if we have intermediary state Payment and want to "cancel" it the only reason we'd ever want to do that is to recover tokens from its funding pot. However for Payments, we can move funds around any way we want including emptying its fundings which defies the point of including extra logic around "cancelling" to reclaim.

contracts/ColonyFunding.sol
```diff
@@ -76,52 +76,55 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
-      // TODO: (post CCv1) If it's a whitelisted token, it goes straight to the metaColony
-      // If it's any other token, goes to the colonyNetwork contract first to be auctioned.
-      ERC20Extended payoutToken = ERC20Extended(_token);
-      payoutToken.transfer(task.roles[_role].user, remainder);
-      payoutToken.transfer(colonyNetworkAddress, fee);
-    }
+  function setPayout(uint256 _id, address _token, uint256 _amount) public
+  auth
+  stoppable
+  validPayoutAmount(_amount)
```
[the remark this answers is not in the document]
elenadimitrova: If we're only going to have 1tx flow only then change functions are obsoleted really.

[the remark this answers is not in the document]
elenadimitrova: okay I've added a `finalizePayment` transaction into the mix following today's research discussion.

contracts/ColonyAuthority.sol
```diff
@@ -57,9 +57,27 @@ contract ColonyAuthority is CommonAuthority {
...
+    setFounderRoleCapability(colony, "setPaymentRecipient(uint256,address)");
+    
+    setAdminRoleCapability(colony, "setPaymentDomain(uint256,uint256)");
+    setFounderRoleCapability(colony, "setPaymentDomain(uint256,uint256)");
+
+    setAdminRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
+    setFounderRoleCapability(colony, "setPaymentSkill(uint256,uint256)");
+
+    setAdminRoleCapability(colony, "setPayout(uint256,address,uint256)");
+    setFounderRoleCapability(colony, "setPayout(uint256,address,uint256)");
```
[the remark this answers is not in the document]
elenadimitrova: That just sounded a bit off that's why I avoided it..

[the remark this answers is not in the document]
kronosapiens: Fair fair. I think the benefit of having a consistent convention is worth it in this case but it's your prerogative.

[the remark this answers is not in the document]
elenadimitrova: On second thought, it doesn't look too bad so I changed it 🤷‍♀️

truffle.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
     }
   },
   mocha: {
-    reporter: "mocha-circleci-reporter",
+    reporter: "eth-gas-reporter",
```
area: 🤔

elenadimitrova: Well, yes, sorry

contracts/ColonyFunding.sol
```diff
@@ -76,52 +76,49 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
+  function claimTaskPayout(uint256 _id, uint8 _role, address _token) public
   stoppable
   taskFinalized(_id)
   {
     Task storage task = tasks[_id];
+    FundingPot storage fundingPot = fundingPots[task.fundingPotId];
     assert(task.roles[_role].user != address(0x0));
 
     uint payout = task.payouts[_role][_token];
+    task.payouts[_role][_token] = 0;
```
area: Reminder about our discussion about this line :)

contracts/ColonyPayment.sol
```diff
@@ -0,0 +1,119 @@
...
+  returns (uint256)
+  {
+    require(_recipient != address(0x0), "colony-payment-invalid-recipient");
+    paymentCount += 1;
+
+    fundingPotCount += 1;
+    fundingPots[fundingPotCount] = FundingPot({
+      associatedType: FundingPotAssociatedType.Payment,
+      associatedTypeId: paymentCount,
+      payoutsWeCannotMake: 1
```
area: Only true if `_amount` is nonzero, right? Otherwise, I think that a payout that starts with a zero _amount can never finalize (because `payoutsWeCannotMake` will never be able to become zero).

test-gas-costs/gasCosts.js
```diff
@@ -86,6 +90,13 @@ contract("All", function(accounts) {
 
     const otherTokenArgs = getTokenArgs();
     otherToken = await DSToken.new(otherTokenArgs[1]);
+
+    oneTxExtension = await OneTxPayment.new();
+    await fundColonyWithTokens(colony, token, INITIAL_FUNDING);
+    // Give a use colony admin rights
```
area: Typo

test/extensions/one-tx-payment.js
```diff
@@ -0,0 +1,93 @@
...
+    await colonyNetwork.initialiseReputationMining();
+    await colonyNetwork.startNextCycle();
+    oneTxExtension = await OneTxPayment.new();
+    globalSkillId = await colonyNetwork.getRootGlobalSkillId();
+  });
+
+  beforeEach(async () => {
+    ({ colony, token } = await setupRandomColony(colonyNetwork));
+    await fundColonyWithTokens(colony, token, INITIAL_FUNDING);
+    // Give a use colony admin rights
```
area: Typo

contracts/ColonyPayment.sol
elenadimitrova: Very good catch, it slipped when I started allowing 0 amounts.. Not that is an excuse.

contracts/ColonyPayment.sol
```diff
@@ -0,0 +1,119 @@
...
+  paymentNotFinalized(_id)
+  skillExists(_skillId)
+  globalSkill(_skillId)
+  {
+    payments[_id].skills[0] = _skillId;
+  }
+
+  function getPayment(uint256 _id) public view returns(address payable, bool, uint256, uint256, uint256[] memory) {
+    Payment storage payment = payments[_id];
+    return (payment.recipient, payment.finalized, payment.fundingPotId, payment.domainId, payment.skills);
```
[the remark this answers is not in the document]
elenadimitrova: Good point. Returning struct now.

contracts/ColonyFunding.sol
```diff
@@ -76,52 +76,49 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs {
...
 
-    uint fee = calculateNetworkFeeForPayout(payout);
-    uint remainder = sub(payout, fee);
+  function setPaymentPayout(uint256 _id, address _token, uint256 _amount) public
+  auth
+  stoppable
+  validPayoutAmount(_amount)
+  paymentNotFinalized(_id)
+  {
+    FundingPot storage fundingPot = fundingPots[_id];
```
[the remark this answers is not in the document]
elenadimitrova: Leftover mistake from the time when there was an idea that payouts were to be managed commonly between Tasks and Payments.

test/colony-payment.js
```diff
@@ -0,0 +1,352 @@
...
+
+    it("can finalize payment when it is fully funded", async () => {
+      const payment = await colony.getPayment(paymentId);
+      await colony.moveFundsBetweenPots(payment.fundingPotId, 1, 1, token.address);
+      expect(payment.finalized).to.be.false;
+
+      await checkErrorRevert(colony.finalizePayment(paymentId), "colony-payment-not-funded");
+    });
+
+    it("cannnot finalize payment if it is NOT fully funded", async () => {
```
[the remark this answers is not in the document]
elenadimitrova: I've switched the 2 test names here. Will fix

PR #560, opened 2019-03-04
Maintenance/solc 0.5

### LINKED ISSUE #459: Upgrade contracts to solc 0.5.0

area:
Closes #459 

Using this partly as a test-bed for the new `solidity-coverage` version I've got brewing up, so until that's done this is inappropriate to be merged. 

The upgrades to solc 0.5 I have done are only to get this branch compiling, and are not the definitive upgrades - in particular, casting addresses through `uint160` to get around `payable` safety isn't a good look.

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -95,7 +95,7 @@ contract ColonyTask is ColonyStorage {
   domainExists(_domainId)
   {
     taskCount += 1;
-    
+
     fundingPotCount += 1;
     fundingPots[fundingPotCount] = FundingPot({
       associatedType: FundingPotAssociatedType.Task,
@@ -297,7 +297,7 @@ contract ColonyTask is ColonyStorage {
     return taskWorkRatings[_id].secret[_role];
   }
 
-  function setTaskManagerRole(uint256 _id, address _user) public
+  function setTaskManagerRole(uint256 _id, address payable _user) public
   stoppable
   self()
   isAdmin(_user)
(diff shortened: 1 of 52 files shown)


truffle.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: "0.4.23",
+      version: "0.5.3",
```
elenadimitrova: Can we also change the CI step step_pull_solc_docker to pull the new image

contracts/EtherRouter.sol
```diff
@@ -48,6 +48,9 @@ contract EtherRouter is DSAuth {
 
     // Make the call
     assembly {
+      let size := extcodesize(destination)
+      if eq(size, 0) { revert(0,0) }
```
elenadimitrova: 👍 
Going though this contract I feel we may want to reword the large comment above here to make it clear which choice we made in the design ( i.e. option 2). I remember hearing at least one contributor being confused by this comment and unclear on what option we went with.

contracts/ColonyTask.sol
```diff
@@ -297,22 +297,22 @@ contract ColonyTask is ColonyStorage {
...
+  function setTaskManagerRole(uint256 _id, address payable _user) public
   stoppable
   self()
   isAdmin(_user)
   {
     setTaskRoleUser(_id, TaskRole.Manager, _user);
   }
 
-  function setTaskEvaluatorRole(uint256 _id, address _user) public stoppable self {
+  function setTaskEvaluatorRole(uint256 _id, address payable _user) public stoppable self {
```
elenadimitrova: Just so we don't forget what we discussed, let's bring these into the IColony interface.

yarn.lock
```diff
@@ -7322,6 +7409,16 @@ web3-core-helpers@1.0.0-beta.35:
     web3-eth-iban "1.0.0-beta.35"
     web3-utils "1.0.0-beta.35"
 
+web3-core-helpers@1.0.0-beta.47:
+  version "1.0.0-beta.47"
+  resolved "https://registry.yarnpkg.com/web3-core-helpers/-/web3-core-helpers-1.0.0-beta.47.tgz#f2c445bce0ca9fe3f0d4e2040dcdde319ae49684"
+  integrity sha512-A/Yh1L1h8Yfd0MJaBcrU1XP2xPeXG1Mphq/zkf9fhom4BgkXagrUjEaPCe5iN98Zgdn9w7MQpibyNu0aUAcxNA==
```
elenadimitrova: For some reason all of these integrity properties are missing when I run `yarn`. What version of `yarn` are you on?

elenadimitrova: ok so this turned out to be a `yarn` version mismatch. We should all now be on 1.13.0

PR #567, opened 2019-03-14
Downgrade global skill tree to global skill tags

### LINKED ISSUE #565: Tags Cloud
Following onto [this document](https://docs.google.com/document/d/1O_fnQWEsIMctQ6ZQ6OXb4I_BJsQTDyzPnrOy8-ZOaL8/edit), we would like to move forward with a "tags cloud" in lieu of a "skills tree", with tags being a more semantically-neutral term for skills, and a cloud being a more flexible structure for storing what we anticipate will be a quite protean set of concepts. 

In this [Slack update](https://joincolony.slack.com/archives/C02R9QLA0/p1551982249010200) we announced that we would like to move forward as follows:

1. We will use a cloud!
2. We will support 1 tag per task at launch...
3. But will figure out the best way to build support for many tags under-the-hood so they can be added easily in the future.
4. For a task with multiple tags, reputation will be divided evenly among the tags.
5. The meta colony will manage the cloud, centrally at first but we will move to more decentralized methods in the future.

As a spec for the network team, this means:

1. We should update our naming conventions, replacing references to "skills" with "tags". As a nice side-effect, we can rename `localSkillId`, our current (and I will suggest somewhat unintuitive) way to refer to domains to something more specific such as `reputationId`. Tags have a reputationId, domains have a reputationId. reputationIds are how we refer to things in the reputation tree. Huzzah.

2. We currently support skills as an array, capped at 1 entry. We have two options: to leave this as-is, or to imagine a new approach to storing skills in a way which avoids looping over an array of skills. The former approach is simpler but will lead to higher gas costs in the future (and serve implicitly to bound the number of possible skills), while the second has a higher up-front development cost but should permit "infinite" skills for free. A related question is whether "infinite" skills is desirable.

3. This is less important now since we only allow one skill, but when we allow support for multiple skills we will need to add logic which divides the rep amongst all the skills (`rep / nSkills`).

~4. We will need to decide how skills are to be added / deprecated / removed. Let's assume we have "administration" or "architecture" or "root" functions which perform this and worry about how to trustless-ify those functions in the future (i.e. via the dispute mechanism, a curated registry, or some other mechanism).~

5. We will need to update our tests, as currently we use skills to test the tree structure for reputation mining (as domains are capped at 2 levels at the moment). Specifically, we will need to use domains to test the reputation miner now. One hacky idea is to make the domain restriction a function of colony version (i.e. `require(domainId == 1 || version >= 100)`) and bump the version in the tests to 100.

area:
Closes #565 

- [x] Remove concept of the global skill tree for contracts
- [x] Update tests that run on each build
- [x] Update nightly E2E tests

Optional:
- [x] Update reputation mining log and clients to allow skills to be tagged with multiple skills, and reputation to be awarded divided up among all tags.


kronosapiens:
Since we're moving past the concept of "skills" in a tree, can we update the nomenclature and drop the distinction between "local" and "global" skills? I've always found that a bit unintuitive... something like `reputationId` would be more clear. Thoughts?

area:
So the 'smart' way of allowing unlimited skills per task with no additional gas cost proved impossible. Fundamentally, the idea was that the log entry would be able to point at the task in question, and then the reputation miners and dispute resolution contracts would be able to reference that task to see what the `nth` tag on the skill was. Unfortunately, this results in an exploit:

* Submit a bad hash during reputation mining, judiciously chosen
* Wait until you get put in to a dispute
* Use recovery mode to change the tags assigned to a task. Your bad hash was judiciously chosen such that it now represents the 'correct' state for the disagreement, so you win the dispute and get to set the reputation state to whatever you want.

We then considered adjusting the reputation log entries such that they could take an array of skill ids, rather than just one. This would be gas limited, but to the tune of 20000 gas per tag.  Unfortunately, this would mean that many of the functions in `reputationMiningCycleRespond` (such as `checkUserOriginReputation`) would have to binary search over the array of based on the relative update number to find the right skill, which would be hard (but not impossible) to make guarantees about the gas required to do so. This admittedly remains a possibility in the future to explore, but on the timescales we're talking about here, I'm unhappy with such a radical change to the reputation mining contracts.

This left the 'dumb' way, where we add an extra entry in the log for each tag in the task, which is what I've implemented here. This is the most gas limited of the options I've discussed here (around 100000 gas per tag), but the easiest to implement. When we enable more than one tag on a task in the contracts, we should have a hard limit in the contracts of around 30, and I feel like messaging in the dApp should heavily encourage people to only put 3-ish tags on a task.

I have included some tests that demonstrate that more than one tag in a task works; once this is properly implemented in the contracts, these tests should be updated to not use the `testHelper/taskSkillEditing.sol` contract.

kronosapiens:
I would even say put the hard cap at 10 and message 3.

test/meta-colony.js
```diff
@@ -243,8 +261,8 @@ contract("Meta Colony", accounts => {
       expect(parentId).to.be.zero;
     });
 
-    it("should NOT be able to add a new root global skill", async () => {
-      await checkErrorRevert(metaColony.addGlobalSkill(0), "colony-invalid-skill-id");
+    it.skip("should NOT be able to add a new root global skill", async () => {
```
elenadimitrova: Why skipped?

test/meta-colony.js
```diff
@@ -93,54 +108,57 @@ contract("Meta Colony", accounts => {
...
-      await metaColony.addGlobalSkill(5);
+      // Why this random addGlobalSkill in the middle? It means we can use effectively the same tests
+      // below, but with skill ID 7 replaced with skill ID 2. While porting everything to the tag cloud
+      // arrangement, I was very interested in changing tests as little as possible.
+      await metaColony.addDomain(1); // Domain ID 2, skill id 4
+      await metaColony.addDomain(1); // Domain ID 3, skill id 5
+      await metaColony.addDomain(3); // Domain ID 4, skill id 6
+      await metaColony.addGlobalSkill(); // Skill id 7
+      await metaColony.addDomain(2); // Domain ID 5, skill id 8
+      await metaColony.addDomain(3); // Domain ID 6, skill id 9
```
elenadimitrova: Phew, the comments were helpful. Thanks

test/meta-colony.js
```diff
@@ -172,10 +190,10 @@ contract("Meta Colony", accounts => {
       const skill3ParentSkillId2 = await colonyNetwork.getParentSkillId(6, 1);
       expect(skill3ParentSkillId2).to.eq.BN(1);
 
-      const skill4 = await colonyNetwork.getSkill(7);
+      const skill4 = await colonyNetwork.getSkill(2);
```
elenadimitrova: What's wrong with testing skillId 7? Skill 2 is the reputation mining skill which is probably tested elsewhere but [here](https://github.com/JoinColony/colonyNetwork/pull/567/files#diff-07c4d8d59a6ece14a120308fcc8f0e1bR147) we are adding the new global skill with Id 7 so it should be that which is tested.

test/meta-colony.js
```diff
@@ -93,54 +108,57 @@ contract("Meta Colony", accounts => {
...
+      await metaColony.addDomain(1);
 
-      await checkErrorRevert(metaColony.addGlobalSkill(6), "colony-invalid-skill-id");
+      await checkErrorRevert(metaColony.addDomain(6), "colony-domain-does-not-exist");
       const skillCount = await colonyNetwork.getSkillCount();
       expect(skillCount).to.eq.BN(5);
     });
 
-    it("should NOT be able to add a child skill to a local skill parent", async () => {
+    it.skip("should NOT be able to add a child skill to a local skill parent", async () => {
```
elenadimitrova: why skipped 🤔

test/meta-colony.js
```diff
@@ -60,23 +73,25 @@ contract("Meta Colony", accounts => {
...
+      expect(rootSkillChild).to.eq.BN(2);
+      expect(rootSkillChild2).to.eq.BN(4);
     });
 
     it("should not allow a non-founder role in the metacolony to add a global skill", async () => {
-      await checkErrorRevert(metaColony.addGlobalSkill(1, { from: OTHER_ACCOUNT }), "ds-auth-unauthorized");
+      await checkErrorRevert(metaColony.addGlobalSkill({ from: OTHER_ACCOUNT }), "ds-auth-unauthorized");
     });
 
     it("should be able to add multiple child skills to the root global skill", async () => {
```
elenadimitrova: Can we rename that test to maybe `should be able to add multiple child skills to the root domain skill`

contracts/testHelpers/NoLimitSubdomains.sol
```diff
@@ -0,0 +1,59 @@
...
+    // Setup new local skill
+    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    uint256 newLocalSkill = colonyNetwork.addSkill(parentSkillId);
+
+    // Add domain to local mapping
+    initialiseDomain(newLocalSkill);
+  }
+
+  // We copy this function from Colony.sol to here because it's private, not internal, so we can't access it even if we inherit
+  function initialiseDomain(uint256 _skillId) private skillExists(_skillId) {
```
elenadimitrova: Can we make `initialiseDomain` internal in `Colony` then?

contracts/ColonyNetwork.sol
```diff
@@ -199,52 +193,58 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+  allowedToAddSkill(_parentSkillId == 0)
   returns (uint256)
   {
     skillCount += 1;
 
     Skill storage parentSkill = skills[_parentSkillId];
-
     // Global and local skill trees are kept separate
-    require(parentSkill.globalSkill == _globalSkill, "colony-global-and-local-skill-trees-are-separate");
+    require(_parentSkillId == 0 || parentSkill.globalSkill == false, "colony-global-and-local-skill-trees-are-separate");
```
kronosapiens: Could use the `!` operator here.

contracts/ColonyNetwork.sol
```diff
@@ -199,52 +193,58 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
-      }
-
-      // Check if we've reached the root of the tree yet (it has no parents)
-      // Otherwise get the next parent
-      if (parentSkill.nParents == 0) {
-        notAtRoot = false;
-      } else {
-        parentSkillId = parentSkill.parents[0];
-        parentSkill = skills[parentSkill.parents[0]];
+    if (_parentSkillId > 0) {
```
kronosapiens: I would say `_parentSkillId != 0` would be slightly more semantic, since it's 0 vs. all the others.

contracts/ColonyNetwork.sol
```diff
@@ -199,52 +193,58 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+
+      s.nParents = parentSkill.nParents + 1;
+      s.globalSkill = false;
```
kronosapiens: This is the default no?

contracts/ColonyTask.sol
```diff
@@ -484,7 +484,20 @@ contract ColonyTask is ColonyStorage {
 
     colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, domains[task.domainId].skillId);
     if (taskRole == TaskRole.Worker) {
-      colonyNetworkContract.appendReputationUpdateLog(role.user, reputation, task.skills[0]);
+      int256 nSkills = 0;
```
kronosapiens: Noting that if we're removing skill penalties we should be able to treat these at `uint`s now.

contracts/testHelpers/TaskSkillEditing.sol
```diff
@@ -0,0 +1,44 @@
...
+  function addTaskSkill(uint256 _taskId, uint256 _skillId) 
+  public
+  {
+    tasks[_taskId].skills.push(_skillId);
+  }
+
+  function removeTaskSkill(uint256 _taskId, uint256 _skillId) 
+  public
+  {
+    tasks[_taskId].skills[_skillId] = 0;
```
kronosapiens: Feels more like this should be `_skillIndex` than `_skillId`, no?

test-system/end-to-end.js
```diff
@@ -62,6 +64,12 @@ contract("End to end Colony network and Reputation mining testing", function(acc
     tokenLocking = await ITokenLocking.at(tokenLockingAddress);
     ({ metaColony, clnyToken } = await setupMetaColonyWithLockedCLNYToken(colonyNetwork));
 
+    // Replace addDomain with the addDomain implementation with no restrictions on depth of subdomains
+    const noLimitSubdomains = await NoLimitSubdomains.new();
+    const resolverAddress = await colonyNetwork.getColonyVersionResolver(1);
+    const resolver = await Resolver.at(resolverAddress);
+    await resolver.register("addDomain(uint256)", noLimitSubdomains.address);
```
kronosapiens: TIL

test-system/end-to-end.js
```diff
@@ -231,62 +240,102 @@ contract("End to end Colony network and Reputation mining testing", function(acc
...
-        });
-      });
+      await Promise.all(
+        colonyTaskProps.map(async taskProp => {
+          const { colony } = colonies[taskProp.colonyIdx];
+          await setupFinalizedTask({
+            colonyNetwork,
+            colony,
+            domainId: taskProp.domainId,
+            skillId: taskProp.skillId,
```
kronosapiens: Wait, isn't this `undefined` now?

test/colony-network.js
```diff
@@ -129,21 +129,18 @@ contract("Colony Network", accounts => {
     it("when meta colony is created, should have the root global and local skills initialised, plus the local mining skill", async () => {
       const skillCount = await colonyNetwork.getSkillCount();
       expect(skillCount).to.eq.BN(3);
-      const rootGlobalSkill = await colonyNetwork.getSkill(1);
-      expect(parseInt(rootGlobalSkill.nParents, 10)).to.be.zero;
-      expect(parseInt(rootGlobalSkill.nChildren, 10)).to.be.zero;
+      const globalSkill = await colonyNetwork.getSkill(3);
```
kronosapiens: I feel like adding a `ROOT_GLOBAL_SKILL = 3` to `constants.js` would be a good idea vs. having that as implicit knowledge.

test/meta-colony.js
kronosapiens: 🤔

test/meta-colony.js
```diff
@@ -479,14 +497,14 @@ contract("Meta Colony", accounts => {
...
+      await metaColony.addGlobalSkill();
+      await metaColony.addGlobalSkill();
       await fundColonyWithTokens(colony, token, INITIAL_FUNDING);
       const taskId = await setupFinalizedTask({ colonyNetwork, colony });
-      await checkErrorRevert(colony.setTaskSkill(taskId, 6), "colony-task-complete");
+      await checkErrorRevert(colony.setTaskSkill(taskId, 5), "colony-task-complete");
 
       const task = await colony.getTask(taskId);
-      expect(task[8][0]).to.eq.BN(1);
+      expect(task.skillIds[0]).to.eq.BN(3);
```
kronosapiens: Great place for a `ROOT_GLOBAL_SKILL` constant.

test/reputation-mining/client-calculations.js
```diff
@@ -28,11 +33,22 @@ const loader = new TruffleLoader({
 
 const realProviderPort = process.env.SOLIDITY_COVERAGE ? 8555 : 8545;
 
+async function customSetupFinalizedTask(args) {
+  const newArgs = Object.assign(args, {
+    skillId: 3,
```
kronosapiens: Ditto the constant.

test/reputation-update.js
```diff
@@ -271,20 +275,26 @@ contract("Reputation Updates", accounts => {
...
-      await metaColony.addGlobalSkill(6);
-
-      await setupFinalizedTask({ colonyNetwork, colony: metaColony, skillId: 5 });
-      let repLogEntryWorker = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(4);
-      expect(repLogEntryWorker.amount).to.eq.BN(WORKER_PAYOUT);
+      // Replace addDomain with the addDomain implementation with no restrictions on depth of subdomains
+      const noLimitSubdomains = await NoLimitSubdomains.new();
+      const resolverAddress = await colonyNetwork.getColonyVersionResolver(1);
+      const resolver = await Resolver.at(resolverAddress);
+      await resolver.register("addDomain(uint256)", noLimitSubdomains.address);
```
kronosapiens: I'm seeing this code block a lot -- breaking this out into a test helper would make this PR slightly more succinct.

test/reputation-update.js
```diff
@@ -271,20 +275,26 @@ contract("Reputation Updates", accounts => {
...
+      expect(repLogEntryWorker.amount).to.eq.BN(MANAGER_PAYOUT);
       expect(repLogEntryWorker.nUpdates).to.eq.BN(6);
 
-      await setupFinalizedTask({ colonyNetwork, colony: metaColony, skillId: 6 });
-      repLogEntryWorker = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(8);
-      expect(repLogEntryWorker.amount).to.eq.BN(WORKER_PAYOUT);
-      expect(repLogEntryWorker.nUpdates).to.eq.BN(8); // Negative reputation change means children change as well.
+      await setupFinalizedTask({ colonyNetwork, colony: metaColony, domainId: 4, managerRating: 1 });
+      repLogEntryWorker = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(5);
+      expect(repLogEntryWorker.amount).to.eq.BN(MANAGER_PAYOUT.mul(new BN(-1)));
```
kronosapiens: Would prefer `.neg()`.

test/reputation-update.js
```diff
@@ -327,5 +337,49 @@ contract("Reputation Updates", accounts => {
...
       expect(numUpdates).to.eq.BN(3);
     });
+
+    it("should add the right log entries for tasks with multiple skills", async () => {
+      // Introduce our ability to add and remove skills from tasks
+      const taskSkillEditing = await TaskSkillEditing.new();
+      const resolverAddress = await colonyNetwork.getColonyVersionResolver(1);
+      const resolver = await Resolver.at(resolverAddress);
+      await resolver.register("addTaskSkill(uint256,uint256)", taskSkillEditing.address);
+      await resolver.register("removeTaskSkill(uint256,uint256)", taskSkillEditing.address);
```
kronosapiens: Ditto above, a helper function which captures this logic would be nice.

contracts/ColonyNetwork.sol
```diff
@@ -38,7 +38,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
   }
 
   modifier skillExists(uint skillId) {
-    require(skillId > 0 && skillCount >= skillId, "colony-invalid-skill-id");
+    require(skillId == 0 || skillCount >= skillId, "colony-invalid-skill-id");
```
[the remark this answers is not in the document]
kronosapiens: Nm Elena said this already.

contracts/ColonyNetwork.sol
```diff
@@ -199,52 +193,58 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
-
-    uint parentSkillId = _parentSkillId;
-    bool notAtRoot = true;
-    uint powerOfTwo = 1;
-    uint treeWalkingCounter = 1;
-
-    // Walk through the tree parent skills up to the root
-    while (notAtRoot) {
-      // Add the new skill to each parent children
-      // TODO: skip this for the root skill as the children of that will always be all skills
```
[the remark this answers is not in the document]
area: Yeah, it can certainly be removed... I'm not even sure it was correct 😟

contracts/ColonyNetwork.sol
```diff
@@ -199,52 +193,58 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+          parentSkill = skills[parentSkill.parents[0]];
+        }
+
+        treeWalkingCounter += 1;
       }
-
-      treeWalkingCounter += 1;
+    } else {
+      // Add a global skill
+      s.globalSkill = true;
```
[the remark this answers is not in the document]
area: The issue is that the skills corresponding to top-level domains in colonies are also not going to have any parents, which I can't immediately see an elegant solution for.

contracts/ColonyTask.sol
```diff
@@ -484,7 +484,20 @@ contract ColonyTask is ColonyStorage {
...
+      for (uint i = 0; i < task.skills.length; i += 1) {
+        if (task.skills[i] > 0 ) {
+          nSkills += 1;
```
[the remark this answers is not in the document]
area: No, simply because we're not supporting multiple skills at this point. That will work, though.

contracts/ColonyTask.sol
```diff
@@ -484,7 +484,20 @@ contract ColonyTask is ColonyStorage {
...
+        }
+      }
+
+      int256 reputationPerSkill = reputation / nSkills;
```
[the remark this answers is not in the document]
area: We can certainly `assert(nSkills >0)`. I don't think `assert(reputation >= nSkills)` is appropriate though. Once we have two skills, it's certainly possible for someone to be paid out 1 wei of a token.

migrations/8_setup_meta_colony.js
```diff
@@ -52,6 +52,8 @@ module.exports = async function(deployer, network, accounts) {
   const tokenLocking = await ITokenLocking.at(tokenLockingAddress);
   await tokenLocking.deposit(clnyToken.address, DEFAULT_STAKE, { from: MAIN_ACCOUNT });
 
+  await metaColony.addGlobalSkill();
```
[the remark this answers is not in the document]
area: Correct.

helpers/test-data-generator.js
```diff
@@ -37,7 +37,7 @@ const ColonyTask = artifacts.require("ColonyTask");
 const IColonyNetwork = artifacts.require("IColonyNetwork");
 const ContractRecovery = artifacts.require("ContractRecovery");
 
-export async function makeTask({ colony, hash = SPECIFICATION_HASH, domainId = 1, skillId = 1, dueDate = 0, manager }) {
+export async function makeTask({ colony, hash = SPECIFICATION_HASH, domainId = 1, skillId = 3, dueDate = 0, manager }) {
```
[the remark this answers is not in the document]
area: The first two are the meta colony root domain skill and the mining skill.

test/reputation-mining/happy-paths.js
```diff
@@ -72,15 +74,21 @@ contract("Reputation Mining - happy paths", accounts => {
     colonyNetwork = await setupColonyNetwork();
     ({ metaColony, clnyToken } = await setupMetaColonyWithLockedCLNYToken(colonyNetwork));
 
+    // Replace addDomain with the addDomain implementation with no restrictions on depth of subdomains
+    const noLimitSubdomains = await NoLimitSubdomains.new();
+    const resolverAddress = await colonyNetwork.getColonyVersionResolver(1);
```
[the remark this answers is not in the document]
area: I don't think I understand how this would break, or how `getCurrentColonyVersion` would help solve the problem (though maybe that becomes obvious once I understand the problem)

test/meta-colony.js
area: There's no such thing as the root global skill any more. Or are you asking why skipped, and not deleted?

test/reputation-mining/happy-paths.js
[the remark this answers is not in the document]
elenadimitrova: `getColonyVersionResolver` returns the Resolver instance of Colony contract version 1 in this instance.
However, any new colonies we create, will always be using the latest Colony contract version, which currently is also 1 but will obviously be incremented over time.

contracts/testHelpers/NoLimitSubdomains.sol
area: I generally don't like changing the codebase to make tests easier to write, but given you're both in favour of it...

test/reputation-mining/happy-paths.js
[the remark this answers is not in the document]
area: I'm going to have to blame being distracted by Brexit stuff for why I was such an idiot here.

test/meta-colony.js
area: I was trying to preempt this comment with [this explanation](https://github.com/JoinColony/colonyNetwork/pull/567/files/985928455386ede5568c68b4347b49610b297828#diff-07c4d8d59a6ece14a120308fcc8f0e1bR141).

contracts/testHelpers/NoLimitSubdomains.sol
elenadimitrova: I agree. But I disagree this is an example of said anti-working pattern.
We are extending the inheritance model here and we are reviewing the functions that trickle through and those that don't and updating accordingly. I can't imagine we could have predicted we'll be fudging domain creation.
Our use of `private` functions has been a bit more liberal, in that we prefer them even for functions only called from within the current contract even if they are potentially useful downstream in future contracts.
I personally prefer this more conservative approach but I'm happy to re-review our `internal`/`private` modifiers use in line with what I think are your expectations.

contracts/ColonyTask.sol
area: Maybe I've misunderstood you, but we're only removing the penalty on the global skill given to the worker for not rating the manager. They might still have a negative reputation change overall if they've been rated -1, so this needs to remain `int` or I need to cast a lot in the following lines. I opted for the former.

test-system/end-to-end.js
area: Sure was. But the helpers set default values, so it all still worked.

test/meta-colony.js
area: This function no longer takes a parent parameter, so there's no way to try and give it an ID corresponding to a local skill as a parent.

test/meta-colony.js
elenadimitrova: We should just delete it really.

test/meta-colony.js
elenadimitrova: The explanation link is now gone..

test/meta-colony.js
elenadimitrova: Yeah, delete it then. I think the practice of skipping tests stayed with us since we were reluctant to delete reputation mining tests that took a while to write yet were obsoleted by added logic. We should clean that all up perhaps.

test/reputation-mining/happy-paths.js
```diff
@@ -319,9 +322,9 @@ contract("Reputation Mining - happy paths", accounts => {
       await badClient.initialise(colonyNetwork.address);
 
       let repCycle = await getActiveRepCycle(colonyNetwork);
-      const rootGlobalSkill = await colonyNetwork.getRootGlobalSkillId();
-      const globalKey = ReputationMinerTestWrapper.getKey(metaColony.address, rootGlobalSkill, ZERO_ADDRESS);
-      const userKey = ReputationMinerTestWrapper.getKey(metaColony.address, rootGlobalSkill, MINER1);
+      const skillId = new BN(3);
```
elenadimitrova: `GLOBAL_SKILL`

test/reputation-mining/happy-paths.js
```diff
@@ -360,9 +363,9 @@ contract("Reputation Mining - happy paths", accounts => {
       const badClient = new MaliciousReputationMinerExtraRep({ loader, realProviderPort, useJsTree, minerAddress: MINER2 }, 1, new BN("10"));
       await badClient.initialise(colonyNetwork.address);
 
-      const rootGlobalSkill = await colonyNetwork.getRootGlobalSkillId();
-      const globalKey = ReputationMinerTestWrapper.getKey(metaColony.address, rootGlobalSkill, ZERO_ADDRESS);
-      const userKey = ReputationMinerTestWrapper.getKey(metaColony.address, rootGlobalSkill, MINER1);
+      const skillId = new BN(3);
```
elenadimitrova: `GLOBAL_SKILL`

test/reputation-mining/happy-paths.js
```diff
@@ -448,18 +451,17 @@ contract("Reputation Mining - happy paths", accounts => {
...
       await advanceMiningCycleNoContest({ colonyNetwork, test: this, client: goodClient });
-      expect(Object.keys(goodClient.reputations).length).to.equal(27);
+      expect(Object.keys(goodClient.reputations).length).to.equal(33);
 
-      const GLOBAL_SKILL = new BN(1);
-      const META_ROOT_SKILL = new BN(2);
-      const MINING_SKILL = new BN(3);
+      const META_ROOT_SKILL = new BN(1);
+      const MINING_SKILL = new BN(2);
+      const GLOBAL_SKILL = new BN(3);
```
elenadimitrova: Deja vu

kronosapiens: CSONTNATS

test/meta-colony.js
area: After some thought, they're both reachable, so I've replaced the tests

test/meta-colony.js
```diff
@@ -122,8 +123,21 @@ contract("Meta Colony", accounts => {
       expect(skillCount).to.eq.BN(5);
     });
 
-    it.skip("should NOT be able to add a child skill to a local skill parent", async () => {
-      await checkErrorRevert(metaColony.addGlobalSkill(2), "colony-global-and-local-skill-trees-are-separate");
+    it("should NOT be able to add a child skill to a local skill parent", async () => {
+      // Put colony in to recovery mode
+      await metaColony.enterRecoveryMode();
```
elenadimitrova: oo clever way we can test all `assert`s! Not in this PR obviously

test/meta-colony.js
area: Let's [try again](https://github.com/JoinColony/colonyNetwork/pull/567/files#diff-07c4d8d59a6ece14a120308fcc8f0e1bR146)...

PR #568, opened 2019-03-14
Remove redundant auth from ReputationMiningCycleStorage

kronosapiens:
Follow-on from #553 , looking into whether the `DSAuth` in `ReputationMiningCycleStorage.sol` is redundant.

After doing some investigation, I have found that `DSAuth` is both redundant _and_ not redundant. `DSAuth` is **redundant** because no functions in `ReputationMiningCycle` use the `auth` modifier. However, it is **not redundant** because under-the-hood, we generate a new `EtherRouter` contract and _cast it to_ `ReputationMiningCycle` before use. In that sense, the actual contract has `owner` and `authority` variables from `EtherRouter`, and so we need to reserve the first two storage variables in `ReputationMiningCycle` so that the storage lines up.

My solution is to _remove_ the DSAuth dependency but to _keep_ the `owner` and `authority` storage variables (although to rename them to dummy names for clarity). These storage variables are not meant for use. While having "dummy" variables is a bit awkward, I would suggest that this does provide more clarity overall, specifically by making it clear that we _do not_ authenticate the `ReputationMiningCycle` functions.

Ultimately it's not a big change so I could see arguments for not merging this and leaving things as-are. But I did the investigation and so I figured I should present the change to the team for consideration.

### DIFF
--- a/contracts/ReputationMiningCycleStorage.sol
+++ b/contracts/ReputationMiningCycleStorage.sol
@@ -22,8 +22,11 @@ import "./ReputationMiningCycleDataTypes.sol";
 
 
 contract ReputationMiningCycleStorage is ReputationMiningCycleDataTypes, DSAuth {
+  // From DSAuth there is authority and owner at storage slots 0 and 1 respectively
+  // These are not used but are necessary for alignment when casting from EtherRouter
+
   // Address of the Resolver contract used by EtherRouter for lookups and routing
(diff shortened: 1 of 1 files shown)


elenadimitrova:
I understand why it's inherited now and personally slightly prefer the current state as the storage variables on the `ReputationMiningCycle` are in fact the `owner` and `authority` that `DSAuth` imposes. Perhaps we should only merge a comment on the `ReputationMiningCycle` to explain your findings above so it doesn't confuse someone else again?

kronosapiens:
No problem.

PR #576, opened 2019-03-26
Do not penalise worker's skill awarded 

### LINKED ISSUE #563: Remove global skill reputation penalty for worker
Currently, we expect when a task is complete, the evaluator and worker to rate the worker and manager respectively. [If one of them doesn't, they get penalised](https://github.com/JoinColony/colonyNetwork/pull/469#issuecomment-448639144). 

We've since deemed unfair for the worker penalty to apply to both their domain _and_ global skill reputation amount, since their ability to perform in a certain skill should be differentiated from their ability to perform in the colony domain, which is more representative of the project / work management aspect of it.

Highlighting that this is specifically and only referring to the circumstance in which the worker has not rated the manager in time and as skill reputation is not used for claiming rewards this seems a reasonable suggestion approved by @jakzilla

area:
Closes #563

### DIFF
--- a/test/reputation-update.js
+++ b/test/reputation-update.js
@@ -196,7 +196,7 @@ contract("Reputation Updates", accounts => {
 
       const repLogEntryWorker2 = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(4);
       expect(repLogEntryWorker2.user).to.equal(WORKER);
-      expect(repLogEntryWorker2.amount).to.eq.BN(WORKER_PAYOUT);
+      expect(repLogEntryWorker2.amount).to.eq.BN(WORKER_PAYOUT.muln(3).divn(2));
     });
 
     it("should set the correct reputation change amount in log when evaluator has failed to rate", async function() {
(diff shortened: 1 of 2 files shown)


elenadimitrova:
Would it make sense to move [this logic](https://github.com/JoinColony/colonyNetwork/pull/576/files#diff-577c88725f9b28dec96aee64a5873f00R478) in `ColonyTask.assignWorkRating`? It feels a little misplaced and lonely there.

elenadimitrova:
Also I feel [this](https://github.com/JoinColony/colonyNetwork/pull/576/files#diff-577c88725f9b28dec96aee64a5873f00R496) should be an `assert` not `require`  (sorry I know it's not exactly part of this work, if you feel too strongly for that, you can log them as separate issues)

test/reputation-update.js
```diff
@@ -196,7 +196,7 @@ contract("Reputation Updates", accounts => {
 
       const repLogEntryWorker2 = await inactiveReputationMiningCycle.getReputationUpdateLogEntry(4);
       expect(repLogEntryWorker2.user).to.equal(WORKER);
-      expect(repLogEntryWorker2.amount).to.eq.BN(WORKER_PAYOUT);
+      expect(repLogEntryWorker2.amount).to.eq.BN(WORKER_PAYOUT.muln(3).divn(2));
```
elenadimitrova: Shouldn't the worker here get 1.5 times their payout because evaluator failed to rate them?

PR #583, opened 2019-04-04
Fixed Restore good state in reputation mining test.

### LINKED ISSUE #524: Restore good state in reputation mining tests
In mining tests, work on restoring good state even after a test fails. Currently the test run exits on test failure.

dev1644:
Moved the logic of `before` block into a function `newNetworkInstance`, if we ever reached to a point from where we can't untangle the disputed states, we will call this function which will reset the network completely. 

Closes #524

### DIFF
--- a/test/reputation-mining/root-hash-submissions.js
+++ b/test/reputation-mining/root-hash-submissions.js
@@ -17,9 +17,7 @@ import {
   submitAndForwardTimeToDispute,
   accommodateChallengeAndInvalidateHash,
   getValidEntryNumber,
-  finishReputationMiningCycleAndWithdrawAllMinerStakes,
-  takeSnapshot,
-  revertToSnapshot
+  finishReputationMiningCycleAndWithdrawAllMinerStakes
 } from "../../helpers/test-helper";
 
 import ReputationMinerTestWrapper from "../../packages/reputation-miner/test/ReputationMinerTestWrapper";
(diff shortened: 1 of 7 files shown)


elenadimitrova:
Thanks @dev1644 . It's a working solution with `ganache-cli` however as @area pointed out to me earlier this won't work with `parity` or `geth` as neither supports `evm_revert`. Sorry this wasn't explicitly in the description but we have a [working item to get our nightly builds to run against a production client here](https://github.com/JoinColony/colonyNetwork/issues/319) and using `evm_revert` will take us a step away from being able to achieve this.

Instead, can we try just replicating what the `before` section in mining tests does in terms of resetting the network. This is what I originally meant with [this](https://github.com/JoinColony/colonyNetwork/issues/524#issuecomment-479848147) comment although I see now that wasn't completely clear.

dev1644:
Ok, So replicating the `before` section will do the trick, right? I am on it.

test/reputation-mining/root-hash-submissions.js
```diff
@@ -95,7 +99,8 @@ contract("Reputation mining - root hash submissions", accounts => {
   });
 
   afterEach(async () => {
-    await finishReputationMiningCycleAndWithdrawAllMinerStakes(colonyNetwork, this);
+    const shouldSetupNewInstance = await finishReputationMiningCycleAndWithdrawAllMinerStakes(colonyNetwork, this);
+    if (!shouldSetupNewInstance) await newNetworkInstance(MINER1, MINER2, MINER3);
```
elenadimitrova: `shouldSetupNewInstance ` isn't representative of what the result is here, and so the next line doesn't read right too in that "if not shouldSetupNewInstance then setupNewInstance"..
Maybe simply rename to `success` or `reputationMiningClean`?

dev1644: You are right, `shouldSetupNewInstance` should be renamed.

elenadimitrova: And another thing `shouldSetupNewInstance` is always coming back as `undefined` irrespective of whether `finishReputationMiningCycleAndWithdrawAllMinerStakes` succeeded or failed. This should be changed to return boolean as the check now requires.

dev1644: All Fixed!

PR #584, opened 2019-04-05
Assorted cleanups and improvements

elenadimitrova:
This is my own personal running tap of bits that bother me all bundled up nicely here as none are big enough to attribute own issue and PR.

- Remove the funding and withdrawing of CLNY to the miner accounts. There's really no need for to-ing-and fro-ing with the CLNY deposit that enable a miner to propose a hash. I've left the helper function (`withdrawAllMinerStakes`) in there just in case we ever need it again.

- Use [ethers constants](https://docs.ethers.io/ethers.js/html/api-utils.html#constants) `HashZero` and `AddressZero`

- Naming is important!
`getReputationHashSubmissions` -> `getReputationHashSubmission`
`getSubmittedHashes` -> `getSubmissionUser`
`getDisputeRounds` -> `getDisputeRoundSubmission`
`getTaskWorkRatings` -> `getTaskWorkRatingSecretsInfo`

### DIFF
--- a/contracts/ColonyTask.sol
+++ b/contracts/ColonyTask.sol
@@ -296,7 +296,7 @@ contract ColonyTask is ColonyStorage {
     return keccak256(abi.encodePacked(_salt, _value));
   }
 
-  function getTaskWorkRatings(uint256 _id) public view returns (uint256, uint256) {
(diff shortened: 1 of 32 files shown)


area:
If we have a list of things to tell ColonyJS about, the function name changes should go on it :)

PR #586, opened 2019-04-10
Split dispute resolution properties away from the Submission struct

elenadimitrova:
Currently both `disputeRounds` and `reputationHashSubmissions` store the entire `Submission` struct. Where `reputationHashSubmissions` is only interested in the `proposedNewRootHash`, `nNodes`, `jrh`, `jrhNNodes` properties of that while `disputeRounds` manages the remainder of the properties through the dispute resolution process.

Here we introduce a `DisputedEntry` struct which is used only for dispute resolution properties while `Submission` is used for miner root hash submissions.

### DIFF
--- a/contracts/ReputationMiningCycleStorage.sol
+++ b/contracts/ReputationMiningCycleStorage.sol
@@ -34,8 +34,12 @@ contract ReputationMiningCycleStorage is ReputationMiningCycleDataTypes, DSAuth 
 
   ReputationLogEntry[] reputationUpdateLog; // Storage slot 6
   mapping (bytes32 => mapping(uint256 => mapping(bytes32 => address[]))) submittedHashes; // Storage slot 7
+  // Maps a submission with the first miner address who submitted it
+  // This address is used as a lookup key from DisputedEntry struct
   mapping (address => Submission) reputationHashSubmissions; // Storage slot 8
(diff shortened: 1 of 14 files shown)


contracts/ReputationMiningCycleStorage.sol
```diff
@@ -33,10 +33,14 @@ contract ReputationMiningCycleStorage is ReputationMiningCycleDataTypes, DSAuth
   address clnyTokenAddress; // Storage slot 5
 
   ReputationLogEntry[] reputationUpdateLog; // Storage slot 6
-  mapping (bytes32 => mapping( uint256 => mapping( bytes32 => address[]))) submittedHashes; // Storage slot 7
+  mapping (bytes32 => mapping(uint256 => mapping(bytes32 => address[]))) submittedHashes; // Storage slot 7
+  // Maps a submission with the first miner address who submitted it
```
area: I don't like this comment. The mapping is also used to see if someone has already made a submission in a cycle, and to check that if they make a second submission, it's for the same submission with a different entry.

What about something like "Maps the addresses of miners making submissions on to the submissions they made", and then going on to note (as the comment currently does) that this is also used to look up a disputed entry to a submission based on who first submitted it?

PR #587, opened 2019-04-11
Fix e2e tests after merging DLP

kronosapiens:
e2e test fixes including:
- updates to invocation of `addDomain` to include domain permissions params
- assign permissions for the task manager to be able to create and fund tasks
- ensure negative reputation updates (via low rated task work) are done after positive updates, as `PromiseAll` doesn't honour order of execution.

This also includes the addition of a convenience function in the miner client to allow the current reputation state to be printed to the console in the following format:
```
colonyAddress 0xd50038a999af0fd28d7a4e70893bb2ec16bdc0aa
userAddress 0x0021cb24d7d4e669120b139030095315dfa6699a
skillId 524
value 700
---------
colonyAddress 0xd50038a999af0fd28d7a4e70893bb2ec16bdc0aa
userAddress 0x27ff0c145e191c22c75cd123c679c3e1f58a4469
skillId 524
value 100
---------
colonyAddress 0x3fe71d088141cea5a213810ef4e18c5a48c840ce
userAddress 0x0000000000000000000000000000000000000000
skillId 504
value 2048
---------
```


helpers/test-data-generator.js
```diff
@@ -162,19 +162,14 @@ export async function setupFundedTask({
...
   const workerPayoutBN = new BN(workerPayout);
   const totalPayouts = managerPayoutBN.add(workerPayoutBN).add(evaluatorPayoutBN);
 
+  const childSkillIndex = await getChildSkillIndex(colonyNetwork, colony, 1, task.domainId);
   await colony.setFundingRole(1, 0, manager, 1, true);
-  let toSkillIndex = 0;
-  if (task.domainId !== 1) {
-    // Then we need to figure out the correct skill indices.
-    toSkillIndex = await getChildSkillIndex(colonyNetwork, colony, 1, task.domainId);
-  }
```
elenadimitrova: Removing this is what forced you into passing `colonyNetwork` in `makeTask` everywhere. This is because most tests use the root domain which has skill index of `0`

kronosapiens: Mm makes sense, but I'd advocate for keeping it the way it is in the PR for the following reasons:

1. It makes the `makeTask` interface consistent with the other task generation functions, all of which take `colonyNetwork` as the first argument.
2. It avoids the duplication of the domain-checking logic, which occurs already in `getChildSkillIndex`. Further, you could remove this check and _still_ pass an undefined `colonyNetwork` in cases where `domainId == 1` since `colonyNetwork` will never be used.
3. Further, I think that the logic which checks if the domains are the same should be hidden under the `getChildSkillIndex` abstraction layer; having that logic in `makeTask` seems like an unnecessary violation of abstraction boundaries.

That said, after merging #567 I'm pretty sure the only thing from this PR that's not elective are the updates to the `addDomain` invocations, so it's whatever you want really.

elenadimitrova:
This wasn't anywhere near ready-to-review as the `addDomain` calls [here](https://github.com/JoinColony/colonyNetwork/pull/587/commits/ad2f011aa551a779877c5018e2b3aaa6604ea0a1#diff-650d66e1f594732736411846489a5851L141) weren't fixed and the permissions [here](https://github.com/JoinColony/colonyNetwork/pull/587/commits/ad2f011aa551a779877c5018e2b3aaa6604ea0a1#diff-650d66e1f594732736411846489a5851R248) were missing.
Also I am going to lower the story points estimate down to `3` as an `8` implies a complexity I don't see here. This took me around 2-3h at most and came down to fixing the 8 `addDomain` calls in end2end tests file plus 2 missing permissions for the task manager in same test. 
Remember that story points estimates are an indicator of _relative_ complexity not an actual representation of how long it took you to do it. Irrespective of who completes the work the estimate should be the same. In the same line of thought we should start estimating collectively in our regular bi-weekly meeting on Tuesday as to avoid this in the future.

kronosapiens:
Mm, those changes came after rebasing onto #567 which is a bit annoying as the `addDomain` calls there should have had the correct invocation in the first place. Also fine, 3 points seems fair although I'll note that the poor isolation of the tests in this suite meant that I spent a whole morning debugging the wrong function. My mistake for not going back and double-checking the e2e tests after getting the build fixed, but I'll note that this whole PR was basically in a race condition with #567.

PR #591, opened 2019-04-16
Add factories for extension contracts

area:
After talking to Chris about what would be useful to exist on-chain regarding extension contracts, we've arrived at this middle ground. In short, there needs to be a way for the dapp to work out if a supported extension is installed on a colony. We could have modified the extensions to be singletons on-chain that all colonies used, but establishing that precedent could prove dangerous if more complicated extensions were developed in the future, as the singleton would potentially have very many permissions across very many colonies, and so a bug in one could prove catastrophic.

The alternative we went for was to create a factory for each of the supported extensions. These provide a way to easily deploy an extension for a colony (just call the `deployExtension` function), as well as an easy way to look up the address of an extension for a particular colony (call the `deployedExtensions` getter).

Notes:

* I have used very generic names for the extension factory functions. This is so that in the future, when there are very many extensions (we hope) and factories, it's easier to develop against - we always know the names of these functions, rather than having to remember it's `deployOneTxExtension`, `deployOldRolesExtension` and so on.

* Adding/removing an extension from a colony is restricted to those with root permissions on the colony.

* I have deployed the factory during migrations so it's easier for people to develop against.

### DIFF
--- a/contracts/extensions/ExtensionFactory.sol
+++ b/contracts/extensions/ExtensionFactory.sol
@@ -1,0 +1,25 @@
+/*
+  This file is part of The Colony Network.
+
+  The Colony Network is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  The Colony Network is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.5.3;
+pragma experimental ABIEncoderV2;
+
+
+interface ExtensionFactory {
+  function deployExtension(address _colony) external;
+  function removeExtension(address _colony) external;
+}
--- a/contracts/extensions/ExtensionFactory.sol
+++ b/contracts/extensions/ExtensionFactory.sol
@@ -22,4 +22,6 @@ pragma experimental ABIEncoderV2;
 interface ExtensionFactory {
   function deployExtension(address _colony) external;
   function removeExtension(address _colony) external;
+  event ExtensionDeployed(string _name, address _colony, address _extension);
+  event ExtensionRemoved(string _name, address _colony);
 }
--- a/contracts/extensions/OldRolesFactory.sol
+++ b/contracts/extensions/OldRolesFactory.sol
@@ -21,13 +21,14 @@ pragma experimental ABIEncoderV2;
 import "./../ColonyDataTypes.sol";
 import "./../IColony.sol";
 import "./../ColonyAuthority.sol";
+import "./ExtensionFactory.sol";
 import "./OldRoles.sol";
 
 
-contract OldRolesFactory is ColonyDataTypes {
+contract OldRolesFactory is ExtensionFactory, ColonyDataTypes {
   mapping (address => OldRoles) public deployedExtensions;	
 
-  function deployExtension(address _colony) public {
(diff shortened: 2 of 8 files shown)


contracts/extensions/OldRolesFactory.sol
```diff
@@ -0,0 +1,48 @@
...
+pragma solidity >=0.5.3;
+pragma experimental ABIEncoderV2;
+
+import "./../ColonyDataTypes.sol";
+import "./../IColony.sol";
+import "./../ColonyAuthority.sol";
+import "./OldRoles.sol";
+
+
+contract OldRolesFactory is ColonyDataTypes {
```
elenadimitrova: I wish we had type inference in Solidity when only one factory would have sufficed. I vaguely remember reading something about this in v0.5.0 breaking changes or a more recent release but couldn't find it this morning.

In the absence of this 👆 if we can at least have an interface which these factories to comply with (my preference would be for the actual `interface` keyword and `external` functions it mandates).

test/extensions/one-tx-payment.js
```diff
@@ -37,14 +41,34 @@ contract("One transaction payments", accounts => {
     await colony.setAdministrationRole(1, 0, COLONY_ADMIN, 1, true);
     await colony.setFundingRole(1, 0, COLONY_ADMIN, 1, true);
 
-    oneTxExtension = await OneTxPayment.new(colony.address);
+    await oneTxExtensionFactory.deployExtension(colony.address);
+    const oneTxExtensionAddress = await oneTxExtensionFactory.deployedExtensions(colony.address);
+    oneTxExtension = await OneTxPayment.at(oneTxExtensionAddress);
```
elenadimitrova: Can we add a test to check the `oneTxExtensionAddress` matches the new entry in Factory

elenadimitrova: Same for `OldRolesFactory` tests as above

area: I'm not really sure how you want me to do this that is different from what this already does - I would just be calling `oneTxExtensionFactory.deployedExtensions` again and checking it gave the same result?

elenadimitrova: It'll be nice to check the contract created in `deployExtension` transaction (via the `contractAddress` property) matches `oneTxExtensionAddress`. Alternatively (as we do elsewhere in the network) we can trigger an event with that address which we check?

contracts/extensions/ExtensionFactory.sol
```diff
@@ -0,0 +1,25 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity >=0.5.3;
+pragma experimental ABIEncoderV2;
+
+
+interface ExtensionFactory {
```
elenadimitrova: 👍

test/extensions/one-tx-payment.js
area: I have ended up going the events route for consistency.

contracts/extensions/OldRolesFactory.sol
```diff
@@ -0,0 +1,51 @@
...
+import "./ExtensionFactory.sol";
+import "./OldRoles.sol";
+
+
+contract OldRolesFactory is ExtensionFactory, ColonyDataTypes {
+  mapping (address => OldRoles) public deployedExtensions;	
+
+  function deployExtension(address _colony) external {
+    require(
+      ColonyAuthority(IColony(_colony).authority()).hasUserRole(msg.sender, 1, uint8(ColonyRole.Root)) == true,
```
kronosapiens: Doesn't matter a ton (also this is merged already) but could also use the `hasUserRole` implemented in the Colony itself, which does basically the same thing. If we're going to have this require in every Factory it would simplify the code to:

```
require(
      IColony(_colony).hasUserRole(msg.sender, 1, ColonyRole.Root),
      "colony-extension-user-not-root"
)
```

Also you don't need the `== true` here, since the return value is already a boolean.

PR #603, opened 2019-05-03
Docs/glider

gichiba:
Well, closed #600 but that's merged already! 

This update adds: 

- a short docs page on extensions 
- a short docs page on modular permissions
- logo for the glider release
- payments to the TL;DR
- an updated colonyNetwork inheritance diagram
-some misc broken link fixes


gichiba:
Needs to be double checked after deployment with addresses added

PR #612, opened 2019-05-14
Fix broken link to whitepaper

gichiba:
Hmm, this is a messy workflow for really tiny docs updates. It would be nice, however, to figure out a way to streamline tiny changes to docs like this, that might come from external contributors. 

colonyJS doesn't really have this problem because it moves so fast, but for the network it seems preferable to keep a running branch of typos and fixes to the docs that doesn't waste circleci resources, and doesn't require such thorough checking before making it into `develop`


elenadimitrova:
Repo admins in GitHub can merge without passing all checks so we can expedite docs changes like this. You can also skip Circle builds with adding `[skip ci]` to the end of the commit message.

PR #615, opened 2019-05-18
Update role change events to include domainId

kronosapiens:
1. Adds `domainId` as argument to role-change events.
2. Make the role-change event generic and add the `uint8(ColonyRole)`.
3. Introduce `RecoveryRoleSet` event for recovery role changes.

### DIFF
--- a/contracts/ColonyDataTypes.sol
+++ b/contracts/ColonyDataTypes.sol
@@ -36,21 +36,12 @@ contract ColonyDataTypes {
   /// @param newVersion The new colony version upgraded to
   event ColonyUpgraded(uint256 oldVersion, uint256 newVersion);
 
-  /// @notice Event logged when a new user is assigned the colony funding role
-  /// @param user The newly added colony funding user address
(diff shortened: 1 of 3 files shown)


contracts/ColonyDataTypes.sol
```diff
@@ -36,21 +36,12 @@ contract ColonyDataTypes {
...
-
-  /// @notice Event logged when a new user is assigned the colony root role
-  /// @param user The newly added colony root user address
-  event ColonyRootRoleSet(address user, bool setTo);
+  /// @notice Event logged when a user/domain/role is granted or revoked
+  /// @param user The address of the user being affected
+  /// @param domainId The damainId of the role
+  /// @param role The role being granted/revoked
+  /// @param setTo A boolean representing the action -- granted (`true`) or revoked (`false`)
+  event ColonyRoleSet(address indexed user, uint256 indexed domainId, uint8 indexed role, bool setTo);
```
elenadimitrova: We should probably emit this event for `setRecoveryRole` and `removeRecoveryRole` as well?

PR #616, opened 2019-05-19
Upgrade to latest solc 0.5.8

elenadimitrova:
Also due to https://github.com/dapphub/ds-note/issues/4 removed `DSToken` from tests and instead use instances of our own `Token`

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -34,7 +34,7 @@ module.exports = {
     }
   },
   mocha: {
-    reporter: "eth-gas-reporter",
+    reporter: "mocha-circleci-reporter",
     reporterOptions: {
       currency: "USD",
       gasPrice: 5,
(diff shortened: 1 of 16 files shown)


truffle.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
     }
   },
   mocha: {
-    reporter: "mocha-circleci-reporter",
+    reporter: "eth-gas-reporter",
```
area: Mistakenly commited?

elenadimitrova: yes :(

elenadimitrova:
Not a bad idea with the `UnlockedToken` although I would rather be explicit with the Token model we have unless you mind terribly?

area:
No, was just wondering. It'd have pros and cons, so happy to leave as-is.

PR #621, opened 2019-05-23
RepCycle Visualizer Improvements

kronosapiens:
- This morning I realized that using the block timestamp to figure out the remaining time in the mining cycle is probably more relevant than the system timestamp. Not a big change but improves the semantics slightly.
- Changed the denomination of the reputation amount in the cycle visualizers from Ether to Wei, to match the tree visualizer.
- Realized I could drop the URI host/port and make requests locally by default (so requests are port-agnostic).
- Finally, I figured out how to cleanly bring up the active cycle on load. 💪


elenadimitrova:
Works ok for me here. Good to merge.

PR #622, opened 2019-05-23
Generate documentation for interface contracts

ryanchristo:
## Description

This pull request adds a script that generates documentation for all the interface contact methods.

Screen Shot 2019-05-23 at 6 21 03 PM


docs/_Interface_IColonyNetwork.md
```diff
@@ -0,0 +1,858 @@
...
+|node|bytes32| The namehash of the ENS address being requested
+
+
+
+**Return Parameters**
+
+
+|Name|Type|Description|
+|---|---|---|
+|_undefined_|address|
```
elenadimitrova: Badly parsed return parameter

docs/_Interface_IColonyNetwork.md
```diff
@@ -0,0 +1,858 @@
...
+|_tokenAddress|address| Address of an ERC20 token to serve as the colony token
+
+
+
+**Return Parameters**
+
+
+|Name|Type|Description|
+|---|---|---|
+|_undefined_|address|
```
elenadimitrova: Badly parsed return parameter

docs/_Interface_IColonyNetwork.md
```diff
@@ -0,0 +1,858 @@
...
+
+
+
+
+**Return Parameters**
+
+
+|Name|Type|Description|
+|---|---|---|
+|_undefined_|address|
```
elenadimitrova: Badly parsed return parameter

docs/_Interface_IColonyNetwork.md
```diff
@@ -0,0 +1,858 @@
...
```
elenadimitrova: Badly parsed return parameter

docs/_Interface_IColonyNetwork.md
ryanchristo: What would you rather have here if there is no name for the variable?

ryanchristo:
How would you like the @dev notes to be included? Maybe something like this?

*Developer Note: Used for testing.*

docs/_Interface_IColonyNetwork.md
elenadimitrova: I would rather have the description from natspec i.e. `The address the supplied node resolves to`

ryanchristo:
```
function addr(bytes32 node) public view returns (address);
```

ryanchristo: Currently it checks the name of the parameter. I can make it check the type if there is no name provided (in this case it would use `address`) or we can add a name to the return parameters that are missing a name.

elenadimitrova:
```
   /// @notice Returns the address the supplied node resolves do, if we are the resolver
  /// @param node The namehash of the ENS address being requested
  /// @return address The address the supplied node resolves to
  function addr(bytes32 node) public view returns (address);
```

elenadimitrova: I was questioning the return parameter description rather than name or type.

ryanchristo: In order to verify that the description is the right description for the given param, it checks the first word in the description (currently the name of the param).

ryanchristo: I can make it check the type if no name is provided. It would solve this particular case.

elenadimitrova: We can add names to all return parameters. Most should be there already so this is more of an exception

docs/_Interface_IColonyNetwork.md
ryanchristo: This one neither matches the name (no name) or type (`address`)

```
  /// @return colonyAddress Address of the newly created colony
  function createColony(address _tokenAddress) public returns (address);
```

docs/_Interface_IColonyNetwork.md
ryanchristo: This one neither matches the name (no name) or type (`address`)

```
  /// @return colonyAddress The Meta colony address, if no colony was found, returns 0x0
  function getMetaColony() public view returns (address payable);
```

docs/_Interface_IColonyNetwork.md
ryanchristo: I think adding names to the return parameters would be nice. It falls back to the type if no name is provided. That resolved this specific case and one other point you pointed out.

ryanchristo:
Ok. I have ensured that all the natspec comments are being accurately parsed including `@dev` and natspec comments that span multiple lines. There are some minor punctation issues that need to be resolved in the interface contracts such as the use of periods at the end of comments and back ticks when referencing a `_variable`.

The script runs and adds the documentation on precommit, so no need to worry about changing your workflow. The script also logs a warning for any methods that do not have natspec comments.

```
WARNING: setResolver is missing a natspec notice
WARNING: setOwner is missing a natspec notice
WARNING: setAuthority is missing a natspec notice
WARNING: respondToChallenge is missing a natspec notice
```

docs/_Interface_IColonyNetwork.md
elenadimitrova: That's a mistake in the natspec really. I will fix

docs/_Interface_IColonyNetwork.md
elenadimitrova: I'll also fix this.

elenadimitrova:
I like the idea of using the commit hooks to process the docs, but I favour the postcommit hook for that rather than the precommit.

docs/_Interface_ITokenLocking.md
```diff
@@ -65,7 +65,7 @@ Get user token lock info (lock count and deposited amount)
 
 |Name|Type|Description|
 |---|---|---|
-|lock|Lock|
+|lock|Lock|Lock object containing
```
elenadimitrova: This is also missing the multiline comment

docs/_Interface_IReputationMiningCycle.md
```diff
@@ -352,22 +352,23 @@ Respond to a binary search step, to eventually discover where two submitted hash
...
-|u|uint256[29]|
-|b32|bytes32[8]|
-|reputationSiblings|bytes32[]|
-|agreeStateSiblings|bytes32[]|
-|disagreeStateSiblings|bytes32[]|
-|previousNewReputationSiblings|bytes32[]|
-|userOriginReputationSiblings|bytes32[]|
-|childReputationSiblings|bytes32[]|
-|adjacentReputationSiblings|bytes32[]|
+|u|uint256[29]|A `uint256[27]` array. The elements of this array, in order are:
```
elenadimitrova: This is missing the multiline comment for the `u` parameter

docs/_Interface_IReputationMiningCycle.md
```diff
@@ -360,8 +360,8 @@ Respond to challenge, to establish which (if either) of the two submissions faci
...
...ated is in the agree state * 21. The amount of reputation that the child reputation for the user being updated is in the agree state * 22. The UID of the child reputation for the user being updated in the agree state * 23. A dummy variable that should be set to 0. If nonzero, transaction will still work but be slightly more expensive. For an explanation of why this is present, look at the corresponding solidity code. * 24. The branchMask of the proof that the reputation adjacent to the new reputation being inserted is in the agree state * 25. The amount of reputation that the reputation adjacent to a new reputation being inserted has in the agree state * 26. The UID of the reputation adjacent to the new reputation being inserted * 27. A dummy variable that should be set to 0. If nonzero, transaction will still work but be slightly more expensive. For an explanation of why this is present, look at the corresponding solidity code. * 28. The value of the reputation that would be origin-adjacent that proves that the origin reputation does not exist in the tree * 29. The value of the reputation that would be child-adjacent that proves that the child reputation does not exist in the tree
```
elenadimitrova: The way this parses is not ideal as new lines aren't honoured but I'm not too fussed about it. At least we have _something_ to show in the docs..

ryanchristo: Yea. This is not ideal. I think we can still make some improvements but it's a good start.

PR #629, opened 2019-05-31
Polish off the mining client and make it more resillient

elenadimitrova:
- Add try/catch-es in mining client
- On restart, recover any submissions made
- Add test resuming submissions after making some
- Save historical states after syncing the client

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,7 +20,8 @@
     "no-await-in-loop": "off",
     "no-trailing-spaces": "error",
     "no-multi-spaces": ["error", {"ignoreEOLComments": true}],
-    "eslint-comments/disable-enable-pair": ["error", {"allowWholeFile": true}]
+    "eslint-comments/disable-enable-pair": ["error", {"allowWholeFile": true}],
+    "space-infix-ops": ["error"]
   },
   "parserOptions": {
     "ecmaVersion": 2017
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -612,8 +612,14 @@ export async function finishReputationMiningCycle(colonyNetwork, test) {
   // Finish the current cycle. Can only do this at the start of a new cycle, if anyone has submitted a hash in this current cycle.
   const repCycle = await getActiveRepCycle(colonyNetwork);
   const nUniqueSubmittedHashes = await repCycle.getNUniqueSubmittedHashes();
+
   if (nUniqueSubmittedHashes.gtn(0)) {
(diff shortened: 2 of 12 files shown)


elenadimitrova:
After adding the one new test of the miner  here builds started failing on `should successfully complete a hash submission if there are 2 good miners` test. Тhis was due to the following redundant`goodClient` calls
```
await goodClient.addLogContentsToReputationTree();
await goodClient.saveCurrentState();
```

When logging more extensively the above two calls appeared to duplicate what had already been done in `beforeEach` so removing them also removed what looked like a race condition to process the log.

area:
> In theory, theory and practise are the same. In practise... 

While running the miner, every so often it would stop silently. After some digging, I realised that this was because the event listener for the `ReputationMiningCycleComplete` event sometimes missed the event, which meant that `twelveBestSubmissions` was empty and never refreshed, leading to the `onBlock` listener not doing anything. I'm not really sure what the cause of this is - it's clearly something to do with how it's talking to parity, and either a problem with how we're using it, or ethers or parity. There is some weirdness going on at `ethers.js` regarding events, as people keep reporting related issues, but they are mostly with ganache rather than a real client/chain as far as I can see. This is one of those things that we'll have to figure out eventually, but for now we can work around it.

Instead of relying on the events, it now tracks the address of the active `reputationMiningCycle` contract, and when that changes deduces that a new mining cycle has started and does all the updates we had contingent on the event.

In order to improve the devops of the reputation miner, I've added two 'sanity checks'. The first checks that a block is seen at least every five minutes. Obviously we should see them much faster than this, and I originally had it at sixty seconds, but we were occasionally seeing the message. I think this was because our parity sometimes does not see new blocks immediately. The second check is that a reputation cycle is completed every twenty four hours (with ten minute grace period to allow for the transaction in question to be mined).

These sanity checks are reported by default in the console, but by passing `--adapter slack` as a command line parameter and setting the `SLACK_BOT_TOKEN` and `SLACK_CHANNEL` env variables, it will report those failed sanity checks in Slack as well as any other errors it comes across. When using the slack adapter, expected (non-error) output is still put in the console.

As well as the above, which are added by my latest commits, and what Elena mentioned in the original message, this PR...

* Optionally kills the miner in the event of an error, which given we're running in kubernetes should restart it.
* Optionally splits up the oracle service and the mining client service, which means that the oracle (exposed to the world) no long has private keys in memory.

area:
I'm pretty confident this gets us to a good place with the miner, so merging for now. Also closes #662.

PR #630, opened 2019-05-31
Arbitration Role in IColony

gichiba:
This PR would add a `setArbitrationRole` function to a colony, allowing users to assign the role even though there are no functions in the current glider release that make use of the ARBITRATION role. 

There are reasons. First, the arbitration role is described as a "placeholder" for dispute resolution. When voting or disputes are implemented, these changes will need to be put in anyways. Not exposing a flaccid function is an understandable design decision, but it leaves the arbitration role in a weird state of being in principle useful on lower levels, but inaccessible. 

More importantly (and more generally), developers working on extensions might want to utilize a colony's permissions system separately with other contracts that are related to the colony (read: as a module). Checking colony permissions before executing some hypothetical extension function should be possible with the other roles, so why not allow the Arbitration role to be utilized for extensions? 

I would argue that exposing the arbitration role for extension contracts is a good devrel-driven design decision, because then devs have some free reign to develop novel rules around a role that (for now) doesn't also have implications elsewhere in the colony imposed by payments, tasks, etc... I can imagine developers wanting to implement their own alternative voting / dispute systems as an extension, and using the arbitration role to connect them to their colony, for example. 

--

Okay, now that I've made my case, I should say that these changes made are ones that I know are necessary, but I'm not sure they're sufficient for a PR. I'm particularly in the dark about what I need to do regarding tests and anything not in the `contracts/` folder. I suppose it's also possible that I've missed something entirely that makes this not work, or don't see some other objection originating from room 3. plz halp.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -76,6 +76,19 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit ColonyRoleSet(_user, _domainId, uint8(ColonyRole.Administration), _setTo);
   }
 
+  function setArbitrationRole(
+    uint256 _permissionDomainId,
+    uint256 _childSkillIndex,
+    address _user,
+    uint256 _domainId,
+    bool _setTo
+  ) public stoppable authDomain(_permissionDomainId, _childSkillIndex, _domainId)
+  {
+    ColonyAuthority(address(authority)).setUserRole(_user, _domainId, uint8(ColonyRole.Arbitration), _setTo);
+
+    emit ColonyRoleSet(_user, _domainId, uint8(ColonyRole.Arbitration), _setTo);
+  }
+
   function hasUserRole(address _user, uint256 _domainId, ColonyRole _role) public view returns (bool) {
     return ColonyAuthority(address(authority)).hasUserRole(_user, _domainId, uint8(_role));
   }
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -34,6 +34,19 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit ColonyRoleSet(_user, 1, uint8(ColonyRole.Root), _setTo);
   }
 
+  function setArbitrationRole(
+    uint256 _permissionDomainId,
+    uint256 _childSkillIndex,
+    address _user,
+    uint256 _domainId,
+    bool _setTo
+  ) public stoppable authDomain(_permissionDomainId, _childSkillIndex, _domainId)
+  {
+    ColonyAuthority(address(authority)).setUserRole(_user, _domainId, uint8(ColonyRole.Arbitration), _setTo);
+
+    emit ColonyRoleSet(_user, _domainId, uint8(ColonyRole.Arbitration), _setTo);
+  }
+
   function setArchitectureRole(
     uint256 _permissionDomainId,
     uint256 _childSkillIndex,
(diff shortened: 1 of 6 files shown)


kronosapiens:
Re: testing, I'd say add the following for now:

Uncomment: https://github.com/JoinColony/colonyNetwork/blob/develop/test/contracts-network/colony-permissions.js#L61

Give User2 the arbitration role somewhere here: https://github.com/JoinColony/colonyNetwork/blob/develop/test/contracts-network/colony-permissions.js#L227

contracts/ColonyAuthority.sol
```diff
@@ -57,16 +57,20 @@ contract ColonyAuthority is CommonAuthority {
     addRoleCapability(ROOT_ROLE, "setArchitectureRole(uint256,uint256,address,uint256,bool)");
     addRoleCapability(ROOT_ROLE, "setFundingRole(uint256,uint256,address,uint256,bool)");
     addRoleCapability(ROOT_ROLE, "setAdministrationRole(uint256,uint256,address,uint256,bool)");
+    addRoleCapability(ROOT_ROLE, "setArbitrationRole(uint256,uint256,address,uint256,bool)");
```
kronosapiens: Put this below Root

test/contracts-network/colony-permissions.js
```diff
@@ -58,13 +59,14 @@ contract("ColonyPermissions", accounts => {
     it("should give colony creator all permissions in root domain", async () => {
       const fundingRole = await colony.hasUserRole(FOUNDER, 1, FUNDING_ROLE);
       const administrationRole = await colony.hasUserRole(FOUNDER, 1, ADMINISTRATION_ROLE);
-      // const arbitrationRole = await colony.hasUserRole(FOUNDER, 1, ARBITRATION_ROLE); Not implemented yet.
+      const arbitrationRole = await colony.hasUserRole(FOUNDER, 1, ARBITRATION_ROLE);
```
kronosapiens: Make these:

```
root
arbitration
architecture
funding
administration
```

test/contracts-network/colony-permissions.js
```diff
@@ -58,13 +59,14 @@ contract("ColonyPermissions", accounts => {
...
       const architectureRole = await colony.hasUserRole(FOUNDER, 1, ARCHITECTURE_ROLE);
       const architectureSubdomainRole = await colony.hasUserRole(FOUNDER, 1, ARCHITECTURE_SUBDOMAIN_ROLE);
       const rootRole = await colony.hasUserRole(FOUNDER, 1, ROOT_ROLE);
 
       expect(fundingRole).to.be.true;
       expect(administrationRole).to.be.true;
+      expect(arbitrationRole).to.be.true;
```
kronosapiens: Same order as above

test/contracts-network/colony-permissions.js
```diff
@@ -223,6 +225,7 @@ contract("ColonyPermissions", accounts => {
 
       // Can create manage permissions in the root domain!
       await colony.setFundingRole(1, 0, USER2, 1, true, { from: USER1 });
+      await colony.setArbitrationRole(1, 0, USER2, 1, true, { from: USER1 });
```
kronosapiens: Same order as above

contracts/IColony.sol
```diff
@@ -88,6 +88,15 @@ contract IColony is ColonyDataTypes, IRecovery {
   /// @param _setTo The state of the role permission (true assign the permission, false revokes it)
   function setAdministrationRole(uint256 _permissionDomainId, uint256 _childSkillIndex, address _user, uint256 _domainId, bool _setTo) public;
 
+  /// @notice Set new colony arbitration role.
```
kronosapiens: Put this below setRootRole

contracts/ColonyNetwork.sol
```diff
@@ -156,6 +156,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
     // Assign all permissions in root domain
     colony.setRecoveryRole(msg.sender);
     colony.setRootRole(msg.sender, true);
+    colony.setArbitrationRole(1, 0, msg.sender, 1, true);
```
kronosapiens: Good ordering 💪

contracts/Colony.sol
```diff
@@ -76,6 +76,19 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit ColonyRoleSet(_user, _domainId, uint8(ColonyRole.Administration), _setTo);
   }
 
+  function setArbitrationRole(
```
kronosapiens: Put below setRootRole

PR #596, opened 2019-04-22
Real world mining client

elenadimitrova:
Implements the following parts of #211 

- Submits a root hash as soon as it is able to, (rather than wait for the end of the submission period which isn't feasible when working with multiple submissions). It does so by checking at every block if there are less than 12 submissions of the current hash/nNodes/jrh and submits if there's still room

- Confirms the last hash after the mining window closes

- Implements a listener for `ReputationMiningCycleComplete` event for when a mining cycle completes so it processes the new update log

PR #631, opened 2019-06-02
Update mining docs for Goerli, better errors and logging for miner

kronosapiens:
@gichiba recently drew my attention to the fact that the current [mining docs](https://docs.colony.io/colonynetwork/docs-reputation-mining-client) don't include instructions for Goerli. After the merge of #596 I figured I'd try and spin up the minimum viable mining client. After a few stumbles I figured it out and have updated the docs with new instructions, as well as added a bit more logging.

At first I thought it would be nice to have the miner sync from block 1 by default, but unfortunately a quirk with event logging means that the first `ReputationMiningCycleComplete` event is emitted before the first mining cycle is deployed, meaning that either an exception is raised for the missing mining cycle, or we ignore something that should really raise an exception, which would be asking for trouble down the road. So this PR adds the exception, which is the Right Thing To Do.

So, since we can't sync from block 1, the next thought was wouldn't it be fun to tell the mining client about the deployment details (such as the Colony Network address, what block to sync from, etc). That seemed like a good idea at first, but after some reflection it became clear that there are lots of ways to start a mining client (with an infura node, with a local ethereum client, with a local testnet), and making the miner smart enough to "know what to do" with all of these different configurations turned into more of a production that we bargained for. Instead, let's keep the client dumb as bricks and update the docs with the information that we otherwise might have baked into the client.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -123,10 +123,9 @@ class ReputationMiner {
 
     this.justificationHashes = {};
     this.reverseReputationHashLookup = {};
-    const addr = await this.colonyNetwork.getReputationMiningCycle(true, { blockTag: blockNumber });
-    const repCycle = new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+    const repCycle = await this.getActiveRepCycle(blockNumber)
 
-    if (addr === `0x${new BN(0).toString(16, 40)}`) {
+    if (repCycle.address === ethers.constants.AddressZero) {
       console.log(`WARNING: No active mining cycle found for block number ${blockNumber}`);
       return;
     }
@@ -467,8 +466,7 @@ class ReputationMiner {
    */
   async getLogEntryNumberForLogUpdateNumber(_i, blockNumber = "latest") {
     const updateNumber = _i;
-    const addr = await this.colonyNetwork.getReputationMiningCycle(true, { blockTag: blockNumber });
-    const repCycle = new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+    const repCycle = await this.getActiveRepCycle(blockNumber)
     const nLogEntries = await repCycle.getReputationUpdateLogLength({ blockTag: blockNumber });
     let lower = ethers.constants.Zero;
     let upper = nLogEntries.sub(1);
@@ -498,8 +496,7 @@ class ReputationMiner {
     }
     // Else it's from a log entry
     const logEntryNumber = await this.getLogEntryNumberForLogUpdateNumber(updateNumber.sub(this.nReputationsBeforeLatestLog), blockNumber);
-    const addr = await this.colonyNetwork.getReputationMiningCycle(true, { blockTag: blockNumber });
-    const repCycle = new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+    const repCycle = await this.getActiveRepCycle(blockNumber)
 
     const logEntry = await repCycle.getReputationUpdateLogEntry(logEntryNumber, { blockTag: blockNumber });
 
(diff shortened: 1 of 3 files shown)


packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -153,12 +153,12 @@ class ReputationMinerClient {
...
         }
       });
-      
+
       this._miner.realProvider.polling = true;
       this._miner.realProvider.pollingInterval = 1000;
 
       // Do the other checks for whether we can submit or confirm a hash
-      lockedForBlockProcessing = false;        
+      lockedForBlockProcessing = false;
```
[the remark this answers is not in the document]
kronosapiens: Fair, I updated both the `eslint` and `solium` linters in #634

kronosapiens:
@elenadimitrova I've updated the PR with https://github.com/JoinColony/colonyNetwork/pull/631/commits/32123ec2c69d8cf175cf8b71750422bf44296834

elenadimitrova:
Posting synchronisation results here for posterity. It took my client just under 2 hours to sync all 28 mining cycles locally (I used the `-- syncFrom 548534 ` option although I don't think that makes a difference to the sync time as noted in this PR.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -123,8 +123,12 @@ class ReputationMiner {
 
     this.justificationHashes = {};
     this.reverseReputationHashLookup = {};
-    const addr = await this.colonyNetwork.getReputationMiningCycle(true, { blockTag: blockNumber });
-    const repCycle = new ethers.Contract(addr, this.repCycleContractDef.abi, this.realWallet);
+    const repCycle = await this.getActiveRepCycle(blockNumber)
+
+    if (repCycle.address === ethers.constants.AddressZero) {
+      console.log(`WARNING: No active mining cycle found for block number ${blockNumber}`);
+      return;
```
elenadimitrova: Continuing [this](https://github.com/JoinColony/colonyNetwork/pull/631#commitcomment-33803015) discussion in context here. 
On the topic of handling control flow when getting the active mining cycle fails, looking through the instances of `getActiveRepCycle` in the miner, I think it might be more appropriate to `throw` an error instead of just `return`. In the Network there should always be an active mining cycle, except for when it is asked for from an old block before mining process has started (i.e. `blockNumber < 548534` on Goerli) which I can't see being possible with the current code, but I need to look harder later.
Was there a specific case you had in mind when you introduced the `return` here?

kronosapiens: Yup! The first time you call `startNextCycle`, it emits a `ReputationMiningCycleComplete` event, but there is actually no mining cycle in the previous block. So if you `sync` from block 1, or even from the block where the network was deployed (which seems like a reasonable behavior), the first block that `sync` looks at has no mining cycle. If we were to `throw` there, then the sync would stop, while the `return` functions more like a `continue`.

An alternative would be to tweak how the mining cycle emits events so `ReputationMiningCycleComplete` is emitted closer to where the previous cycle is destroyed, maybe even in `confirmNewHash` if possible.

kronosapiens: It may not be worth looking at this close to the finish line but glancing at `initialiseReputationMining` and `startNextCycle` it does seem like there's similar logic which could be merged / re-arranged.

elenadimitrova: Considering the mining cycle has already been initialised on Goerli and very close to being initialised on mainnet, reengineering the initialisation of the mining process is probably a very unlikely option.
Instead perhaps we can change the default block number for syncing to be the correct one when mining was initialised, i.e. `548534` for goerli.

kronosapiens: Ok, I can live with that.

kronosapiens: Here's what I've got: https://github.com/JoinColony/colonyNetwork/pull/631/commits/f332c8d18fd75a616415ee922c4721851185e1b8

I'm certain that there is a more elegant way to handle the error messages in `index.js` than what I did, if anyone has a suggestion.

kronosapiens:
So since we cannot sync from block 1, and further since there are many different ways people can start the mining client, I now feel that the best way forward is to keep the client dumb and robust, and include specific invocation details (such as network-specific `colonyNetworkAddress` and `syncFrom` details) in the docs.

PR #633, opened 2019-06-04
Preserve ENS root node ownership

elenadimitrova:
Here we remove the ownership transfer of the `joincolony.eth` root ENS node to the colonyNetwork. We'd like to preserve the ownership to allow us to set other subdomains. Ultimately, the colonyNetwork can own the root node, but we decided it is premature ceding control in the first release.

Note that after this change, the subdomains for `user` and `colony` will need to be explicitly called by the owner.


test/contracts-network/colony-network.js
```diff
@@ -287,9 +288,14 @@ contract("Colony Network", accounts => {
     let ensRegistry;
 
     beforeEach(async () => {
+      const USER_HASH = await soliditySha3("user");
+      const COLONY_HASH = await soliditySha3("colony");
```
kronosapiens: These constants can probably be bumped one scope higher, into the `describe` block.

elenadimitrova: `describe` block is not `async` and these are. Making the `describe` async is problematic I think because of some default mocha behaviour.

kronosapiens: TIL.

area: `soliditySha3` surely isn't async?

kronosapiens: Yes good point, the other usages don't `await`.

https://github.com/JoinColony/colonyNetwork/blob/fa71fe33cf34e65ecd0f67cb622c504aa3736f9e/helpers/constants.js#L44

elenadimitrova: Yes you're right, I must be assuming everything in `web3` is `async` since v1. 
I've gone further here and extracted the common logic of setting up a new `ENSRegistrar` into a helper which is reused.
I've also removed the other one reference where we await the `soliditySha3` call.

PR #667, opened 2019-07-03
Allow users and colonies to update their OrbitDB store address

area:
Because some usernames have been registered already on mainnet passing dummy orbitDB values in to the `registerUserLabel`/`registerColonyLabel` functions, in order for those users / colonies to be able to use the dApp correctly, we need to let them update the orbitDB store values to correct ones. How the users extract the correct values from the dApp to pass to these functions is outside the scope of this PR and, indeed, repository.

The functionality itself was quite straightforward to implement, as were the 'normal' tests to check that the functionality works. The bulk of my effort in this PR was around the tests that check that an upgrade of an existing colony works as expected. These tests are run separately from the 'normal' tests via `yarn run:test:contracts:upgrade:2to3`. They:

* Start up ganache as a fork of mainnet at a recent(ish) block (this relies on Infura, so if they go down, the test breaks...)
* Check functionality doesn't exist before upgrading Colony / ColonyNetwork
* Upgrade ColonyNetwork / Colony and check new functionality works.

Because of the bug preventing us deploying new contracts and interacting with them in a single transaction on a ganache-forked chain, I skip migration 8 in these tests by renaming it to not be picked up by the truffle regex. It is moved back on successful completion of the tests. Once the bug is fixed, this dance can be removed.

No tests are performed on functionality that shouldn't be touched by the upgrade - that would be in the scope of the regression tests which we have in mind, though they will also rely on a ganache fork from mainnet.


contracts/ColonyNetworkENS.sol
```diff
@@ -88,6 +88,30 @@ contract ColonyNetworkENS is ColonyNetworkStorage {
     emit ColonyLabelRegistered(msg.sender, subnode);
   }
 
+  function updateColonyOrbitDB(string memory orbitdb)
```
elenadimitrova: Do we care about protecting against an empty `orbitdb` string?

test/contracts-network/colony-network.js
```diff
@@ -486,5 +486,35 @@ contract("Colony Network", accounts => {
...
+
+    it("should allow a user to update their orbitDBAddress", async () => {
+      const hash = namehash.hash("test.user.joincolony.eth");
+      await colonyNetwork.registerUserLabel("test", orbitDBAddress, { from: accounts[1] });
+      await colonyNetwork.updateUserOrbitDB("anotherstring", { from: accounts[1] });
+      const retrievedOrbitDB = await colonyNetwork.getProfileDBAddress(hash);
+      expect(retrievedOrbitDB).to.equal("anotherstring");
+    });
+
+    it("should allow a user to set an orbitDBAddress if they've not got a label", async () => {
```
elenadimitrova: "should _not_ allow a ..."

contracts/Colony.sol
```diff
@@ -277,6 +281,15 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
   }
 
+  function finishUpgrade2To3() public always {
+    ColonyAuthority(address(authority)).setRoleCapability(
+      uint8(ColonyDataTypes.ColonyRole.Root),
+      address(this),
+      bytes4(keccak256("updateColonyOrbitDB(string)")),
+      true
+    );
+  }
```
elenadimitrova: For atomicity I feel this function should be internal and called from within "upgrade"

contracts/Colony.sol
```diff
@@ -277,6 +281,15 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+
```
elenadimitrova: Again for the `upgrade` function we should probably limit that to sequential version upgrades only.

test-upgrade/2to3/2to3.js
```diff
@@ -0,0 +1,93 @@
...
+    metaColony = await IMetaColony.at(metaColonyAddress);
+    await metaColony.addNetworkColonyVersion(3, newColonyVersionResolverAddress, { from: COLONYNETWORK_OWNER });
+
+    // These tests run against the developers colony. We assume we have forked main chain at 8046998
+    colony = await IColony.at(TESTCOLONY_ADDRESS);
+  });
+
+  describe("check behaviour before colonyNetwork upgrade", function() {
+    it("should not be able to update user orbit db", async function() {
+      await colonyNetwork.registerUserLabel("username", "before update");
```
elenadimitrova: For clarity, maybe bring this line out in the overall `before`? It is relied on in the [second test further down](https://github.com/JoinColony/colonyNetwork/pull/667/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.sol#diff-6cb8c134a81b9080e08341493e6a02c3R61) and it feels cleaner to have it at the top.

contracts/Colony.sol
area: I I don't think it can be internal and called from within `upgrade` because it doesn't exist on v2 Colony.sol, which we'll still be `delegatecall`'d in to even after calling `.setResolver`. We can however keep it external, perhaps even make it colony only, and do an external call after setting the new resolver.

contracts/ColonyNetworkENS.sol
area: I don't think so, though I don't feel strongly about it. An empty string vs a string that isn't an `orbitdb` address seems like a negligible difference for the dApp to deal with.

contracts/Colony.sol
elenadimitrova: Called conditionally when `currentColonyVersion = 2`?

PR #678, opened 2019-07-16
Add Arbitration role setup in colony upgrade

elenadimitrova:
Adds two missing steps in upgrading a colony from v2 to v3 for adding the newly introduced `Arbitration` role permissions to colonies that upgrade.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -285,8 +285,24 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   }
 
   function finishUpgrade2To3() public always {
-    ColonyAuthority(address(authority)).setRoleCapability(
-      uint8(ColonyDataTypes.ColonyRole.Root),
+    ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
+
+    colonyAuthority.setRoleCapability(
+     uint8(ColonyDataTypes.ColonyRole.ArchitectureSubdomain),
+      address(this),
+      bytes4(keccak256("setArbitrationRole(uint256,uint256,address,uint256,bool)")),
+      true
+    );
+
+    colonyAuthority.setRoleCapability(
+     uint8(ColonyDataTypes.ColonyRole.Root),
+      address(this),
+      bytes4(keccak256("setArbitrationRole(uint256,uint256,address,uint256,bool)")),
+      true
+    );
+
+    colonyAuthority.setRoleCapability(
+     uint8(ColonyDataTypes.ColonyRole.Root),
       address(this),
       bytes4(keccak256("updateColonyOrbitDB(string)")),
       true
--- a/contracts/ColonyAuthority.sol
+++ b/contracts/ColonyAuthority.sol
@@ -48,14 +48,12 @@ contract ColonyAuthority is CommonAuthority {
 
     // Add permissions for the Architecture role
     addRoleCapability(ARCHITECTURE_ROLE, "addDomain(uint256,uint256,uint256)");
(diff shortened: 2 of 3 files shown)


contracts/Colony.sol
```diff
@@ -285,8 +285,24 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+
+    colonyAuthority.setRoleCapability(
+     uint8(ColonyDataTypes.ColonyRole.Root),
+      address(this),
+      bytes4(keccak256("setArbitrationRole(uint256,uint256,address,uint256,bool)")),
+      true
+    );
+
+    colonyAuthority.setRoleCapability(
+     uint8(ColonyDataTypes.ColonyRole.Root),
```
kronosapiens: Interesting -- is the idea here that every version of colony contains it's own custom upgrade function? Seems reasonable if a colony must be upgraded in sequence (kind of like a migration) -- what sorts of things will end up here? Looks like just authority changes?

I'm  assuming that the next release will delete this function and introduce a `finishUpgrade3to4`? Also, why isn't this function called in the `finishUpgrade` function? Are colonies expected to call this manually?

contracts/ColonyAuthority.sol
```diff
@@ -80,6 +78,8 @@ contract ColonyAuthority is CommonAuthority {
 
     // Added in colony v3
     addRoleCapability(ROOT_ROLE, "updateColonyOrbitDB(string)");
+    addRoleCapability(ROOT_ROLE, "setArbitrationRole(uint256,uint256,address,uint256,bool)");
+    addRoleCapability(ARCHITECTURE_SUBDOMAIN_ROLE, "setArbitrationRole(uint256,uint256,address,uint256,bool)");
```
kronosapiens: I don't see the benefit of grouping this functionality in chronological order -- it seems better for readability if these stay with the original grouping.

test-upgrade/2to3/2to3.js
```diff
@@ -81,6 +84,13 @@ contract("Colony contract upgrade", () => {
       await colony.finishUpgrade2To3();
     });
 
+    it("should be able to assign arbitration role", async function() {
+      let hasRole = await colony.hasUserRole(accounts[3], 2, 2);
+      await colony.setArbitrationRole(1, 0, accounts[3], 2, true, { from: TESTCOLONY_OWNER });
+      hasRole = await colony.hasUserRole(accounts[3], 2, 2);
```
kronosapiens: Nitpick but it would slightly more readable if there was a `const ARBITRATION_ROLE = 2` somewhere -- `colony.hasUserRole(accounts[3], 2, 2)` is hard to parse :(

contracts/Colony.sol
elenadimitrova: This upgrade function was not introduced here. Please read PR https://github.com/JoinColony/colonyNetwork/pull/667 and discussions below it to understand the upgrade process we went for there.

contracts/ColonyAuthority.sol
elenadimitrova: The benefit is understanding what domain role capabilities are there for each colony version, which arguably is rather important when for example troubleshooting a specific colony version issues related to permissions.

test-upgrade/2to3/2to3.js
elenadimitrova: Sure, we have a constant for that in the helper.

contracts/Colony.sol
kronosapiens: Ah I see -- the currently-deployed version do not have the `finishUpgrade` function, so we can't leverage it in this version. Ok makes sense.

contracts/ColonyAuthority.sol
kronosapiens: Mm ok, I can buy that.

PR #686, opened 2019-07-24
Add ability to get all domain roles of a user

### LINKED ISSUE #677: Expose getUserRoles on IColony interface
As well as allowing `hasUserRole` to be called on IColony interface, it'd be useful to let people call `getUserRole`s as well.

elenadimitrova:
Closes #677

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -93,6 +93,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return ColonyAuthority(address(authority)).hasUserRole(_user, _domainId, uint8(_role));
   }
 
+  function getUserRoles(address who, uint256 where) public view returns (bytes32) {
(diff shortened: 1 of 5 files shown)


contracts/Colony.sol
```diff
@@ -93,6 +93,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return ColonyAuthority(address(authority)).hasUserRole(_user, _domainId, uint8(_role));
   }
 
+  function getUserRoles(address who, uint256 where) public view returns (bytes32) {
+    return DomainRoles(address(authority)).getUserRoles(who, where);
```
kronosapiens: Why not `ColonyAuthority`? `DomainRoles` is more specific but it seems more consistent to refer to `ColonyAuthority` everywhere for interacting with the permissions.

Issue #687, opened 2019-07-25
Update Greenkeeper config to update `yarn.lock` correctly


kronosapiens:
Looking at #682 , we see how Greenkeeper updated `yarn.lock` in an incorrect way. Rather than have to overwrite the `yarn.lock` commits for every GK update, let's see if we can update the Greenkeeper bot to update the lockfile correctly.

kronosapiens:
This issue may have resolved itself... looking at a recent GK commit, `yarn.lock` seems to have been updated fine: https://github.com/JoinColony/colonyNetwork/pull/738

kronosapiens:
Closing this because the issue seems to have been resolved.

Issue #569, opened 2019-03-15
Make use of external functions


elenadimitrova:
External functions can be used to [save gas when large arrays of data are being received](https://solidity.readthedocs.io/en/v0.5.6/contracts.html?#visibility-and-getters). This is especially true for the reputation mining process where such large input arrays are accepted. An [experimental branch](https://github.com/JoinColony/colonyNetwork/compare/maintenance/use-external-functions) was pushed that started testing this which shows 1,000 gas savings in `confirmJustificationRootHash` (down from `139140` to `138377`).

Note that to enable multi-dimensional array to be accepted in an external function, the ABIEncoderV2 has to be enabled. See [documentation note here](https://solidity.readthedocs.io/en/v0.5.6/contracts.html?#function-parameters)

Currently we use dynamic size input array of `bytes` in some external candidate reputation functions which blocks their switch to `external`, however this will potentially be unblocked when #393 is done.

area:
While doing #393 with an eye on unblocking this issue, I refactored to a point where in principle, we should have been able switch to `external` on `respondToChallenge`. However, doing so gave us the dreaded 'stack too deep' error. Elena found [this](https://github.com/ethereum/solidity/issues/5751#issuecomment-451972436), and it seems logical to assume the same issue applies to `bytes32[]` which... we have an awful lot of in that function call. The only solution at this juncture would seem to be compressing all of our `bytes32[]`s in to one giant `bytes32[]`, but we've opted to not do that for intelligibility purposes right now. We're going to be the only entity reputation mining at the start, so shouldn't need to use this function so we can hold off such drastic refactoring action for now.

(Note that we also need to remove the lines

```
u[U_DECAY_TRANSITION] = 0;
    u[U_GLOBAL_CHILD_UPDATE] = 0;
    u[U_NEW_REPUTATION] = 0;
```

as `calldata` is read only. Two of these are fairly clean to the remove, but the third is less so and should warrant more consideration when this is being done).

dev1644:
Does this issue require some work or is it fixed?

elenadimitrova:
As per @area 's comment above it's iceboxed for now.

dev1644:
Are you guys waiting for the `ABIEncoderv2` to be released for this issue?

elenadimitrova:
No, we have a "stack too deep" issue, like I said, read the last comment for details.

PR #690, opened 2019-08-05
Fix nightly security build

kronosapiens:
Fix for failing nightly security build


area:
The ESLint upgrade was the issue here. Has already been reported [https://github.com/ConsenSys/truffle-security/issues/228].

If we downgrade ESLint, we could drop the `--ignore-engines` again, but there's still the chance that's what fixed #687...

area:
Okay, because this hasn't been run for a while, Slither have added some things they look for that we fall foul to:

* [Public functions should be declared as external](https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-as-external) 1646 new detections. We were aware of this (see the discussion in #569). This is legitimate, and has a tangible benefit for users, so it's not clear we should just ignore here.
* [State variables that could be constant](https://github.com/crytic/slither/wiki/Detector-Documentation#state-variables-that-could-be-declared-constant). 163 new detections. I think these are mostly incorrect - while we don't have any code that changes the value of variables declared in `*Storage.sol` contracts, we definitely do change them elsewhere.
* [Dangerous strict equalities](https://github.com/crytic/slither/wiki/Detector-Documentation#dangerous-strict-equalities). 9 detections. Essentially, using `==` in a `require` with a `(u)int`. This is _very_ opinionated, and while I can see the logic, we are mostly using that pattern to check state (e.g. auction hasn't started, by checking `startTime==0`). In theory, we could refactor in a couple of places to pointedly work around this (`startTime < 1`), or introduce `bool`s for these states.
* [Constant functions changing the state](https://github.com/crytic/slither/wiki/Detector-Documentation#constant-functions-changing-the-state) 5 detections. In our case, the functions in question are declared `view` but contain assembly, which Slither flags because (I think) it can't tell if the code is changing state. Probably able to be ignored - we should be thinking very carefully about assembly, and any 'normal' code that changes state in a `view` function is caught by the compiler.
* [Functions allow anyone to self-destruct contracts](https://github.com/crytic/slither/wiki/Detector-Documentation#suicidal) Two detections, in `ReputationMiningCycle.sol` and `DutchAuction.sol`. Slither is correct, but this is by design in our case.
* Various issues in library contracts. We can exclude these now with `--filter-paths lib` which seems like a no brainer (as these aren't under our control)

area:
I've added `--ignore-optimization` to the slither configuration. We should remove it down the line, but for now, we just want the nightly build back and useful. At some point when we're removing technical debt, we can circle back and see what the optimisation results are telling us to do. For now, this covers the public/external function issue.

I've added `--filter-paths lib` to remove issues in contracts that aren't under our control.

The rest we were actually already intending to ignore via `--triage` mode and `slither.db.json`, but I'm guessing we did something to mess up the line / character locations at some point?

area:
Now MythX have launched, in theory, we can remove these blacklisted SWCs later this week when our pro trial runs out (see [the plan details page](https://mythx.io/swc-coverage)) because the free plan doesn't look for them. I either don't properly understand SWC-123 and/or we can't work around it - it flags [this line](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/EtherRouter.sol#L47) in EtherRouter which shoudn't ever throw as far as I can tell (but might get `0x00` back from Resolver).

area:
[Green nightly build](https://circleci.com/gh/JoinColony/colonyNetwork/10015?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) from before dropping the commit that runs it...

area:
[Green nightly build](https://circleci.com/gh/JoinColony/colonyDapp/21801)

PR #712, opened 2019-09-16
Colony setup bundle

### LINKED ISSUE #607: Colony creation bundle
In order to minimise the number of transactions which are part of the "Colony creation" flow in the dApp, associated with setting up a new Token, Colony, extension contracts and associated wiring (see list below) we should investigate how much of these contract calls we can bundle in a single call as part of an external contract. It is likely that the entire set of transaction will exceed the block gas limit, so find a way around that too. 

- Create token
- Create colony
- Register colony ENS name
- Create `TokenAuthority`
- Set token authority on the token
- Deploy `OneTxPayment` for colony
- Give that permissions
- Deploy `OldRoles` for colony
- Give that permissions

kronosapiens:
Closes #607 

NOTE: this PR complements #714 and should be coordinated with that one, although it is not strictly a dependency.

This PR introduces the following overload function, allowing users to choose their version, set an ENS name, or configure the Extension Manager when creating their colonies:

```
  function createColony(
    address _tokenAddress,
    uint256 _version,
    string memory _colonyName,
    string memory _orbitdb,
    bool _useExtensionManager
  ) public stoppable returns (address)
```

When combined with the Extension Manager, the colony creation flow can be reduced to between 3-5 tx, [depending on when](https://joincolony.slack.com/archives/C0YNG3M8B/p1569268565092900) the Token Authority is deployed:

1. deploy Token
2. create Colony with ENS name and Extension Manager
3. install OneTxPayment
4. deploy TokenAuthority (or elsewhere)
5. configure TokenAuthority (or elsewhere)

The corresponding updates in ColonyJS can be found [here](https://github.com/JoinColony/colonyJS/pull/458).

### DIFF
--- a/contracts/ColonyNetwork.sol
+++ b/contracts/ColonyNetwork.sol
@@ -142,15 +142,16 @@ contract ColonyNetwork is ColonyNetworkStorage {
   ) public stoppable returns (address)
   {
     uint256 version = (_version == 0) ? currentColonyVersion : _version;
-    address colonyAddress = deployColony(_tokenAddress version);
+    address colonyAddress = deployColony(_tokenAddress, version);
 
     if (bytes(_colonyName).length > 0) {
       IColony(colonyAddress).registerColonyLabel(_colonyName, _orbitdb);
     }
 
-    if (_useExtensionManager) {
-      IColony(colonyAddress).setRootRole(extensionManagerAddress, true);
-    }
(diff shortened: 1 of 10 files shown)


kronosapiens:
@area very reasonable. Also I had thought I'd unskipped that test but I suppose not...

I'll update the comments and send this back around for you after AIRCS.

PR #713, opened 2019-09-17
Immutable domains

### LINKED ISSUE #707: Make domain on a task/payment immutable
After consulting with the product team, they are happy for us to remove the ability for users to change the domain for a task / payment (as this capability can be used to break assumptions made elsewhere about funding).

kronosapiens:
Closes #707 

Not much to say, this PR removes both `setTaskDomain` and `setPaymentDomain` functionality. As a best practice we also remove the `setPaymentDomain` function from our `authority`, freeing up some storage (and getting a sweet gas refund).

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -312,8 +312,13 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     );
   }
 
+  // v3 to v4
   function finishUpgrade() public always {
-    // Nothing here for v2 to v3, but it needs to be defined.
+    ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
+
+    // Remove function
+    bytes4 sig1 = bytes4(keccak256("setPaymentDomain(uint256,uint256,uint256,uint256)"));
+    colonyAuthority.setRoleCapability(uint8(ColonyRole.Administration), address(this), sig1, false);
   }
 
   function checkNotAdditionalProtectedVariable(uint256 _slot) public view recovery {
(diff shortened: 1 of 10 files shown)


contracts/Colony.sol
```diff
@@ -115,7 +115,6 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     setFunctionReviewers(bytes4(keccak256("setTaskBrief(uint256,bytes32)")), TaskRole.Manager, TaskRole.Worker);
     setFunctionReviewers(bytes4(keccak256("setTaskDueDate(uint256,uint256)")), TaskRole.Manager, TaskRole.Worker);
     setFunctionReviewers(bytes4(keccak256("setTaskSkill(uint256,uint256)")), TaskRole.Manager, TaskRole.Worker);
-    setFunctionReviewers(bytes4(keccak256("setTaskDomain(uint256,uint256)")), TaskRole.Manager, TaskRole.Worker);
```
area: The upgrade function should also remove this from the `reviewers` mapping

contracts/Colony.sol
```diff
@@ -313,8 +312,13 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     );
   }
 
+  // v3 to v4
   function finishUpgrade() public always {
-    // Nothing here for v2 to v3, but it needs to be defined.
+    ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
+
+    // Remove function
+    bytes4 sig1 = bytes4(keccak256("setPaymentDomain(uint256,uint256,uint256,uint256)"));
```
area: This could be a constant rather than being calculated on-chain during the upgrade.

PR #714, opened 2019-09-19
Extension Manager

### LINKED ISSUE #699: Improve framework for Colony extensions
While working on the "task as extension" part of #693, it became clear that extension development could be simplified, i.e. by extracting what will likely become common patterns. This includes:

- [x] Creating a public `validateDomainInheritence` function to allow extensions to easily validate domain proofs. Extensions should not have to implement [this type of logic](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/extensions/OneTxPayment.sol#L154).

- [x] Potentially a `ColonyExtension` base contract which provides a standard interface, including a `deprecate` and `uninstall` function.

- [x] Relatedly, think about uninstall logic which tracks the number of "active" objects in the extension (i.e. the number of active tasks) -- an extension can be uninstalled only when there are no active objects. "Deprecating" an extension would amount to preventing the creation of new objects, allowing the mount of active objects to be slowly drawn down.

- [x] A convenience function on the Colony allowing for multiple roles to be assigned in one transaction, ideally bundled into the `*Factory` `install` function. I.e. a "one-click" install which deploys an extension and assigns it the correct roles.

- [x] Using a library or etherrouter pattern to reduce the gas costs of individual extension deployments.

- Potentially more, I will add more notes as I continue.

kronosapiens:
Closes #699 
Implements part of #607 

Introduce `ColonyNetworkExtensions` which functions as a type of "app store" for Colony extensions:

```
// MetaColony only
addExtensionToNetwork(
    bytes32 _extensionId, // Extension identifier, i.e. soliditySha3("ExtensionName")
    address _resolver, // Resolver address for extensionId:version
)

installExtension(
    bytes32 _extensionId,
    uint256 _version,
)

upgradeExtension(
    bytes32 _extensionId,
    uint256 _newVersion
)

uninstallExtension(
    bytes32 _extensionId,
)
```

### Design decisions

- All functions are root-permissioned

- Extension resolvers, like Colony resolvers, can be set at any time by the Metacolony

- Versions must start at 1 and be introduced in sequence.


contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,167 @@
...
+    uint256 _version,
+    bool _setTo
+  )
+    internal
+  {
+    IColony colony = IColony(_colony);
+    address extension = installations[_extensionId][_version][_colony][_domainId];
+    uint8[] storage extensionRoles = roles[_extensionId][_version];
+
+    for (uint256 i; i < extensionRoles.length; i++) {
```
area: I do not like any of this loop one bit. I'm not sure how to make it better though.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,167 @@
...
+  )
+    public
+    isAuthorized(_colony, _permissionDomainId, _childSkillIndex, _domainId)
+  {
+    require(
+      installations[_extensionId][_version][_colony][_domainId] == address(0x0),
+      "extension-manager-already-installed"
+    );
+
+    ColonyExtension extension = new ColonyExtension();
```
area: An extension deployed like this will have the default `install`, `uninstall`, `uninstallable` functions, regardless of what's defined in the resolver for that extension. Incoming transactions with those signatures will never reach the fallback function. This is exactly what I was talking about last Thursday. This has to be an EtherRouter.

contracts/Colony.sol
```diff
@@ -93,6 +94,33 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+  function userCanSetRoles(
+    address _user,
+    uint256 _domainId,
+    uint256 _childSkillIndex,
+    uint256 _childDomainId
+  ) public view returns (bool)
+  {
+    return (
+      hasUserRole(_user, 1, ColonyRole.Root) ||
+      (_domainId != _childDomainId && hasInheritedUserRole(_user, _domainId, ColonyRole.Architecture, _childSkillIndex, _childDomainId))
```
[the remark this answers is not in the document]
kronosapiens: Mm I'm partial to treating ArchitectureSubdomain as a very low-level implementation detail and really never referring to it separately from Architecture -- it can't be assigned separately and so I don't really like the idea of referring to both Architecture and Subdomain in external calls.

contracts/Colony.sol
```diff
@@ -323,6 +358,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     // Remove from authority
     ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Administration), address(this), SIG2, false);
+    colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), SIG3, true);
```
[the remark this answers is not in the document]
kronosapiens: Fair point.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,167 @@
...
+  modifier isAuthorized(address _colony, uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _domainId) {
+    require(IColony(_colony).userCanSetRoles(msg.sender, _permissionDomainId, _childSkillIndex, _domainId), "extension-manager-unauthorized");
+    _;
+  }
+
+  function addExtension(bytes32 _extensionId, uint256 _version, address _resolver, uint8[] memory _roles)
+    public
+    onlyMetaColony
+  {
+    require(_version == 0 || resolvers[_extensionId][_version - 1] != address(0x0), "extension-manager-bad-version");
```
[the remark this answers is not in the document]
kronosapiens: Urgh ok that's fair.

contracts/extensions/ColonyExtension.sol
```diff
@@ -18,10 +18,28 @@
...
+import "../EtherRouter.sol";
+import "../IColony.sol";
 
-interface ExtensionFactory {
-  function deployExtension(address _colony) external;
-  function removeExtension(address _colony) external;
-  event ExtensionDeployed(string _name, address _colony, address _extension);
-  event ExtensionRemoved(string _name, address _colony);
+
+contract ColonyExtension is EtherRouter {
```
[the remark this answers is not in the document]
kronosapiens: Mm that doesn't bother me so much -- you can still send any other ERC20 to an extension with no way to recover it, so does it really matter if Ether behaves the same? I like being able to call `setResolver` directly without doing the cast and it makes the storage layout more straightforward.

contracts/extensions/OldRoles.sol
```diff
@@ -18,20 +18,14 @@
 pragma solidity 0.5.8;
 pragma experimental ABIEncoderV2;
 
-import "./../ColonyDataTypes.sol";
```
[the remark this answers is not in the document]
kronosapiens: I suppose we could remove it entirely? But maybe in a separate PR?

contracts/extensions/OneTxPayment.sol
```diff
@@ -18,24 +18,16 @@
...
 
-contract OneTxPayment {
+contract OneTxPayment is ColonyExtension {
   bytes4 constant ADD_PAYMENT_SIG = bytes4(keccak256("addPayment(uint256,uint256,address,address,uint256,uint256,uint256)"));
   bytes4 constant MOVE_FUNDS_SIG = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,address)"));
 
-  IColony colony;
-  IColonyNetwork colonyNetwork;
-
-  constructor(address _colony) public {
```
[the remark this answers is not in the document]
kronosapiens: Mm that's fair, I hadn't realized the overhead was so expensive.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,167 @@
...
+      installations[_extensionId][_version][_colony][_domainId] == address(0x0),
+      "extension-manager-already-installed"
+    );
+
+    ColonyExtension extension = new ColonyExtension();
+    installations[_extensionId][_version][_colony][_domainId] = address(extension);
+
+    assignRoles(_colony, _rootChildSkillIndex, _domainId, _extensionId, _version, true);
+    address resolver = resolvers[_extensionId][_version];
+    extension.install(_colony, resolver);
```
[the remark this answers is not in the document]
kronosapiens: Hmm, that's a reasonable point. Alternatively, the extension could include functions for setting these additional values and restrict them to Root/Architecture users (i.e. `setToken()`), which works out to the same thing but keeps the common interface smaller.

contracts/extensions/ExtensionManager.sol
kronosapiens: My reasoning was that since the only one who can input the array is the metaColony, there's very little risk.

contracts/extensions/ExtensionManager.sol
kronosapiens: That is an excellent point.

kronosapiens:
@area from your earlier comment I'm taking that to mean you'd like to add an `upgradeExtension` function which performs one-version-at-a-time resolver updates. Is that right?

contracts/extensions/OneTxPayment.sol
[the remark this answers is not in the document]
kronosapiens: Mm so overriding `install` doesn't work since that is called by the ExtensionManager, which does not know anything about the extension apart from the resolver address. But I've done it in a different (and mildly hacky, but I think acceptable way).

contracts/Colony.sol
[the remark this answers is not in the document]
area: Why do we even have it then? Why did we not just have `_domainId != _childDomainId` in functions that are currently limited to `ArchitectureSubdomain`?

area:
Yeah, that's certainly a good place to start. We probably also want an `upgrade()` function that is called, like we do for colony updates.

contracts/Colony.sol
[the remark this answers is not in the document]
kronosapiens: Mm that's a fair question. At the time it seemed like the only way to implement the subdomain-only behavior but perhaps that was an oversight. Let me think about it some more.

contracts/extensions/OneTxPayment.sol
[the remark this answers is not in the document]
kronosapiens: Problem solved by setting the resolver _before_ calling `install`, which allows for the resolver's version to be called (with any custom behavior).

contracts/extensions/OneTxPayment.sol
```diff
@@ -18,24 +18,40 @@
...
+  /// @notice Configures the extension
+  /// @param _colony The colony in which the extension holds permissions
+  function install(address _colony) public auth {
+    require(address(colony) == address(0x0), "extension-already-installed");
+
     colony = IColony(_colony);
     colonyNetwork = IColonyNetwork(colony.getColonyNetwork());
   }
 
+  /// @notice Called when upgrading the extension (currently a no-op since this OneTxPayment does not support upgrading)
```
area: I mean, this not being a noop doesn't mean upgrades aren't possible. I wouldn't expect any first version of an extension to have anything here, because for a version 1 there's no way this can be called.

test/extensions/extension-manager.js
```diff
@@ -0,0 +1,326 @@
...
+  });
+
+  beforeEach(async () => {
+    ({ colony } = await setupRandomColony(colonyNetwork));
+    await colony.setRootRole(extensionManager.address, true);
+    await colony.setArchitectureRole(1, 0, ARCHITECT, 1, true);
+  });
+
+  describe("without the extension manager", () => {
+    it("allows a colony to install an extension manually (and without a resolver)", async () => {
```
area: I guess we can't stop this...

contracts/extensions/ColonyExtension.sol
```diff
@@ -18,10 +18,15 @@
...
+  IColony colony;
+
+  function version() public pure returns (uint256);
+  function install(address _colony) public;
+  function upgrade() public;
```
area: Can we make this `finishUpgrade` for consistency with Colony upgrades?

test/extensions/extension-manager.js
kronosapiens: Nor should we, I like that users can install the same extension without needing the extension manager.

test/extensions/extension-manager.js
```diff
@@ -0,0 +1,326 @@
...
+
+      // Check that fakeRole is not part of the role bit array
+      const extensionAddress = await extensionManager.getExtension(extensionId, 1, colony.address, 1);
+      const userRoles = await colony.getUserRoles(extensionAddress, 1);
+      expect(new BN(parseInt(userRoles, 16)).and(new BN(1).shln(fakeRole))).to.be.zero;
+    });
+  });
+
+  describe("upgrading extensions", () => {
+    it("allows authorized users to upgrade an extension", async () => {
```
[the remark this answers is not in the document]
kronosapiens: I sort of tried cover all these tests in the `install` block (since the same logic does authorization for installation, upgrading, and uninstalling), but I can explicitly replicate the tests for upgrading and uninstalls as well.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,215 @@
...
+    address payable extension = installations[_extensionId][_version][_colony][_domainId];
+    require(extension != address(0x0), "extension-manager-not-installed");
+
+    uint256 newVersion = _version + 1;
+    require(resolvers[_extensionId][newVersion] != address(0x0), "extension-manager-bad-version");
+
+    installations[_extensionId][_version][_colony][_domainId] = address(0x0);
+    installations[_extensionId][newVersion][_colony][_domainId] = extension;
+
+    assignRoles(_colony, _rootChildSkillIndex, _domainId, _extensionId, _version, false);
```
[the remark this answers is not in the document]
kronosapiens: Thinking more, I'd like to disallow updating roles for future extension versions -- required roles can only be set when version 1 of the extension is added. Doing set operations on two arrays in solidity does not seem like it's in the spirit of the language.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,215 @@
...
+
+    extension.setResolver(resolvers[_extensionId][_version]);
+    ColonyExtension(address(extension)).install(_colony);
+
+    emit ExtensionInstalled(_extensionId, _version, _colony, _domainId);
+  }
+
+  function upgradeExtension(
+    bytes32 _extensionId,
+    uint256 _version,
```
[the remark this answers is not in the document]
kronosapiens: Yup, that can be removed once we start reading `version` from the resolver.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,215 @@
...
+
+    require(
+      resolvers[_extensionId][_version] != address(0x0),
+      "extension-manager-bad-version"
+    );
+    require(
+      installations[_extensionId][_version][_colony][_domainId] == address(0x0),
+      "extension-manager-already-installed"
+    );
+
```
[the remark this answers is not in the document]
kronosapiens: I think they should be able to install any version.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,215 @@
...
+contract ExtensionManager {
+  IMetaColony metaColony;
+
+  // [_extensionId][version] => resolver
+  mapping(bytes32 => mapping(uint256 => address)) resolvers;
+  // [_extensionId][version] => roles
+  mapping(bytes32 => mapping(uint256 => uint8[])) roles;
+  // [_extensionId][version][colony][domainId] => address
+  mapping(bytes32 => mapping(uint256 => mapping(address => mapping(uint256 => address payable)))) installations;
```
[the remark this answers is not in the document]
kronosapiens: Nope, you're right, we can omit it.

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,215 @@
...
+  function authorized(address _colony, uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _domainId)
+    internal
+    view
+    returns (bool)
+  {
+    return IColony(_colony).userCanSetRoles(msg.sender, _permissionDomainId, _childSkillIndex, _domainId);
+  }
+
+  function getResolverVersion(address _resolver) internal returns (uint256) {
+    EtherRouter extension = new EtherRouter();
```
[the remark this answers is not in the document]
kronosapiens: Makes sense, I didn't know that was possible.

test/extensions/extension-manager.js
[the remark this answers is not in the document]
area: You're right, but you still have to test that the auth code is correctly wired up to the function in question, which is essentially what was missing. Otherwise you can remove the `require` that does the auth at the start of upgrade the code base still passes all tests, but is clearly broken.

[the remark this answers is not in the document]
kronosapiens: Yup, I can appreciate that distinction.

area:
A thought looking at this PR from a broader perspective:

Installing the same extension over and over again in a colony in different domains seems inelegant. Intuitively, if I have installed an extension in a root domain, I probably expect it to be installed/available in all child domains (indeed, this is how the tasks-as-extension contract has been implemented in #693). Moreover, you could effectively end up installing an extension in the same domain multiple times if you installed it in the domain, the parent domain, the grandparent domain... ("It's a task in domain 3" "I'm looking at tasks in domain 3, but I'm not seeing it" "Oh, it's a task in domain three in the task extension installed in domain 1, not the task extension installed in domain 3")

Would a better approach be to just add permissions to the existing contract for all domains installations after the first? If two sibling domains installed an extension, and then their parent did, you could remove the explicit permissions in the child domains and everything would continue to function.

Downsides: It might make it harder to track where exactly it's been installed, but equally it will be 'more accurate' on the dapp side (if someone removes the permissions an extension requires... is it installed or not? Surely a lot closer to 'not'). It also means that all domains would be forced to use the same version of the extension, and might be forced in to doing an upgrade when they didn't want to?

kronosapiens:
I spent some time thinking about this, and while I think this makes sense in principle I don't think these requirements are enforceable on-chain.

Consider the case where we have an extension installed in two sibling domains, and we then install the same extension in a shared parent -- how can we verify at install time the two inheritance relationships necessary? I'm assuming that the user will pass in domain inheritance information at install time, but unlike the single-ancestry domains proofs we use for function calls (which have a fixed number of inputs and can be verified in constant time), the situation here can contain an unbounded number of inputs (since there can be many possible affected children), each with a potentially different ancestries (i.e. the parent might be the first parent of one domain, the third parent of another, the fourth parent of a third, etc) -- and further we have no way of knowing if every child domain is being included (I could easily omit information about the twentieth affected child domain, which would then retain its own installation).

Ultimately, while I agree that this type of consistency is desirable, I am skeptical it can be enforced at the contract level. Much more likely is that these requirements should be supported at the ColonyJS (or similar) level, in which it is possible to maintain a full colony domain hierarchy in memory and identify inconsistencies and encourage users to take appropriate action.

area:
> If two sibling domains installed an extension, and then their parent did, you could remove the explicit permissions in the child domains and everything would continue to function.

This sentence of mine was misleading - I wasn't talking about all these permission changes happening at install time (which I agree can't happen), but only saying that when the extension was added to a parent domain, down the line, the children's permissions could in principle be removed and everything would still continue to function (modulo users changing the permission proofs they supplied).

I could even believe it was desirable to leave the permissions in place, in some cases, in case the parent domain removed the permissions but a child domain wanted to keep using the extension. Indeed, perhaps that's another argument for the sort of change I'm proposing - if a child domain uses an extension, there's no way currently for a parent domain to remove it from itself but still let the child domain keep using it.

kronosapiens:
Mm I think I understand what you're driving at -- orienting installation around permissions rather than EtherRouter deployments (and the Extension Manager could use something like a semaphore to keep track of "installed domains" for purposes of uninstallation).

My main critique then would be that is seems to complicate the question of lifecycle management (i.e. upgrading/uninstalling, which you mention) -- who decides about upgrades/deprecations? The domain closest to the root? What happens in the case of two equal siblings?

kronosapiens:
I see where you're coming from, but returning to the theme of "consistent semantics", I think that a situation where someone in a sibling domain can make changes in my domain represents a violation of the roles semantics and is basically a non-starter.

The data loss may be an inconvenience, but I would argue that this possibility further incentivizes treating extensions as ephemeral -- this may boil down to good or bad planning on the parts of individual colonies. I think our goal should be to make well-planned colonies a pleasure to use and poorly-planned colonies able to recover from their mistakes, not to prevent poorly-planned colonies from inconveniencing themselves.

I'm happy to take a step back and bring product folks in for a larger process if you think that will help, although I'll say again that I think the most important thing is to ensure a consistent semantics of domain-level permissions, even if it creates some inconvenient edge-cases.

kronosapiens:
Noting that after a call with product to discuss [this doc](https://docs.google.com/document/d/1hsWU11KXu076knW6htmdk-o8_8R3sdU0dBYvKL-imq0/edit), we will move forward with a permissions-based approach to installations.

kronosapiens:
@area as we circle back around to this, I think the design where the extension manager is a separate contract is perhaps over-engineered. Keeping the extension resolvers on the colony network seems totally appropriate and leads to a simpler feature.

area:
Hm, I'm not sure I agree. From the outset, we said that the core contracts shouldn't be aware of extensions at all, and this definitely seems like it'd be going in that direction.

kronosapiens:
@area sure, but fundamentally we are making the colony network aware of the extension manager, so that is already occurring. If that is the case, why not skip a level of indirection? The "awareness" is minimal, simply deploying contracts using resolvers. The specifics of the extensions are still obscured to the network.

area:
I think we arrived at this pattern after thinking about permissions as well - the manager needs root permissions in your colony, and giving a contract root permissions when that can be upgraded to have arbitrary functionality without your permission will rightly spook people. This way the extension manager is not upgradable, so assuming you're happy with the manager code you don't have to worry about it going forward.

kronosapiens:
Noting that the rebase is complete and this branch has now been updated.

contracts/extensions/ColonyExtension.sol
```diff
@@ -18,10 +18,16 @@
...
-  event ExtensionRemoved(string _name, address _colony);
+
+contract ColonyExtension is DSAuth {
+  address resolver; // Align storage with EtherRouter
+  IColony colony;
+
+  function version() public pure returns (uint256);
+  function install(address _colony) public;
+  function finishUpgrade() public;
+  function uninstall() public;
```
area, line 32: I think you noted it down, but for posterity we talked about having a `deprecate()` / `undeprecate` here.

contracts/extensions/FundingQueue.sol
```diff
@@ -327,39 +345,51 @@ contract FundingQueue is DSMath, PatriciaTreeProofs {
...
-      wmul(decayRates[lowerBin + 1], lowerPct)
+      wmul(getDecayRateFromBin(lowerBin), sub(WAD, lowerPct)),
+      wmul(getDecayRateFromBin(lowerBin + 1), lowerPct)
     );
   }
 
+  function getDecayRateFromBin(uint256 bin) internal pure returns (uint256) {
+    // Used for mapping backing percent to the appropriate decay rate (10 second intervals)
+    // Result of evaluating ((1 - backingPercent / 2) ** (1 / (7 * 24 * 60 * 6)))
+    //  at the following points: [0, .1, .2, .3, .4, .5, .6, .7, .8, .9, 1]
```
area, line 367: I think this one's my fault, but to truly make this a functional refactor only, we should check `bin<=10`, which should be assertable.

contracts/colonyNetwork/IColonyNetwork.sol
```diff
@@ -139,7 +139,18 @@ contract IColonyNetwork is ColonyNetworkDataTypes, IRecovery {
   /// @param _version The version of colony to deploy (pass 0 for the current version)
   /// @param _colonyName The label to register (if null, no label is registered)
   /// @param _orbitdb The path of the orbitDB database associated with the user profile
-  /// @param _useExtensionManager If true, give the ExtensionManager the root role in the colony
+  /// @return colonyAddress Address of the newly created colony
+  function createColony(address _tokenAddress, uint256 _version, string memory _colonyName, string memory _orbitdb)
```
area, line 143: If we're cleaning this function up, should we not get rid of `_orbitdb` as well?

contracts/extensions/ExtensionManager.sol
```diff
@@ -0,0 +1,209 @@
...
+    bool _setTo
+  )
+    internal
+  {
+    IColony colony = IColony(_colony);
+    address extension = installations[_extensionId][_colony];
+    uint8[] storage extensionRoles = roles[_extensionId];
+
+    for (uint256 i; i < extensionRoles.length; i++) {
+      if (extensionRoles[i] == uint8(ColonyDataTypes.ColonyRole.Root)) {
```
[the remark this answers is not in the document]
kronosapiens: This code is no longer there, we've moved this over to `ColonyRoles.sol` already.

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,135 @@
...
+    public
+    view
+    returns (address)
+  {
+    return installations[_extensionId][_colony];
+  }
+
+  // Internal functions
+
+  function root(address _colony) internal view returns (bool) {
```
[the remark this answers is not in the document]
kronosapiens: Sure! I'll also make it a modifer.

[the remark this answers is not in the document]
kronosapiens: Although... it would be more consistent with our conventions to add these functions to the IColony interface, which does the auth, and then have the colony make the call out to ColonyNetwork. What do you think of that?

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function upgradeExtension(bytes32 _extensionId, address _colony, uint256 _newVersion)
+    public
+    stoppable
+  {
+    require(root(_colony), "extension-manager-unauthorized");
+    require(installations[_extensionId][_colony] != address(0x0), "extension-manager-not-installed");
+
+    address payable extension = installations[_extensionId][_colony];
+    require(_newVersion == ColonyExtension(extension).version() + 1, "extension-manager-bad-increment");
```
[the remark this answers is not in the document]
kronosapiens: Why do we have `_newVersion` for colony upgrades? If you can only upgrade one at a time. Same reason, as a bit of a check to ensure people don't upgrade by accident.

contracts/colonyNetwork/IColonyNetwork.sol
kronosapiens: Is that not being used anymore?

contracts/extensions/ExtensionManager.sol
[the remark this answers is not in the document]
area: Oh, wow, I think this was a holdover from a review I started a while ago....

contracts/colonyNetwork/IColonyNetwork.sol
area: It is not. He's only been gone a handful of days, and you've already forgotten Chris' exasperation with Orbit and friends? :stuck_out_tongue:

kronosapiens: Oh yes, I suppressed that memory.

contracts/colonyNetwork/ColonyNetworkExtensions.sol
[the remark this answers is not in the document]
area: Do we do that for things in `IColony`? I know we do in `IMetaColony`, but that's special :smile:

contracts/colonyNetwork/ColonyNetworkExtensions.sol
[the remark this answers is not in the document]
area: I guess you've got me :sweat_smile:

contracts/colonyNetwork/ColonyNetworkExtensions.sol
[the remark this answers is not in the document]
kronosapiens: For example here:

```
  function registerColonyLabel(string memory colonyName, string memory orbitdb) public stoppable auth {
    IColonyNetwork(colonyNetworkAddress).registerColonyLabel(colonyName, orbitdb);
  }
```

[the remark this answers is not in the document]
kronosapiens: I say let's be consistent.

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,135 @@
...
+  function addExtension(bytes32 _extensionId, address _resolver, bytes32 _roles)
+    public
+    stoppable
+    calledByMetaColony
+  {
+    require(_resolver != address(0x0), "extension-manager-bad-resolver");
+
+    uint256 version = getResolverVersion(_resolver);
+    require(version == 1 || _roles == 0, "extension-manager-nonempty-roles");
+    require(version == 1 || resolvers[_extensionId][version - 1] != address(0x0), "extension-manager-bad-version");
```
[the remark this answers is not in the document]
kronosapiens: I would actually leave out that check, since that would preclude updating resolvers, which we allow for Colony versions:

```
  function addColonyVersion(uint _version, address _resolver) public
  always
  calledByMetaColony
  {
    require(currentColonyVersion > 0, "colony-network-not-intialised-cannot-add-colony-version");

    if (_version > currentColonyVersion) {
    }

    emit ColonyVersionAdded(_version, _resolver);
  }
```

Thinking more, perhaps even requiring resolvers to be added in sequence is unecessary.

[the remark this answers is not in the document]
kronosapiens: Further, storing the roles is also not really necessary since they're never used for anything consensus-relevant, it's really just documentation.

contracts/colonyNetwork/IColonyNetwork.sol
kronosapiens: So we still pass it into the ENS reservation, but I'm guessing we can just pass in the empty string?

contracts/colonyNetwork/ColonyNetworkExtensions.sol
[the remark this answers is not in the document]
area: > Thinking more, perhaps even requiring resolvers to be added in sequence is unecessary.

Surely it definitely is if we only allow upgrading one-at-a-time?

I think allowing overwriting of a version's resolving is asking for a whole load of trouble to be honest, and it probably shouldn't be allowed on the colony version side either.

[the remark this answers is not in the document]
kronosapiens: I would agree, but while the logic for colony resolvers is as permissive as it is, I don't think it makes sense to have the logic for extensions be more restrictive. My preference would be to keep them at the same level of security and update them both in parallel once we're ready to do so.

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,128 @@
...
+  // Public functions
+
+  function addExtension(bytes32 _extensionId, address _resolver)
+    public
+    stoppable
+    calledByMetaColony
+  {
+    uint256 version = getResolverVersion(_resolver);
+    require(version > 0, "colony-network-extension-bad-version");
+    resolvers[_extensionId][version] = _resolver;
```
area, line 38: This shouldn't be able to be changed, and should probably also check that the previous version is set.

contracts/extensions/ColonyExtension.sol
```diff
@@ -18,10 +18,24 @@
...
-  event ExtensionDeployed(string _name, address _colony, address _extension);
-  event ExtensionRemoved(string _name, address _colony);
+
+contract ColonyExtension is DSAuth {
+  address resolver; // Align storage with EtherRouter
+
+  IColony colony;
+  bool deprecated;
+
+  modifier undeprecated() {
```
area, line 31: Can we call this `notDeprecated`?

docs/_Docs_Extensions.md
```diff
@@ -6,34 +6,42 @@ order: 6
...
-
-// Instantiate the new extension
-const oneTxExtensionAddress = await oneTxExtensionFactory.deployedExtensions(colony.address)
-const oneTxExtension = await OneTxPayment.at(oneTxExtensionAddress)
+const extensionAddress = await colonyNetwork.getExtensionInstallation(oneTxExtensionId, colony.address);
+const oneTxExtension = await OneTxPayment.at(extensionAddress)
 ```
 
-Once the extension is setup, it will need `Admininstration` and `Funding` permissions in your colony to function. Permissions can be given is any domain you require the payment extension to work in, here is an example of permitting it those in the root with domainId 1.
+Once the extension is setup, it will need the `Funding` permission in your colony to function. Permissions can be given is any domain you require the extension to work in, here is an example of permitting it those in the root with domainId 1.
```
area, line 25: 'given in any domain'

docs/_Interface_IColony.md
```diff
@@ -1519,6 +1562,18 @@ Transfer some amount of obligated tokens. Can be called by the arbitration role.
 |_recipient|address|Recipient of the transferred tokens.
 
 
+### `uninstallExtension`
+
+Uninstall an extension from a colony. Secured function to authorised members.
```
area, line 1567: I'd add that this is not reversible. Even if you reinstall, it's a different instance of the extension. Perhaps also mention the intended flow for removing an extension (deprecating it, waiting for everything to complete, and then uninstalling it).

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,128 @@
...
+import "../colonyNetwork/IColonyNetwork.sol";
+import "../extensions/ColonyExtension.sol";
+import "./ColonyNetworkStorage.sol";
+
+
+contract ColonyNetworkExtensions is ColonyNetworkStorage {
+
+  // Public functions
+
+  function addExtension(bytes32 _extensionId, address _resolver)
```
area, line 31: Should there be a `removeExtension` function? I haven't thought through potential consequences...

contracts/colonyNetwork/ColonyNetworkExtensions.sol
kronosapiens: I can make this change, but I want to be clear that this actually does nothing from a security perspective, since the upstream Colony resolvers are not similarly locked, so this guarantees nothing. I would advocate for leaving this as-is and changing how we handle resolvers for both extensions and colony versions together. Otherwise, someone reading the code could get a false sense of security. Do you agree?

contracts/colonyNetwork/ColonyNetworkExtensions.sol
kronosapiens: I don't feel strongly, I'd say no need for now

contracts/colonyNetwork/ColonyNetworkExtensions.sol
area: > the upstream Colony resolvers are not similarly locked

I'm not sure what you mean by this? Can you explain?

Do you mean that the extension resolvers themselves are still able to have the functions they have implemented / the particular implementations that they point at changed still? That's only been the case while we've been finding our feet and I've kept those permissions, and won't be the case at some point in the future where we feel comfortable. We'll certainly be comfortable giving up those permissions for extensions sooner than we have been for the colony resolvers themselves.

Even if that were always the case, it is not true to say that my proposed changes guarantee nothing - they give us two useful guarantees to lean on:

A colony that has version x of an extension installed is using the resolver at a particular address.
Existing colonies using the extension are able to upgrade to the new version that has been added.

kronosapiens: Alright, let's do it.

contracts/colonyNetwork/ColonyNetworkExtensions.sol
```diff
@@ -0,0 +1,135 @@
...
+  {
+    return resolvers[_extensionId][_version];
+  }
+
+  function getExtensionInstallation(bytes32 _extensionId, address _colony)
+    public
+    view
+    returns (address)
+  {
+    return installations[_extensionId][_colony];
```
area, line 124: I'm going back and forth on whether this should return the version as well.

contracts/colonyNetwork/IColonyNetwork.sol
```diff
@@ -285,6 +295,45 @@ contract IColonyNetwork is ColonyNetworkDataTypes, IRecovery {
   /// @return miningResolverAddress The address of the mining cycle resolver currently used by new instances
   function getMiningResolver() public view returns (address miningResolverAddress);
 
+  /// @notice Add a new extension resolver to the Extensions repository.
+  /// @dev Can only be called by the MetaColony.
+  /// @dev The extension version is queried from the resolver itself.
+  /// @dev The roles array can be set only for version == 1 (must be empty otherwise).
```
area, line 301: No longer exists

contracts/colonyNetwork/IColonyNetwork.sol
```diff
@@ -285,6 +295,45 @@ contract IColonyNetwork is ColonyNetworkDataTypes, IRecovery {
...
+  /// @param extensionId keccak256 hash of the extension name, used as an indentifier
+  /// @param version Version of the extension to install
+  function installExtension(bytes32 extensionId, uint256 version) public;
+
+  /// @notice Upgrade an extension in a colony. Can only be called by a Colony.
+  /// @param extensionId keccak256 hash of the extension name, used as an indentifier
+  /// @param newVersion Version of the extension to upgrade to (must be one greater than current)
+  function upgradeExtension(bytes32 extensionId, uint256 newVersion) public;
+
+  /// @notice Deprecate an extension in a colony. Can only be called by a Colony.
```
area, line 316: Or undeprecate. I note the ability to undeprecate isn't really reflected in the name of the function, but I'm not really sure what we'd rename it to to improve that.

contracts/extensions/FundingQueue.sol
```diff
@@ -327,39 +351,53 @@ contract FundingQueue is DSMath, PatriciaTreeProofs {
...
+      return 999994102614216063;
+    } else if (bin == 7) {
+      return 999992877291965621;
+    } else if (bin == 8) {
+      return 999991553844799874;
+    } else if (bin == 9) {
+      return 999990115177810890;
+    } else {
+      return 999988539298800050;
+    }
```
area, line 398: Every time I look at this function I implemented, I still don't like it :P. If we nested if statements to do a binary search, we'd improve the gas cost of it being called, but it'd probably look awful... :disappointed:

contracts/testHelpers/TestExtensions.sol
```diff
@@ -0,0 +1,62 @@
...
+
+
+contract TestExtension0 is TestExtension {
+  function version() public pure returns (uint256) { return 0; }
+}
+
+
+contract TestExtension1 is TestExtension {
+  function version() public pure returns (uint256) { return 1; }
+  function sendEther() external payable {}
```
area, line 50: `receiveEther`, surely?

helpers/test-data-generator.js
```diff
@@ -384,7 +384,7 @@ export async function setupRandomColony(colonyNetwork) {
 }
 
 export async function setupColony(colonyNetwork, tokenAddress) {
-  const { logs } = await colonyNetwork.createColony(tokenAddress, 0, "", "", false);
+  const { logs } = await colonyNetwork.createColony(tokenAddress, 0, "", "", true);
```
area, line 387: Not sure what the point of this particular change is?

contracts/colonyNetwork/ColonyNetworkExtensions.sol
kronosapiens: I don't think it's necessary -- you can get the version from the extension itself.

contracts/extensions/FundingQueue.sol
kronosapiens: I hear you, but I think it's best to leave it alone for now. It works, its fine, its out of the way.

helpers/test-data-generator.js
kronosapiens: Gosh me neither.

contracts/colonyNetwork/IColonyNetwork.sol
kronosapiens: We could call it "setExtensionDeprecation" but I think that passing in a boolean is sufficiently clear.

area (commented):
I'm also a little perturbed by the fact that a rebase required updating the metaColony state but not a normal colony state... can you remember what you were rebasing on to?

kronosapiens:
@area I rebased onto the tokenLocking address change, and during the rebase I may have chosen the "wrong" hash and fixed it later (it was a bit hard to tell).

area (approved):

area:
Yeah, that's the conclusion I've come to - it's not changed in my 'Update smoke test storage hashes commit' shortly before, when I think it probably should have been.

If you're happy with the extra require I've added, I'm now happy with this, I think.

PR #717, opened 2019-09-24
Introduce secure vs. managed tasks

kronosapiens:
Part of the ongoing "task types" project.

WIP


collinvine:
@kronosapiens is this still being worked on, and is it still planned for LWSS? I'm thinking that this no longer a relevant feature. Thoughts?

kronosapiens:
@collinvine Alex and I discussed this today during sprint planning -- we can pick it up, but wanted to make sure it was still relevant. You're saying it's not? What's the reasoning?

collinvine:
TBD, but plan is to simplify the task feature. Meaning, natively supporting expenditures & also support the 1tx payment via the extension manager. In the UI, the "task" feature would be drastically reduced. cc @jakzilla

kronosapiens:
@collinvine Ok, works for us. The nice thing is we can toss this in really whenever we want. The only consideration is when/how we deprecate the built-in task functionality, since the longer we leave it lying around the harder it will be to remove it later (that said, I doubt it's getting a lot of use currently).

PR #718, opened 2019-09-25
Do not use user-supplied branchmasks if possible

area:
While working through the whitepaper updates around dispute resolution, I realised it was possible to confirm a justification root hash where RH_N was not the last leaf in the tree, but the tree had N leaves still. The correct justification tree would look like:

```
       / \
      /\ /\
     A B C D
```
 And an incorrect one could look like:
```
       / \
      /\ /\
     A B D E
```

The proofs for A and D would still have the same number of siblings, which was all we were checking during the confirmation. Such a submission would ultimately be found out, because the submitter would find themselves unable to respond during the search for the first key where there was a disagreement - either when the missing key was asked for as a `targetNode` during the search, or when a proof of the missing key was asked for during `confirmBinarySearchResult`. However, we can stop it getting to that point in this case.

Instead, now we calculate the branchmask we expect to see for a given response, rather than using one provided by the user. This means we can now guarantee that RH_N is the last key in the tree (because we know what the branchmask should look like in that case) along with RH_0 being the first (though this is trivial for key 0 in a tree where the keys are not hashed). We are still not able to guarantee that all keys between 0 and N are present, but that would only be possible by proving the existence of every second key in the tree, which is not feasible on chain.

I've tried to explain in the comments of the `expectedBranchMask` function what is going on in that function, but I appreciate it could easily look like black magic like all good bit twiddles do. Note that it assumes that the tree is 'full' i.e. all keys from 0 to N are present, but otherwise, happy to talk it through some.

We are only able to do this for the justification tree, where the keys are not hashed, not for the reputation state tree.

So long as we are still the only reputation miner when this is deployed, we can afford to be 'lazy' with the deployment, as the only changes to the mining contract interfaces are to do with the dispute resolution - which if we're the only miner, cannot happen.

### DIFF
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -402,9 +402,8 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     uint256 reputationRootHashNNodes = IColonyNetwork(colonyNetworkAddress).getReputationRootHashNNodes();
     uint256 nLogEntries = reputationUpdateLog.length;
 
-    uint256 nUpdates = reputationUpdateLog[nLogEntries-1].nUpdates +
-      reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes;
-    submission.jrhNNodes = nUpdates + 1; // This is the number of nodes we expect in the justification tree
+    submission.jrhNNodes = reputationUpdateLog[nLogEntries-1].nUpdates +
+      reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes + 1; // This is the number of nodes we expect in the justification tree
 
     uint256 expectedLength = expectedProofLength(submission.jrhNNodes, 0);
     require(expectedLength == siblings1.length, "colony-reputation-mining-invalid-jrh-proof-1-length");
--- a/contracts/ReputationMiningCycle.sol
+++ b/contracts/ReputationMiningCycle.sol
@@ -319,7 +319,6 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     uint256 round,
     uint256 idx,
     bytes memory jhIntermediateValue,
-    uint256 branchMask,
     bytes32[] memory siblings
   ) public
   {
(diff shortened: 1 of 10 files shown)


kronosapiens:
Reading this, my comprehension is:

1. A user cannot successfully defend a false branch mask, but they can use it to extend the length of the dispute process, which represents a type of griefing attack.

2. Because in the special case of the JRH (with plaintext keys), we can infer deterministically the exact tree structure from the number of nodes only, and thus infer the only possible branch mask for any given node.

3. Using (2), we can prevent (1).

Is that right?

area:
Broadly speaking, yes. It doesn't remove entirely this type of attack (it is still possible to construct a JT that is missing nodes that will pass these checks), but reduces the 'wiggle room' we're allowing on such submissions.

contracts/ReputationMiningCycle.sol
```diff
@@ -399,27 +398,32 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     // Require we've not submitted already.
     require(submission.jrhNNodes == 0, "colony-reputation-jrh-hash-already-verified");
 
-    // Get reputation root hash NNodes, which we need in both of the following checkJRHProofs
+    // Calculate how many updates we're expecting in the justification tree
     uint256 reputationRootHashNNodes = IColonyNetwork(colonyNetworkAddress).getReputationRootHashNNodes();
+    uint256 nLogEntries = reputationUpdateLog.length;
+
+    uint256 nUpdates = reputationUpdateLog[nLogEntries-1].nUpdates +
```
kronosapiens: Where is this being used?

packages/reputation-miner/test/MaliciousReputationMinerWrongJRHRightNNodes.js
```diff
@@ -0,0 +1,82 @@
...
+        .mul(constant);
+
+      if (ethers.utils.bigNumberify(entryHash).lt(target)) {
+        entryIndex = i;
+        break;
+      }
+    }
+    if (!entryIndex) {
+      return new Error("No valid entry for submission found");
+    }
```
kronosapiens: I feel like all of the above could probably be extracted into a `miner-utils.js` file somewhere.

packages/reputation-miner/test/MaliciousReputationMinerWrongJRHRightNNodes.js
```diff
@@ -0,0 +1,82 @@
...
+    // Mess up the JRH
+
+    const jt2 =  new PatriciaTreeNoHash();
```
kronosapiens: Lint meeeeeeeeeeeeeeeee

contracts/ReputationMiningCycle.sol
area: Only in the next line, so it can just be all done at once. This was refactored out of `checkJRHProof2` where it was used a little more, but not much.

packages/reputation-miner/test/MaliciousReputationMinerWrongJRHRightNNodes.js
area: The relevant functionality is already in `getEntryIndex` in the class. The miner I copied this from just hadn't been updated since that refactoring. Both this file and that file are a lot smaller now they're using that function 😄

packages/reputation-miner/test/MaliciousReputationMinerWrongJRHRightNNodes.js
area: This wasn't invalid according to our linter setup. I've added the rule [`no-multi-spaces`](https://eslint.org/docs/rules/no-multi-spaces), with an allowance for spaces to be used to pad the start of comments.

area:
The test I've added the timeout to was fixed as part of #715 - it makes sense it can take longer now,  depending on what exactly Circle spool up for us / the structure of the tree, and so is much more prone to hitting the test timeout.

PR #721, opened 2019-09-30
Say leaves when we mean leaves

area:
This is very embarrassing! For clarity:

```
      .
     / \
    .   .
   / \ / \
  .  . .  .
```

This tree has seven nodes, and four leaves (also known as leaf nodes). The number of leaves is the number of reputations in the reputation state, not the number of nodes, as we have been so cheerfully using. I think there's only one change to the interface (`getReputationRootHashNNodes()` -> `getReputationRootHashNNLeaves`), aside from the naming of parameters in functions (not a problem) and events (possibly a problem, but I think it's okay).

Given the change in the interface, not sure what we should do with this release wise...


area:
We could leave `getReputationRootHashNNodes()` on the interface for a version, and mark as deprecated?

kronosapiens:
If that's the only public function which is affected, I think that's a fine idea. I would include a `console.log` message in `getReputationRootHashNNodes` which loudly prints something along the lines of

```
DEPRECATION WARNING: getReputationRootHashNNodes is being deprecated in favor of the more accurate getReputationRootHashNLeaves, and will be removed in a future release.
```

contracts/ColonyNetwork.sol
```diff
@@ -78,8 +78,8 @@ contract ColonyNetwork is ColonyNetworkStorage {
     return reputationRootHash;
   }
 
-  function getReputationRootHashNNodes() public view returns (uint256) {
-    return reputationRootHashNNodes;
+  function getReputationRootHashNLeaves() public view returns (uint256) {
```
kronosapiens: So something like:

```
 function getReputationRootHashNNodes() public view returns (uint256) {
    console.log("DEPRECATED: use `getReputationRootHashNNodes` instead.")
    return reputationRootHashNLeaves;
  }
```

contracts/reputationMiningCycle/ReputationMiningCycle.sol
```diff
@@ -712,15 +713,15 @@ contract ReputationMiningCycle is ReputationMiningCycleCommon {
...
-    nNodes -= 1;
-    uint256 nextPowerOfTwo = nextPowerOfTwoInclusive(nNodes + 1);
+  function expectedProofLength(uint256 nLeaves, uint256 leaf) private pure returns (uint256) { // solium-disable-line security/no-assign-params
+    nLeaves -= 1;
+    uint256 nextPowerOfTwo = nextPowerOfTwoInclusive(nLeaves + 1);
     uint256 layers = 0;
-    while (nNodes != 0 && (node+1 > nextPowerOfTwo / 2)) {
-      nNodes -= nextPowerOfTwo/2;
-      node -= nextPowerOfTwo/2;
+    while (nLeaves != 0 && (leaf+1 > nextPowerOfTwo / 2)) {
```
kronosapiens, line 720: Spacing is inconsistent but not gonna worry about it.

PR #723, opened 2019-10-04
Add endpoint to reputation oracle to list addresses with skill in a colony

area:
Adds an API endpoint `/:rootHash/:colonyAddress/:skillId/` that returns 
```
{ addresses: [] }
```
Where the array contains a list of addresses that have the skill `skillId` in `colonyAddress` in the reputation state represented with the root `rootHash`.

As part of this, I've also sanitised user inputs for this and the proof endpoint, rather than just passing the user-supplied parameters in to the SQL query. Nothing ever went wrong doing that, right? 😓 

Marking this as draft for now so some discussion with the dapp team can occur as to what they might like changed.

### DIFF
--- a/packages/reputation-miner/ReputationMinerClient.js
+++ b/packages/reputation-miner/ReputationMinerClient.js
@@ -136,7 +136,11 @@ class ReputationMinerClient {
         }
 
         try {
-          const [branchMask, siblings, value] = await this._miner.getHistoricalProofAndValue(req.params.rootHash, key);
+          const historicalProof = await this._miner.getHistoricalProofAndValue(req.params.rootHash, key);
+          if (historicalProof instanceof Error) {
+            return res.status(400).send({ message: historicalProof.message.replace("Error: ") });
+          }
+          const [branchMask, siblings, value] = historicalProof;
           const proof = { branchMask: `${branchMask.toString(16)}`, siblings, key, value };
           proof.reputationAmount = ethers.utils.bigNumberify(`0x${proof.value.slice(2, 66)}`).toString();
           return res.status(200).send(proof);
--- a/packages/reputation-miner/ReputationMinerClient.js
+++ b/packages/reputation-miner/ReputationMinerClient.js
@@ -95,6 +95,16 @@ class ReputationMinerClient {
         return res.status(200).send({ active: activeAddr, inactive: inactiveAddr });
       });
 
+      // Query users who have given reputation in colony
+      this._app.get("/:rootHash/:colonyAddress/:skillId/", async (req, res) => {
+        const addresses = await this._miner.getAddressesWithReputation(req.params.rootHash, req.params.colonyAddress, req.params.skillId);
+        try {
+          return res.status(200).send({ addresses });
+        } catch (err) {
+          return res.status(400).send({ message: "Requested reputation does not exist or invalid request" });
+        }
+      });
+
       // Query specific reputation values
       this._app.get("/:rootHash/:colonyAddress/:skillId/:userAddress", async (req, res) => {
         const key = ReputationMiner.getKey(req.params.colonyAddress, req.params.skillId, req.params.userAddress);
--- a/packages/reputation-miner/ReputationMinerClient.js
+++ b/packages/reputation-miner/ReputationMinerClient.js
@@ -102,31 +102,13 @@ class ReputationMinerClient {
           !ethers.utils.isHexString(req.params.colonyAddress) ||
           !ethers.utils.bigNumberify(req.params.skillId)
         ) {
-          return res.status(400).send({ message: "Requested reputation does not exist or invalid request" });
+          return res.status(400).send({ message: "One of the parameters was incorrect" });
         }
         const addresses = await this._miner.getAddressesWithReputation(req.params.rootHash, req.params.colonyAddress, req.params.skillId);
         try {
           return res.status(200).send({ addresses });
         } catch (err) {
-          return res.status(400).send({ message: "Requested reputation does not exist or invalid request" });
-        }
-      });
-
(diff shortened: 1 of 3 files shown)


sprusr:
Yeah this looks great 👍

I can't think of a use case in the dApp right now, but being able to query without a `skillId` seems like it could be useful (return users with rep from any skill) and easy to add on to this.

area:
I thought about that too, but couldn't come up with a scenario where that would result in a different list than querying with the `skillId` associated with the root domain? i.e. I'm saying that anyone that has earned reputation in a global-skill has also earned reputation in the root domain - can you think of a counterexample?

sprusr:
Ah I see, you're quite right. Then I think this PR does everything we'd want.

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -95,8 +95,52 @@ class ReputationMinerClient {
...
       });
 
+      // Query users who have given reputation in colony
+      this._app.get("/:rootHash/:colonyAddress/:skillId/", async (req, res) => {
+        if (
+          !ethers.utils.isHexString(req.params.rootHash) ||
+          !ethers.utils.isHexString(req.params.colonyAddress) ||
+          !ethers.utils.bigNumberify(req.params.skillId)
+        ) {
+          return res.status(400).send({ message: "Requested reputation does not exist or invalid request" });
```
kronosapiens: Seems as though you could be more specific with the error messages here. In this case, the request is invalid, while in the next case, the reputation does not exist. Using the same message for both just withholds information from the user.

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -95,8 +95,52 @@ class ReputationMinerClient {
...
+        }
+        const addresses = await this._miner.getAddressesWithReputation(req.params.rootHash, req.params.colonyAddress, req.params.skillId);
+        try {
+          return res.status(200).send({ addresses });
+        } catch (err) {
+          return res.status(400).send({ message: "Requested reputation does not exist or invalid request" });
+        }
+      });
+
+      // Query users who have given reputation in colony
```
kronosapiens: Am I mad or is this appearing twice?

kronosapiens:
Regarding the earlier discussion, two points:

1. Technically, this endpoint returns "users who have non-zero reputation for the given skillId", since in theory all users have reputation in all skills, but the value is often zero. Right?

2. Regarding giving the root domain skill to get all users with reputation, that wouldn't include skills reputation, would it? I don't recall whether skills reputation is included in root domain rep, but it would be quirky if it did. Although from @area's second comment maybe that is the case?

packages/reputation-miner/ReputationMinerClient.js
area: I agree this one could be better. I don't believe the reputation not existing would trigger the next one, though - should just return an empty list. I guess a failed `try/catch` should return a 500 error code though. Have tried to improve the error messages.

area:
> Technically, this endpoint returns "users who have non-zero reputation for the given skillId", since in theory all users have reputation in all skills, but the value is often zero. Right?

It returns users who have the reputation in the given `skillId`. If I've never earned reputation in a skill, it's not in the tree. If I earn reputation, and then it decays away to 0 or I lose it through misbehaviour, I will still be returned in this list (because I am still in the tree with 0 reputation).

> Regarding giving the root domain skill to get all users with reputation, that wouldn't include skills reputation, would it? I don't recall whether skills reputation is included in root domain rep, but it would be quirky if it did. Although from @area's second comment maybe that is the case?

They are not directly related, but I was trying to say that (I think) in all cases where you earn reputation in a skill, you have also earned reputation in a domain, and therefore in the root domain. Is there a scenario where you can only earn reputation in a skill I've overlooked?

kronosapiens:
> It returns users who have the reputation in the given skillId. If I've never earned reputation in a skill, it's not in the tree. If I earn reputation, and then it decays away to 0 or I lose it through misbehaviour, I will still be returned in this list (because I am still in the tree with 0 reputation).

Makes sense, my point was more to do with the formalism where we said that "all users have reputation even if 0", but that's more to do with the notation than practice. Makes sense that, inasmuch as here we are querying an actual db, we would only return reputation that has ever been non-zero.

> They are not directly related, but I was trying to say that (I think) in all cases where you earn reputation in a skill, you have also earned reputation in a domain, and therefore in the root domain. Is there a scenario where you can only earn reputation in a skill I've overlooked?

Yup that makes sense, I see what you mean now.

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -136,7 +136,11 @@ class ReputationMinerClient {
         }
 
         try {
-          const [branchMask, siblings, value] = await this._miner.getHistoricalProofAndValue(req.params.rootHash, key);
+          const historicalProof = await this._miner.getHistoricalProofAndValue(req.params.rootHash, key);
+          if (historicalProof instanceof Error) {
+            return res.status(400).send({ message: historicalProof.message.replace("Error: ") });
```
kronosapiens: Is this the intended error message?

area: I'm not sure what you're asking?

kronosapiens: Is the user going to see `Error: ` as the message?

area: No, it's stripping the 'Error' in the message that is returned from `getHistoricalProofAndValue` (my reasoning being that getting a 400 status code already conveys there was an error, and the other error messages returned by this endpoint don't prefix the message with 'Error:'.

PR #731, opened 2019-10-30
Remove 'newest reputation' proofs from mining disputes

area:
As part of the whitepaper work, @kronosapiens asked why we bothered having the users prove an existing UID in the tree when adding a new reputation leaf, when we should know the number of leaves in the last agree state and therefore the UID that the new node should have. I didn't have an answer,[1] so here we are with a pleasingly large removal of code from the dispute side (can we do this more often?)

It's a fairly straightforward change - the key change is [here](https://github.com/JoinColony/colonyNetwork/compare/develop...maint/mining-dispute-efficiency#diff-abc40266312ec550436f1217f4be3299R502), everything else is really a knock-on of that. If we're not using the `PREVIOUS_NEW_REPUTATION`, we get to remove it and everything required for its proof (five variables in `respondToChallenge`), we remove tests involving getting that reputation wrong and remove the bookkeeping in the clients around tracking which reputation was the previous newest reputation for each update.

With #718, we assumed that we were the only reputation miner until it was deployed, so I've taken the liberty of doing that here - there is a storage change in `ReputationMiningCycle`, but as that is a disposable contract, so long as after this is deployed we are the only miner until the old contract has been removed (i.e. two cycles are completed post deployed) there should not be an issue.

[1] My best working guess is that I'm an idiot


kronosapiens:
https://en.wikipedia.org/wiki/Linus%27s_law

packages/reputation-miner/test/MaliciousReputationMinerWrongResponse.js
```diff
@@ -62,62 +62,55 @@ class MaliciousReputationMinerWrongResponse extends ReputationMinerTestWrapper {
...
           this.responseValue :
           soliditySha3(lastAgreeJustifications.childAdjacentReputationProof.key)
       ],
-      this.responseToFalsify === 37 ? this.responseValue : firstDisagreeJustifications.justUpdatedProof.siblings,
-      this.responseToFalsify === 38 ? this.responseValue : agreeStateSiblings,
-      this.responseToFalsify === 39 ? this.responseValue : disagreeStateSiblings,
-      this.responseToFalsify === 40 ? this.responseValue : lastAgreeJustifications.newestReputationProof.siblings,
-      this.responseToFalsify === 41 ? this.responseValue : lastAgreeJustifications.originReputationProof.siblings,
-      this.responseToFalsify === 42 ? this.responseValue : lastAgreeJustifications.childReputationProof.siblings,
-      this.responseToFalsify === 43 ? this.responseValue : lastAgreeJustifications.adjacentReputationProof.siblings,
```
kronosapiens: My one question is -- if there are a total of 4 arguments removed (3 + 1), why has the number of options here been reduced by 5?

kronosapiens:
Also, I pulled the branch and found some leftover references to `newestReputationProof` -- mostly in `disputes-over-child-reputation.js` but one in `MaliciousReputationMinerUnsure.js`. Are those meant to have been removed?

packages/reputation-miner/test/MaliciousReputationMinerWrongResponse.js
area: By 3+1, I assume you mean `uint256[29] -> uint256[26]` and `bytes32[8] -> bytes32[7]` but the array of siblings for the relevant proof [is also removed](https://github.com/JoinColony/colonyNetwork/pull/731/files/22f2f4c47b3f69dd40667ae717bfebbcd97f5e37..23aa9d62470e8596cf8515369080be3afa95bdfa#diff-38a06e49a953fabbbea435027d827918L157)

area:
The instance in `...Unsure` can be removed, that's completely unnecessary now. The other test is skipped because contract changes meant the thing it was testing for could no longer happen. We could just remove it, I suppose, rather than updating a test that's impossible to execute...

PR #732, opened 2019-10-31
Update contracts to Solidity 0.6.0

erak:
As part of Solidity's testing strategy, a few external projects are compiled and tested with nightly builds of the upcoming breaking release 0.6.0. The Colony contracts are part of this and would need some updates such that they compile.

This PR mainly updates fixed version pragmas to also allow 0.6.0 and adjusts to the defaulting strict inline assembly.

Target branch is up for discussion.

### DIFF
--- a/contracts/Colony.sol
+++ b/contracts/Colony.sol
@@ -15,7 +15,7 @@
   along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
 */
 
-pragma solidity 0.5.8;
+pragma solidity >=0.5.8 <0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "./ColonyStorage.sol";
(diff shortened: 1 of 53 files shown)


area:
Nothing major then, I was briefly worried when I saw 46 files changed! Doesn't look like there should be any issues from our side (having also looked over the [`Changelog.md`](https://github.com/ethereum/solidity/blob/develop_060/Changelog.md) in the solidity 0.6.0 branch.

Target branch is fine as-is, will watch with interest! Is there even a vague ETA for the 0.6.0 release or are we just sitting tight for now?

kronosapiens:
Hip hip hooray for `payable(...)`!

area:
Closing in favour of #870

PR #748, opened 2019-12-05
Make --no-auto function correctly

area:
When I split up the oracle and the mining client, I failed to do it correctly. Even if we're not mining, we still need to follow updates as they happen and update our local state. So this rearranges a few things so that `--no-auto` just disables the mining, but still follows updates as they happen.

EDIT: Marking this as `in-progress` because I'm still testing with the goerli side, but I think the changes are done and would appreciate a review anyway to see if there's something obvious I've missed, just not a merge yet.

### DIFF
--- a/packages/reputation-miner/ReputationMinerClient.js
+++ b/packages/reputation-miner/ReputationMinerClient.js
@@ -34,6 +34,7 @@ class ReputationMinerClient {
     this._oracle = oracle;
     this._exitOnError = exitOnError;
     this.submissionIndex = 0;
+    this.blocksSinceCycleCompleted = 0;
     this.best12Submissions = [];
     this.lockedForBlockProcessing;
     this._adapter = adapter;
@@ -256,7 +257,7 @@ class ReputationMinerClient {
 
       this.lockedForBlockProcessing = true;
       // DO NOT PUT ANY AWAITS ABOVE THIS LINE OR YOU WILL GET RACE CONDITIONS
-      // If you leave this function after this line, make sure to call this.endDoBlockChecks() to unlock
+      // When you leave this function, make sure to call this.endDoBlockChecks() to unlock
 
       if (this.blockTimeoutCheck) {
         clearTimeout(this.blockTimeoutCheck);
@@ -264,9 +265,14 @@ class ReputationMinerClient {
 
       const block = await this._miner.realProvider.getBlock(blockNumber);
       const addr = await this._miner.colonyNetwork.getReputationMiningCycle(true);
-
       if (addr !== this.miningCycleAddress) {
-        // Then the cycle has completed since we last checked. Let's process the reputation log
+        // Then the cycle has completed since we last checked. Let's process the reputation log if it's been ten blocks
+        if (this.blocksSinceCycleCompleted < 10) {
+          this.blocksSinceCycleCompleted += 1;
+		  if (this.blocksSinceCycleCompleted === 1) { this._adapter.log("⏰ Waiting for ten blocks before processing next log") };
+          this.endDoBlockChecks();
+          return;
+        }
         await this.processReputationLog();
 
         // And if appropriate, sort out our potential submissions for the next cycle.
@@ -275,6 +281,8 @@ class ReputationMinerClient {
         }
 
         this.miningCycleAddress = addr;
+        this.blocksSinceCycleCompleted = 0;
+
         if (this.confirmTimeoutCheck) {
           clearTimeout(this.confirmTimeoutCheck);
         }
--- a/packages/reputation-miner/ReputationMinerClient.js
+++ b/packages/reputation-miner/ReputationMinerClient.js
@@ -8,6 +8,8 @@ const ReputationMiner = require("./ReputationMiner");
 
 const minStake = ethers.utils.bigNumberify(10).pow(18).mul(2000); // eslint-disable-line prettier/prettier
 const miningCycleDuration = ethers.utils.bigNumberify(60).mul(60).mul(24); // 24 hours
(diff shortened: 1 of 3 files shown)


packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -225,24 +227,46 @@ class ReputationMinerClient {
...
+        // Then the cycle has completed since we last checked.
         if (this.confirmTimeoutCheck) {
           clearTimeout(this.confirmTimeoutCheck);
         }
-        // If we don't see this cycle completed in the next day and ten minutes, then report it
+        // If we don't see this next cycle completed in the next day and ten minutes, then report it
         this.confirmTimeoutCheck = setTimeout(this.reportConfirmTimeout.bind(this), (24 * 3600 + 600) * 1000);
+
+        // Let's process the reputation log if it's been ten blocks
+        if (this.blocksSinceCycleCompleted < 10) {
```
kronosapiens: Remind me why we process the reputation log every ten blocks? If it's the active mining cycle, it's no longer being updated, so why keep checking? And if it's the inactive one, why process it before it's active?

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -372,6 +396,9 @@ class ReputationMinerClient {
       if (this._exitOnError) {
         process.exit(1);
       }
+      // Note we don't call this.endDoBlockChecks here... this is a deliberate choice on my part; depending on what the error is,
```
kronosapiens: Should this comment go in the `exitOnError` block?

test/reputation-system/reputation-mining-client/client-auto-functionality.js
```diff
@@ -80,7 +80,8 @@ process.env.SOLIDITY_COVERAGE
           realProviderPort,
           minerAddress: MINER1,
           useJsTree: true,
-          auto: true
+          auto: true,
+          oracle: false
```
kronosapiens: Maybe I missed it but what's the purpose of this flag? Seems like it's the same as setting `auto = false`, but maybe not? Does it tell the miner to run the express server? I guess you'd want to be able to auto-mine without exposing the server so it does make sense to have them be orthogonal features.

test/reputation-system/reputation-mining-client/client-auto-functionality.js
```diff
@@ -164,6 +165,102 @@ process.env.SOLIDITY_COVERAGE
...
+              } else {
+                await mineBlock();
+              }
+            });
+
+            // After 30s, we throw a timeout error
+            setTimeout(() => {
+              reject(new Error("ERROR: timeout while waiting for 12 hash submissions"));
+            }, 30000);
+          });
```
kronosapiens: This section doesn't involve the oracleClient at all and looks a bit boilerplate-y -- any way this can be pulled out into a helper function? Or let me know if I'm mistaken.

test/reputation-system/reputation-mining-client/client-auto-functionality.js
```diff
@@ -164,6 +165,102 @@ process.env.SOLIDITY_COVERAGE
...
+
+          // Forward time to the end of the mining cycle and since we are the only miner, check the client confirmed our hash correctly
+          await forwardTime(MINING_CYCLE_DURATION * 0.1, this);
+          await miningCycleComplete;
+
+          await oracleUpdated;
+
+          clearTimeout(oracleCheckInterval);
+          await oracleClient.close();
```
kronosapiens: Should this end with an `expect` or something to confirm that the oracleClient is synced up with the miner?

packages/reputation-miner/ReputationMinerClient.js
area: So if this line is running, a mining cycle has completed, and we've not processed the next log yet. My intention here is to wait until it has been ten blocks to process the next log to avoid falling foul of a block reorganisation that doesn't include the transaction that completes the cycle; if that happens, then the log processing will fail as after the reorganisation, it will be attempting to query a contract that doesn't exist.

Once the log has been processed, it doesn't get processed again - it is certainly not meant to be running every ten blocks (and if it is, that's very wrong, as you say!)

This is my best guess as to the intermittent error we were seeing on goerli, which did coincide with reorgs each time it was seen.

test/reputation-system/reputation-mining-client/client-auto-functionality.js
area: Yeah, that's exactly what we're doing, and is probably ultimately what we should recommend people do - no server exposed means no private keys exposed if there's an express exploit.

packages/reputation-miner/ReputationMinerClient.js
kronosapiens: Ah I see, because the miner isn't set up to correct itself for a reorg, you're basically waiting for 10 blocks as a type of probabilistic finality.

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -8,6 +8,8 @@ const ReputationMiner = require("./ReputationMiner");
 
 const minStake = ethers.utils.bigNumberify(10).pow(18).mul(2000); // eslint-disable-line prettier/prettier
 const miningCycleDuration = ethers.utils.bigNumberify(60).mul(60).mul(24); // 24 hours
+const MINUTE_IN_SECONDS = 60;
+const DAY_IN_SECONDS = 3600 * 24;
```
kronosapiens: Can we keep the convention from `constants`, i.e. `SECONDS_PER_DAY`?

PR #750, opened 2019-12-06
Fix the nightly

area:
New version of slither since the last PR meant the ignores changed again (though I don't really understand how), and I'm finally tired of chasing consensys/truffle-security#245, so I've given up and disabled SWC101 (which I admit isn't a great one to disable...) until it gets sorted.


kronosapiens:
Noting that bumping truffle-security created a lot of changes to `yarn.lock`

Issue #754, opened 2019-12-10
Remove reward payout time restrictions


kronosapiens:
Currently there is a restriction where only one reward payout can be active at a time, and each reward payout is active for 60 days (meaning that only one payout can be initiated every 60 days).

This now appears to be an overly-restrictive restriction, as there are use-cases where reward payouts may need to be initiated more frequently. Given that now reward payouts are secured with the `root` permission, any time restriction at all seems less necessary. We should remove the time restriction and add support for multiple _simultaneous payouts_.

Regarding implementation, one approach would be to track activePayoutTotal and payout.totalClaimed and use that to determine the amount available for any particular payout cycle:

```
// When starting a new payout:
payout.totalPayout = fundingPots[0].balance[token] - activePayouts[token]; // the inactive balance

// When finalizing a payout:
activePayouts[token] -= (payout.totalPayout - payout.totalClaimed) // the unclaimed surplus
```

PR #770, opened 2020-01-09
Prevent 'finalisation' of old reward payouts

### LINKED ISSUE #763: Fix Rewards Payouts
Unfortunately rewards payouts are broke. In two ways:

1. In `finalizeRewardPayout`, [this line](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyFunding.sol#L348) does not check to see if the _given payout_ is still active, only that _any_ payout for that token is active. Which means a malicious user could repeatedly "finalize" an old rewards payout, allowing them to initiate arbitrarily many rewards payouts and claim the funds. What _should_ happen is that any individual payout can only be finalized once, 60 days after it is created.

~2. In `startNextRewardPayout`, we use [the total number of locks](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/ColonyFunding.sol#L278) _for a token_ as the key for the reward payout _for a colony_. This is bad because these two numbers mean different things -- both tokens A and B can have the same number of locks, which would mean the reward payout for one will be replaced by the reward payout for the other. Instead, what we should do is key the reward payout by _both_ the number of locks and the token, as every token/lock combination is guaranteed to be unique.~

Regarding the above strikeout: upon second look, it turns out that the total number of locks is not of the token in question (being distributed), but rather of the colony's internal token. This means that the issue -- two different tokens could have the same number of locks -- is a non-issue, since every lock will be unique (since only one token is being locked).

area:
Closes #763 

The fix here is arguably controversial, which hinges on changing

```  
mapping (address => bool) activeRewardPayouts;
```
to
```
mapping (address => uint256) activeRewardPayouts;
```

This is a cardinal sin of upgradable contracts, after all, so how is this justifiable? It's because no-one has started a reward payout yet, and the functionality is disabled on the deployed versions - so we know that this mapping will be empty for all colonies that have not upgraded to this version. Changing it from a `bool` to a `uint256` therefore is fine, as far as I am aware.

The token lock count is a network-wide lock count for the colony's token, which means payouts in `rewardPayoutCycles` will not necessarily share adjacent keys. That lock count is [1-indexed](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/TokenLocking.sol#L69), which means we can use a 0 in `activeRewardPayouts` to indicate that there is no active reward for a token currently.

Of course, some of this logic becomes invalid as soon as I start working on #754 again but best to start from a place that we think works as intended 😓 .

### DIFF
--- a/contracts/ColonyFunding.sol
+++ b/contracts/ColonyFunding.sol
@@ -277,7 +277,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
     uint256 totalLockCount = tokenLocking.lockToken(token);
 
-    require(!activeRewardPayouts[_token], "colony-reward-payout-token-active");
+    require(activeRewardPayouts[_token] == 0, "colony-reward-payout-token-active");
 
     uint256 totalTokens = sub(ERC20Extended(token).totalSupply(), ERC20Extended(token).balanceOf(address(this)));
     require(totalTokens > 0, "colony-reward-payout-invalid-total-tokens");
@@ -294,7 +294,7 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     );
     require(colonyWideReputation > 0, "colony-reward-payout-invalid-colony-wide-reputation");
 
-    activeRewardPayouts[_token] = true;
+    activeRewardPayouts[_token] = totalLockCount;
 
     rewardPayoutCycles[totalLockCount] = RewardPayoutCycle(
       rootHash,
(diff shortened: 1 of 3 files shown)


contracts/ColonyFunding.sol
```diff
@@ -344,11 +344,12 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
 
   function finalizeRewardPayout(uint256 _payoutId) public stoppable {
     RewardPayoutCycle memory payout = rewardPayoutCycles[_payoutId];
-
-    require(activeRewardPayouts[payout.tokenAddress], "colony-reward-payout-token-not-active");
+    require(payout.reputationState != 0x00, "colony-reward-payout-does-not-exist");
+    require(activeRewardPayouts[payout.tokenAddress] > 0, "colony-reward-payout-token-not-active");
+    require(_payoutId == activeRewardPayouts[payout.tokenAddress], "colony-reward-payout-not-most-recent");
```
kronosapiens: Alternatively, couldn't this function take `_token` instead, and then set `payoutId = activeRewardPayouts[_token]`? Seems like that would eliminate the need for some of this input validation.

area: Yeah, but given that #754 is coming down the pipe very shortly which will have to use `_payoutId`, I didn't think I'd bother!

kronosapiens: Fair!

PR #784, opened 2020-01-23
Allow creation of colonies with earlier version

### LINKED ISSUE #783: Add support for creating colonies at any version
In order to properly decouple the network and the (d)app, we need to be able to do two things:

1. Release new versions of the network without requiring the (d)app to support them, and
2. Allow the (d)app to create new colonies at older version, which it supports.

This issue is to add support for creating colonies at any version, not simply the latest. Will likely look something like:

```
  function createColony(address _tokenAddress, uint256 _version)
    public
    stoppable
    returns (address)
  {
    require(_tokenAddress != address(0x0), "colony-token-invalid-address");

    address resolver = colonyVersionResolver[_version];
    require(resolver != address(0x0), "colony-resolver-invalid-address");

    EtherRouter etherRouter = new EtherRouter();
    etherRouter.setResolver(resolver);

     ...
  }
```

area:
Closes #783 

I've overloaded `createColony` rather than replacing it here (to aid our how-to-upgrade-everything process until we have it down pat), but otherwise, fairly straightforward. There are no limits here other than 'the version exists'; it's possible that in the future we'd have a minimum-supported version (especially as the app evolves) but for now this seems sufficient.

### DIFF
--- a/contracts/ColonyNetwork.sol
+++ b/contracts/ColonyNetwork.sol
@@ -129,7 +129,7 @@ contract ColonyNetwork is ColonyNetworkStorage {
   returns (address)
   {
     require(colonyVersionResolver[_version] != address(0x00), "colony-network-invalid-version");
-    return createColonyFunctionality(_tokenAddress, _version);
+    return createColonyFromVersion(_tokenAddress, _version);
   }
 
   function createColony(address _tokenAddress) public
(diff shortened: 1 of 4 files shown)


test/contracts-network/colony-network.js
```diff
@@ -152,6 +152,54 @@ contract("Colony Network", accounts => {
     });
   });
 
+  describe("when creating new colonies at a specific version", () => {
+    beforeEach(async () => {
```
kronosapiens: I don't get it -- why are we manually setting up a resolver here?

contracts/ColonyNetwork.sol
```diff
@@ -124,16 +124,31 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
+
   function createColony(address _tokenAddress) public
   stoppable
   returns (address)
   {
     require(currentColonyVersion > 0, "colony-network-not-initialised-cannot-create-colony");
+    return createColonyFunctionality(_tokenAddress, currentColonyVersion);
+  }
+
+  function createColonyFunctionality(address _tokenAddress, uint256 _version) internal
```
kronosapiens: I don't like this name, it's too abstract -- can we use `createColonyFromVersion` instead?

test/contracts-network/colony-network.js
area: I wanted a separate resolver that would allow `createColony` to be called. I didn't just want to reuse the existing one so I could check it was set up with the old version rather than the new version (though I do now see I never actually put that assert in 😅

PR #812, opened 2020-02-19
Reward users for responding during dispute

area:
Okay, this has been long enough coming! First off, this is based on #810, so marking as draft until then is merged - it is probably otherwise ready for review.

Before this PR, none of the calls during dispute resolution provided rewards, so there was going to be little incentive for responses to occur; while those that made the good submission are incentivised to defend their submission (because they don't want to lose their stake), if more than one person has made the submission then there is incentive to just wait for other people to defend the submission (and not pay gas) and still claim the rewards at the end. 

So, the fundamental question is how large the reward is. The reward is being paid for by slashed stakes (if a submission you made is found invalid, you lose `MIN_STAKE`), so the minimum pool of rewards we have available is `(NSubmissions - 1)*MIN_STAKE`.

The next question is how many responses we need to award. We want to reward people who call:

* `confirmJustificationRootHash`
* `respondtoBinarySearchForChallenge`
* `confirmBinarySearchResult`
* `eliminateHash`

I opted to not reward people who call `confirmNewHash` - this is the transaction that rewards the people who staked originally, so they are already incentivised to call it. I have also made the assumption we are going to reward all of the above calls equally, even though they have different gas costs.

For each pairwise dispute, `confirmJustificationRootHash`, `confirmBinarySearchResult` are each called up to twice, `eliminateHash` is called once, and `respondToBinarySearchForChallenge` is called up to a number of times based on how long the binary search can take, which in turn is based on the number of leaves in the `justificationTree`. Specifically, it can take `ceil(log_2(JRHLeaves))` transactions for each party.

Unfortunately, we cannot just consider the pairwise disputes. Some submissions can also be awarded byes, which involves a call to `eliminateHash`, and should get a reward, but there isn't a corresponding opposing submission being slashed. We therefore have to consider these rewards across the whole reputation cycle, not just on a pairwise basis. At most, a bye can be awarded in each round of disputes, so the total number of byes (`nByes`) we have to be able to pay out is `ceil(log_2(nSubmissions))`.

As this is based on the number of submissions, we have to know how many submissions there are before any rewards are paid out - so we now prevent any disputes from starting before the submission window is closed. Pairwise disputes can still operate independently of each other, however.

How many pairwise disputes can there be? In each pairwise dispute, either one or two submissions can be eliminated, and every submission bar one has to lose one pairwise dispute. So there can be at most `nSubmissions-1` pairwise disputes.

Putting this all together, this means we have to be able to reward:
```
nByes + (nSubmissions-1) * (2 * (3 + ceil(log_2(JRHLeaves))) + 1)
```

responses during the dispute cycle. `(NSubmissions - 1)*MIN_STAKE` divided by this number gives us the largest reward we are able to guarantee we will be able to cover with the slashed stakes.

The rewards given are locked until the end of the mining cycle because the stakes that are paying for them haven't been slashed at the time they're being paid out. This requires the changes to `TokenLocking` around who has been 'involved' in the mining cycle. There's also been a separatation of `reward` and `punish` functions in Token Locking as a result of these changes.

### DIFF
--- a/contracts/ReputationMiningCycleCommon.sol
+++ b/contracts/ReputationMiningCycleCommon.sol
@@ -1,0 +1,105 @@
+/*
+  This file is part of The Colony Network.
+
+  The Colony Network is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  The Colony Network is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+import "../lib/dappsys/math.sol";
+import "./PatriciaTree/PatriciaTreeProofs.sol";
+import "./ReputationMiningCycleStorage.sol";
+import "./ITokenLocking.sol";
+import "./IColonyNetwork.sol";
+
+
+contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTreeProofs, DSMath {
+  /// @notice Minimum reputation mining stake in CLNY
+  uint256 constant MIN_STAKE = 2000 * WAD;
+
+  function rewardResponder(address _responder) internal returns (bytes32) {
+    respondedToChallenge[_responder] = true;
+    uint256 reward = disputeRewardIncrement();
+    ITokenLocking(tokenLockingAddress).reward(
+      _responder,
+      reward
+    );
+    rewardsPaidOut += reward;
+  }
+
+  function disputeRewardIncrement() internal view returns (uint256) {
+    // TODO: Is this worth calculating once, and then saving? Seems quite likely.
+    uint256 nLogEntries = reputationUpdateLog.length;
+
+    // If there's no log, it must be one of the first two reputation cycles - no reward.
+    if (nLogEntries == 0) {
+      return 0;
+    }
+
+    // No dispute, so no dispute reward
+    if (nUniqueSubmittedHashes <= 1) {
+      return 0;
+    }
+
+    uint256 reputationRootHashNNodes = IColonyNetwork(colonyNetworkAddress).getReputationRootHashNNodes();
+    uint jrhNnodes = reputationUpdateLog[nLogEntries-1].nUpdates +
+      reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes + 1; // This is the number of nodes we expect in the justification tree
+
+    uint256 nByes = log2Ceiling(nUniqueSubmittedHashes); // We can have at most one bye per round, and this is the maximum number of rounds
+
+    // The maximum number of responses that need rewards
+    uint256 rewardDenominator = nByes + (nUniqueSubmittedHashes-1)*(2*(3 + log2Ceiling(jrhNnodes)) + 1);
+
+    // The minimum amount of stake to be lost
+    uint256 rewardNumerator = MIN_STAKE * (nUniqueSubmittedHashes - 1);
+    uint256 reward = rewardNumerator / rewardDenominator;
+    return reward;
+  }
+
+  // https://ethereum.stackexchange.com/questions/8086/logarithm-math-operation-in-solidity
+  // Some impressive de Bruijn sequence magic here...
+  function log2Ceiling(uint x) internal pure returns (uint y) {
+    assembly {
+        let arg := x
+        x := sub(x,1)
+        x := or(x, div(x, 0x02))
+        x := or(x, div(x, 0x04))
+        x := or(x, div(x, 0x10))
+        x := or(x, div(x, 0x100))
+        x := or(x, div(x, 0x10000))
+        x := or(x, div(x, 0x100000000))
+        x := or(x, div(x, 0x10000000000000000))
+        x := or(x, div(x, 0x100000000000000000000000000000000))
+        x := add(x, 1)
+        let m := mload(0x40)
+        mstore(m,           0xf8f9cbfae6cc78fbefe7cdc3a1793dfcf4f0e8bbd8cec470b6a28a7a5a3e1efd)
+        mstore(add(m,0x20), 0xf5ecf1b3e9debc68e1d9cfabc5997135bfb7a7a3938b7b606b5b4b3f2f1f0ffe)
+        mstore(add(m,0x40), 0xf6e4ed9ff2d6b458eadcdf97bd91692de2d4da8fd2d0ac50c6ae9a8272523616)
+        mstore(add(m,0x60), 0xc8c0b887b0a8a4489c948c7f847c6125746c645c544c444038302820181008ff)
+        mstore(add(m,0x80), 0xf7cae577eec2a03cf3bad76fb589591debb2dd67e0aa9834bea6925f6a4a2e0e)
+        mstore(add(m,0xa0), 0xe39ed557db96902cd38ed14fad815115c786af479b7e83247363534337271707)
+        mstore(add(m,0xc0), 0xc976c13bb96e881cb166a933a55e490d9d56952b8d4e801485467d2362422606)
+        mstore(add(m,0xe0), 0x753a6d1b65325d0c552a4d1345224105391a310b29122104190a110309020100)
+        mstore(0x40, add(m, 0x100))
+        let magic := 0x818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff
+        let shift := 0x100000000000000000000000000000000000000000000000000000000000000
+        let a := div(mul(x, magic), shift)
+        y := div(mload(add(m,sub(255,a))), shift)
+        y := add(y, mul(256, gt(arg, 0x8000000000000000000000000000000000000000000000000000000000000000)))
+    }
+  }
+
+
+}
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -210,8 +210,7 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
   function unstakeForMining(uint256 _amount) public stoppable {
     address clnyToken = IMetaColony(metaColony).getToken();
     // Prevent those involved in a mining cycle withdrawing stake during the mining process.
-    bool involved = IReputationMiningCycle(activeReputationMiningCycle).userInvolvedInMiningCycle(msg.sender);
-    require(!involved, "colony-token-locking-hash-submitted");
+    require(!IReputationMiningCycle(activeReputationMiningCycle).userInvolvedInMiningCycle(msg.sender), "colony-token-locking-hash-submitted");
     ITokenLocking(tokenLocking).deobligateStake(msg.sender, _amount, clnyToken);
     miningStakes[msg.sender].amount = sub(miningStakes[msg.sender].amount, _amount);
   }
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -196,6 +196,18 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     }
   }
 
+  function reward(address _recipient, uint256 _amount) public stoppable onlyReputationMiningCycle {
+    // TODO: Gain rep?
+    pendingMiningRewards[_recipient] = add(pendingMiningRewards[_recipient], _amount);
+  }
+
+  function claimMiningReward(address _recipient) public stoppable {
+    address clnyToken = IMetaColony(metaColony).getToken();
+    uint256 amount = pendingMiningRewards[_recipient];
+    pendingMiningRewards[_recipient] = 0;
+    ITokenLocking(tokenLocking).transfer(clnyToken, amount, _recipient, true);
+  }
+
   function stakeForMining(uint256 _amount) public stoppable {
     address clnyToken = IMetaColony(metaColony).getToken();
     uint256 existingObligation = ITokenLocking(tokenLocking).getObligation(msg.sender, clnyToken, address(this));
@@ -210,13 +222,18 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
   function unstakeForMining(uint256 _amount) public stoppable {
     address clnyToken = IMetaColony(metaColony).getToken();
     // Prevent those involved in a mining cycle withdrawing stake during the mining process.
-    require(!IReputationMiningCycle(activeReputationMiningCycle).userInvolvedInMiningCycle(msg.sender), "colony-token-locking-hash-submitted");
+    require(!IReputationMiningCycle(activeReputationMiningCycle).userInvolvedInMiningCycle(msg.sender), "colony-network-hash-submitted");
     ITokenLocking(tokenLocking).deobligateStake(msg.sender, _amount, clnyToken);
     miningStakes[msg.sender].amount = sub(miningStakes[msg.sender].amount, _amount);
   }
 
   function getMiningStake(address _user) public stoppable returns (MiningStake memory) {
     return miningStakes[_user];
+  }
+
+  function burnUnneededRewards(uint256 _amount) public stoppable {
+    ITokenLocking(tokenLocking).claim(address(this), true);
+    ITokenLocking(tokenLocking).burn(_amount);
   }
 
   uint256 constant UINT192_MAX = 2**192 - 1; // Used for updating the stake timestamp
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -231,8 +231,8 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     return miningStakes[_user];
   }
 
-  function burnUnneededRewards(uint256 _amount) public stoppable {
-    ITokenLocking(tokenLocking).claim(address(this), true);
+  function burnUnneededRewards(uint256 _amount) public stoppable onlyReputationMiningCycle() {
+    ITokenLocking(tokenLocking).claim(IMetaColony(metaColony).getToken(), true);
     ITokenLocking(tokenLocking).burn(_amount);
   }
 
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -183,18 +183,16 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     );
   }
 
-  function punishStakers(address[] memory _stakers, address _beneficiary, uint256 _amount) public stoppable onlyReputationMiningCycle {
+  function punishStakers(address[] memory _stakers, uint256 _amount) public stoppable onlyReputationMiningCycle {
     address clnyToken = IMetaColony(metaColony).getToken();
     uint256 lostStake;
     // Passing an array so that we don't incur the EtherRouter overhead for each staker if we looped over
     // it in ReputationMiningCycle.invalidateHash;
     for (uint256 i = 0; i < _stakers.length; i++) {
       lostStake = min(ITokenLocking(tokenLocking).getObligation(_stakers[i], clnyToken, address(this)), _amount);
-      ITokenLocking(tokenLocking).transferStake(_stakers[i], lostStake, clnyToken, _beneficiary);
-
+      ITokenLocking(tokenLocking).transferStake(_stakers[i], lostStake, clnyToken, address(this));
       // TODO: Lose rep?
-
-      emit ReputationMinerPenalised(_stakers[i], _beneficiary, lostStake);
+      emit ReputationMinerPenalised(_stakers[i], lostStake);
     }
   }
 
@@ -211,9 +209,9 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
 
   function unstakeForMining(uint256 _amount) public stoppable {
     address clnyToken = IMetaColony(metaColony).getToken();
-    // Prevent reputation miners from withdrawing stake during the mining process.
-    bytes32 submissionHash = IReputationMiningCycle(activeReputationMiningCycle).getReputationHashSubmission(msg.sender).proposedNewRootHash;
-    require(submissionHash == 0x0, "colony-network-hash-submitted");
+    // Prevent those involved in a mining cycle withdrawing stake during the mining process.
+    bool involved = IReputationMiningCycle(activeReputationMiningCycle).userInvolvedInMiningCycle(msg.sender);
+    require(!involved, "colony-token-locking-hash-submitted");
     ITokenLocking(tokenLocking).deobligateStake(msg.sender, _amount, clnyToken);
     miningStakes[msg.sender].amount = sub(miningStakes[msg.sender].amount, _amount);
   }
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -236,6 +236,4 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
 
     return add(mul(prevWeight, _prevTime), mul(currWeight, _currTime)) / add(prevWeight, currWeight);
   }
-
-
 }
--- a/contracts/reputationMiningCycle/ReputationMiningCycle.sol
+++ b/contracts/reputationMiningCycle/ReputationMiningCycle.sol
@@ -22,14 +22,10 @@ import "./../../lib/dappsys/math.sol";
 import "./../colonyNetwork/IColonyNetwork.sol";
 import "./../patriciaTree/PatriciaTreeProofs.sol";
 import "./../tokenLocking/ITokenLocking.sol";
-import "./ReputationMiningCycleStorage.sol";
 import "./ReputationMiningCycleCommon.sol";
 
 
 contract ReputationMiningCycle is ReputationMiningCycleCommon {
-  /// @notice Size of mining window in seconds
-  uint256 constant MINING_WINDOW_SIZE = 60 * 60 * 24; // 24 hours
-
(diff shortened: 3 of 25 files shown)


kronosapiens:
> The rewards given are locked until the end of the mining cycle because the stakes that are paying for them haven't been slashed at the time they're being paid out. This requires the changes to TokenLocking around who has been 'involved' in the mining cycle. There's also been a separatation of reward and punish functions in Token Locking as a result of these changes.

Does that really matter? If we _know_ that the stakes will be slashed, then why can't we pay out the tokens immediately, and then slash the tokens later? It's still zero-sum.

contracts/ReputationMiningCycleCommon.sol
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    uint256 reputationRootHashNNodes = IColonyNetwork(colonyNetworkAddress).getReputationRootHashNNodes();
+    uint jrhNnodes = reputationUpdateLog[nLogEntries-1].nUpdates +
+      reputationUpdateLog[nLogEntries-1].nPreviousUpdates + reputationRootHashNNodes + 1; // This is the number of nodes we expect in the justification tree
+
+    uint256 nByes = log2Ceiling(nUniqueSubmittedHashes); // We can have at most one bye per round, and this is the maximum number of rounds
+
+    // The maximum number of responses that need rewards
+    uint256 rewardDenominator = nByes + (nUniqueSubmittedHashes-1)*(2*(3 + log2Ceiling(jrhNnodes)) + 1);
```
kronosapiens: Spacing is inconsistent :(

contracts/ReputationMiningCycleCommon.sol
```diff
@@ -0,0 +1,105 @@
...
+        mstore(0x40, add(m, 0x100))
+        let magic := 0x818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff
+        let shift := 0x100000000000000000000000000000000000000000000000000000000000000
+        let a := div(mul(x, magic), shift)
+        y := div(mload(add(m,sub(255,a))), shift)
+        y := add(y, mul(256, gt(arg, 0x8000000000000000000000000000000000000000000000000000000000000000)))
+    }
+  }
+
+
```
kronosapiens: Too many newlines...

contracts/ReputationMiningCycleCommon.sol
```diff
@@ -0,0 +1,105 @@
...
+    respondedToChallenge[_responder] = true;
+    uint256 reward = disputeRewardIncrement();
+    ITokenLocking(tokenLockingAddress).reward(
+      _responder,
+      reward
+    );
+    rewardsPaidOut += reward;
+  }
+
+  function disputeRewardIncrement() internal view returns (uint256) {
```
kronosapiens: This name to me suggests state mutation (i.e. "increment the dispute reward"). Can we specify `calculateDisputeRewardIncrement`?

contracts/TokenLocking.sol
```diff
@@ -46,12 +46,12 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
     _;
   }
 
-  // Prevent reputation miners from withdrawing stake during the mining process.
-  modifier hashNotSubmitted(address _token) {
+  // Prevent those involved in a mining cycle withdrawing stake during the mining process.
+  modifier notInvolvedInMiningCycle(address _token) {
```
kronosapiens: Not critical now but worth wondering if this stake management functionality could be subsumed by the more general stake management grammar proposed in #757 ? Would be nice if so.

contracts/TokenLocking.sol
```diff
@@ -46,12 +46,12 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
...
     address clnyToken = IMetaColony(IColonyNetwork(colonyNetwork).getMetaColony()).getToken();
     if (_token == clnyToken) {
-      bytes32 submissionHash = IReputationMiningCycle(IColonyNetwork(colonyNetwork).getReputationMiningCycle(true)).getReputationHashSubmission(msg.sender).proposedNewRootHash;
-      require(submissionHash == 0x0, "colony-token-locking-hash-submitted");
+      bool involved = IReputationMiningCycle(IColonyNetwork(colonyNetwork).getReputationMiningCycle(true)).userInvolvedInMiningCycle(msg.sender);
+      require(involved != true, "colony-token-locking-hash-submitted");
```
kronosapiens: Ahem...

```
require(
!IReputationMiningCycle(IColonyNetwork(colonyNetwork).getReputationMiningCycle(true)).userInvolvedInMiningCycle(msg.sender),
"colony-token-involved-in-cycle"
)
```

contracts/TokenLocking.sol
```diff
@@ -134,21 +134,31 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
...
       // TODO: Lose rep?
 
-      emit ReputationMinerPenalised(_stakers[i], _beneficiary, lostStake);
+      emit ReputationMinerPenalised(_stakers[i], lostStake);
     }
   }
 
+  function reward(address _recipient, uint256 _amount) public onlyReputationMiningCycle {
+    address clnyToken = IMetaColony(IColonyNetwork(colonyNetwork).getMetaColony()).getToken();
+    // TODO: Gain rep?
```
kronosapiens: Meh not sure it's necessary -- rep comes from submitting the right hash, we don't need to give rep for defending, that can just be cash.

contracts/TokenLocking.sol
```diff
@@ -134,21 +134,31 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
...
       userLocks[clnyToken][_stakers[i]].balance = sub(userLocks[clnyToken][_stakers[i]].balance, lostStake);
-      userLocks[clnyToken][_beneficiary].balance = add(userLocks[clnyToken][_beneficiary].balance, lostStake);
       // TODO: Lose rep?
 
-      emit ReputationMinerPenalised(_stakers[i], _beneficiary, lostStake);
+      emit ReputationMinerPenalised(_stakers[i], lostStake);
     }
   }
 
+  function reward(address _recipient, uint256 _amount) public onlyReputationMiningCycle {
```
kronosapiens: Hmm... I wonder if there's a way to do this within the existing grammar -- i.e. give the reputation mining cycle CLNY at the start and deposit it in tokenlocking, and then use existing stake management interface to manage these distributions... although maybe let's look at this once we come back to #757.

scripts/generate-test-contracts.sh
```diff
@@ -35,7 +35,7 @@ sed -i.bak "s/ColonyDataTypes/UpdatedColonyDataTypes/g" ./contracts/IUpdatedColo
 sed -i.bak "s/contract IUpdatedColony is UpdatedColonyDataTypes, IRecovery {/contract IUpdatedColony is UpdatedColonyDataTypes, IRecovery {function isUpdated() public pure returns(bool);/g" ./contracts/IUpdatedColony.sol
 # Modify UpdatedReputationMiningCycle contract
 sed -i.bak "s/contract ReputationMiningCycle/contract UpdatedReputationMiningCycle/g" ./contracts/UpdatedReputationMiningCycle.sol
-sed -i.bak "s/WAD;/WAD;function isUpdated() public pure returns(bool) {return true;}/g" ./contracts/UpdatedReputationMiningCycle.sol
+sed -i.bak "s|// 24 hours|// 24 hours\nfunction isUpdated() public pure returns(bool) {return true;}|g" ./contracts/UpdatedReputationMiningCycle.sol
```
kronosapiens: Remind me what's happening here?

test/reputation-system/reputation-mining-client/client-auto-functionality.js
```diff
@@ -381,6 +382,10 @@ process.env.SOLIDITY_COVERAGE
           });
 
           await badClient.submitRootHash();
+
+          // Forward time again so clients can start responding to challenges
+          await forwardTime(MINING_CYCLE_DURATION * 0.5, this);
```
kronosapiens: Why not `/ 2` like everywhere else?

test/reputation-system/root-hash-submissions.js
```diff
@@ -573,21 +574,38 @@ contract("Reputation mining - root hash submissions", accounts => {
...
-      const invalidatorLock = await tokenLocking.getUserLock(clnyToken.address, accounts[0]);
-      expect(invalidatorLock.balance, "Account was not rewarded properly").to.eq.BN(new BN(invalidatorLockBefore.balance).add(MIN_STAKE.muln(2)));
+      const miner1NonceBefore = await web3.eth.getTransactionCount(MINER1, blockBeforeChallenge.number);
+      const miner1NonceAfter = await web3.eth.getTransactionCount(MINER1);
+
+      const miner2NonceBefore = await web3.eth.getTransactionCount(MINER2, blockBeforeChallenge.number);
+      const miner2NonceAfter = await web3.eth.getTransactionCount(MINER2);
+      const miner1SuccessfulDefences = miner1NonceAfter - miner1NonceBefore;
+      // The bad miner has one of these transactions that fails
+      const miner2SuccessfulDefences = miner2NonceAfter - miner2NonceBefore - 1;
```
kronosapiens: I'm sure the order of operations here is right but... a parentheses would help since subtraction is not associative...

test/reputation-system/root-hash-submissions.js
```diff
@@ -573,21 +574,38 @@ contract("Reputation mining - root hash submissions", accounts => {
...
       await accommodateChallengeAndInvalidateHash(colonyNetwork, this, goodClient, badClient, {
         client2: { respondToChallenge: "colony-reputation-mining-increased-reputation-value-incorrect" }
       });
 
-      const invalidatorLock = await tokenLocking.getUserLock(clnyToken.address, accounts[0]);
-      expect(invalidatorLock.balance, "Account was not rewarded properly").to.eq.BN(new BN(invalidatorLockBefore.balance).add(MIN_STAKE.muln(2)));
+      const miner1NonceBefore = await web3.eth.getTransactionCount(MINER1, blockBeforeChallenge.number);
+      const miner1NonceAfter = await web3.eth.getTransactionCount(MINER1);
+
+      const miner2NonceBefore = await web3.eth.getTransactionCount(MINER2, blockBeforeChallenge.number);
```
kronosapiens: Interesting, never knew about that function.

contracts/ReputationMiningCycle.sol
```diff
@@ -399,9 +409,15 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     bytes32[] memory siblings2
   ) public
   {
+    require(submissionWindowClosed(), "colony-reputation-mining-cycle-submissions-not-closed");
     require(index < disputeRounds[round].length, "colony-reputation-mining-index-beyond-round-length");
+
+    // We reward the submitter for this, so we better have an opponent so that someone gets slashed and can pay for it
+    uint256 opponentIdx = (index % 2 == 1 ? index-1 : index + 1);
```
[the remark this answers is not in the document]
kronosapiens: Also throwing this into a `getOpponentIndex` helper would be nice, I've seen this a few places.

contracts/ReputationMiningCycle.sol
```diff
@@ -274,7 +278,7 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
       require(disputeRounds[round][opponentIdx].challengeStepCompleted >= disputeRounds[round][idx].challengeStepCompleted, "colony-reputation-mining-less-challenge-rounds-completed");
 
       // Require that it has failed a challenge (i.e. failed to respond in time)
-      require(now - disputeRounds[round][idx].lastResponseTimestamp >= 600, "colony-reputation-mining-not-timed-out"); // Timeout is ten minutes here.
+      require(disputeRounds[round][idx].lastResponseTimestamp < now && sub(now, disputeRounds[round][idx].lastResponseTimestamp) >= 600, "colony-reputation-mining-not-timed-out"); // Timeout is ten minutes here.
```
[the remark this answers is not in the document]
area: `lastResponseTimestamp` is set to a value in the future when each submission is made - specifically, when the window is going to close. By doing that, and having these checks, we make sure we're not accidentally paying anyone out before we know how many submissions there have been (and therefore knowing how much the reward should be for each). I will admit that I have not properly considered whether this particular one can be hit - if that's your specifically the question here, let me know and I'll look properly.

contracts/ReputationMiningCycle.sol
```diff
@@ -297,27 +301,30 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
...
         // Our opponent completed the same number of challenge rounds, and both have now timed out.
         nInvalidatedHashes += 2;
+
         // Punish the people who proposed our opponent
         ITokenLocking(tokenLockingAddress).punishStakers(
           submittedHashes[opponentSubmission.proposedNewRootHash][opponentSubmission.nNodes][opponentSubmission.jrh],
-          msg.sender,
           MIN_STAKE
         );
+        stakeLost += submittedHashes[opponentSubmission.proposedNewRootHash][opponentSubmission.nNodes][opponentSubmission.jrh].length * MIN_STAKE;
```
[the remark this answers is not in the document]
area: Yeah, I think at twice I'm okay with it. If a third pops up, then we can talk :wink:

contracts/ReputationMiningCycle.sol
```diff
@@ -399,9 +409,15 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     bytes32[] memory siblings2
   ) public
   {
+    require(submissionWindowClosed(), "colony-reputation-mining-cycle-submissions-not-closed");
     require(index < disputeRounds[round].length, "colony-reputation-mining-index-beyond-round-length");
+
+    // We reward the submitter for this, so we better have an opponent so that someone gets slashed and can pay for it
+    uint256 opponentIdx = (index % 2 == 1 ? index-1 : index + 1);
+    require(opponentIdx < disputeRounds[round].length, "colony-reputation-mining-no-opponent-yet");
```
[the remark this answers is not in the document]
area: I thought the comment would help explain this, but I guess not. This is checking that we have an opponent that we can punish later . `disputeRounds[round]` is zero indexed, so if the `opponentIdx` calculated is not less than the `length`, it's out-of-range and so doesn't exist (yet). This will occur if this function is called for the last submission currently in `disputeRounds` if the index of that submission is even. Or another way of putting that, the last submission if the total number of submissions in that round is odd.

Now I've typed this out, I suspect this is a holdover from when I thought I could do each dispute separately, pairwise. This is probably not necessary now, because no-one is actually slashed in this call.

contracts/ReputationMiningCycleCommon.sol
area: `disputeRewardSize`?

contracts/TokenLocking.sol
area: I mean, maybe, but I think that's definitely out of the purview of this PR given it's not really related to rewarding users and that functionality isn't merged yet. If this PR added significant functionality that was affected by that, I'd be much keener but it'll require changes to the existing functionality so should probably be its own PR.

contracts/TokenLocking.sol
area: I mean, I'm being paid in the colony's native token for work I'm doing...

scripts/generate-test-contracts.sh
area: We're generating an 'upgraded' `reputationMiningCycle` to use in the upgrade tests. Because the constant that was at the start of this contract has been moved to `reputationMiningCycleCommon`, we need to change the regex used here.

test/reputation-system/reputation-mining-client/client-auto-functionality.js
area: T'was just how it already was in this test but I will change it.

contracts/TokenLocking.sol
kronosapiens: Fair fair

contracts/TokenLocking.sol
kronosapiens: I guess... I think the difference is that those tokens already paid out rep once (when they were paid out initially), so not sure if a slashed stake (which is what this is) should lead to rep being generated a second time.

contracts/TokenLocking.sol
kronosapiens: Discussed above.

contracts/ReputationMiningCycle.sol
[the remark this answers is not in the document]
kronosapiens: Hmm, ok that makes sense, I didn't realize it was a future cutoff, I thought it was the time of a prior submission. But re-working the arithmetic so that `now` is always on the same side of the inequality would be clarifying, I think. So instead of:

```
require(
  disputeRounds[round][idx].lastResponseTimestamp < now &&
  sub(now, disputeRounds[round][idx].lastResponseTimestamp) >= 600,
  "colony-reputation-mining-not-timed-out"
); // Timeout is ten minutes here.
```

we have the equivalent (since `now - ts > x <=> now > ts + x`):

```
require(
  disputeRounds[round][idx].lastResponseTimestamp < now &&
  add(disputeRounds[round][idx].lastResponseTimestamp, 600) <= now,
  "colony-reputation-mining-not-timed-out"
); // Timeout is ten minutes here.
```

So we're checking that the submission window has closed (`lastResponseTimestamp < now`) AND that the 10-minute timeout for response has elapsed (`add(lastResponseTimestamp, 600) <= now`). Looking at it this way, it does seem like the first check is redundant -- if the second passes, the first one will also.

contracts/ReputationMiningCycleCommon.sol
kronosapiens: I like a verb prefix like `calculate...` or `get...` since it implies that we're fetching some value for use elsewhere. `disputeRewardSize` is fine though.

scripts/generate-test-contracts.sh
kronosapiens: Mkay

contracts/ReputationMiningCycle.sol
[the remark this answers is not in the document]
area: To be clear, once a dispute has started, it does represent the last time that submission was defended as part of that process. I agree with the rest of this though :smile:

test/reputation-system/root-hash-submissions.js
area: Prettier doesn't like it I'm afraid! The calculation is right though.

area:
So with the addition of the staking functionality, the implementation here has been forced to change. We don't have a guarantee that stakes will be slashed before payouts are required, and because `TokenLocking` no longer knows about mining we can't payout anyway and just leave balances alone until we know who to slash (which had other issues, anyway) so rewards are tracked and only made available at the end of the mining cycle, and have to be explicitly claimed by miners. There is no time limit on being able to make this claim, and it is added to their (pending, if appropriate) balance. One claim covers all mining cycles the user has pending rewards for, so this can be a very occasional 'as needed' transaction.

I believe `reward` has to be left on `TokenLocking` because whenever we do the deployment, existing `reputationMiningCycles` will expect to be able to call it. I've changed it to a No-op to make this as obvious as possible, and it should be removed for the deployment after that.

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -183,21 +183,31 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
...
 
+  function reward(address _recipient, uint256 _amount) public stoppable onlyReputationMiningCycle {
+    // TODO: Gain rep?
+    pendingMiningRewards[_recipient] = add(pendingMiningRewards[_recipient], _amount);
+  }
+
+  function claimMiningReward(address _recipient) public stoppable {
+    address clnyToken = IMetaColony(metaColony).getToken();
+    uint256 amount = pendingMiningRewards[_recipient];
+    pendingMiningRewards[_recipient] = 0;
```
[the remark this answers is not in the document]
area: Here? Exactly the same. It gets a bit more complicated for things that aren't elementary types (arrays, structs), which the [docs](https://solidity.readthedocs.io/en/v0.6.8/types.html#delete) explain pretty well.

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -222,6 +231,11 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     return miningStakes[_user];
   }
 
+  function burnUnneededRewards(uint256 _amount) public stoppable {
```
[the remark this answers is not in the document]
area: Haha, yes, the annotation in the interface claims this can only be called by the active mining cycle, but that is demonstrably not true! :facepalm:

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -222,6 +231,11 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
...
+    ITokenLocking(tokenLocking).claim(address(this), true);
```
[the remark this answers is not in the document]
area: Yikes...

I guess this really illustrates that `claim` isn't needed in any of the tests. I'll engineer a situation where it is.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -0,0 +1,99 @@
...
+
+  function rewardResponder(address _responder) internal returns (bytes32) {
+    respondedToChallenge[_responder] = true;
+    uint256 reward = disputeRewardSize();
+    IColonyNetwork(colonyNetworkAddress).reward(_responder, reward);
+    rewardsPaidOut += reward;
+  }
+
+  function disputeRewardSize() internal view returns (uint256) {
+    // TODO: Is this worth calculating once, and then saving? Seems quite likely.
```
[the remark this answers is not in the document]
area: It couldn't be on initialisation, it'd have to be the first time this was called during a dispute resolution (because we don't know how big the reward is going to be until the window is closed, and we know how many submissions there are).

contracts/tokenLocking/ITokenLocking.sol
```diff
@@ -84,6 +84,14 @@ contract ITokenLocking is TokenLockingDataTypes {
   /// @param _force Pass true to forcibly unlock the token
   function withdraw(address _token, uint256 _amount, bool _force) public;
 
+  /// @notice This function is deprecated and only exists to aid upgrades.
+  function reward(address _recipient, uint256 _amount) public;
+
+  /// @notice Function called to burn CLNY tokens held by TokenLocking.
+  /// @dev While public, it can only be called successfully by colonyNetwork and is only used for reputation mining.
+  /// @param _amount Amount of CLNY to burn
+  function burn(uint256 _amount) public;
```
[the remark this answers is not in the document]
area: The problem is that this is actually burning tokens via `.burn`, which isn't ERC20, so I wouldn't expect most tokens to have it.

contracts/reputationMiningCycle/ReputationMiningCycle.sol
```diff
@@ -534,6 +548,24 @@ contract ReputationMiningCycle is ReputationMiningCycleStorage, PatriciaTreeProo
     return reputationMiningWindowOpenTimestamp;
   }
 
+  function getDisputeRewardSize() public view returns (uint256) {
+    return disputeRewardSize();
```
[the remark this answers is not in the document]
area: Okay, so the trouble here is in our EtherRouter setup scripts. By putting a public function in the `Common` file, it appears in the ABI of each contract, which the scripts balk at, claiming we've defined functions with the same name in multiple contracts (which, technically, we have). In theory, there's nothing wrong with that, but it's just a guard we've been keeping ourselves to to avoid confusions.

In theory, we could add an exception to any functions that appear in the Common ABI, but we'd have to change deployment scripts and so forth too, which use the same type of logic to check what's happening during deployment.

contracts/tokenLocking/ITokenLocking.sol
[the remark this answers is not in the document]
kronosapiens: Ok fair

contracts/reputationMiningCycle/ReputationMiningCycle.sol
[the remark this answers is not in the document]
kronosapiens: No it's probably best not to mess with those conventions. I do like the idea of having it as a storage variable which is set when the submission window is closed.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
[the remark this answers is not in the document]
kronosapiens: See comment above, I like the idea of calculating this once and saving it to a storage variable once the submission window is closed.

contracts/reputationMiningCycle/ReputationMiningCycleStorage.sol
```diff
@@ -63,5 +63,11 @@ contract ReputationMiningCycleStorage is ReputationMiningCycleDataTypes, DSAuth
...
   int256 constant MIN_INT128 = (2**127)*(-1);
 
-  uint256 firstIncompleteRound;
+  uint256 firstIncompleteRound; // Storage slot 15
+
+  // Tracks whether the address in question has helped during a challenge / response process
+  mapping (address => bool) respondedToChallenge;  // Storage slot 16
+
+  uint256 stakeLost;// Storage slot 17
+  uint256 rewardsPaidOut;// Storage slot 18
```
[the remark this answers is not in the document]
kronosapiens: Related to above comments

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -49,6 +55,13 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
       return 0;
     }
 
+    if (cachedDisputeRewardSize != 0 ) {
+      // Then we've already calculated it! Just return it
+      return cachedDisputeRewardSize;
```
kronosapiens, line 60: This isn't my favorite -- I'd rather have it calculated once at a known spot and then just return the storage variable, vs this memoization approach, but I can live.

area: If you can point out where that spot should be, I'd gladly reconsider it. But it can't be calculated until the submission window closes, which happens at a particular time, not via a transaction. The next transaction could then be a submission claiming a bye, or one of the submissions confirming their JRH. This is the best spot that captures all those possibilities, as far as I could see.

kronosapiens: That makes sense... if it's a time-based condition with multiple entry points then I don't see any obvious way to go about it.

Issue #811, opened 2020-02-17
PM (d)app support for new network releases


kronosapiens:
Currently, the Network and the (d)app are tightly coupled -- releasing a new version of the Network without having support in the (d)app will currently lead to problems. In order to de-couple these components, we need to do the following:

 - [x] 1. Make it possible to create new colonies at any prior version,

 - [x] 2. Have the (d)app disable upgradability until it has added support for `version >= 4`

 - [x] 3. (Ideally) have ColonyJS also know what version it can support.

(1), (2), and (3) have been completed: (#784), (https://github.com/JoinColony/colonyDapp/pull/2106), and (https://github.com/JoinColony/colonyJS/pull/458).

kronosapiens:
Release plan:

1. (d)app updates to remove the "upgrade button"
2. Network deploys burgundy-glider to Goerli
3. (d)app tests to make sure (d)app doesn't break in the presence of burgundy-glider
4. Test burgundy-glider using truffle console
5. Once satisfied, release to mainnet

...

6. When the time is right, update ColonyJS & (d)app to use the new burgundy-glider functionality

kronosapiens:
We live!

PR #818, opened 2020-03-09
Pin deprecated createColony version to 3

kronosapiens:
Part of #811, pins the version used for new colonies to 3 when created using the now-deprecated `createColony`.

This makes it possible for ColonyJS + (d)app to work with the upgraded network without needing to be upgraded themselves.


area:
The comments in `IColonyNetwork` need to be changed as well to say that the one-parameter `createColony` is deprecated, but otherwise this seems like it does what we need.

kronosapiens:
@area already done?

https://github.com/JoinColony/colonyNetwork/blob/maint/fixed-version/contracts/IColonyNetwork.sol#L130

PR #825, opened 2020-03-30
Make sure we test deprecated createColony

area:


area:
Can we do something about the docs switching the order of stuff apparently at random? Or have I done something here to cause that somehow?

kronosapiens:
@area yeah I'm not sure what's causing that -- could be the doc generator changing behavior, could be that I re-organized the functions at some point and forgot about it... I don't know if it's worth worrying about though really.

Issue #72, opened 2017-09-28
Completing a BFP


elenadimitrova:
### 8.2.10 Completing a BFP
If an update finds that a proposal is fully funded (i.e. `TotalPaid` = `TotalRequested`), it is removed from this queue to allow the next-most-popular funding proposal to accrue funds. Explicitly, the following steps need to happen:
1. The time at which the funding proposal was fully funded is calculated.
2. `TotalPaid` is set to `TotalRequested`.
3. The BFP is removed from the queue.
4. The next BFP in the queue is promoted to the top of the queue, and its `LastUpdated` time is set as the time calculated in 1.
Once the the BFP has been fully funded, for a period of three days anyone can make a proposal that `PunishCreator` should be set to ‘true’, and the creator lose their stake. This proposal takes the form of an Objection (Section 9). If no such proposal is made, or the proposal fails to pass, then the creator can reclaim their stake. Once the fate of the stake is decided, and the creator either reclaims the stake or loses it and a matching amount of reputation, the BFP is set to the `completed` state.

Issue #71, opened 2017-09-28
The Rate of funding for BFPs


elenadimitrova:
### 8.2.9 The Rate of funding for BFPs
The more reputation backs a proposal, the faster it is funded. The rate scales linearly, and at the limit, if 100% of the reputation in the source domain backs a basic funding proposal, then that funding proposal will be funded at a rate of 50% of the domain’s holdings (of the `TokenType`) per week. The goal is a steady and predictable allocation of resources directed collectively by the domain’s (reputation weighted) priorities.
When a user backs a proposal, both the user and their reputation at the time are recorded.
Consequently the user is able to update their backing at a later date. However, we note that such an update is not automatic and even if a user loses reputation due to bad behaviour, their backing level remains unchanged. To rectify this, we will allow users to update another user’s backing to reflect their updated reputation scores, but we don’t expect this functionality to be used often.
We would only anticipate it being used if a user lost a lot of reputation due to some very bad behaviour, and other users wanted to prevent a bad funding proposal backed by the same user from being completed before it could be cancelled by other means (i.e. via dispute, described in Section
9).
We emphasised that a user could back a proposal with their reputation at the time of backing because the reputation backing a proposal will not change when that user’s reputation does so. If by a quirk in this system, the reputation recorded as backing a funding proposal ends up higher than 100% of the total of that reputation in the colony, then the funding occurs no quicker than it would at 100%.

Issue #70, opened 2017-09-28
Ordering of BFPs


elenadimitrova:
### 8.2.8 Ordering of BFPs
Basic funding proposals are ordered in the _Funding Queue_. Only one of them can receive funds at any one time. The proposals are ordered by ‘amount of reputation backing the proposal’. 
When created, a basic funding proposal gets placed at the bottom of the queue. Users can give a proposal ‘backing’ weighted by their reputation in the source domain* at the time of backing**. There are no costs to backing a proposal (other than gas costs) and the users obtain no direct benefits; it does not represent them putting their earned reputation at risk, nor any tokens — it merely helps the proposal achieve funding in a more timely fashion.
The more reputation backs a proposal, the higher up the queue it is placed. Every transaction that adds backing to a proposal (or otherwise updates the backing level) inserts the proposal in the correct place in the queue. Only the funding proposal at the top of the queue accrues funds.

* The source domain of a BFP is the domain of the pot that the funding proposal is `From`.
** A user’s reputation may change, but the backing weight is recorded at the time of backing and does not change without further user action.

Issue #69, opened 2017-09-28
Submitting a Basic Funding Proposal


elenadimitrova:
### 8.2.7 Basic Funding Proposals
A basic funding proposal (BFP) is a funding proposal from some domain’s pot to one of its children’s. It starts out in the active state and is thus immediately eligible for funding. It may be cancelled at any time by the `Creator`, or through the dispute mechanism. In either case, the stake is only able to be reclaimed if `PunishCreator` has not been set to true via a dispute by the end of a timeout period.

PR #828, opened 2020-04-03
Funding Queues

### LINKED ISSUE #807: Make a funding queue extension contract
This should be an extension contract that implements basic funding proposals and the queue thereof as described in the [whitepaper §8.2](https://colony.io/whitepaper.pdf), and functions when it has been given the funding permission in a domain. Priority funding proposals do not need to be included.

kronosapiens:
Closes #807, #60, #62, #63, #64, #65, #66, #67, #68, #69, #70, #71, #72 

## Features

- [x] Creating & staking basic proposals
- [x] Backing them with rep (and moving them into the right location in the queue)
- [x] Pinging them for funding transfers
- [x] Completing & cancelling proposals
- [ ] Creating priority proposals (maybe later?)
- [ ] Updating backing as rep changes (maybe later?)

### DIFF
--- a/contracts/extensions/FundingQueue.sol
+++ b/contracts/extensions/FundingQueue.sol
@@ -30,8 +30,26 @@ contract FundingQueue is DSMath, PatriciaTreeProofs {
 
   // Constants
   uint256 constant HEAD = 0;
-  uint256 constant UINT256_MAX = 2**256 - 1;
-  uint256 constant FUNDING_MULTIPLE = WAD / 2;
+  uint256 constant UINT256_MAX = (2 ** 256) - 1;
+  uint256 constant COMPOUNDING_FREQUENCY = 168;
+  uint256 constant BASE_DECAY_RATE = 691719225406032925; // TODO: use RAYs?
+  // From Wolfram Alpha:           0.691719225406032924664669179186644...
+
+  //  BASE_DECAY_RATE is determined via the following equation:
+  //   r = (z ** (1/n) - 1) * n
+  //
+  //  With the following variables:
+  //   z = target decay rate (currently: -0.5 per week, see whitepaper)
+  //   n = COMPOUNDING_FREQUENCY (currently: 168, i.e. hours per week)
+  //   r = -BASE_DECAY_RATE
+  //
+  //  Basically, we take a target decay rate (-0.5 per week), and a compounding
+  //   frequency (1 hour), and figure out what "interest rate" is needed to
+  //   achieve that (about -0.69). This is necessary because more frequent
+  //   compounding requires a higher decay rate to keep weekly decay constant.
+  //  This equation was derived by taking the regular compound interest
+  //   formula, z = (1 + r/n) ** nt, setting t = 1, and solving for r = f(z,n).
+  //  To keep everything uint, we store the negative and use `sub()` later.
 
   // Initialization data
   IColony colony;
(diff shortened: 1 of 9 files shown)


contracts/extensions/FundingQueue.sol
```diff
@@ -0,0 +1,349 @@
...
+  event ProposalStaked(uint256 indexed id, uint256 domainTotalRep);
+  event ProposalBacked(uint256 indexed id, uint256 indexed newPrevId, uint256 userRep);
+  event ProposalPinged(uint256 indexed id, uint256 amount);
+  event ProposalCompleted(uint256 indexed id);
+  event ProposalCancelled(uint256 indexed id);
+
+  // Constants
+  uint256 constant HEAD = 0;
+  uint256 constant UINT256_MAX = (2 ** 256) - 1;
+  uint256 constant STAKE_PCT = WAD / 1000; // 0.1%
```
area, line 42: Contender for worst-named variable I've seen this year, I have to say. It's called 'Stake Percent', the comment suggests it represents 0.1%, and it holds a value of `1000000000000000` :P

contracts/extensions/FundingQueue.sol
```diff
@@ -0,0 +1,391 @@
...
+    uint256 nextId = queue[_newPrevId];
+    if (queue[HEAD] == nextId && nextId != 0) {
+      pingProposal(nextId);
+    }
+    // Does this proposal have less than or equal support to the previous proposal?
+    require(
+      proposals[_newPrevId].totalSupport >= proposal.totalSupport,
+      "funding-queue-excess-support"
+    );
+    // Does this proposal have more than or equal support to the next proposal?
```
area, line 223: Only more than, not equal. The code is right, and having the inequalities this way around between the two requires is good as it means that proposals with the same backing can't jostle for position (effectively, the first proposal to reach a particular amount of backing has priority).

contracts/extensions/FundingQueue.sol
```diff
@@ -0,0 +1,391 @@
...
+    require(
+      nextId == HEAD || proposals[nextId].totalSupport < proposal.totalSupport,
```
area, line 225: This `nextId==HEAD` really bugs me, which after staring at it for a while I think it's because this isn't a queue, it's a circular linked list. Can we characterise it more honestly in the code as such, or at least own up to that fact in the comments, and very explicitly call out this special case? Alternatively, we could make it more queue like with a `TAIL` that signifies the end of the queue.

contracts/extensions/FundingQueue.sol
```diff
@@ -0,0 +1,391 @@
...
+    assert(proposal.totalPaid <= proposal.totalRequested);
+
+    if (proposal.totalPaid == proposal.totalRequested) {
+      proposal.state = ProposalState.Completed;
+
+      queue[HEAD] = queue[_id];
+      delete queue[_id];
+
+      // May be the null proposal, but that's ok
+      proposals[queue[HEAD]].lastUpdated = now;
```
area, line 259: Should this be set to sometime in the past, when funding would have finished for the current proposal if everything was continuously updated?

kronosapiens: I'm not sure if we can figure that out exactly on-chain but I can do something approximate.

contracts/extensions/FundingQueue.sol
kronosapiens: Mm that's fair.

contracts/extensions/FundingQueue.sol
kronosapiens: Right right, I went back and forth on this a bit yesterday and decided the way I had it was best. I changed the comments and seems I forgot to change them back.

contracts/extensions/FundingQueue.sol
kronosapiens: Ok my vote is to infer the timestamp based on the amount of funding to be transferred and the amount remaining -- i.e. if we are set to transfer 100 tokens but the amount requested is only 50, then we set the `lastUpdated` to the 50% mark between the current proposal's `lastUpdated` and `now`. This will put the timestamp slightly later than if we were to figure out the exact timestamp that would have satisfied the first proposal, but I think that's fine since it's not exploitable and creates an incentive for people to ping the leading proposal to put their's at the head, while also limiting the downside if they forget.

area: Yeah, that sounds like a sensible middle ground. I agree it's under-selling what the timestamp should be, but it's the 'right side' of wrong as far as I'm concerned and if people care enough, they should be paying attention as you say.

PR #830, opened 2020-04-08
Fix blockchain start script error on process termination

igorline:
I was always getting an error like this when running tests.

```
npm ERR! code ELIFECYCLE
npm ERR! errno 143
npm ERR! @colony.io/colony-network-contracts@1.0.0 start:blockchain:client: `bash ./scripts/start-blockchain-client.sh`
npm ERR! Exit status 143
```

These changes fixes it.

### DIFF
--- a/scripts/stop-blockchain-client.sh
+++ b/scripts/stop-blockchain-client.sh
@@ -3,4 +3,4 @@
 pid=$(lsof -i:8545 -sTCP:LISTEN -t); 
 
 echo "Killing blockchain client process $pid on port 8545"
-kill -TERM $pid || kill -KILL $pid
+kill -INT $pid


area:
Thanks for taking the time to open this PR. I agree that getting rid of that error message is a good idea, but I don't think this is the way to go about it - making sure scripts actually throw errors if things appear to not work is always a good idea, and makes debugging things easier.

Here, I think the appropriate fix is in `stop-blockchain-client`, which should send `SIGINT` and not `SIGTERM` to the process in order to bring it down cleanly. While we're in there, I'm pretty sure the alternative `-KILL` is useless. I think the intention there was that if bringing the process down cleanly fails, then `-KILL` it. But `kill` doesn't wait to see if the process comes down, it just sends the signal, which will always succeed (unless the PID is wrong, in which case the `-KILL` command will also fail...). So I think we were wrong there, and could remove the `KILL` command.

igorline:
Thanks for the feedback @area! I'll take a look at stop script

igorline:
Hey @area! Thanks for suggestion! 
SIGINT expectedly is terminating node started ganache process correctly!

PR #835, opened 2020-04-22
Coin Machine

kronosapiens:
Coin Machine is an experimental token-distributing mechanism which came out of a #research brainstorm. 

Essentially, it takes a duration of time, a number of periods, and a target token sale volume, and executes a series of token sales, with the price adjusting as a function of prior sale demand. 

Advantages of this mechanism are: stability of price during the sale period; a straightforward user experience (tokens are either available or "sold out"); the ability to adjust price to the market, albeit with a time delay.

Disadvantages of this mechanism are: periods of potential token unavailability; prices which may substantially lag the market.

Ultimately, we feel that as the originator of these tokens, we can accept the arbitrage costs introduced by price lags, in return for the increase in demand created by the under-pricing. The goal is to put tokens into circulation, not to capture 100% of their value in the sale. We hope that the simple semantics of the mechanism make it appealing to users.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -196,6 +196,13 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     ERC20Extended(token).mint(address(this), _wad); // ignore-swc-107
   }
 
+  function mintTokensFor(address _guy, uint _wad) public
+  stoppable
+  auth
+  {
+    ERC20Extended(token).mint(_guy, _wad); // ignore-swc-107
+  }
+
   function mintTokensForColonyNetwork(uint _wad) public stoppable {
     // Only the colony Network can call this function
     require(msg.sender == colonyNetworkAddress, "colony-access-denied-only-network-allowed");
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -38,7 +38,7 @@ contract CoinMachine is DSMath {
   address purchaseToken; // The token in which we receive payments, 0x0 for eth
 
   uint256 periodLength; // Duration of a sale period in seconds (e.g. 3600 = 1 hour)
-  uint256 windowSize; // Number of periods in the price averaging window in integers (e.g. 10)
+  uint256 alpha; // WAD-denominated float between 0 and 1 that controls how quickly the EMA adjusts.
 
   uint256 targetPerPeriod; // Target number of tokens to sell in a period
   uint256 maxPerPeriod; // Maximum number of tokens sellable in a period
@@ -49,7 +49,7 @@ contract CoinMachine is DSMath {
   uint256 tokensSold; // Tokens sold in the active period
   uint256 activeIntake; // Payment received in the active period
 
-  mapping (uint256 => uint256) pastIntakes; // Payment received in past periods
+  uint256 emaIntake;
 
   // Public
 
@@ -75,7 +75,9 @@ contract CoinMachine is DSMath {
     purchaseToken = _purchaseToken;
 
     periodLength = _periodLength;
-    windowSize = _windowSize;
+    alpha = wdiv(2, _windowSize + 1); // Two ints enter, 1 WAD leaves
+    // In the long-term, 86% of the weighting will be in this window size. This is a 'standard' conversion
+    // used between SMAs and EMAs, as the center of mass of the weightings is the same if this is used.
 
     targetPerPeriod = _targetPerPeriod;
     maxPerPeriod = _maxPerPeriod;
@@ -83,10 +85,7 @@ contract CoinMachine is DSMath {
     activePrice = _startingPrice;
     activePeriod = getCurrentPeriod();
 
-    uint256 startingIntake = wmul(targetPerPeriod, _startingPrice);
-    for (uint256 i; i < _windowSize; i++) {
-      pastIntakes[i] = startingIntake;
-    }
+    emaIntake = wmul(targetPerPeriod, _startingPrice);
   }
 
   function buyTokens(uint256 _numTokens) public payable {
@@ -116,47 +115,29 @@ contract CoinMachine is DSMath {
   function updatePeriod() public {
     uint256 currentPeriod = getCurrentPeriod();
 
-    // If we've missed an entire window, skip ahead
-    uint256 periodGap = currentPeriod - activePeriod;
-    activePeriod = currentPeriod - min(periodGap, windowSize);
-
-    bool newPeriod;
-
-    // Handle the recently elapsed period
+    // We need to update the price if the active period is not the current one.
     if (activePeriod < currentPeriod) {
-      newPeriod = true;
 
-      pastIntakes[toBin(activePeriod)] = activeIntake;
+      emaIntake = wmul((WAD - alpha), emaIntake) + wmul(alpha, activeIntake); // wmul(wad, int) => int
       activeIntake = 0;
 
       tokensSold = 0;
-      activePeriod += 1;
-    }
 
-    // Handle any additional missed periods
-    while (activePeriod < currentPeriod) {
-      pastIntakes[toBin(activePeriod)] = 0;
-      activePeriod += 1;
-    }
-
-    // Update the price
-    if (newPeriod) {
-      uint256 sum;
-
-      for (uint256 i; i < windowSize; i++) {
-        sum += pastIntakes[i];
+      // Handle any additional missed periods
+      uint256 periodGap = currentPeriod - activePeriod - 1;
+      if (periodGap != 0) {
+        emaIntake = wmul(wpow((WAD - alpha), periodGap), emaIntake);
       }
 
-      activePrice = wdiv(sum / windowSize, targetPerPeriod);
+      activePeriod = currentPeriod;
+
+      // Update the price
+      activePrice = wdiv(emaIntake, targetPerPeriod);
     }
   }
 
   function getPeriodLength() public view returns (uint256) {
     return periodLength;
-  }
-
-  function getWindowSize() public view returns (uint256) {
-    return windowSize;
   }
 
   function getTargetPerPeriod() public view returns (uint256) {
@@ -175,19 +156,16 @@ contract CoinMachine is DSMath {
 
     // Otherwise, infer the new price
     } else {
-      uint256 sum;
-      uint256 period = currentPeriod - windowSize;
+      uint256 newIntake = emaIntake;
 
-      for (; period <= currentPeriod; period++) {
-        if (period < activePeriod) {
-          sum += pastIntakes[toBin(period)];
-        } else if (period == activePeriod) {
-          sum += activeIntake;
-        }
-        // No `else`, any intakes between activePeriod and currentPeriod are 0
-      }
+      // Accommodate the activePeriod
+      newIntake = wmul((WAD - alpha), newIntake) + wmul(alpha, activeIntake);
 
-      return wdiv(sum / windowSize, targetPerPeriod);
+      // Accommodate periods between the activePeriod and the currentPeriod
+      uint256 periodGap = currentPeriod - activePeriod - 1;
+      emaIntake = wmul(wpow((WAD - alpha), periodGap), emaIntake);
+
+      return wdiv(newIntake, targetPerPeriod);
     }
   }
 
@@ -202,7 +180,8 @@ contract CoinMachine is DSMath {
     return now / periodLength;
   }
 
-  function toBin(uint256 period) internal view returns(uint256) {
-    return period % windowSize;
+  function wpow(uint256 x, uint256 n) internal pure returns (uint256) {
+    // Must convert WAD (10 ** 18) to RAY (10 ** 27) and back
+    return rpow(x * (10 ** 9), n) / (10 ** 9);
   }
 }
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -34,7 +34,7 @@ contract CoinMachine is DSMath {
   }
 
   // Storage
-  ERC20Extended purchaseToken; // The token in which we receive payments
+  address purchaseToken; // The token in which we receive payments, 0x0 for eth
 
   uint256 periodLength; // Duration of a sale period (i.e. one hour)
   uint256 windowSize; // Number of periods in the price averaging window (i.e. 10)
@@ -63,9 +63,9 @@ contract CoinMachine is DSMath {
   )
     public
   {
-    require(address(purchaseToken) == address(0x0), "coin-machine-already-initialised");
+    require(activePeriod == 0, "coin-machine-already-initialised");
 
-    purchaseToken = ERC20Extended(_purchaseToken);
+    purchaseToken = _purchaseToken;
 
     periodLength = _periodLength;
     windowSize = _windowSize;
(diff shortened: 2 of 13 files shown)


contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,260 @@
...
+
+    uint256 numTokens = min(_numTokens, getNumAvailable());
+    uint256 totalPrice = wmul(numTokens, currPrice);
+
+    currIntake += totalPrice;
+    tokensSold += numTokens;
+    assert(tokensSold <= maxPerPeriod);
+
+    require(
+      purchaseToken.transferFrom(msg.sender, address(this), totalPrice),
```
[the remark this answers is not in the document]
kronosapiens: Nope, not deliberate. Can add.

kronosapiens:
Regarding your initial comment, the reason I liked the sort of "memoryless" circular-buffer approach is that it requires less storage and bookeeping -- rather than have a "deployed" timestamp, which would have periods ending at random times in perpetuity, using the system clock meant that periods would end, for example, on the hour, which seems more sensible. There doesn't seem to be a real need to keep track of the total number of periods, since it doesn't provide any value once you go past the number of periods we need to price calculation. The only real advantage I could see is that it allows for more precise behavior in the edge cases of a) first initialisation, where we haven't yet seen enough periods to create a proper average, and b) long inactivity, in which we might cycle entirely through the circular buffer and get weird prices. In my mind those scenarios aren't much to worry about, but I'll look at it some more. I definitely like the idea of periods ending at regular intervals independent of deployment time, which means using the system clock rather than a "deployed" timestamp.

area:
Frankly, from my perspective, not ending on the hour and one extra storage slot used seems like a very worthwhile price to pay for much simpler code. It's not like this implementation is totally regular either - by forcing everything to end on the hour (or whatever), the first period is a length based on when the transaction is mined.

kronosapiens:
Regarding your point about alternative implementations, I don't see how an ongoing cumulative total gets you the same behavior as a fixed averaging window. With the cumulative total, the price will become less and less flexible as the cumulative total gets larger, whereas with a fixed averaging window, the price remains sensitive in perpetuity.

I do think an approach in which the period is a treated as an ongoing counter, rather than a modulus, but then goes through a modulus to look up historical prices, could be a good middle approach, as it lets us handle situations of initialization and long periods of inactivity with more control.

area:
`numPeriods` is the number of periods that are used to work out the price, not the number of periods a coin machine runs. I can now at least understand why this approach was taken given that specification!

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    // Update the price
+    uint256 sum;
+
+    for (uint256 i; i < windowSize; i++) {
+      sum += pastIntakes[i];
+    }
+
+    activePrice = wdiv(sum / windowSize, targetPerPeriod);
```
area, line 135: Maybe this more concretely expresses my distaste for `wdiv` / `wmul`. Here, we're using `wdiv`, passing in two regular `int`s, and getting a `wad` out - without comment! In fact, I do not believe there is a single use in this contract of either function where two `wad`s are passed in (which is how they are intended to be used).

I'm not suggesting a change here necessarily, just trying to explain further my position :smile:

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,197 @@
...
+
+  constructor(address _colony) public {
+    colony = IColony(_colony);
+  }
+
+  // Storage
+  address purchaseToken; // The token in which we receive payments, 0x0 for eth
+
+  uint256 periodLength; // Duration of a sale period (i.e. one hour)
+  uint256 windowSize; // Number of periods in the price averaging window (i.e. 10)
```
area, line 40: e.g., not i.e. for these comments.

contracts/colony/Colony.sol
```diff
@@ -196,6 +196,13 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     ERC20Extended(token).mint(address(this), _wad); // ignore-swc-107
   }
 
+  function mintTokensFor(address _guy, uint _wad) public
```
area, line 199: Is there not a worry this could be used to pay out tokens without paying the network fee?

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,197 @@
...
+
+  uint256 targetPerPeriod; // Target number of tokens to sell in a period
+  uint256 maxPerPeriod; // Maximum number of tokens sellable in a period
+  int256 tokenSurplus; // Total tokens undersold (i.e. below target), negative means deficit
```
area, line 44: Why are we spending gas to track this on-chain? Perhaps I'm being blind, but I can't see how it's used.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,197 @@
...
+    assert(tokensSold <= maxPerPeriod);
+
+    if (purchaseToken == address(0x0)) {
+      require(msg.value >= totalPrice, "coin-machine-insufficient-funds");
+      msg.sender.transfer(msg.value - totalPrice);
+    } else {
+      ERC20Extended(purchaseToken).transferFrom(msg.sender, address(this), totalPrice);
+    }
+
+    colony.mintTokensFor(msg.sender, numTokens);
```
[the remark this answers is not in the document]
kronosapiens: I'd counter by asking the benefit of a coin selling mechanism for BYOT tokens which are ostensibly already available on secondary markets.

We _could_ include some "deprecate" function which stops the process, but I'm not sure it really adds anything above and beyond just undeploying the extension, which holds on to no funds and tracks no data apart from its own price.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,197 @@
...
+  function buyTokens(uint256 _numTokens) public payable {
+    updatePeriod();
+
+    uint256 numTokens = min(_numTokens, maxPerPeriod - tokensSold);
+    uint256 totalPrice = wmul(numTokens, activePrice);
+
+    activeIntake += totalPrice;
+    tokensSold += numTokens;
+
+    assert(tokensSold <= maxPerPeriod);
```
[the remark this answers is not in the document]
kronosapiens: Can it? How?

contracts/extensions/CoinMachine.sol
kronosapiens: Unfortunately you are mistaken regarding what units are WADs, but that's my fault for not providing sufficient documentation.

contracts/extensions/CoinMachine.sol
kronosapiens: Lol nice catch.

contracts/colony/Colony.sol
kronosapiens: Not particularly -- when creating a colony normally, the founder is the token owner until they manually transfer it to the colony. So if the founder can mint tokens for free, it's not clear why they would give the token ownership to the colony if that ability would be permanently lost.

contracts/extensions/CoinMachine.sol
kronosapiens: It's not, it's just a convenience. I suppose we could remove it.

contracts/extensions/CoinMachine.sol
kronosapiens: Although I suppose I could re-organize the math around Wei instead of Ether as units and perhaps then the wad math can be removed.

contracts/extensions/CoinMachine.sol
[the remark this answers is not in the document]
area: Gagh, caught out by `_` again.

contracts/extensions/CoinMachine.sol
kronosapiens: Although if the num tokens is Wei and the price is Wei, then that means the price can never drop below 1:1 (i.e. 1 Wei of price for every 1 Wei of token). So I think keeping it 18-decimal denominated is good, since then the price is more flexible (i.e. 1 WAD of token could cost as little as 1 Wei, and vice versa)

contracts/extensions/CoinMachine.sol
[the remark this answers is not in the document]
area: I think that's a pretty large assumption. If we're going to go down that road ('this functionality doesn't work with BYOT'), I think it needs to be after a conversation with more people involved. 

I'd prefer extensions to be able to properly delete themselves. I think it's likely we'll add something like that in to #714, so doesn't really need to be here.

contracts/extensions/CoinMachine.sol
area: I think there probably needs to be some, as regardless of your base unit, you need `0<price<1` being possible, I assume.

contracts/extensions/CoinMachine.sol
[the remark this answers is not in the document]
area: After talking with Jack, this is fine for now at least. We do at least need a way to establish whether a colony (token) is compatible with CoinMachine for the frontend though.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+  uint256 activePrice; // The active sale price (a WAD-denominated ratio)
+
+  uint256 tokensSold; // Tokens sold in the active period
+  uint256 activeIntake; // Payment received in the active period
+
+  mapping (uint256 => uint256) pastIntakes; // Payment received in past periods
+
+  // Public
+
+  function initialise(
```
area, line 55: Does this function need to be permissioned somehow? As-is, someone could watch for new CoinMachines being deployed and initialise them with silly numbers.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    targetPerPeriod = _targetPerPeriod;
+    maxPerPeriod = _maxPerPeriod;
+
+    activePrice = _startingPrice;
+    activePeriod = getCurrentPeriod();
+
+    uint256 startingIntake = wmul(targetPerPeriod, _startingPrice);
+    for (uint256 i; i < _windowSize; i++) {
+      pastIntakes[i] = startingIntake;
```
area, line 86: I'm still not a huge fan of this. But on the plus side, it does mean that the gas bottleneck is here, and not anywhere else, which is a good situation to be in (if it could be initialised to introduce a bottleneck later on in the flow, it could get stuck).

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+    colony.mintTokensFor(msg.sender, numTokens);
+  }
+
+  // Make sure this is called at least once during the averaging period
+  function updatePeriod() public {
+    uint256 currentPeriod = getCurrentPeriod();
+
+    // If we've missed an entire window, skip ahead
+    uint256 periodGap = currentPeriod - activePeriod;
+    if (periodGap % windowSize < periodGap) {
```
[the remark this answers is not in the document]
kronosapiens: No clue! Made sense at the time.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+      activePeriod = currentPeriod - (periodGap % windowSize + windowSize);
+    }
+
+    // Handle the recently elapsed period
+    if (activePeriod < currentPeriod) {
+      pastIntakes[toBin(activePeriod)] = activeIntake;
```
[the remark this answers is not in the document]
kronosapiens: Not necessarily -- the skipping just moves us through whole increments of the averaging window. We end up at the same number of periods behind, simply modulus the window. We treat the whole windows as a "loss" but still want to update any missed periods (i.e. if the window is 10 and we missed 12, we throw out the first 10 but still process the remaining 2).

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  // Make sure this is called at least once during the averaging period
+  function updatePeriod() public {
+    uint256 currentPeriod = getCurrentPeriod();
+
+    // If we've missed an entire window, skip ahead
+    uint256 periodGap = currentPeriod - activePeriod;
+    if (periodGap % windowSize < periodGap) {
+      activePeriod = currentPeriod - (periodGap % windowSize + windowSize);
```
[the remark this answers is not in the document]
kronosapiens: Yes, this was on purpose -- if we've missed more than a window, then we should zero out the entire window. If the max is `windowSize-1`, then we're not going to clear out at least one entry.

I.e. if we have [10, 30, 20] for the past intakes, and we miss 5 periods, we should zero out all three entries, not simply two (which would occur with 5 % 3 logic).

contracts/extensions/CoinMachine.sol
kronosapiens: Good point, let's do that.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+  IColony colony;
+
+  constructor(address _colony) public {
+    colony = IColony(_colony);
+  }
+
+  // Storage
+  address purchaseToken; // The token in which we receive payments, 0x0 for eth
+
+  uint256 periodLength; // Duration of a sale period in seconds (e.g. one hour)
```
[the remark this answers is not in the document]
kronosapiens: Sure!

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+    uint256 _maxPerPeriod,
+    uint256 _startingPrice
+  )
+    public
+  {
+    require(activePeriod == 0, "coin-machine-already-initialised");
+
+    require(_periodLength > 0, "coin-machine-period-too-small");
+    require(_windowSize > 0, "coin-machine-window-too-small");
+    require(_windowSize <= 256, "coin-machine-window-too-large");
```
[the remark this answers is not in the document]
kronosapiens: It was due to the way we handle the catch-up -- I was getting out-of-gas errors when we had too many periods (512-1024). This seemed like a conservative number that still left a lot of room. I agree that 255 is a better number.

contracts/extensions/CoinMachine.sol
```diff
@@ -0,0 +1,202 @@
...
+      activePeriod += 1;
+    }
+
+    // Handle any additional missed periods
+    while (activePeriod < currentPeriod) {
+      pastIntakes[toBin(activePeriod)] = 0;
+      activePeriod += 1;
+    }
+
+    // Update the price
```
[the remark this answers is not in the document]
kronosapiens: Correct, I left it this way to avoid needing a conditional (seemed cleaner) but I guess it doesn't really make sense from a gas perspective.

contracts/extensions/CoinMachine.sol
kronosapiens: Yup I agree, I think one-off setup costs aren't the end of the world.

kronosapiens:
A simple moving average was chosen because of its simplicity -- easy to understand, easy to reason about, etc. I don't see a compelling reason to use a more complex averaging system, so I'd advocate a "wait and see" approach.

test/extensions/coin-machine.js
```diff
@@ -186,24 +188,55 @@ contract("Coin Machine", (accounts) => {
...
 
-      // Expect price to decrease by 10% to baseline since we sold 0% in 1/10 periods
+      emaIntake = emaIntake.mul(WAD.sub(alphaAsWad)).divRound(WAD);
+      expectedPrice = emaIntake.divRound(WAD.muln(100));
       currentPrice = await coinMachine.getCurrentPrice();
-      expect(currentPrice).to.eq.BN(WAD.divn(10).muln(9));
+      expect(currentPrice).to.eq.BN(expectedPrice);
+    });
+
+    it("it monotonically adjusts prices according to demand", async () => {
```
kronosapiens, line 216: Nice test 👍

PR #842, opened 2020-05-07
Stagger allowed miner responses during a dispute

area:
First off, this is (notionally) based on #812 hence draft status for now.

We obviously hope that multiple people will run reputation mining clients. For submissions, we have a window over which it gets easier and easier to submit new hashes, with the difficulty being randomised per user, and with it being easier to submit if you have more staked. We did this because we wanted rewards to be 'fairly' distributed and for miners to not be wasting gas, making submissions that wouldn't succeed because all the slots filled up before their transaction was mined (which wasn't the case when they broadcast it). This PR adds similar functionality to dispute resolutions.

In this implementation, there is a window during a dispute resolution where only those that have submitted a hash are able to respond to a challenge, and the point in this window where a submitter is able to respond is randomised per-submitter for each stage. This additional window lasts 10 minutes, so in the event of a 'full' mining cycle submission, each submitter can expect to get  (*waves hands vaguely*) a minute after they are eligible to respond to a dispute before someone else becomes eligible, which seems adequate. This increases the potential severity of a DDOS attack, because each dispute round can take twice as long to resolve if bad submissions are defended judiciously, but the well-behaving miners shouldn't step on each others toes during submissions, which is well worth the cost. 

There's one subtlety in the implementation worth drawing attention to; in the case of all but one of these windows, the window is based on the last response timestamp of the submission being defended. The one case that's not true is the `invalidateHash` call where an entry is being invalidated (i.e. not the case where a bye is awarded), where the window is based on the last response timestamp of the submission being invalided.


area:
Further comments as this hopefully nears being done:

`ReputationMiningCycleBinarySearch.sol` and `ReputationMiningCycleCommon.sol` have been introduced as the contract sizes were getting too large.

`lastResponseTimestamp` is now set on submission to the time the window closes. This is fairer (even-numbered submissions now pay the same gas) and also makes working out when a valid window is easier in the case of the last submission being able to get a bye in the first round (when previously, that timestamp wouldn't be set).

I have added a `BinarySearchStep` event - every other stage of a mining dispute had a corresponding event.

I have added a helper function `makeTxAtTimestamp` which does what it says on the tin. It stops mining, sends a transaction, and then mines a block at the specified timestamp before re-enabling mining. This helps massively in making tests reproduceable (and was necessary here) for the main test added (`should not allow miners to respond immediately during a dispute, but they should be able to some time before the end of the window`). It arguably should be used instead of `forwardTime` wherever possible, but that is much more of a maintenance PR than something that should be here I think.

contracts/reputationMiningCycle/ReputationMiningCycleDataTypes.sol
```diff
@@ -62,12 +62,15 @@ contract ReputationMiningCycleDataTypes {
                                           // Justification tree.
   }
 
+
+  enum disputeStages { CONFIRM_JRH, BINARY_SEARCH_RESPONSE, BINARY_SEARCH_CONFIRM, RESPOND_TO_CHALLENGE, INVALIDATE_HASH, CONFIRM_NEW_HASH }
```
kronosapiens, line 66: Our enum convention is `enum PascalCase { camelCase, camelCase }` -- any reason to depart here?

contracts/reputationMiningCycle/ReputationMiningCycle.sol
```diff
@@ -563,82 +490,14 @@ contract ReputationMiningCycle is ReputationMiningCycleCommon {
     return reputationHashSubmissions[_user].proposedNewRootHash != 0x00 || respondedToChallenge[_user];
   }
 
+  function getResponsePossible(disputeStages stage, uint256 since) external view returns (bool) {
+    return responsePossible(stage, since);
```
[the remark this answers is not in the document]
area: I assume you're referring to [this](https://github.com/JoinColony/colonyNetwork/pull/812#discussion_r430766055) - the same applies here.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -115,4 +124,64 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
         y := add(y, mul(256, gt(arg, 0x8000000000000000000000000000000000000000000000000000000000000000)))
     }
   }
+
+  uint256 constant UINT256_MAX = 2**256 - 1;
+  uint256 constant SUBMITTER_ONLY_WINDOW = 60*10;
+  uint256 constant Y = UINT256_MAX / SUBMITTER_ONLY_WINDOW;
+
+  function responsePossible(disputeStages stage, uint256 since) internal view returns (bool) {
+    if (since > now) {
```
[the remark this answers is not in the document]
area: Either:

* A reputation miner is calling `getResponsePossible`
* A transaction is being called against a transaction that has had its `lastResponseTimestamp` [set to a value in the future](https://github.com/JoinColony/colonyNetwork/pull/842/files#diff-5e6ea9a45310f313334f5c3e8ba335cdR168) (expand `ReputationMiningCycle.sol` to view the highlighted lint)

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -115,4 +124,64 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
...
+      // require user made a submission
+      if (reputationHashSubmissions[msg.sender].proposedNewRootHash == bytes32(0x00)) {
+        return false;
+      }
+      uint256 target = delta * Y;
+      if (uint256(keccak256(abi.encodePacked(msg.sender, stage))) > target) {
+        return false;
+      }
+    }
+    return true;
```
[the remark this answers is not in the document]
area: That's my intention, yep.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -115,4 +124,64 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
...
+  uint256 constant UINT256_MAX = 2**256 - 1;
+  uint256 constant SUBMITTER_ONLY_WINDOW = 60 * 10;
+  uint256 constant Y = UINT256_MAX / SUBMITTER_ONLY_WINDOW;
+
+  function responsePossible(DisputeStages stage, uint256 since) internal view returns (bool) {
+    if (since > now) {
+      return false;
+    }
+
+    uint256 delta = now - since;
```
kronosapiens, line 137: Reading through this and I'm still getting confused around which way the window works. One one level it seems like it should be `>=`, i.e. more restrictive when the delta is large, but we have the opposite. Can we rename `delta` to `responseTimeRemaining` or something? `since` could also be clearer.

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -367,30 +377,44 @@ class ReputationMinerClient {
 
         // If we're here, we do have an opponent.
         // Has our opponent timed out?
+        // TODO: Remove these magic numbers
```
kronosapiens, line 380: Does the client not have a `constants.js` file?

test/contracts-network/reputation-basic-functionality.js
```diff
@@ -156,6 +156,24 @@ contract("Reputation mining - basic functionality", (accounts) => {
...
+
+    it("should update nSubmissionsForHash as submissions are made", async () => {
+      await giveUserCLNYTokensAndStake(colonyNetwork, MINER1, MIN_STAKE);
+      await advanceMiningCycleNoContest({ colonyNetwork, test: this });
+
+      const repCycle = await getActiveRepCycle(colonyNetwork);
+      await forwardTime(MINING_CYCLE_DURATION, this);
+
+      let nSubmissionsForHash = await repCycle.getNSubmissionsForHash("0x12345678", 10, "0x00");
+      expect(nSubmissionsForHash).to.eq.BN(0);
```
kronosapiens, line 168: `to.be.zero`

test/reputation-system/reputation-mining-client/client-auto-functionality.js
```diff
@@ -427,12 +463,22 @@ process.env.SOLIDITY_COVERAGE
...
-          // Good client should now realise it can timeout bad submission
+          disputeRound = await repCycle.getDisputeRound(0);
+          badEntry = disputeRound[badIndex];
+          goodEntry = disputeRound[goodIndex];
 
-          await goodClientInvalidateOpponent;
+          const noChallengeCompleted = noEventSeen(repCycleEthers, "ChallengeCompleted");
+          const noHashInvalidated = noEventSeen(repCycleEthers, "HashInvalidated");
+
+          await noChallengeCompleted;
```
kronosapiens, line 473: Remind me why we don't just `await` these promises directly?

test/reputation-system/root-hash-submissions.js
```diff
@@ -793,5 +830,37 @@ contract("Reputation mining - root hash submissions", (accounts) => {
...
+      await goodClient.addLogContentsToReputationTree();
+      await badClient.addLogContentsToReputationTree();
+      await badClient2.addLogContentsToReputationTree();
+      await badClient3.addLogContentsToReputationTree();
+
+      const repCycle = await getActiveRepCycle(colonyNetwork);
+
+      await goodClient.submitRootHash();
+      let reward = await repCycle.getDisputeRewardSize();
+      expect(reward).to.eq.BN("0");
```
kronosapiens, line 850: `to.be.zero`

contracts/reputationMiningCycle/ReputationMiningCycleBinarySearch.sol
```diff
@@ -0,0 +1,187 @@
...
+import "../colonyNetwork/IColonyNetwork.sol";
+import "../patriciaTree/PatriciaTreeProofs.sol";
+import "../tokenLocking/ITokenLocking.sol";
+import "./ReputationMiningCycleStorage.sol";
+import "./ReputationMiningCycleCommon.sol";
+
+
+contract ReputationMiningCycleBinarySearch is ReputationMiningCycleCommon {
+  function respondToBinarySearchForChallenge(
+    uint256 round,
```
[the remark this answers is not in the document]
area: It's barely used in the interface contract either for what it's worth, but I guess it's best to be consistent across all contracts.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
area: Rearranged this a bit, that hopefully makes it clearer to the reader.

packages/reputation-miner/ReputationMinerClient.js
area: It does not. We could add one, ideally though we'd want these values being picked up off the contracts, which I think is what I had in mind when I wrote this comment.

test/reputation-system/reputation-mining-client/client-auto-functionality.js
area: Because it's testing the auto functionality. If we awaited directly where they're created, the actions to trigger them have not yet taken place on chain so the test would time out. If we moved the creation to where they're awaited, then the tests become spotty, as the thing the promises are waiting for might have happened before they've been set up.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -126,21 +126,21 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
...
 
-    uint256 delta = now - _since;
-    if (delta <= SUBMITTER_ONLY_WINDOW) {
+    if (now <= _responseWindowOpened + SUBMITTER_ONLY_WINDOW_DURATION) {
       // require user made a submission
       if (reputationHashSubmissions[msg.sender].proposedNewRootHash == bytes32(0x00)) {
         return false;
       }
-      uint256 target = delta * Y;
+      uint256 windowOpenFor = now - _responseWindowOpened;
```
kronosapiens, line 142: If you want to make the most of this variable I'd be happy using it up in the conditional as

```
if (windowOpenFor <= SUBMITTER_ONLY_WINDOW_DURATION) {
```

PR #844, opened 2020-05-14
Use staking functionality for mining

area:
With staking merged, it makes sense to use it wherever it's appropriate - and a big place where it would be appropriate and is not currently used is reputation mining. 

This PR moves all 'knowledge' of mining up to the `ColonyNetwork` contract which is much better than having that on `TokenLocking`. So to mine, you obligate CLNY through `ColonyNetwork`, and otherwise everything is the same as before. Instead of not being allowed to withdraw during mining if you've submitted, you're not allowed to unstake if you've submitted. 

The only thing I'm not happy with here is the timestamp stuff; are we planning on using that anywhere other than with mining? If not, it should just be elevated to `ColonyNetwork` as well, specific to staking for mining.

I've taken the liberty of adding `getApproval` and `getObligation` to `TokenLocking` as well.

### DIFF
--- a/contracts/colony/ColonyFunding.sol
+++ b/contracts/colony/ColonyFunding.sol
@@ -421,10 +421,8 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
 
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
-    uint256 userDepositLockCount = tokenLocking.getUserLock(token, msg.sender).lockCount;
     uint256 userTokens = tokenLocking.getUserLock(token, msg.sender).balance;
 
-    require(userDepositLockCount < payoutId, "colony-reward-payout-lock-count-too-high");
     require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
     require(userReputation > 0, "colony-reward-payout-invalid-user-reputation");
 
--- a/contracts/colony/ColonyFunding.sol
+++ b/contracts/colony/ColonyFunding.sol
@@ -421,10 +421,10 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
 
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
-    uint256 userDepositTimestamp = tokenLocking.getUserLock(token, msg.sender).timestamp;
+    uint256 userDepositLockCount = tokenLocking.getUserLock(token, msg.sender).lockCount;
     uint256 userTokens = tokenLocking.getUserLock(token, msg.sender).balance;
 
-    require(userDepositTimestamp < payout.blockTimestamp, "colony-reward-payout-deposit-too-recent");
+    require(userDepositLockCount < payoutId, "colony-reward-payout-lock-count-too-high");
     require(userTokens > 0, "colony-reward-payout-invalid-user-tokens");
     require(userReputation > 0, "colony-reward-payout-invalid-user-reputation");
 
--- a/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
+++ b/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
@@ -124,4 +124,9 @@ contract ColonyNetworkDataTypes {
     uint128 nUpdates;
     uint128 nPreviousUpdates;
   }
+
+  struct MiningStakeInfo {
+    uint256 amount;
+    uint256 timestamp;
+  }
 }
--- a/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
+++ b/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
@@ -125,7 +125,7 @@ contract ColonyNetworkDataTypes {
     uint128 nPreviousUpdates;
   }
 
-  struct MiningStakeInfo {
+  struct MiningStake {
     uint256 amount;
     uint256 timestamp;
   }
--- a/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
+++ b/contracts/colonyNetwork/ColonyNetworkDataTypes.sol
@@ -94,6 +94,8 @@ contract ColonyNetworkDataTypes {
   /// @param label The label registered
   event ColonyLabelRegistered(address indexed colony, bytes32 label);
 
+  event ReputationMinerPenalised(address miner, address beneficiary, uint256 tokensLost);
+
   struct Skill {
     // total number of parent skills
     uint128 nParents;
--- a/contracts/tokenLocking/ITokenLocking.sol
+++ b/contracts/tokenLocking/ITokenLocking.sol
@@ -84,36 +84,29 @@ contract ITokenLocking is TokenLockingDataTypes {
   /// @param _force Pass true to forcibly unlock the token
   function withdraw(address _token, uint256 _amount, bool _force) public;
 
-  /// @notice Function called to punish people who staked against a new reputation root hash that turned out to be incorrect.
-  /// @dev While public, it can only be called successfully by the current ReputationMiningCycle.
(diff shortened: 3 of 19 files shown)


kronosapiens:
You raise a good point about tracking the timestamp on the lock itself vs. in the colony network. On one hand, I could imagine that there are other use-cases for this type of timestamp, and making it available means that more use-cases are possible than would be otherwise. On the other hand, the ability to deposit for others makes the semantics of the timestamp murky, and further it is extra gas to record the timestamp for every token, when in many cases it will not be used.

area:
I think I mostly don't like that the timestamp is for the deposit as a whole, whereas I think the number that might actually prove useful for the type of functionality we're thinking of is the timestamp that a stake was made. Indeed, it's possible that one contract would have different timestamps for stakes for different reasons, so I think I'm leaning towards taking it out of `TokenLocking`, and adding it where it's needed, per-stake.

kronosapiens:
I'm supportive of that, and it should be possible for extensions, etc to implement whatever bookkeeping they need (i.e. alongside any `obligateStake` calls).

contracts/colony/ColonyFunding.sol
```diff
@@ -421,10 +421,10 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     require(block.timestamp - payout.blockTimestamp <= 60 days, "colony-reward-payout-not-active");
 
     ITokenLocking tokenLocking = ITokenLocking(IColonyNetwork(colonyNetworkAddress).getTokenLocking());
-    uint256 userDepositTimestamp = tokenLocking.getUserLock(token, msg.sender).timestamp;
+    uint256 userDepositLockCount = tokenLocking.getUserLock(token, msg.sender).lockCount;
     uint256 userTokens = tokenLocking.getUserLock(token, msg.sender).balance;
 
-    require(userDepositTimestamp < payout.blockTimestamp, "colony-reward-payout-deposit-too-recent");
+    require(userDepositLockCount < payoutId, "colony-reward-payout-lock-count-too-high");
```
kronosapiens, line 427: We really mean that `userDepositLockCount + 1 == payoutId` right? It's meant to be exactly one less, not just arbitrarily less.

contracts/colonyNetwork/ColonyNetworkDataTypes.sol
```diff
@@ -122,4 +124,9 @@ contract ColonyNetworkDataTypes {
     uint128 nUpdates;
     uint128 nPreviousUpdates;
   }
+
+  struct MiningStakeInfo {
```
kronosapiens, line 128: IMO this could just be `MiningStake` but it's not a big deal.

contracts/tokenLocking/ITokenLocking.sol
```diff
@@ -84,36 +84,29 @@ contract ITokenLocking is TokenLockingDataTypes {
...
-
   /// @notice Allow the colony to obligate some amount of tokens as a stake.
-  /// Can only be called by a colony.
-  /// @param _colony Address of the colony we are willing to let obligate us.
+  /// @dev Can only be called by a colony or colonyNetwork
+  /// @param _user Address of the user that is allowing their holdings to be staked by the caller
   /// @param _amount Amount of that colony's internal token up to which we are willing to be obligated.
   /// @param _token The colony's internal token address
-  function approveStake(address _colony, uint256 _amount, address _token) public;
+  function approveStake(address _user, uint256 _amount, address _token) public;
```
kronosapiens, line 92: Mm nice catch thank you.

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -112,7 +98,7 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
     makeConditionalDeposit(_token, _amount, msg.sender);
 
     Lock storage lock = userLocks[_token][msg.sender];
-    emit UserTokenDeposited(_token, msg.sender, lock.balance, lock.timestamp);
+    emit UserTokenDeposited(_token, msg.sender, lock.balance);
```
kronosapiens, line 101: Do we have to worry about changing the event signatures here?

contracts/colony/ColonyFunding.sol
area: Yeah, we do, I _think_ this can just be straight up removed though as the appropriate checks are done in `unlockTokenForUser`. I think that realisation also solves the problem I've been worrying about existing colonies that would check the timestamp here with these changes, and find zero (because new deposits don't get timestamps / timestamps aren't updated with the new TokenLocking), but unlockToken covers us there.

I still don't know why we do all this in `calculateRewardForUser` though :sweat_smile:

contracts/tokenLocking/TokenLocking.sol
area: I don't think it's a problem. I don't think anything is looking for / using these signatures?

contracts/colony/ColonyFunding.sol
kronosapiens: Agreed, that check already exists in TokenLocking... although I could see it clarifying for an error to be here as well (otherwise it seems sort of hidden). And yup, we've definitely discussed how `calculateRewardForUser` should be a `view` function and all the state changes should be elsewhere, but /shrug.

contracts/tokenLocking/TokenLocking.sol
kronosapiens: Just checking -- not ColonyJS or anything?

contracts/colony/ColonyFunding.sol
area: We get `colony-token-locking-has-previous-active-locks` instead of `colony-token-locking-has-previous-active-locks` which seems pretty much the same to me.

contracts/tokenLocking/TokenLocking.sol
area: I mean, they're visible there, so the definitions etc will need updating, but I don't know of anything actually consuming them - they're just there for completeness so far, as far as I know.

contracts/colony/ColonyFunding.sol
kronosapiens: You mean instead of `colony-reward-payout-lock-count-too-high`?

test/contracts-network/colony-reward-payouts.js
```diff
@@ -514,7 +514,7 @@ contract("Colony Reward Payouts", (accounts) => {
 
       await checkErrorRevert(
         colony.claimRewardPayout(payoutId, initialSquareRoots, ...userReputationProof1, { from: userAddress1 }),
-        "colony-reward-payout-lock-count-too-high"
+        "colony-token-already-unlocked"
```
kronosapiens: Mm this seems like the wrong error message -- don't we want `colony-token-locking-has-previous-active-locks`?

kronosapiens: What about this?

```
    require(lockCountDelta <= 1, "colony-token-locking-has-previous-active-locks");
    require(lockCountDelta != 0, "colony-token-already-unlocked");
```

The first one will trigger for all diffs except 0 and 1, indicating that there are existing locks. The second then triggers only in the remaining case of a diff of 0, indicating the token is unlocked. If both fail to trigger, it means we have 1, the correct value.

area: > Mm this seems like the wrong error message -- don't we want colony-token-locking-has-previous-active-locks?

I am definitely not up to scratch today, but I don't think so? When this error message is being checked, we have already waived the payout, and we're checking that we can't claim a payout that we've waived. So there should not be previous active locks, otherwise we've not correctly waived this payout.

> Seems like the issue is that the requires as they are currently can't differentiate between a diff of 0 and a diff of >1.

The requires are currently:
```
    require(lockCountDelta != 0, "colony-token-already-unlocked");
    require(lockCountDelta == 1, "colony-token-locking-has-previous-active-locks");
```
So surely it can differentiate? If the diff is 0, the first one triggers and if the diff is >1, the second one triggers. I think I must just not be understanding something!

contracts/colony/ColonyFunding.sol
area: I do, yes, got a bit trigger happy with my pasting there.

test/contracts-network/colony-reward-payouts.js
kronosapiens: Ah my mistake, I didn't see the entire test in the diff view and for some reason I assumed these tests were for locked tokens. You're right, looks like things are fine as they are.

PR #848, opened 2020-06-05
Change how coverage runs are done

area:
A year ago, [some idiot](https://github.com/JoinColony/colonyNetwork/pull/610#issuecomment-494373474) put in a short-term fix that was never properly resolved, and now it's come to collect.

Due to (I assume) some changes on circle, our monolithic coverage job fails with [exit code 137](https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory), and 4Gb is the limit on the free circle instances we get, so we can't simply up the memory. This PR splits contract coverage in to two separate runs (one dealing with reputation, the other dealing with everything else), then merges those results in a third run and compares to the thresholds we've set. This means that our builds will take up to three instances on Circle concurrently, but should take less time.

Things I'm not super happy with:

A whole new `solcover.js` just to change the coverage output directory. The alternative was to output coverage to the same directory in each job, and then move them to a new directory, but in the rare cases when we run coverage locally, this is more predictable.

The glob I had to go with in `test:contracts:coverage`. It tests all directories that don't end in `m`, which works for now. I couldn't find a cleaner way of doing this with the glob syntax and what is allowed for the `--path` parameter (weirdly, doing `--path ./test/contracts-network/* --path ./test/extensions/*` also ended up testing `reputation-system` which is probably a bug in `truffle` or the `solidity-coverage` extension, but I couldn't face trying to debug that, especially as our builds would still be broken while we waited for it to be merged upstream).


.circleci/config.yml
```diff
@@ -126,13 +126,47 @@ jobs:
...
       - run:
-          name: "Running unit tests with coverage"
+          name: "Running contract unit tests with coverage"
           command: yarn run test:contracts:coverage
           environment:
             NODE_OPTIONS: --max_old_space_size=4096
+      - persist_to_workspace:
+          root: ./
+          paths:
+            - coverage
```
kronosapiens, line 136: Why not `coverage-contracts` for symmetry?

package.json
```diff
@@ -23,12 +24,13 @@
...
     "test:contracts:upgrade:parity": "npm run start:blockchain:client parity & npm run generate:test:contracts && truffle migrate --reset --compile-all && truffle test ./test-upgrade/* --network integration",
     "test:contracts:upgrade:ganache": "npm run start:blockchain:client & npm run generate:test:contracts && truffle migrate --reset --compile-all && truffle test ./test-upgrade/* --network development",
     "test:contracts:gasCosts": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle test test-gas-costs/gasCosts.js --network development",
     "test:contracts:patricia": "npm run start:blockchain:client parity & truffle migrate --reset --compile-all && truffle test packages/reputation-miner/patricia-test.js --network development",
-    "test:contracts:coverage": "SOLIDITY_COVERAGE=1 truffle run coverage --network coverage --temp build && istanbul check-coverage --statements 99 --branches 94 --functions 99 --lines 99",
+    "test:contracts:coverage": "SOLIDITY_COVERAGE=1 truffle run coverage --network coverage --temp build --file='./test/*[!m]/*'",
```
kronosapiens, line 33: Can we do better than `m`? Would `reputation-system` work? It's more verbose but also more clear.

area: This matches all directories that do not end in an `m` i.e. not `reputation-system`. As I say, I was unable to come up with a better way of doing this matching within the limitations of the glob syntax.

kronosapiens: Sure, but I mean can we select directories that don't end in `reputation-system` instead?

area: If you can see a way of doing that with the [glob](https://www.man7.org/linux/man-pages/man7/glob.7.html) syntax I'd much prefer that, but I can't, at least.

PR #849, opened 2020-06-07
Use payments rather than tasks in reputation tests where appropriate

area:
As part of trying to make our tests better (because they have started randomly failing on Circle right now which is really cramping our style), I decided to take a look at some of the heavier tests in the hope that by making some tests lighter, Circle would be happier.

Far and away the biggest / most intensive tests are some of the reputation ones with many clients, with single tests taking nearly 30 minutes on Circle - clearly dominating the reputation coverage test runs. In these tests, many (up to fourteen) reputation miners are created, and they all process the reputation log generated for that test. While for some of these tests, we do need this many miners, this log was four times longer than it needed to be for the tests to work, as we created a task (and four reputation log entries) for each miner, when we only need one reputation log entry for each miner for the tests to behave correctly.

This PR creates payments rather than tasks, which create one log entry each, and dramatically reduces the runtime of these tests. 

On CircleCI [before](https://circleci.com/gh/JoinColony/colonyNetwork/12087):

```
✓ should prevent a hash from advancing if it might still get an opponent,
     even if that opponent is from more than one round ago (1737172ms)
```

and with [this change](https://circleci.com/gh/JoinColony/colonyNetwork/12108):
```
      ✓ should prevent a hash from advancing if it might still get an opponent,
     even if that opponent is from more than one round ago (404934ms)
```

Intending to run multiple instances of this workflow to see if this results in consistently green tests. [First one was green](https://circleci.com/workflow-run/3d036eb7-54f3-4d9f-a831-7f6d505301ab).

### DIFF
--- a/test/reputation-system/dispute-resolution-misbehaviour.js
+++ b/test/reputation-system/dispute-resolution-misbehaviour.js
@@ -282,11 +282,16 @@ contract("Reputation Mining - disputes resolution misbehaviour", (accounts) => {
     async function setUpNMiners(n) {
       expect(accounts.length, "Not enough accounts for test to run").to.be.at.least(n + 3);
       const accountsForTest = accounts.slice(3, n + 3);
-
       await fundColonyWithTokens(metaColony, clnyToken, INITIAL_FUNDING.muln(n));
       for (let i = 0; i < n; i += 1) {
         await giveUserCLNYTokensAndStake(colonyNetwork, accountsForTest[i], DEFAULT_STAKE);
-        await setupFinalizedTask({ colonyNetwork, colony: metaColony, worker: accountsForTest[i] });
+        // await setupFinalizedTask({ colonyNetwork, colony: metaColony, worker: accountsForTest[i] });
+        await metaColony.addPayment(1, UINT256_MAX, accountsForTest[i], clnyToken.address, 40, 1, 0);
+        const paymentId = await metaColony.getPaymentCount();
+        const payment = await metaColony.getPayment(paymentId);
+        await metaColony.moveFundsBetweenPots(1, UINT256_MAX, UINT256_MAX, 1, payment.fundingPotId, INITIAL_FUNDING, clnyToken.address);
+        await metaColony.finalizePayment(1, UINT256_MAX, paymentId);
+
         // These have to be done sequentially because this function uses the total number of tasks as a proxy for getting the
         // right taskId, so if they're all created at once it messes up.
       }
--- a/test/reputation-system/dispute-resolution-misbehaviour.js
+++ b/test/reputation-system/dispute-resolution-misbehaviour.js
@@ -479,7 +479,8 @@ contract("Reputation Mining - disputes resolution misbehaviour", (accounts) => {
       await repCycle.confirmNewHash(4);
     });
 
-    it("should not allow stages to be skipped even if the number of updates is a power of 2", async () => {
+    it("should not allow stages to be skipped even if the number of updates is a power of 2", async function powerOfTwoTest() {
+      this.timeout(600000);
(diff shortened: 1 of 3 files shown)


area:
We could also split up some of the tests that run in the `lint-and-unit-test` job in to separate jobs.

area:
Second two builds were red, so I've split up the non-coverage build in two.

[First build](https://circleci.com/workflow-run/2eb8fe8e-a66a-43d8-8e81-bb78f6d53ec7)
[Second build](https://circleci.com/workflow-run/cac99f1a-6121-478b-b567-3fe332005b5f)
[Third build](https://circleci.com/workflow-run/c52d6f0b-96d7-44b7-b01f-1eeb50f59f7c)

area:
So that second build failed on the reputation job, with two tests timing out. If we compare a [bad run](https://circleci.com/gh/JoinColony/colonyNetwork/12128) to a [good run](https://circleci.com/gh/JoinColony/colonyNetwork/12126) the tests that passed in the bad run take twice as long:

```
      ✓ should prevent a hash from advancing if it might still get an opponent (226135ms)
```
compared to 
```
      ✓ should prevent a hash from advancing if it might still get an opponent (113314ms)
```

Which suggests we're just getting a lower spec'd circle instance, and just need to make ourselves more resilient to that?

area:
So I imagine the thing that kicked this all off was [this update](https://discuss.circleci.com/t/kernel-update-increase-in-out-of-memory-job-failures/36225) on Circle, which they claim they've rolled back but... _gestures at test runs_

test/reputation-system/dispute-resolution-misbehaviour.js
```diff
@@ -282,11 +282,16 @@ contract("Reputation Mining - disputes resolution misbehaviour", (accounts) => {
...
       await fundColonyWithTokens(metaColony, clnyToken, INITIAL_FUNDING.muln(n));
       for (let i = 0; i < n; i += 1) {
         await giveUserCLNYTokensAndStake(colonyNetwork, accountsForTest[i], DEFAULT_STAKE);
-        await setupFinalizedTask({ colonyNetwork, colony: metaColony, worker: accountsForTest[i] });
+        // await setupFinalizedTask({ colonyNetwork, colony: metaColony, worker: accountsForTest[i] });
+        await metaColony.addPayment(1, UINT256_MAX, accountsForTest[i], clnyToken.address, 40, 1, 0);
+        const paymentId = await metaColony.getPaymentCount();
+        const payment = await metaColony.getPayment(paymentId);
+        await metaColony.moveFundsBetweenPots(1, UINT256_MAX, UINT256_MAX, 1, payment.fundingPotId, INITIAL_FUNDING, clnyToken.address);
+        await metaColony.finalizePayment(1, UINT256_MAX, paymentId);
```
kronosapiens, line 293: Worth pointing out that using expenditures could actually be cheaper here if you used a single expenditure across all miners, something along the lines of:

```
await metaColony.makeExpenditure(1, UINT256_MAX, 1);
const expenditureId = await metaColony.getExpenditureCount();
const expenditure = await metaColony.getExpenditure(expenditureId);

await metaColony.moveFundsBetweenPots(1, UINT256_MAX, UINT256_MAX, 1, expenditure.fundingPotId, 40 * n, clnyToken.address);
  
for (let i = 0; i < n; i += 1) {
        await giveUserCLNYTokensAndStake(colonyNetwork, accountsForTest[i], DEFAULT_STAKE);
        await metaColony.setExpenditureRecipient(expenditureId, i, accountsForTest[i]);
        await metaColony.setExpenditurePayout(expenditureId, i, clnyToken.address, 40);
}

await metaColony.finalizeExpenditure(expenditureId);

for (let i = 0; i < n; i += 1) {
        await metaColony.claimExpenditurePayout(expenditureId, i, clnyToken.address);
}
```

PR #855, opened 2020-07-01
Factor in @colony/colony-js-contract-loader-fs

chmanie:
This PR removes the `@colony/colony-js-contract-loader-fs` package and factors it in as a file in the reputation miner


chmanie:
This is going to be needed in any case as we're retiring `@colony/colony-js-contract-loader-fs`.

PR #862, opened 2020-07-22
Add CLNY issuance control / automation mechanisms

area:
CLNY will be issued via three mechanisms:

* CoinMachine
* A stipend for the metacolony to pay for tasks
* Rewards for reputation miners.

Coinmachine is being dealt with in #835; this PR deals with the other two mechanisms.

### Metacolony stipend
Set via `setAnnualMetaColonyStipend`. An absolute value, given that none of these mechanisms react to each other - we trust the metacolony to set the values appropriately. This function is permissioned, but the `issueMetaColonyStipend` that awards the stipend can be called by anyone, as often as is desired, which will result in the appropriate pro-rata'd amount being awarded.

### Reputation mining cycle reward
Set via `setReputationMiningCycleReward` on the metacolony. This reward will be split between all entries that were submitted for the reputation hash that was ultimately accepted. The natural place to store this is on `colonyNetwork`, where previously the implication was that it was stored in `ReputationMiningCycle` (though has been set to zero until now). As a result, there are now two `setReputationRootHash` functions on `ColonyNetworkMining` to ease the transition - the one with `reward` as a parameter is deprecated, and will be removed after whatever upgrade these changes were included in.

### Notes

This PR also includes the `makeTxAtTimestamp` functionality I wrote for colonyNetwork#842, but I needed it here in order to write some of the tests. Obviously only one of these PRs will ultimately contain that functionality :smile:

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -245,6 +245,22 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return colonyNetwork.setFeeInverse(_feeInverse); // ignore-swc-107
   }
 
+  function setAnnualMetaColonyStipend(uint256 _amount) public
+  stoppable
+  auth
+  {
+    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    return colonyNetwork.setAnnualMetaColonyStipend(_amount); // ignore-swc-107
+  }
+
+  function setReputationMiningCycleReward(uint256 _amount) public
+  stoppable
+  auth
+  {
+    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    return colonyNetwork.setReputationMiningCycleReward(_amount); // ignore-swc-107
+  }
+
   function addNetworkColonyVersion(uint256 _version, address _resolver) public
   stoppable
   auth
@@ -339,6 +355,12 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Arbitration), address(this), sig, true);
     sig = bytes4(keccak256("emitSkillReputationPenalty(uint256,address,int256)"));
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Arbitration), address(this), sig, true);
+
+    // Add CLNY issuance functionality
+    sig = bytes4(keccak256("setAnnualMetaColonyStipend(uint256)"));
+    colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
+    sig = bytes4(keccak256("setReputationMiningCycleReward(uint256)"));
+    colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
   }
 
   function checkNotAdditionalProtectedVariable(uint256 _slot) public view recovery {
--- a/contracts/colonyNetwork/ColonyNetwork.sol
+++ b/contracts/colonyNetwork/ColonyNetwork.sol
@@ -19,8 +19,10 @@ pragma solidity 0.5.8;
 pragma experimental "ABIEncoderV2";
 
 import "./../common/EtherRouter.sol";
+import "./../common/ERC20Extended.sol";
 import "./../colony/ColonyAuthority.sol";
 import "./../colony/IColony.sol";
+import "./../colony/IMetaColony.sol";
 import "./../reputationMiningCycle/IReputationMiningCycle.sol";
 import "./ColonyNetworkStorage.sol";
 
@@ -298,6 +300,33 @@ contract ColonyNetwork is ColonyNetworkStorage {
     feeInverse = _feeInverse;
 
     emit NetworkFeeInverseSet(_feeInverse);
+  }
+
+  function issueMetaColonyStipend() public stoppable {
+    // Can be called by anyone
+    require(lastMetaColonyStipendIssued > 0, "colony-network-metacolony-stipend-not-set");
+    // When was the last one issued
+    // How much in total should have been issued since then
+    uint256 amountToIssue = mul(annualMetaColonyStipend, sub(now,lastMetaColonyStipendIssued)) / (60*60*24*365);
+    lastMetaColonyStipendIssued = now;
+    // amountToIssue = 0;
+    IMetaColony meta = IMetaColony(metaColony);
+
+    // mintTokensFor is coming in #835, use that instead of this.
+    meta.mintTokensForColonyNetwork(amountToIssue);
+    ERC20Extended clnyToken = ERC20Extended(IColony(metaColony).getToken());
+    clnyToken.transfer(metaColony, amountToIssue);
+  }
+
+  function setAnnualMetaColonyStipend(uint256 amount) public stoppable
+  calledByMetaColony
+  {
+    if (lastMetaColonyStipendIssued == 0) { lastMetaColonyStipendIssued = now; }
+    annualMetaColonyStipend = amount;
+  }
+
+  function getAnnualMetaColonyStipend() public view returns (uint256) {
+    return annualMetaColonyStipend;
   }
 
   function deployColony(address _tokenAddress, uint256 _version) internal returns (address) {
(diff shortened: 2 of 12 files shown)


contracts/colonyNetwork/IColonyNetwork.sol
```diff
@@ -195,13 +195,20 @@ contract IColonyNetwork is ColonyNetworkDataTypes, IRecovery {
...
   /// @param stakers Array of users who submitted or backed the hash, being accepted here as the new reputation root hash
-  /// @param reward Amount of CLNY to be distributed as reward to miners
+  /// @param reward Amount of CLNY to be distributed as reward to miners (not used)
   function setReputationRootHash(bytes32 newHash, uint256 newNLeaves, address[] memory stakers, uint256 reward) public;
 
+  /// @notice Set a new Reputation root hash and starts a new mining cycle. Can only be called by the ReputationMiningCycle contract.
+  /// @param newHash The reputation root hash
+  /// @param newNLeaves The updated leaves count value
+  /// @param stakers Array of users who submitted or backed the hash, being accepted here as the new reputation root hash
+  function setReputationRootHash(bytes32 newHash, uint256 newNLeaves, address[] memory stakers) public;
```
kronosapiens, line 210: Is this new?

contracts/tokenLocking/ITokenLocking.sol
```diff
@@ -131,8 +131,10 @@ contract ITokenLocking is TokenLockingDataTypes {
   /// @param _user Address of the user
   /// @return lock Lock object containing:
   ///   `lockCount` User's token lock count,
-  ///   `amount` User's deposited amount,
-  ///   `timestamp` Timestamp of deposit.
+  ///   `balance` User's deposited amount,
+  ///   `DEPRECATED_timestamp` Timestamp of deposit (deprecated)
+  ///   `pendingBalance` Tokens that have been sent to them, but are inaccessible until all locks are cleared and then these
+  ///                    tokens are claimed
   function getUserLock(address _token, address _user) public view returns (Lock memory lock);
```
kronosapiens, line 138: 👍

test/contracts-network/meta-colony.js
```diff
@@ -539,4 +544,68 @@ contract("Meta Colony", (accounts) => {
       await checkErrorRevert(metaColony.mintTokensForColonyNetwork(100), "colony-access-denied-only-network-allowed");
     });
   });
+
+  describe("when setting the metaColony stipend", () => {
+    it("should allow the metacolony stipend to be set", async () => {
+      const stipendBefore = await colonyNetwork.getAnnualMetaColonyStipend();
+      expect(stipendBefore).to.eq.BN(0);
```
kronosapiens, line 551: Channeling Elena: `to.be.zero`

test/contracts-network/meta-colony.js
```diff
@@ -539,4 +544,68 @@ contract("Meta Colony", (accounts) => {
...
+      ({ colony } = await setupRandomColony(colonyNetwork));
+      const colonyAsMetaColony = await IMetaColony.at(colony.address);
+      await checkErrorRevert(colonyAsMetaColony.setAnnualMetaColonyStipend(0), "colony-caller-must-be-meta-colony");
+    });
+
+    it("should allow the metacolony stipend to be claimed, and the amount minted changes based on when it is last claimed", async () => {
+      await metaColony.setAnnualMetaColonyStipend(WAD);
+      // In a year, we should get the full reward
+      let time = await currentBlockTime();
+      time = new BN(time).addn(24 * 3600 * 365);
```
kronosapiens, line 571: Could use `SECONDS_PER_DAY` to be consistent

test/reputation-system/root-hash-submissions.js
```diff
@@ -219,18 +222,31 @@ contract("Reputation mining - root hash submissions", (accounts) => {
...
+
+      const r2 = await WAD.muln(10)
+        .mul(mw2.mul(WAD).divRound(mw1.add(mw2)))
+        .divRound(WAD);
+
       // Check that they will be getting the reputation owed to them.
       let repLogEntryMiner = await inactiveRepCycle.getReputationUpdateLogEntry(0);
       expect(repLogEntryMiner.user).to.equal(MINER1);
-      expect(repLogEntryMiner.amount).to.be.zero; // Reward is 0 for now
+      expect(repLogEntryMiner.amount).to.eq.BN(r1); // Reward is 0 for now
```
kronosapiens, line 241: Comment seems no longer accurate

contracts/colony/Colony.sol
```diff
@@ -238,6 +238,22 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return colonyNetwork.setFeeInverse(_feeInverse); // ignore-swc-107
   }
 
+  function setAnnualMetaColonyStipend(uint256 _amount) public
+  stoppable
+  auth
+  {
+    IColonyNetwork colonyNetwork = IColonyNetwork(colonyNetworkAddress);
+    return colonyNetwork.setAnnualMetaColonyStipend(_amount); // ignore-swc-107
```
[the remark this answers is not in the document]
area: The MC does own the token (the issuance process where someone calls `colonyNetwork.issueMetaColonyStipend` goes through `mintTokensForColonyNetwork` on the metaColony). My reasoning here was that we didn't want to pollute the storage of every colony with a slot that would only be used in the metaColony - we can have metacolony-only functions in a metacolony-only interface that 'normal' colonies won't see, but the same isn't true for storage. This also makes it a bit more obvious that yes, while the metaColony is in control of its own issuance rate, it should be thinking about it in terms of the network as a whole, and could plausibly have that control removed in the future.

This also mirrors the pattern used for setting the network fee, which is also stored on the network. The location is a little more obvious in that case but it felt similar enough.

contracts/colonyNetwork/IColonyNetwork.sol
area: Yep, the original (deprecated in this PR) has the `reward` parameter, which at some point in the past I thought would come from the `ReputationMiningCycle`. We have to keep both for transition purposes (as whenever we deploy, the already-deployed ReputationMiningCycle contracts will expect to be able to call the one with the `reward` parameter). The deprecated one acts as a passthrough for this one.

contracts/colony/Colony.sol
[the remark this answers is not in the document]
kronosapiens: Ok, that storage argument is satisfying.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -300,6 +302,32 @@ contract ColonyNetwork is ColonyNetworkStorage {
...
   }
 
+  function issueMetaColonyStipend() public stoppable {
+    // Can be called by anyone
+    require(lastMetaColonyStipendIssued > 0, "colony-network-metacolony-stipend-not-set");
+    // When was the last one issued
+    // How much in total should have been issued since then
+    uint256 amountToIssue = mul(annualMetaColonyStipend, sub(now, lastMetaColonyStipendIssued)) / (365 days);
+    lastMetaColonyStipendIssued = now;
+    // amountToIssue = 0;
```
kronosapiens, line 312: Delete

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -300,6 +302,32 @@ contract ColonyNetwork is ColonyNetworkStorage {
     emit NetworkFeeInverseSet(_feeInverse);
   }
 
+  function issueMetaColonyStipend() public stoppable {
+    // Can be called by anyone
+    require(lastMetaColonyStipendIssued > 0, "colony-network-metacolony-stipend-not-set");
+    // When was the last one issued
+    // How much in total should have been issued since then
```
kronosapiens, line 309: Not sure if all these comments are necessary

test/reputation-system/root-hash-submissions.js
```diff
@@ -196,7 +198,8 @@ contract("Reputation mining - root hash submissions", (accounts) => {
       );
     });
 
-    it("should allow a user to back the same hash more than once in a same cycle with different entries, and be rewarded", async () => {
+    it.only("should allow a user to back the same hash more than once in a same cycle with different entries, and be rewarded", async () => {
```
kronosapiens, line 201: `it.only` ?? 😱

chmanie: https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-exclusive-tests.md

Hint hint

PR #863, opened 2020-07-22
Never restart if miner out of gas

area:
If we restart when the miner is out of gas, we end up in a restart loop, because it's almost certainly still out of gas a few seconds later. This PR means that it just sits patiently until it is manually restarted (presumably, after it has been sent some ether).


cobordism:
how about letting it restart when it receives gas money?

area:
It certainly is, it's just harder to do properly and this was an adequate quick fix to stop the problem (the restart loop, which I think was having a knock-on effect on the oracle that sits in a separate container, but the same pod, as the miner, causing the slow responses).

PR #866, opened 2020-08-26
Add storage layout smoke tests

area:
Trying to give myself the confidence to do a solidity upgrade; a lot of this hinges on the fact that when upgrading solidity, our storage layout / representation of existing data should not change. So this PR adds a test that looks for that. 

This test will probably need to updated reasonably often, as I explain in the comments - anything that changes an address that is stored in these tests (e.g. an extra transaction during migrations) or anything that stores something in a new storage slot will cause them to change. But they shouldn't change (I think) when just doing a solidity upgrade. If they do, that might suggest that there is a breaking change that we have to work around. Or it might suggest that there is something I don't understand (probably more likey).

Note that I am not checking the `codehash` as well - this very well could change between solidity versions as the compiler compiles code that does the same thing in different ways (introducing optimisations etc). This *could* backfire if we're relying on some 'wrong' behaviour, but for a contract as simple as EtherRouter it seems quite unlikely that would not have been found.


test-smoke/colony-storage-consistent.js
```diff
@@ -0,0 +1,163 @@
...
+    await colony.setArbitrationRole(1, UINT256_MAX, ARBITRATOR, 1, true);
+    await fundColonyWithTokens(colony, token, UINT256_MAX);
+    domain1 = await colony.getDomain(1);
+
+    otherToken = await Token.new("otherName", "otherSymbol", 18);
+    await otherToken.unlock();
+    await fundColonyWithTokens(colony, otherToken, UINT256_MAX);
+  });
+
+  // x.engine.manager.state.blockchain.stateTrie.get(utils.toBuffer(y.address), (err, res)=>console.log(res.toString('hex')))
```
kronosapiens, line 65: Delete?

test-smoke/colony-storage-consistent.js
```diff
@@ -0,0 +1,163 @@
...
+
+      const contractEditing = await ContractEditing.new();
+      const networkAsER = await EtherRouter.at(colonyNetwork.address);
+      const colonyNetworkResolverAddress = await networkAsER.resolver();
+      const colonyNetworkResolver = await Resolver.at(colonyNetworkResolverAddress);
+      await colonyNetworkResolver.register("setStorageSlot(uint256,bytes32)", contractEditing.address);
+      const editableNetwork = await ContractEditing.at(colonyNetwork.address);
+
+      let hashable = `0x000000000000000000000000${accounts[5].slice(2)}${new BN(33).toString(16, 64)}`;
+      let hashed = web3.utils.sha3(hashable);
```
kronosapiens, line 113: Not `soliditySha3`?

test-smoke/colony-storage-consistent.js
```diff
@@ -0,0 +1,163 @@
...
+      const colonyNetworkAccount = new Account(colonyNetworkStateHash);
+
+      // We did a whole expenditure above, so let's take out the finalized timestamp
+
+      hashable = `0x${new BN(1).toString(16, 64)}${new BN(25).toString(16, 64)}`;
+      hashed = web3.utils.sha3(hashable);
+      slot = new BN(hashed.slice(2), 16);
+      slot = slot.addn(3);
+      // const valueBefore = await web3.eth.getStorageAt(colony.address, slot);
+      // console.log(valueBefore)
```
kronosapiens, line 129: Delete

test-smoke/colony-storage-consistent.js
```diff
@@ -0,0 +1,163 @@
...
+
+      const colonyNetworkStateHash = await getAddressStateHash(colonyNetwork.address);
+      const colonyNetworkAccount = new Account(colonyNetworkStateHash);
+
+      // We did a whole expenditure above, so let's take out the finalized timestamp
+
+      hashable = `0x${new BN(1).toString(16, 64)}${new BN(25).toString(16, 64)}`;
+      hashed = web3.utils.sha3(hashable);
+      slot = new BN(hashed.slice(2), 16);
+      slot = slot.addn(3);
```
kronosapiens, line 127: Can we add a comment describing what's going on up here?

test-smoke/colony-storage-consistent.js
area: Originally didn't seem to work with `soliditySha3` due to (I think) the way I was expressing the variables, as that function close-packs them before hashing. I've since changed to a big string, however, so I can switch back to `soliditySha3` without issue.

PR #870, opened 2020-08-28
Update to solidity 0.7.0

area:
Supercedes #732, dependent on #866, so a draft PR until the latter is merged.

### Non-solidity changes:

* Moves from `solium` to `solhint`, as the former does not support 0.7.0
* Moves from `solidity-parser-antlr` to `solidity-parser-diligence`, maintained by Consensys (following `solidity-coverage`)
* Moves to our own fork of `dappsys-monolithic`, as the original does not appear to be maintained any more, and changes were required to let overloads etc work.

### Solidity changes:
* Various small language changes - `block.timestamp` instead of `now`, some things in assembly are now functions, and so on.
* Overloading functions you've inherited now has to be declared both for the overrider and the overidee (I'm sure there are better CS names :P). This actually messed up some of our inheritance things, where we inherited from both an Interface and a Base contract with functions defined and so had to add functions that just call `super` in the Patricia Tree contracts to get the same behaviour.
* You can no longer instantiate structs in memory that contain mappings (the mappings were previously ignored, which the solidity team deemed confusing). So where we did that, we now directly assign to each element, rather than declaring it in memory and copying it in.
* Where it seemed appropriate, I made contracts 'interface`s, particularly as this comes with some assumptions about overrides that you don't have to mark.


hrkrshnn:
BTW solidity maintains a version of colony upgraded to 0.7.0: https://github.com/solidity-external-tests/colonyNetwork/tree/develop_070 However it branches off at https://github.com/solidity-external-tests/colonyNetwork/commit/d6fc347c83230ef95e1ac622008f217bd9ecf47b, so it cannot be directly merged. Regardless, I hope it will give an idea of what needs to be upgraded.

I think most of the difficult upgrades are from 0.5.0 to 0.6.0. From 0.6.0 to 0.7.0, the [solidity upgrade tool](https://solidity.readthedocs.io/en/v0.7.1/using-the-compiler.html?highlight=solidity%20upgrade#solidity-upgrade) can upgrade automatically without any issues, I think.

area:
Look ma, no changes to the storage smoke tests!

contracts/common/DomainRoles.sol
```diff
@@ -15,27 +15,27 @@
...
 
-pragma solidity 0.5.8; // ignore-swc-103
+pragma solidity 0.7.3; // ignore-swc-103
 
 import "./../../lib/dappsys/roles.sol";
 
 
 contract DomainRoles is DSRoles {
-  mapping(address=>mapping(uint256=>bytes32)) internal _user_domain_roles;
+  mapping(address=>mapping(uint256=>bytes32)) internal userDomainRoles;
```
kronosapiens, line 24: I followed the snake case convention here because that's what `DSRoles` used and this is basically an override of that, but I could also see how this crosses the threshold into the "Colony" contracts and so we should adopt the Colony style.

kronosapiens (approved):

PR #876, opened 2020-09-24
Add support for arbitrary transactions

### LINKED ISSUE #808: Allow colonies to make arbitrary transactions
In order to talk to other contracts directly, colonies need to support making arbitrary transactions - but we need to do them in a way that doesn't allow for malicious behaviour against the rest of the network. Such transactions should therefore not be allowed to interact with the TokenLocking contract, the ColonyNetwork, and so forth. Only the `ROOT` permission should be allowed to perform these transactions.

kronosapiens:
Closes #808

~Depends on #881~
~Depends on #870~

Following the discussion in #881 , we will add an `identifier()` function to every extension, which returns its `bytes32` identifier. This can be used to verify that an address is a network-managed extension, allowing us to prevent colonies making arbitrary transactions to these extensions.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -56,7 +56,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     require(sig != APPROVE_SIG, "colony-cannot-call-erc20-approve");
     require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
 
-    // Prevent transactions to network-managed extensions
+    // Prevent transactions to network-managed extensions installed in this colony
     try ColonyExtension(_to).identifier() returns (bytes32 extensionId) {
       require(
         IColonyNetwork(colonyNetworkAddress).getExtensionInstallation(extensionId, address(this)) != _to,
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -37,8 +37,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
+  bytes4 constant APPROVE_SIG = bytes4(keccak256("approve(address,uint256)"));
   bytes4 constant TRANSFER_SIG = bytes4(keccak256("transfer(address,uint256)"));
-  bytes4 constant TRANSFER_FROM_SIG = bytes4(keccak256("transferFrom(address,address,uint256)"));
 
   function makeArbitraryTransaction(address _to, uint256 _value, bytes memory _action)
   public stoppable auth
@@ -50,8 +50,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     bytes4 sig;
     assembly { sig := mload(add(_action, 0x20)) }
 
+    require(sig != APPROVE_SIG, "colony-cannot-call-erc20-approve");
     require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
-    require(sig != TRANSFER_FROM_SIG, "colony-cannot-call-erc20-transfer-from");
 
     return executeCall(_to, _value, _action);
   }
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -41,7 +41,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   bytes4 constant APPROVE_SIG = bytes4(keccak256("approve(address,uint256)"));
   bytes4 constant TRANSFER_SIG = bytes4(keccak256("transfer(address,uint256)"));
 
-  function makeArbitraryTransaction(address _to, uint256 _value, bytes memory _action)
+  function makeArbitraryTransaction(address _to, bytes memory _action)
   public stoppable auth
   returns (bool)
   {
@@ -64,7 +64,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
       );
     } catch {}
 
-    return executeCall(_to, _value, _action);
+    return executeCall(_to, 0, _action);
   }
 
   function emitDomainReputationPenalty(
@@ -369,7 +369,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Architecture), address(this), sig, true);
 
     // Add arbitrary tx functionality
-    sig = bytes4(keccak256("makeArbitraryTransaction(address,uint256,bytes)"));
+    sig = bytes4(keccak256("makeArbitraryTransaction(address,bytes)"));
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
   }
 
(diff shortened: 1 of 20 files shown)


contracts/colony/Colony.sol
```diff
@@ -37,6 +37,15 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
+  function makeArbitraryTransaction(address _to, uint256 _value, bytes memory _action)
+  public stoppable auth
+  returns (bool)
+  {
+    require(_to != colonyNetworkAddress, "colony-cannot-target-network");
+    require(_to != tokenLockingAddress, "colony-cannot-target-token-locking");
```
area, line 45: It can't be a call to ERC20 `transfer()` otherwise internal bookkeeping will be messed up (strictly, it could be, if it hasn't been claimed yet, but seems all to easy to grief).

Do we need to prevent it making calls to extensions?

kronosapiens: Great point, that's a good restriction. Extensions don't have any `calledByColony` modifiers, so I don't think that's something we need to worry about.

kronosapiens:
Following up with the conversation @area and I had during sprint planning, we should ask whether we should restrict calls to network-managed extensions. Doing so would require introducing a reverse-mapping from extension address to colony/identifier, something difficult to add later. Currently, no extension uses a `calledByColony` modifier, but we cannot properly guarantee that one might not be added in the future.

It may be a good idea to include this reverse-mapping now, just in case it becomes valuable in the future. It would add a cost of 1-2 words of storage per extension installation, which seems not a deal-breaker.

contracts/colony/Colony.sol
```diff
@@ -37,6 +37,25 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+
+    bytes4 sig;
+    assembly { sig := mload(add(_action, 0x20)) }
+
+    require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
+    require(sig != TRANSFER_FROM_SIG, "colony-cannot-call-erc20-transfer-from");
```
area, line 54: I don't follow why `transferFrom` needs to be restricted? Did you mean `approve`?

kronosapiens: Couldn't a colony call `transferFrom` and pass itself in as an argument, to get around the restriction on `transfer`?

area: Only if it's given itself approval, assuming it's a well behaved ERC20. At the very least, we need to prevent it giving other addresses approval to take from it.

kronosapiens: Ah fair point. I'll add the restriction on `approve`. Part of me likes leaving in the restriction on `transferFrom` for completeness, but I suppose it's not strictly necessary.

area: I don't think it's right. A colony might well want to call `transferFrom` if it is given control over another's tokens.

contracts/colony/Colony.sol
```diff
@@ -37,6 +38,35 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+    require(_to != tokenLockingAddress, "colony-cannot-target-token-locking");
+
+    // Prevent transactions to transfer held tokens
+    bytes4 sig;
+    assembly { sig := mload(add(_action, 0x20)) }
+
+    require(sig != APPROVE_SIG, "colony-cannot-call-erc20-approve");
+    require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
+
+    // Prevent transactions to network-managed extensions
```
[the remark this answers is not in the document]
kronosapiens: I would agree, inasmuch as any sort of `calledByColony` modifier would only consider the extension's base colony.

contracts/extensions/ColonyExtension.sol
```diff
@@ -32,6 +32,7 @@ abstract contract ColonyExtension is DSAuth {
     _;
   }
 
+  function identifier() public pure virtual returns (bytes32);
```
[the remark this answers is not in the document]
kronosapiens: Good point!

contracts/colony/Colony.sol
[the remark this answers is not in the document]
area: So we should say as much in the comment :wink:

test/contracts-network/colony.js
```diff
@@ -118,6 +123,53 @@ contract("Colony", (accounts) => {
...
+
+    it("should not be able to make arbitrary transactions to transfer tokens", async () => {
+      const action1 = await encodeTxData(token, "approve", [USER0, WAD]);
+      const action2 = await encodeTxData(token, "transfer", [USER0, WAD]);
+
+      await checkErrorRevert(colony.makeArbitraryTransaction(token.address, 0, action1), "colony-cannot-call-erc20-approve");
+      await checkErrorRevert(colony.makeArbitraryTransaction(token.address, 0, action2), "colony-cannot-call-erc20-transfer");
+    });
+
+    it("should not be able to make arbitrary transactions to extensions", async () => {
```
area, line 161: There should be a test demonstrating that it can make arbitrary transactions to extensions installed in other colonies, or (equivalently) something that has the `identifier()` function but isn't an extension.

contracts/colony/Colony.sol
```diff
@@ -37,6 +38,35 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return token;
   }
 
+  bytes4 constant APPROVE_SIG = bytes4(keccak256("approve(address,uint256)"));
+  bytes4 constant TRANSFER_SIG = bytes4(keccak256("transfer(address,uint256)"));
+
+  function makeArbitraryTransaction(address _to, uint256 _value, bytes memory _action)
```
area, line 44: `_value` surely must have to be zero, for the same reason that we have to disable `transfer()` on ERC20s?

area: Alternatively, we could assume it has to come out of the pot associated with the root domain, though we could do that with ERC20s as well.

kronosapiens: Mm I suppose you're right. In my mind I wanted to allow for arbitrary transactions which require value transfer, but I see now how that just open the can of worms associated with feeless payouts, which we're not ready to handle yet.

kronosapiens:
@area review meeeeeeeeeeeeee

contracts/colony/Colony.sol
```diff
@@ -37,6 +38,35 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+
+    // Prevent transactions to transfer held tokens
+    bytes4 sig;
+    assembly { sig := mload(add(_action, 0x20)) }
+
+    require(sig != APPROVE_SIG, "colony-cannot-call-erc20-approve");
+    require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
+
+    // Prevent transactions to network-managed extensions installed in this colony
+    try ColonyExtension(_to).identifier() returns (bytes32 extensionId) {
```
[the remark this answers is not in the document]
kronosapiens: Mm initially my reaction was “eh it’s a user supplied parameter why waste the gas” but ultimately I agree it’s The Right Thing To Do

area (approved):

PR #880, opened 2020-10-01
Update setUserRoles to add and remove roles simultaneously

### LINKED ISSUE #878: Be able to simultaenously add and remove roles
It would make product's life much easier if `setUserRoles` were able to set and unset roles at the same time.

kronosapiens:
Closes #878

Removes the `_setTo` argument and updates the user's permissions to match the given `bytes32`.

### DIFF
--- a/contracts/colony/ColonyRoles.sol
+++ b/contracts/colony/ColonyRoles.sol
@@ -86,26 +86,24 @@ contract ColonyRoles is ColonyStorage {
     uint256 _childSkillIndex,
     address _user,
     uint256 _domainId,
-    bytes32 _roles,
-    bool _setTo
+    bytes32 _roles
   ) public stoppable authDomain(_permissionDomainId, _childSkillIndex, _domainId) archSubdomain(_permissionDomainId, _domainId)
   {
     // This is not strictly necessary, since these roles are never used in subdomains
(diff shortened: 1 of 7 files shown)


contracts/colony/ColonyRoles.sol
```diff
@@ -86,26 +86,24 @@ contract ColonyRoles is ColonyStorage {
...
     uint256 _domainId,
-    bytes32 _roles,
-    bool _setTo
+    bytes32 _roles
   ) public stoppable authDomain(_permissionDomainId, _childSkillIndex, _domainId) archSubdomain(_permissionDomainId, _domainId)
   {
     // This is not strictly necessary, since these roles are never used in subdomains
     require(_roles & ROOT_ROLES == 0 || _domainId == 1, "colony-bad-domain-for-role");
-
     require(uint256(_roles) < 2 ** uint256(ColonyRole.NUMBER_OF_ROLES), "colony-roles-do-not-exist");
```
area, line 94: Is this still necessary?

area (commented):
The docs don't appear to have updated?

contracts/colony/ColonyRoles.sol
kronosapiens: No it is not, although I think it's worth keeping the test, which I can rework.

area (approved):

PR #881, opened 2020-10-01
Add reverse mapping for extension installations

kronosapiens:
Add a reverse-mapping allowing the lookup of the `extensionId` for an address, allowing for verification of the identity of an address.

This type of reverse-lookup allows for a wide range of functionality, including:

- ~Letting extensions verify that calls are coming from specific extensions (such as reputation voting vs. hybrid voting), see #875~
- Preventing extensions from making arbitrary transactions, see #876 

In gas terms, this is one additional storage slot per installation, not a big deal.

### DIFF
--- a/contracts/colonyNetwork/ColonyNetworkExtensions.sol
+++ b/contracts/colonyNetwork/ColonyNetworkExtensions.sol
@@ -36,13 +36,13 @@ contract ColonyNetworkExtensions is ColonyNetworkStorage {
     require(_resolver != address(0x0), "colony-network-extension-bad-resolver");
 
     uint256 version = getResolverVersion(_resolver);
-    require(resolvers[_extensionId][version] == address(0x0), "colony-network-extension-already-set");
+    require(extensionResolvers[_extensionId][version] == address(0x0), "colony-network-extension-already-set");
     require(
-      version == 1 || resolvers[_extensionId][version - 1] != address(0x0),
+      version == 1 || extensionResolvers[_extensionId][version - 1] != address(0x0),
       "colony-network-extension-bad-version"
     );
 
-    resolvers[_extensionId][version] = _resolver;
+    extensionResolvers[_extensionId][version] = _resolver;
 
     emit ExtensionAddedToNetwork(_extensionId, version);
   }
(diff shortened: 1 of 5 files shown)


area:
> Letting extensions verify that calls are coming from specific extensions (such as reputation voting vs. hybrid voting),

This is already possible, no? If a contract wants to check something is coming from a particular extension, it can use `getExtensionInstallation`, can't it?

kronosapiens:
@area that's a good point -- if we know the `extensionId` in advance, we can figure out if it's installed. So in that case these two mappings are functionally equivalent.

area:
We could avoid adding this mapping at all by putting `extensionId()` as a callable function on extensions, and upgrading to solidity 0.6 which would allow using `try`/`catch`...

kronosapiens:
Mmm... at first I thought this wouldn't work, since we can't trust extensions to be honest (they could lie about their identifier), but since we would ultimately pass the extension-supplied value in the network mapping to verify the addresses match, I don't see how it could be a problem, something like:

```
bytes32 extensionId = extension.getId();
require(colonyNetwork.getExtensionInstallation(extensionId, address(this)) == msg.sender, "bad-extension");
```

Why would we need try/catch?

area:
Are we able to tell that the arbitrary transaction destination address is an extension?

kronosapiens:
Mm that's fair, we need to run this check during every call, and the check would throw even for valid calls.

kronosapiens:
I think this is ok. The fact that the check is negative (`colonyNetwork.getExtensionInstallation(extensionId, address(this)) != address(extension)`) was initially concerning (since this check will pass for all but one address, ultimately), but as long as we guarantee that all official extensions implement `getId` properly, then it'll be ok:

- The call to `getId` will either fail (for non-extensions, assuming they don't implement `getId`) or return an identifier,
- For all contracts apart from the installed extension with that identifier (the contract we must block), the check will pass.

If a malicious contract implemented `getId` to try and spoof the check (unclear what the benefit would be, but ok), then the address returned by ColonyNetwork would not match, and so the check will pass.

kronosapiens:
This PR is no longer necessary, following #870's introduction of `try/catch`.

PR #882, opened 2020-10-01
Add support for Annotations

### LINKED ISSUE #877: Add justification strings to events
After a discussion w/ @area  and @jakzilla , we like the following approach:

- Add a general purpose `emitJustificationString` event to the Colony contracts, which optionally takes a transaction hash to be justified.
- Develop a `JustifiedAction` extension, which will take arbitrary actions on behalf of users and emit justification strings in a bundled transaction.
- This functionality will exist in parallel with the Motions & Disputes extension, which will also take justification strings (and likely emit them when the action is executed).

kronosapiens:
Closes #877

### Design

This PR adds support for transaction annotations -- arbitrary strings which provide additional metadata about a transaction. These annotations can be plaintext, or some other structure (such as JSON).

Annotations are emitted via the `annotateTransaction` function, which is public:

```
  function annotateTransaction(bytes32 _txHash, string memory _metadata) public stoppable {
    emit Annotation(_txHash, msg.sender, _metadata);
  }
```

In principle, anyone can add an annotation to any transaction, via an additional transaction.

## Open Question

Currently, the ability to emit annotations doesn't distinguish between any sort of type. Specifically, if annotations can be used to emit descriptions of transactions, as well as _comments_ on transactions, how can we distinguish a comment from the initial `msg.sender` and an updating of the description?

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -72,6 +72,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return executeCall(_to, 0, _action);
   }
 
+  function annotateTransaction(bytes32 _txHash, string memory _metadata) public stoppable {
+    emit Annotation(_txHash, msg.sender, _metadata);
+  }
+
   function emitDomainReputationPenalty(
     uint256 _permissionDomainId,
     uint256 _childSkillIndex,
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -73,7 +73,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   }
 
   function annotateTransaction(bytes32 _txHash, string memory _metadata) public stoppable {
-    emit Annotation(_txHash, msg.sender, _metadata);
+    emit Annotation(msg.sender, _txHash, _metadata);
   }
 
   function emitDomainReputationPenalty(
--- a/contracts/colony/ColonyDataTypes.sol
+++ b/contracts/colony/ColonyDataTypes.sol
@@ -198,6 +198,12 @@ interface ColonyDataTypes {
   /// @param fundingPotId Id of the newly-created FundingPot
   event FundingPotAdded(uint256 fundingPotId);
 
+  /// @notice Emit a metadata string for a transaction
+  /// @param txHash Hash of transaction being annotated (0x0 for current tx)
+  /// @param user User emitting the annotation
+  /// @param metadata String of metadata for tx
+  event Annotation(bytes32 indexed txHash, address indexed user, string metadata);
+
   struct RewardPayoutCycle {
     // Reputation root hash at the time of reward payout creation
     bytes32 reputationState;
--- a/contracts/colony/ColonyDataTypes.sol
+++ b/contracts/colony/ColonyDataTypes.sol
@@ -199,10 +199,10 @@ interface ColonyDataTypes {
   event FundingPotAdded(uint256 fundingPotId);
 
   /// @notice Emit a metadata string for a transaction
+  /// @param agent Agent emitting the annotation
   /// @param txHash Hash of transaction being annotated (0x0 for current tx)
-  /// @param user User emitting the annotation
   /// @param metadata String of metadata for tx
-  event Annotation(bytes32 indexed txHash, address indexed user, string metadata);
+  event Annotation(address indexed agent, bytes32 indexed txHash, string metadata);
 
   struct RewardPayoutCycle {
     // Reputation root hash at the time of reward payout creation
--- a/contracts/colony/IColony.sol
+++ b/contracts/colony/IColony.sol
@@ -66,6 +66,11 @@ interface IColony is ColonyDataTypes, IRecovery {
   /// @return success Boolean indicating whether the transaction succeeded
   function makeArbitraryTransaction(address _to, bytes memory _action) external returns (bool success);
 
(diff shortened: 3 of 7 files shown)


area:
I'd like to not be so opinionated as 'justification' - can we call it 'metadata' or something similarly general?

kronosapiens:
@area works for me. How about we call it `annotateTransaction` and call it a `note`? "Metadata" is a little too generic IMO, since it's ultimately meant to be a human-readable string, not _really_ arbitrary metadata.

area:
I mean, we're planning on putting a human-readable string in it, but there's no guarantee, with no validation or similar, that someone else wouldn't put is a JSON blob or something even less human-readable.

kronosapiens:
Ok, we can call it `_metadata` if you prefer. I still like `annotateTransaction` as the function name though.

area:
That's fine by me.

I'm starting to wonder if we need more though, like a shorter 'type' field so that we can filter types of annotations, rather than just get all annotations that are (claimed to be) associated with a transaction.

There's also the fact that this function is public, but presumably we want to permission some annotations. At first glance, it seems like we would want to use this functionality for e.g. naming domains, but obviously that type of annotation shouldn't be public. Are we handing that permissioning off to the client side? That presumably means clients require access to an archive node, which we've tried to steer clear of in the past... I'm not sure this approach is the 'be and and end all' that we thought it was.

kronosapiens:
I hear that, but I don't think this needs to be the _only_ entry point for emitting metadata strings, just the most public / general one. A separate, permissioned `setDomainName` function for renaming domains, for instance, is still appropriate.

Regarding a `type` field... I could see the value, but since there's no validation it's hard to guarantee proper typing anyway.

area:
Well, if we had a separate type field, some types could be forbidden to be emitted via this public function if we're also adding extra, permissioned versions of this function.

kronosapiens:
I suppose, but that could get complicated fast. For instance, if we want to use this to update a domain's description, we'd have to find some way to pass the domainId. If all we have are `type` and `metadata` arguments, then we'd need to include both the domainId and the description in some custom format, and then parse it in the frontend. It's not clear why this is better than having separate functions for specific purposes, which allows for more straightforward permissioning and a much easier coding of arguments.

area:
> then we'd need to include both the domainId and the description in some custom format, and then parse it in the frontend
Yep, this is exactly what I have in mind
>  It's not clear why this is better than having separate functions for specific purposes, which allows for more straightforward permissioning and a much easier coding of arguments.
Because this is general-purpose, would be the argument.

kronosapiens:
I mean, why not just have a single function called `executeColonyTx` which takes a function signature as an argument and is ultimately just a giant conditional? We already have machinery for parsing arguments and authenticating functions, so it doesn't really make sense why we would throw that out the window and duplicate it for the sake of "generality". We add functions basically whenever we want for many reasons, so I'm not sure why in this case there's any value in doing something different.

area:
Just want to check I understand where you're coming from:

1. We should have a general purpose, permissionless function and corresponding event for adding annotations.
2. Where we want to permission who can emit events like this (containing human readable content, to some degree), we create a new function with a new event?

kronosapiens:
Broadly yes. We have a general-purpose, public function for adding comments to transactions. If we need more specific functionality (such as renaming domains, which are not a transaction, and is privileged functionality), we create a function to do so. I see no benefit to collapsing all labeling actions into a single function, as different labeling actions have different subjects and different permissioning, distinctions which are well-handled through separate functions and poorly handled through ad-hoc permissioning and string parsing.

kronosapiens:
As per @cobordism 's comment in the team meeting, I did some research into adding annotation strings directly into the call data (see this [stackoverflow](https://ethereum.stackexchange.com/questions/48095/assign-metadata-to-a-transaction) issue). Seems as though appending annotations to a transaction's `data` field won't mess up function calls, so the question becomes the feasibility of parsing out these strings after-the-fact. TBD.

area:
After talking over several things with Raul and Jack this morning, I have realised that there is a problem with that approach as it makes 'permissioning' the annotations tricky for the app (i.e. in the future, if the same address annotates a transaction, show that one instead of the 'same-transaction' annotation. In short, the only address that the app will have access to on that 'original' transaction is `tx.origin`, but all subsequent annotations done in the same way will be emitted by `msg.sender`. I don't particularly want to start using `tx.origin` in the extension either...

kronosapiens:
@area I hear that, but I would just as easily go ahead and say that annotations are now simply bound to the tx. Adding annotations after-the-fact was never really a part of the spec, it was just something we could do if annotations were done via events. But there's no reason we can't walk that back and say annotations are now linked to their tx only.

area:
Having checked with Jack, it is his (and my own) recollection that being able to update annotations was part of the original desired functionality, so I don't think saying that annotations are now only linked to their transactions is really a runner :sweat_smile:

contracts/colony/Colony.sol
```diff
@@ -72,6 +72,10 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     return executeCall(_to, 0, _action);
   }
 
+  function annotateTransaction(bytes32 _txHash, string memory _metadata) public stoppable {
+    emit Annotation(_txHash, msg.sender, _metadata);
```
area, line 76: As discussed, these arguments should be reordered (msg.sender coming first, and named agent in anticipation of #911

area (approved):

PR #884, opened 2020-10-16
Make extension manager development easier for app

area:
After a call with Chris about developing the extension manager, a couple of minor things came up that will help him. One is to install the extensions when the migrations are run, and the other is a minor functional change, making the `deprecated` boolean available via `getDeprecated`.


kronosapiens:
Could we not just make the `deprecated` boolean public? And get it via `deprecated()`?

area:
We've avoided doing that to date with anything that's an `EtherRouter`. However, in theory, the getter is there if we made it public so if we just added it to the interface, it _should_ work.

kronosapiens:
Mm, I'd say let's just stick with our convention.

kronosapiens (approved):

kronosapiens (approved):

PR #891, opened 2020-11-02
Add ExtensionInitialised events to extensions

kronosapiens:


contracts/extensions/ColonyExtension.sol
```diff
@@ -22,6 +22,9 @@ import "./../common/EtherRouter.sol";
 import "./../colony/IColony.sol";
 
 abstract contract ColonyExtension is DSAuth {
+
+  event ExtensionInitialised(address colony);
```
area, line 26: Do you not think that this should have `colony` indexed? If not then I'm not sure what the point of this argument is.

kronosapiens: Yup, agreed. I threw this up quickly to Chris could work off of it.

area: Thinking about this some more, I don't think this argument is actually useful, and putting it here would encourage behaviour that could be exploited down the line. 

The reason we're adding this event is to provide a way for a client to tell whether an extension that needs initialisation, has been. It already knows the extension contract address, which it got from the colony, so it already knows the colony address.

Adding this encourages someone to implement 'give me all extensions that have been initialised in a colony'. That sounds good and like something that might be useful, but the problem with *that* is that any contract can emit this event, and so potentially could DDOS the dapp client by emitting lots of these events in irrelevant contracts.

kronosapiens: Mm that's a fair point -- you're saying this event is not really meant to be the first source of information about extensions, but rather the last (once you know the extension exists, you simply look for this event coming from it). If you don't know the extension exists, you shouldn't be able to scan for it, since anyone can deploy their own extensions and emit this event.

area (approved):

PR #893, opened 2020-11-03
Order addresses by total reputation in oracle endpoint

area:
Previously, the addresses were in... some unknown order. The UI would like to be able to display addresses in order of reputation, so let's sort them here rather than having the app request each reputation in turn.


rdig:
Yay! :100:

kronosapiens (approved):

PR #895, opened 2020-11-10
Add missing view to getDeprecated to indicate it is read-only

### LINKED ISSUE #894: getDeprecated not marked as a view function

area:
Closes #894

### DIFF
--- a/contracts/extensions/ColonyExtension.sol
+++ b/contracts/extensions/ColonyExtension.sol
@@ -42,7 +42,7 @@ abstract contract ColonyExtension is DSAuth {
   function deprecate(bool _deprecated) public virtual;
   function uninstall() public virtual;
 
-  function getDeprecated() public returns (bool deprecated) {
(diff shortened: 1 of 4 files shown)


contracts/extensions/ColonyExtension.sol
```diff
@@ -42,7 +42,7 @@ abstract contract ColonyExtension is DSAuth {
   function deprecate(bool _deprecated) public virtual;
   function uninstall() public virtual;
 
-  function getDeprecated() public returns (bool deprecated) {
+  function getDeprecated() public view returns (bool) {
```
area, line 45: The fact that we had to remove calling the return variable `deprecated` here took me a long time to find. Just highlighting it as something we fell in to and should keep an eye out to avoid in the future. Really, the takeaway here is to always test our functions, no matter how simple they look :smile:

kronosapiens (approved):

PR #901, opened 2020-11-19
Add needed events to recovery mode; Improve event tests

area:
Based on the designs for the recovery mode screens, we need these events, so I've added them.

While writing the tests it also became apparent that we were not rigorously testing our events. So I've improved `expectEvent` to check the parameters passed by the event as well, and updated all relevant tests (we probably need more, but not in this PR).

### DIFF
--- a/contracts/common/ContractRecovery.sol
+++ b/contracts/common/ContractRecovery.sol
@@ -90,6 +90,7 @@ contract ContractRecovery is ContractRecoveryDataTypes, CommonStorage { // ignor
     uint numRequired = totalAuthorized / 2 + 1;
     require(recoveryApprovalCount >= numRequired, "colony-recovery-exit-insufficient-approvals");
     recoveryMode = false;
+
     emit RecoveryModeExited(msg.sender);
   }
 
--- a/contracts/common/ContractRecovery.sol
+++ b/contracts/common/ContractRecovery.sol
@@ -44,7 +44,9 @@ contract ContractRecovery is ContractRecoveryDataTypes, CommonStorage { // ignor
     // Make recovery edit
     uint x = _slot;
     bytes32 y = _value;
+    bytes32 oldValue;
     assembly {
+      oldValue := sload(x)
       sstore(x, y) // ignore-swc-124
     }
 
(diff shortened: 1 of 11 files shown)


contracts/common/ContractRecovery.sol
```diff
@@ -82,6 +88,7 @@ contract ContractRecovery is ContractRecoveryDataTypes, CommonStorage { // ignor
     uint numRequired = totalAuthorized / 2 + 1;
     require(recoveryApprovalCount >= numRequired, "colony-recovery-exit-insufficient-approvals");
     recoveryMode = false;
+    emit RecoveryModeExited();
```
kronosapiens: Can we add a newline above as per convention?

contracts/common/ContractRecovery.sol
```diff
@@ -68,7 +70,7 @@ contract ContractRecovery is ContractRecoveryDataTypes, CommonStorage { // ignor
     recoveryApprovalCount = 0;
     recoveryEditedTimestamp = block.timestamp;
 
-    emit RecoveryModeEntered();
+    emit RecoveryModeEntered(msg.sender);
```
kronosapiens, line 73: Isn't this information already available on the tx?

helpers/test-helper.js
```diff
@@ -251,9 +251,21 @@ export async function getBlockTime(blockNumber) {
   return p;
 }
 
-export async function expectEvent(tx, eventName) {
+function hexlifyAndPad(input) {
```
kronosapiens, line 254: `hexify`?

test/contracts-network/colony-task-work-rating.js
```diff
@@ -354,7 +354,7 @@ contract("Colony Task Work Rating", (accounts) => {
       );
     });
 
-    it("should log a TaskWorkRatingRevealed event", async () => {
+    it.only("should log a TaskWorkRatingRevealed event", async () => {
```
kronosapiens, line 357: 🚨 🚨 🚨

contracts/common/ContractRecovery.sol
area: I originally didn't have this, but ended up including it - `tx.origin` is available, but if the permission holder is a contract, it is not easily available.

helpers/test-helper.js
area: It's called `hexlify` in ethers.

test/contracts-network/colony-task-work-rating.js
area: :sob:

kronosapiens (approved):

PR #902, opened 2020-11-23
Bump colony contract version number

area:
With more integration with the app team happening, this has now caught out multiple people, so I'm proposing we move to incrementing the version number immediately after a release. This should mean app people are more easily to just pull from our `develop` to see how things are going.

I recall a conversation I had with Chris a long time ago about this, and was on the side of 'only update immediately before a release', which was where we settled. Perhaps it's time for a change though. Thoughts?


kronosapiens:
I would agree with this -- the develop version number an be seen as indicating "version in progress", as any differences between develop and the previous release are part of the new version.

kronosapiens (approved):

PR #904, opened 2020-11-24
Add payoutWhitelist to ColonyNetwork

### LINKED ISSUE #903: Do not auction stable coins
## Requirements
Whitelist is managed with the same permissions as Network Fee
Whitelist applies to all colonies when it is updated, like the network fee.
Whitelisted tokens go straight to the metacolony and are not auctioned.

## Starting list of tokens:

ETH, DAI, USDT, USDC

This should be respected on both chains, with the bridged versions of the above on xDAI having the same behaviour. This seems most likely to be achieved by a different configuration on each side, rather than the contracts being network-aware.

The most natural implementation in my eyes involves a callout to ColonyNetwork around [this line](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/colony/ColonyFunding.sol#L569) to check the whitelist, but open to other suggestions.

kronosapiens:
Closes #903

Adds a `payoutWhitelist` mapping to the Colony Network, configurable from the Metacolony. Used to determine if a token fee should be sent to auction or directly to the Metacolony.

The intention is to whitelist stablecoins, but who knows!

### DIFF
--- a/contracts/colony/ColonyAuthority.sol
+++ b/contracts/colony/ColonyAuthority.sol
@@ -86,6 +86,7 @@ contract ColonyAuthority is CommonAuthority {
     addRoleCapability(ARBITRATION_ROLE, "setExpenditureClaimDelay(uint256,uint256,uint256,uint256,uint256)");
 
     // Added in colony v5 (cerulean-lwss)
+    addRoleCapability(ROOT_ROLE, "setPayoutWhitelist(address,bool)");
     addRoleCapability(ROOT_ROLE, "mintTokensFor(address,uint256)");
     addRoleCapability(ROOT_ROLE, "setAnnualMetaColonyStipend(uint256)");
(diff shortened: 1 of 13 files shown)


area (commented):
Two very minor things so far, otherwise looking just how I imagined it!

contracts/colony/ColonyAuthority.sol
```diff
@@ -86,6 +86,7 @@ contract ColonyAuthority is CommonAuthority {
     addRoleCapability(ARBITRATION_ROLE, "setExpenditureClaimDelay(uint256,uint256,uint256,uint256,uint256)");
 
     // Added in colony v5 (cerulean-lwss)
+    addRoleCapability(ROOT_ROLE, "setPayoutWhitelist(address,bool)");
```
area, line 89: Needs to be added to `finishUpgrade` in `Colony.sol` as well.

area (approved):

PR #905, opened 2020-11-26
Add further events and functions required by app

area:
Work on the Universal Action Creator is to begin soon, so Raul and I had a call to figure out what the contracts have to change or add in order for that to work. Some of that work is taking place in #882, and this PR is intended to contain the remainder. The resulting additional work can broadly be split in to two categories.

## Allowing the app to identify actions

The type of screens we're thinking of look like this (once an action has been created):

screenshot-2020-11-26_15-39-41

The example here is moving funds, but in principle this applies to all actions in the UAC. This type of screen will be navigated to via a path that involves the transaction id. The app will get the transaction in question, and attempt to work out what is going on in it. There are four possibilities 

1. The transaction doesn't exist
2. The transaction exists, but isn't mined yet
3. The transaction is mined, but doesn't do anything the app recognises
4. The transaction is mined, and does something we recognise

We're are purely talking about possibility 4 here, and how the app 'recognises' what is going on.

The natural starting point is looking at the `data` field of the transaction and examine the function being called (the first four bytes of the `data` field). If the function signature is one we recognise, great! Otherwise we fall in the category 3 above. Unfortunately, this is quite a fragile approach. It would require a much more detailed examination of the transaction data if it were performed through the annotation extension in one transaction (see #882), and would also be unable to recognise things done through other multisigs unless they were explicitly added. This is because in both of these cases, the function being called would not be `moveFundsBetweenPots`, but `executeMultisigTransaction` or similar, requiring a much more involved examination of the transaction. Not wishing to put that burden on the app team, we discounted this approach unless it turns out absolutely necessary.

A more robust approach is to use events, which the app team are already fairly comfortable with querying and interpreting. If that transaction contains an event that corresponds to an action we recognise, the app can display the relevant screen and interpret any other events that are present appropriately. This means that every one-transaction action has to have a unique event, and every multi-transaction action down the line (such as the full task flow) will have to have a unique event in the first transaction that starts that flow. Subsequent transactions that interact with that flow will have to be able to be identified via events as well, but we have no such examples in the UAC work for now.

To this end, I add the events 

* `OneTxPaymentMade` 
* `TokensMinted`

And for any future functionality / extensions, we should keep this 'first transaction must contain a unique event' requirement in mind.

~~_I have not yet checked other existing extensions, but I would probably do so in this PR down the line if/when we've settled on this approach_~~ Now done

_A different approach to identifying proposals of actions will be required, despite the very similar interface. I have a plan for that though, and does not need to be worried about here. Essentially, we will emit the function signature of whatever is being proposed as an event in the motions and disputes contract._

## Allowing the app to rely less on a backend

The UAC also contains some new actions that require new functions. These are based around 'metadata' being added to colonies and domains (containing names, colours, a list of 'active' tokens, and so on), stored on-chain in events in order to remove some reliance on the backend. We propose that the contracts are as agnostic as possible, and just accept IPFS hashes of whatever the app team deems needed to be in them (likely JSON blobs), and validating and consuming them falls on the app team, as does handling failure cases where the corresponding data cannot be retrieved from IPFS. Daniel and I loosely agreed to adding relevant permissioned functions last time we discussed it in #882, though not their contents.

To this end, I have added extra `metadata` parameters to `createColony` and `createDomain` with corresponding events, and the new functions `editColony` and `editDomain` that only emit the same events, used to update the details.

We currently propose that the list of tokens relevant to the colony is stored in this metadata along with the colony name etc, with root permission allowed to edit both the list of tokens and the colony description etc. If that separation of permissions is incorrect, now is the time to let us know! Similarly, I have assumed that the permissions required to edit the domain information are the same permissions required to create the domain in the first place. If that is incorrect, similarly, this is the time to let us know!

No tests yet, pending agreement between all parties on the above, but they will exist eventually, whatever we decide.

Finally, the app team is starting the UAC work in around a week, so we should try and have this hashed out by then otherwise we risk blocking them.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -297,7 +297,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     // Add domain to local mapping
     initialiseDomain(newLocalSkill);
 
-    emit DomainMetadata(domainCount, _metadata);
+    if (keccak256(abi.encodePacked(_metadata)) != keccak256(abi.encodePacked(""))) {
+      emit DomainMetadata(domainCount, _metadata);
+    }
   }
 
   function editDomain(uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _domainId, string memory _metadata) public
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -307,7 +307,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   stoppable
   authDomain(_permissionDomainId, _childSkillIndex, _domainId)
   {
-    emit DomainMetadata(_domainId, _metadata);
+    if (keccak256(abi.encodePacked(_metadata)) != keccak256(abi.encodePacked(""))) {
+      emit DomainMetadata(_domainId, _metadata);
+    }
   }
 
   function getDomain(uint256 _id) public view returns (Domain memory domain) {
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -128,6 +128,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
   function initialiseColony(address _colonyNetworkAddress, address _token, string memory _metadata) public stoppable {
     initialiseColony(_colonyNetworkAddress, _token);
+
     emit ColonyMetadata(_metadata);
   }
 
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -126,6 +126,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit ColonyInitialised(_colonyNetworkAddress, _token);
   }
 
+  function initialiseColony(address _colonyNetworkAddress, address _token, string memory _metadata) public stoppable {
+    initialiseColony(_colonyNetworkAddress, _token);
+    emit ColonyMetadata(_metadata);
+  }
+
+  function editColony(string memory _metadata) public
+  stoppable
+  auth {
+    emit ColonyMetadata(_metadata);
+  }
+
   function bootstrapColony(address[] memory _users, int[] memory _amounts) public
   stoppable
   auth
@@ -151,6 +162,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   auth
   {
     ERC20Extended(token).mint(address(this), _wad); // ignore-swc-107
+
+    emit TokensMinted(address(this), _wad);
   }
 
   function mintTokensFor(address _guy, uint _wad) public
@@ -158,6 +171,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
   auth
   {
     ERC20Extended(token).mint(_guy, _wad); // ignore-swc-107
+
+    emit TokensMinted(_guy, _wad);
   }
 
   function mintTokensForColonyNetwork(uint _wad) public stoppable {
@@ -167,6 +182,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     require(address(this) == IColonyNetwork(colonyNetworkAddress).getMetaColony(), "colony-access-denied-only-meta-colony-allowed");
     ERC20Extended(token).mint(_wad);
     assert(ERC20Extended(token).transfer(colonyNetworkAddress, _wad));
+
+    emit TokensMinted(colonyNetworkAddress, _wad);
   }
 
   function registerColonyLabel(string memory colonyName, string memory orbitdb) public stoppable auth {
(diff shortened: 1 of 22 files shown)


area:
Additional thoughts as they come to me:

* Should we be clever with IPFS hashes?

contracts/colony/Colony.sol
```diff
@@ -271,6 +295,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+    if (keccak256(abi.encodePacked(_metadata)) != keccak256(abi.encodePacked(""))) {
+      emit DomainMetadata(domainCount, _metadata);
+    }
+  }
+
+  function editDomain(uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _domainId, string memory _metadata) public
+  stoppable
+  authDomain(_permissionDomainId, _childSkillIndex, _domainId)
+  {
+    emit DomainMetadata(_domainId, _metadata);
```
kronosapiens, line 308: For instance, what if they pass `""` here?

contracts/colony/IColony.sol
```diff
@@ -189,6 +189,11 @@ interface IColony is ColonyDataTypes, IRecovery {
   /// @param _token Address of the colony ERC20 Token
   function initialiseColony(address _colonyNetworkAddress, address _token) external;
 
+  /// @notice Called to change the metadat associated with a colony. Expected to be a IPFS hash of a
```
kronosapiens, line 192: `metadata`

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -163,7 +173,12 @@ contract ColonyNetwork is ColonyNetworkStorage {
       IColony(colonyAddress).registerColonyLabel(_colonyName, "");
     }
 
+    if (keccak256(abi.encodePacked(_metadata)) != keccak256(abi.encodePacked(""))) {
```
kronosapiens, line 176: Ditto, this feels excessive.

contracts/extensions/OneTxPayment.sol
```diff
@@ -24,6 +24,8 @@ import "./ColonyExtension.sol";
 
 
 contract OneTxPayment is ColonyExtension {
+  event OneTxPaymentMade();
```
kronosapiens, line 27: No arguments?

helpers/test-helper.js
```diff
@@ -260,13 +260,46 @@ function hexlifyAndPad(input) {
   return ethers.utils.hexZeroPad(ethers.utils.hexlify(i), 32);
 }
 
-export async function expectEvent(tx, eventName, args) {
-  const { logs } = await tx;
-  const event = logs.find((e) => e.event === eventName);
+export async function expectEvent(tx, nameOrSig, args) {
+  const re = /\((.*)\)/;
+  let event;
+  if (nameOrSig.match(re)) {
```
kronosapiens, line 266: Why is this being added?

contracts/colony/Colony.sol
```diff
@@ -126,6 +126,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     emit ColonyInitialised(_colonyNetworkAddress, _token);
   }
 
+  function initialiseColony(address _colonyNetworkAddress, address _token, string memory _metadata) public stoppable {
```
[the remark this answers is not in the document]
area: You're not wrong, but that's not what the app is expecting to be able to do, unfortunately.

contracts/colony/Colony.sol
```diff
@@ -271,6 +295,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
     // Add domain to local mapping
     initialiseDomain(newLocalSkill);
+
+    if (keccak256(abi.encodePacked(_metadata)) != keccak256(abi.encodePacked(""))) {
```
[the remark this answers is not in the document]
area: Partly philosophical (if someone uses the no-metadata function call on chain, they shouldn't see an event emitted that they don't care about) and partly practical (I believe this is cheaper than just emitting the event every time).

contracts/colony/Colony.sol
area: Fair, I'll put the check here too

contracts/colonyNetwork/ColonyNetwork.sol
area: Ditto :stuck_out_tongue:

contracts/extensions/OneTxPayment.sol
area: What arguments would you expect here? Everything relevant is already emitted in events during the transfer, I believe. This is just indicating that the extension was used for the payment rather than it being 'native'.

contracts/colony/Colony.sol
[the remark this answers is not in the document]
kronosapiens: Alright.

contracts/extensions/OneTxPayment.sol
kronosapiens: Yeah that's true actually -- funding, expenditure, etc already do their events.

helpers/test-helper.js
area: In the case of events being emitted in a transaction but not in the interface of the contract being called (as is done here for `createColony` on `colonyNetwork`, but checking to see `ColonyMetadata` has been emitted during the transaction), this additional functionality means we can specify the full signature of the event rather than just the name to check it has been emitted.

This specific line is checking for `()` in `nameOrSig` to work out if the full signature or just the name has been passed.

contracts/colony/Colony.sol
kronosapiens: But, what if they want to delete the metadata?

area: Then they should be passing the IPFS hash of the empty JSON object.

area: Actually, I'm not sure I should put the check here. There's no reason to call this function if you're not emitting metadata, which is the case for the functions where I have the check.

kronosapiens: I don't feel so strongly I suppose, I'll just say that if they can pass the empty string, then the app must be able to handle the empty string, which makes me less worried about emitting it in the other cases. I see the argument for gas saving, but it makes the code more complex... it's whatever you like I guess.

area: I came down on the side of adding the guard here, for consistency more than anything else. You're absolutely right that the app has to sanitize the inputs here. I certainly expect a `validIPFSHash` or similar to appear in the app, alongside other validation functions though, which would be able to easily handle the special case of the empty string being passed.

kronosapiens (approved):

PR #907, opened 2020-12-03
Revisit TokenSupplier permissions

area:
(PR already on this branch installs `TokenSupplier` on migration, which will be useful for the app team down the line)

We need to radically revisit the permissions on TokenSupplier. I started looking at this after a talk with Jack, with the intention of adding what I thought was going to be a simple change, but the more I fiddled the more I realised we were way off base here. After talking with Jack, going forward for all extensions, we need to make sure that all meaningful interactions with an extension are possible with just a permission so that the app can use them (at least until the voting contracts are implemented). That is not the case here - even if you are root, if the request is coming from a not-contract (i.e. an externally owned account), then the `setTokenSupplyCeiling` and `setTokenIssuanceRate` functions are inaccessible.

So, we could (re)move the check for whether the sender is a contract, I thought. But then it seems weird that a root external account can do stuff, but proxying through an identity contract or a multisig would be disallowed, even if you had root permissions. So maybe we allow non-voting contracts that have the root permission to change these variables. But then the voting contracts could give another contract root permissions, and proxy their requests through that, bypassing the restrictions this extension is intended to enforce. 

More fundamentally, it is legitimate to claim that as-is, this contract does not practically enforce _any_ restrictions! If the voting contracts have root permissions, they can mint any amount of tokens whenever they want. So we need to key their permissions off of something else.

So, yeah, in conclusion, we need to take another swing at the permissions here. Which isn't terribly surprising, given this was the first time we tried something like this :smile:


kronosapiens:
Why don't we remove the contract check for now and reintroduce it once those contracts are ready? I don't understand the issue where multisigs couldn't be allowed -- just give the multisig root permission.

area:
> Why don't we remove the contract check for now and reintroduce it once those contracts are ready? I don't understand the issue where multisigs couldn't be allowed -- just give the multisig root permission.

So you're proposing:

* Remove the `onlyContract` modifier
* Remove the failure case where a contract calling the function has no extension identifier

I mention the issue with the second element of this in the PR description - if contracts with root permissions are allowed to call these functions, then the voting contracts could give a proxy contract the permissions, and then make calls through that.

In either case, the biggest issue here still remains, where the voting contracts with root permissions can call these functions directly anyway (i.e. not through tokenSupplier).

kronosapiens:
After a discussion with @area, I'd like to note that _fundamentally_ I think the issue at hand is that reputation votes intrinsically represent a _subset_ of the full `root` permission. As such, our current treatment of giving reputation votes unfettered root power is a design flaw and should be addressed, perhaps by introducing function-level restrictions, similar to how we do it in `TokenSupplier`.

kronosapiens:
After spending some more time on this (and talking to Jack a bit about the product vision), here are some notes:

- Jack would prefer _not_ to bake "policy" decisions into the contracts, meaning that the contracts should not circumscribe use-cases, even if we think they aren't a great idea. In this case, managing tokens via a reputation vote would be a use case we personally wouldn't advise but we shouldn't prevent users from doing so if they feel strongly.
- Our proposed direction of putting in extension-level restrictions into our  contracts is a new and introduces some risk, specifically around the assumption of who is developing these extensions. The restrictions are a bit brittle in that they require the "official' extensions to stick to a convention. It is feasible that simple human error would lead to these restrictions being broken.
- Ultimately, the issue stems from the fact that up until this point, we have been giving reputation voting the `root` permission; it is becoming clear that reputation voting in fact should not be able to make root-level decisions.

My suggestion is that rather than include extension-level restrictions into our contracts, we should reframe reputation voting as non-root functionality, used for making domain-level (and thus non-root) decisions: managing finances and payments, arbitration, and so on _in domains_. For colony-wide (and thus root-level) decisions, a hybrid vote should be preferred. The nice thing about this approach is that we can achieve it without explicitly restricting extensions in the contracts -- we simply do not give reputation voting the root permission.

Here is a list of all specifically root-permissioned functions in colony (some functionality shared with architecture is excluded):

```
    addRoleCapability(ROOT_ROLE, "setRootRole(address,bool)");
    addRoleCapability(ROOT_ROLE, "setRecoveryRole(address)");
    addRoleCapability(ROOT_ROLE, "removeRecoveryRole(address)");

    addRoleCapability(ROOT_ROLE, "startNextRewardPayout(address,bytes,bytes,uint256,bytes32[])");
    addRoleCapability(ROOT_ROLE, "bootstrapColony(address[],int256[])");
    addRoleCapability(ROOT_ROLE, "registerColonyLabel(string,string)");
    addRoleCapability(ROOT_ROLE, "setRewardInverse(uint256)");
    addRoleCapability(ROOT_ROLE, "mintTokens(uint256)");
    addRoleCapability(ROOT_ROLE, "upgrade(uint256)");

    // Metacolony only
    addRoleCapability(ROOT_ROLE, "addNetworkColonyVersion(uint256,address)");
    addRoleCapability(ROOT_ROLE, "setNetworkFeeInverse(uint256)");
    addRoleCapability(ROOT_ROLE, "addGlobalSkill()");
    addRoleCapability(ROOT_ROLE, "deprecateGlobalSkill(uint256)");

    addRoleCapability(ROOT_ROLE, "updateColonyOrbitDB(string)");

    addRoleCapability(ROOT_ROLE, "setPayoutWhitelist(address,bool)");
    addRoleCapability(ROOT_ROLE, "mintTokensFor(address,uint256)");
    addRoleCapability(ROOT_ROLE, "setAnnualMetaColonyStipend(uint256)");
    addRoleCapability(ROOT_ROLE, "setReputationMiningCycleReward(uint256)");
    addRoleCapability(ROOT_ROLE, "addExtensionToNetwork(bytes32,address)");
    addRoleCapability(ROOT_ROLE, "installExtension(bytes32,uint256)");
    addRoleCapability(ROOT_ROLE, "upgradeExtension(bytes32,uint256)");
    addRoleCapability(ROOT_ROLE, "deprecateExtension(bytes32,bool)");
    addRoleCapability(ROOT_ROLE, "uninstallExtension(bytes32)");
    addRoleCapability(ROOT_ROLE, "makeArbitraryTransaction(address,bytes)");
    addRoleCapability(ROOT_ROLE, "editColony(string)");
    addRoleCapability(ROOT_ROLE, "burnTokens(address,uint256)");
```

This root-specific functionality pertains to managing the entire colony (upgrades, configuration, reward rates, extensions, managing tokens, making arbitrary tx, and so on). It is reasonable to say that these decisions require a hybrid vote, while non-root decisions can be made by reputation holders only. This approach is easily implemented given our current design, which is very convenient.

contracts/colonyNetwork/ColonyNetworkStorage.sol
```diff
@@ -89,16 +89,14 @@ contract ColonyNetworkStorage is CommonStorage, ColonyNetworkDataTypes, DSMath {
   mapping (address => uint256) pendingMiningRewards; // Storage slot 34
 
   uint256 totalMinerRewardPerCycle; // Storage slot 35
-  uint256 annualMetaColonyStipend; // Storage slot 36
-  uint256 lastMetaColonyStipendIssued; // Storage slot 37
```
area, line 93: We need to leave these slots in, unfortunately, because this has already been deployed to goerli. We can repurpose them in the future though.

test/contracts-network/meta-colony.js
```diff
@@ -565,48 +564,6 @@ contract("Meta Colony", (accounts) => {
...
-      await metaColony.claimColonyFunds(clnyToken.address);
-      potAfter = await metaColony.getFundingPotBalance(1, clnyToken.address);
-      expect(potAfter.sub(potBefore)).to.eq.BN(WAD.divn(365 * 24 * 60));
-    });
-
-    it("the metacolony stipend cannot be claimed before it is set", async () => {
-      await checkErrorRevert(colonyNetwork.issueMetaColonyStipend(), "colony-network-metacolony-stipend-not-set");
-    });
-  });
-
```
area, line 609: I was expecting to see these tests rewritten so that they used the extension, rather than just being deleted outright :cry:

area (commented):
A lovely amount of code removed :smiling_face_with_three_hearts:

test/contracts-network/meta-colony.js
kronosapiens: I don't think thats necessary -- the functionality is tested in the token-supplier tests, and there's nothing fundamentally unique about using the extension in the metaColony. The tests would be logically redundant.

contracts/extensions/TokenSupplier.sol
```diff
@@ -111,15 +96,8 @@ contract TokenSupplier is ColonyExtension {
 
   /// @notice Update the tokenSupplyCeiling, cannot set below current tokenSupply
```
area, line 97: Has preventing it being set below tokenSupply been lost somewhere? Does this matter?

test/contracts-network/meta-colony.js
area: My worry here was around subtle differences in functionality or permissions that we would want to be double checking that we were happy with. After looking at this for a while, the only thing I can see is that we have to declare an upper limit for the supply, which I guess is fine.

contracts/extensions/TokenSupplier.sol
kronosapiens: I don't think it matters, really. Tokens can be minted at any time for any reason, so the totalSupply may always become spontaneously larger than the ceiling. But for sanity's sake maybe best to prevent users from doing it here.

kronosapiens (approved):
Wow great PR whoever made this contract changes is an incredible programmer

PR #908, opened 2020-12-03
Remove pendingBalance from TokenLocking

kronosapiens:
Being able to "claim" pending tokens in TokenLocking introduces an unnecessary complexity to the UI. This PR removes this concept, storing tokens with the user directly until deposited.


area:
This PR will also need to accommodate anyone who has a `pendingBalance` when the upgrade is done.

kronosapiens:
@area actually not, since this is part of the LWSS release. `pendingBalance` hasn't actually been deployed yet.

area:
It's been deployed to both Goerli and XDai

kronosapiens:
@area Interesting! I've added a solution which I think is pretty decent.

area (commented):
I have stopped this review, because I think I am missing a lot of information. I've a comment that is relevant (I believe) regardless, but to do I proper review I think I need to know more. I know you've been having difficulty with it, but you've squashed all the commits together, so I can't see the particular subtle problems you encountered, or how the thought process evolved as a result to solve them, and to my unfamiliar eye it just looks like there are a lot of unexpected changes.

The biggest question that needs resolving for me: what was the actual specification here? When I came in to review the PR, I was expecting a minor change to `makeConditionalDeposit`, where transfers to `pendingBalance` were replaced by transfers to the user's address. I recall discussing that essentially, this would replace a `claim()` transaction with a `deposit` transaction for the user, and would be more convenient. But instead *all* transfers now just go to the user, rather than remaining inside `tokenLocking` where possible - which in many cases now replaces no transaction for the user (where their tokens would be inside `tokenLocking` already) with a deposit transaction.

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -240,8 +241,8 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
   }
 
   function burnUnneededRewards(uint256 _amount) public stoppable onlyReputationMiningCycle() {
-    ITokenLocking(tokenLocking).claim(IMetaColony(metaColony).getToken(), true);
-    ITokenLocking(tokenLocking).burn(_amount);
+    address clnyToken = IMetaColony(metaColony).getToken();
+    ERC20Extended(clnyToken).transfer(address(0x0), _amount);
```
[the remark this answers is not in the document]
kronosapiens: Mm that's accurate, I had forgotten about the explicit `burn` function and had fallen back on the `0x0` convention, but I'll update.

kronosapiens:
@area that's good feedback. The spec was simply to remove `pendingBalance`, but that change led to other changes downstream -- for instance, the need for an overridden `deposit` which takes `_force` (otherwise a deposit would just bounce back to the sender). I removed `transfer` because it seemed sort of like residual functionality -- you can't `_force` a transfer, so withdrawing and depositing seemed cleaner... but I see your point that in many cases, the transfer would be desirable, as long as it follows the same conditional logic.

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -240,8 +241,8 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
   }
 
   function burnUnneededRewards(uint256 _amount) public stoppable onlyReputationMiningCycle() {
-    ITokenLocking(tokenLocking).claim(IMetaColony(metaColony).getToken(), true);
-    ITokenLocking(tokenLocking).burn(_amount);
+    address clnyToken = IMetaColony(metaColony).getToken();
+    ERC20Extended(clnyToken).burn(_amount);
```
area, line 245: I am overly suspicious of this change, but haven't investigated fully. I can't immediately see why the tokens are now in a different place.

area (commented):
Not finished, but done for now until I can get back to it.

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -172,19 +172,14 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
     // Transfer the the tokens
     Lock storage userLock = userLocks[_token][_user];
     userLock.balance = sub(userLock.balance, _amount);
-    makeConditionalDeposit(_token, _amount, _recipient);
+
+    require(ERC20Extended(_token).transfer(_recipient, _amount), "colony-token-locking-transfer-failed");
```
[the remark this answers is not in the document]
kronosapiens: The change is to avoid indeterminacy in the destination of the tokens. If we had `makeConditionalDeposit`, then depending on the state of the external world (the locks), the claimed stake would either remain in the user's tokenLocking account, or be held by the address directly. This means that if a contract flow involves sending those tokens elsewhere, the contract would need to check to see where the tokens went (involving some `balanceBefore`, `balanceAfter` ugliness). Previously, the tokens would always be _somewhere_ in TokenLocking, meaning a call to `claim` could sort it out in either case. But there's no one-click reconciliation available now, so I thought it best to provide determinacy by always sending tokens to the obligator.

contracts/colonyNetwork/ColonyNetworkMining.sol
kronosapiens: It is due to the above change removing `makeConditonalDeposit` -- now all reclaimed stakes are held by the mining cycle directly.

contracts/tokenLocking/TokenLocking.sol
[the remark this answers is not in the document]
area: We want it to be easy for users to leave their tokens in this contract though. They shouldn't be forced to redeposit unless they have to. That was the whole point of the original design.

You're right that there's indeterminacy in the destination of the tokens, but I don't think that removing it by always having the tokens leave the contract is the correct solution - I think any contracts interacting with this should be able to cope. Is `deposit(token, 0, true)` prior to making the corresponding call not equivalent to `claim(token, true)` after the corresponding call?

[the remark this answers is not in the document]
kronosapiens: Hey! That's a fine idea.

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -92,12 +92,21 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
...
-    makeConditionalDeposit(_token, _amount, msg.sender);
+  function deposit(address _token, uint256 _amount, bool _force) public tokenNotLocked(_token, _force) {
+    require(ERC20Extended(_token).transferFrom(msg.sender, address(this), _amount), "colony-token-locking-transfer-failed"); // ignore-swc-123
 
     Lock storage lock = userLocks[_token][msg.sender];
+    lock.balance = add(lock.balance, _amount);
+
+    // Handle the deprecated pendingBalance, if any (idempotent operation)
+    lock.balance = add(lock.balance, lock.DEPRECATED_pendingBalance);
+    delete lock.DEPRECATED_pendingBalance;
```
area, line 108: This should only happen if `_force` is true, I think? But otherwise, yeah, this is a nice solution.

test/contracts-network/token-locking.js
```diff
@@ -118,22 +132,12 @@ contract("Token Locking", (addresses) => {
...
       await token.approve(tokenLocking.address, usersTokens, { from: userAddress });
-      await tokenLocking.deposit(token.address, usersTokens / 2, { from: userAddress });
-      await tokenLocking.deposit(token.address, usersTokens / 2, { from: userAddress });
+      await tokenLocking.methods["deposit(address,uint256,bool)"](token.address, usersTokens / 2, true, { from: userAddress });
+      await tokenLocking.methods["deposit(address,uint256,bool)"](token.address, usersTokens / 2, true, { from: userAddress });
       const info = await tokenLocking.getUserLock(token.address, userAddress);
       expect(info.balance).to.eq.BN(usersTokens);
     });
 
-    it("should be able to deposit tokens while they are locked, into pendingBalance", async () => {
```
area, line 127: This test should be turned in to `should not be able to deposit tokens while they are locked`, rather than being deleted.

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -208,11 +203,11 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
   // Internal functions
 
   function makeConditionalDeposit(address _token, uint256 _amount, address _user) internal {
-    Lock storage userLock = userLocks[_token][_user];
     if (isTokenUnlocked(_token, _user)) {
+      Lock storage userLock = userLocks[_token][_user];
       userLock.balance = add(userLock.balance, _amount);
     } else {
-      userLock.pendingBalance = add(userLock.pendingBalance, _amount);
+      require(ERC20Extended(_token).transfer(_user, _amount), "colony-token-locking-transfer-failed");
```
area, line 210: I'm still sceptical about this being a straight `require`. It opens a window for tokens to mess with any flow that involves transfer of tokens e.g. if the option that's winning in a vote is the one I want, I could lock the token and prevent anyone from revealing any more votes until the reveal period has expired.

Unfortunately, I can't come up with a solution that isn't just `pendingBalance` so I'm not sure how to square this away other than burying our heads in the sand.

contracts/tokenLocking/TokenLocking.sol
kronosapiens: I'll go with that. An alternative would be to check if the tokens is locked, but that's more gas for a conditional which will only be flipped once.

contracts/tokenLocking/TokenLocking.sol
kronosapiens: How so? This is an ERC20 transfer, so it can't fail if the token is locked, since it doesn't know about token locking.

area: Ah, I've not been clear - locking means several things here!

ERC20 only says that this function exists, it makes no guarantees about what it does or any additional functionality. The token itself could implement functionality that makes this function fail (what I was referring to as 'locking' in the above comment). [Our token](https://github.com/JoinColony/colonyToken) is ERC20 compliant, but certainly can stop people transferring with its permission system, for example.

kronosapiens: Ah, fair. Well, one idea is to hang on to `pendingBalance` for this (rare) scenario, and allow the balance to be tipped back into the main balance whenever someone deposits with `_force`. The require would be turned into an `if` statement.

contracts/tokenLocking/TokenLocking.sol
area: Oh, you're right, that is a better trigger. But gassier, as you say. I agree it's not worth it.

contracts/tokenLocking/TokenLocking.sol
area: Indeed, that is the only approach that I can think of that works.

contracts/tokenLocking/TokenLocking.sol
area: This still needs to be inside a conditional.

area:
I'm guessing a rebase has gone wrong here...

kronosapiens:
Regarding this logic:

```
      try ERC20Extended(_token).transfer(_user, _amount) returns (bool success) {
        if (!success) {
          userLock.pendingBalance = add(userLock.pendingBalance, _amount);
        }
      } catch {
        userLock.pendingBalance = add(userLock.pendingBalance, _amount);
      }
``` 

I do think this may be overkill. The `transfer` here is not from an external account to token locking, but rather from token locking out, so we can guarantee that the funds are there (so no underflow errors to worry about). You mention that a token "may" implement `transfer` in an unconventional way and we should defend against that, but I'm not sure that's really our responsibility, and further that token will have already been transferred _into_ token locking successfully before it's ever transferred out here. The only way to test it it seems would be to construct a special malicious token contract, which seems like overkill. So on balance I think going back to the simple conditional would be more than sufficient here, or frankly doing away with this depositing into `pendingBalance` entirely for what sort of seems like a very hypothetical failure case.

area:
I've added a test for the other branch where `pendingBalance` can be set (i.e. where a `.transfer()`) call returns false. Like the other test, not my best work, and can maybe be improved, but it does the job.

area:
The test following `should be able to process a large reputation update log` is repeatedly timing out across multiple branches it seems :cry:

area (approved):

PR #909, opened 2020-12-04
Test extensions function as intended when installed in subdomains only

### LINKED ISSUE #890: Enabling extensions in specific domains after they are installed in root
While root permissions are required to install an extension, in theory, permissions could only be awarded to an extension in a subdomain. We must make sure our extensions work in that scenario, and while I believe they should, as far as I can see we don't have tests along those lines.

area:
Closes #890

### DIFF
--- a/contracts/extensions/FundingQueue.sol
+++ b/contracts/extensions/FundingQueue.sol
@@ -25,7 +25,7 @@ import "./../tokenLocking/ITokenLocking.sol";
 import "./ColonyExtension.sol";
 import "./../colony/ColonyDataTypes.sol";
 
-contract FundingQueue is ColonyExtension, PatriciaTreeProofs, ColonyDataTypes {
+contract FundingQueue is ColonyExtension, PatriciaTreeProofs {
 
   // Events
   event ProposalCreated(uint256 id, uint256 indexed fromPot, uint256 indexed toPot, address indexed token, uint256 amount);
@@ -41,7 +41,6 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs, ColonyDataTypes {
   uint256 constant UINT256_MAX = (2 ** 256) - 1;
   uint256 constant STAKE_FRACTION = WAD / 1000; // 0.1%
   uint256 constant COOLDOWN_PERIOD = 14 days;
-  uint8 constant FUNDING_ROLE = 5;
 
   // Initialization data
   IColonyNetwork colonyNetwork;
@@ -283,11 +282,10 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs, ColonyDataTypes {
     proposal.lastUpdated = updateTime;
 
     assert(proposal.totalPaid <= proposal.totalRequested);
-    address colonyAuthority = colony.authority();
 
     // Check if the extension has the permissions to do this
     // If not, cancel the proposal so others aren't blocked
-    if (!colony.hasUserRole(address(this), proposal.domainId, ColonyRole.Funding)) {
+    if (!colony.hasUserRole(address(this), proposal.domainId, ColonyDataTypes.ColonyRole.Funding)) {
         emit ProposalPinged(_id, 0);
         cancelProposal(_id, HEAD);
         return;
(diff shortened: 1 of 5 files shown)


contracts/extensions/FundingQueue.sol
```diff
@@ -283,25 +283,45 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs {
...
 
-    if (proposal.totalPaid == proposal.totalRequested) {
-      proposal.state = ProposalState.Completed;
-
-      queue[HEAD] = queue[_id];
-      delete queue[_id];
-
-      //  May be the null proposal, but that's ok
-      proposals[queue[HEAD]].lastUpdated = updateTime;
+    try
```
kronosapiens, line 286: As per our discussion, we discussed not using `try/catch` and instead checking the permission explicitly:

```
if (!colony.hasUserRole(address(this), proposal.domainId, FUNDING_ROLE)) {
  // Cancel proposal
}
```

I think that'll be more succinct and readable.

area: Hm, I was writing a big long answer saying how that's how I started, but doing this check requires six external calls, but while writing it I've realised you are right, this is enough. That domain does indeed have to be the domain we have an explicit permission in, and so we don't have to check the inheritance etc, which can be left to be done in the main call.

kronosapiens: Fab u lous

test/extensions/funding-queue.js
```diff
@@ -793,4 +810,91 @@ contract("Funding Queues", (accounts) => {
...
+
+      // Proposal is cancelled
+      const proposal = await fundingQueue.getProposal(proposalId);
+      expect(proposal.state).to.eq.BN(STATE_CANCELLED);
+
+      // No tokens transferred
+      const balanceAfter = await colony.getFundingPotBalance(2, token.address);
+      const amountTransferred = balanceBefore.sub(balanceAfter);
+      const expectedTransferred = new BN("0");
+      expect(amountTransferred).to.eq.BN(expectedTransferred);
```
kronosapiens, line 897: Just do `to.be.zero;`

contracts/extensions/FundingQueue.sol
```diff
@@ -282,6 +283,15 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs {
     proposal.lastUpdated = updateTime;
 
     assert(proposal.totalPaid <= proposal.totalRequested);
+    address colonyAuthority = colony.authority();
+
+    // Check if the extension has the permissions to do this
+    // If not, cancel the proposal so others aren't blocked
+    if (!colony.hasUserRole(address(this), proposal.domainId, ColonyRole.Funding)) {
```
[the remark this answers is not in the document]
area: No, that's absolutely the best way. I'd simply forgotten it was possible :sweat:

contracts/extensions/FundingQueue.sol
```diff
@@ -41,6 +41,7 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs {
   uint256 constant UINT256_MAX = (2 ** 256) - 1;
   uint256 constant STAKE_FRACTION = WAD / 1000; // 0.1%
   uint256 constant COOLDOWN_PERIOD = 14 days;
+  uint8 constant FUNDING_ROLE = 5;
```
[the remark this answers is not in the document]
area: Nothing at all, a leftover from when I was trying to do this without importing `ColonyDataTypes`.

contracts/extensions/FundingQueue.sol
```diff
@@ -282,6 +283,15 @@ contract FundingQueue is ColonyExtension, PatriciaTreeProofs {
     proposal.lastUpdated = updateTime;
 
     assert(proposal.totalPaid <= proposal.totalRequested);
+    address colonyAuthority = colony.authority();
```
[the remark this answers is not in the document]
area: Another leftover from trying to avoid using `ColonyDataTypes` :smile:

kronosapiens (approved):

PR #911, opened 2020-12-10
Events for the event god

area:
Explanation / reasoning for events here can be split up:

## New events for payment lifecycle

We just didn't have events you would ordinarily expect. `PaymentFinalized` and things. So I've added them.

## New arguments for existing events

Mostly `agent` i.e. `msg.sender` for the event. There's no other clean way to get this as far as I can see if we're restricting ourselves to events. We _might_ be able to get clever here with TheGraph's call handlers, and not need these in most cases. Note that `FundingPotAdded` doesn't get one, because there is always another event emitted with it, and similarly the events related to signed changes to tasks don't get them, because those are triggered by the contract making an external call to itself, so would always give the colony address. To work around that, I've added `TaskChangedViaSignatures` emitting the addresses that agreed to change the property. Making that work with `expectEvent` is why the helper has changed.

### DIFF
--- a/contracts/colony/ColonyDataTypes.sol
+++ b/contracts/colony/ColonyDataTypes.sol
@@ -22,21 +22,25 @@ interface ColonyDataTypes {
   // Events
 
   /// @notice Event logged when Colony is initialised
+  /// @param agent The address that is responsible for triggering this event
   /// @param colonyNetwork The Colony Network address
   /// @param token The Colony Token address
   event ColonyInitialised(address agent, address colonyNetwork, address token);
 
   /// @notice Event logged when Colony is initially bootstrapped
+  /// @param agent The address that is responsible for triggering this event
   /// @param users Array of address bootstraped with reputation
   /// @param amounts Amounts of reputation/tokens for every address
   event ColonyBootstrapped(address agent, address[] users, int[] amounts);
 
   /// @notice Event logged when colony is upgraded
+  /// @param agent The address that is responsible for triggering this event
   /// @param oldVersion The previous colony version
   /// @param newVersion The new colony version upgraded to
   event ColonyUpgraded(address agent, uint256 oldVersion, uint256 newVersion);
 
   /// @notice Event logged when a user/domain/role is granted or revoked
+  /// @param agent The address that is responsible for triggering this event
   /// @param user The address of the user being affected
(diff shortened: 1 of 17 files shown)


kronosapiens:
Loving these commit messages!

contracts/colony/ColonyDataTypes.sol
```diff
@@ -147,57 +147,81 @@ interface ColonyDataTypes {
...
   /// @notice Event logged when a payout is claimed, either from a Task or Payment
   /// @param fundingPotId Id of the funding pot where payout comes from
   /// @param token Token of the payout claim
   /// @param amount Amount of the payout claimed, after network fee was deducted
-  event PayoutClaimed(uint256 indexed fundingPotId, address token, uint256 amount);
+  event PayoutClaimed(address agent, uint256 indexed fundingPotId, address token, uint256 amount);
 
   /// @notice Event logged when a task has been canceled
   /// @param taskId Id of the canceled task
   event TaskCanceled(uint256 indexed taskId);
```
kronosapiens, line 187: Why no change here?

contracts/extensions/OneTxPayment.sol
```diff
@@ -24,8 +24,8 @@ import "./ColonyExtension.sol";
 
 
 contract OneTxPayment is ColonyExtension {
-  event OneTxPaymentMade();
-  
+  event OneTxPaymentMade(address madeBy);
```
kronosapiens, line 27: I don't like this argument name for some reason -- it seems unconventional. Can we go with `caller` or `agent` or something?

contracts/colony/ColonyDataTypes.sol
area: Cancelling a task goes through the signed-by-contributors process that results in `msg.sender` in that call always being the colony. The transaction now [emits](https://github.com/JoinColony/colonyNetwork/pull/911/files#diff-22a3ddcee839f8d7c9a853e0226611bf413c43506cc778a24d701c420f3b1871R1447) the new `TaskChanvedViaSignatures` event as well with the relevant addresses.

area:
This has everything in it now that I think the frontend team will expect to have access to, event wise, and so is ready for a proper review.

kronosapiens (approved):

PR #913, opened 2020-12-16
OneTxPayment event changes

area:
This branch is the one Raul and I were using to get The Graph working with the contracts / the dev environment, so merging this is sort of obligatory. There's a cherry-picked commit from #911 and some other changes.

`OneTxPaymentMade` event gets `agent` (like many things are going to), I add some events to the payment lifecycle. I added a getter to extensions so that we can work out what colony they're from. Raul added functionality to a migration that writes out the colonynetwork address to a file, which gets used by the graph dev environment to set itself up programmatically.

### DIFF
--- a/contracts/colony/ColonyPayment.sol
+++ b/contracts/colony/ColonyPayment.sol
@@ -55,12 +55,16 @@ contract ColonyPayment is ColonyStorage {
 
     payments[paymentCount] = payment;
 
-    if (_skillId > 0) {
-      setPaymentSkill(_permissionDomainId, _childSkillIndex, paymentCount, _skillId);
-    }
-
     emit FundingPotAdded(fundingPotCount);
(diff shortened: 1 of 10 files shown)


contracts/colony/ColonyPayment.sol
```diff
@@ -55,12 +55,16 @@ contract ColonyPayment is ColonyStorage {
 
     payments[paymentCount] = payment;
 
+    emit FundingPotAdded(fundingPotCount);
+    emit PaymentAdded(paymentCount);
+
     if (_skillId > 0) {
       setPaymentSkill(_permissionDomainId, _childSkillIndex, paymentCount, _skillId);
+      emit PaymentSkillSet(paymentCount, _skillId);
```
kronosapiens, line 63: Can we add a newline before this?

kronosapiens (approved):

PR #914, opened 2020-12-17
Prevent burning of tokens by colony without appropriate bookkeeping

### LINKED ISSUE #910: Prevent colony burning transactions with an arbitrary transaction
While we guard against transfers, we should also guard against a colony burning its own tokens, which would mess up the internal bookkeeping.

We should, however, allow a colony to burn its own tokens through a function, though, doing the appropriate bookkeeping.

area:
Closes #910.

I've added the two burn transaction function signatures we have on the token we provide to the 'forbidden' list of transactions, though I'm wondering if we need to take a different approach to this. Maybe preventing arbitrary transactions to a contract that (appears to be) an ERC20 that the colony has a balance in, or something? I've had to add `transferFrom` too despite arguing against that last time we looked at this, because in some ERC20 implementations (including the one the token we provide uses) you don't have to give yourself permission to `transferFrom` from yourself. That's pretty annoying, because there is a legitimate reason for a colony to want to be able to call `transferFrom`. Are we going to have to inspect a 'transferFrom` transaction further? This is probably involved enough to warrant 

It's also possible there's a more gas-efficient way to do this check (bloom-filter-esqe?), but I haven't worked out what that is or whether it's worthwhile.

I've then added the `burnTokens` functionality to provide a mechanism to burn tokens owned by the colony. You have to burn tokens out of the root funding pot. While this is arguably tokens leaving the colony, I've opted to not put in a network tax.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -18,11 +18,11 @@
 pragma solidity 0.7.3;
 pragma experimental ABIEncoderV2;
 
+import "./../common/ERC20Extended.sol";
 import "./../common/IEtherRouter.sol";
 import "./../extensions/ColonyExtension.sol";
 import "./../tokenLocking/ITokenLocking.sol";
 import "./ColonyStorage.sol";
-
 
 contract Colony is ColonyStorage, PatriciaTreeProofs {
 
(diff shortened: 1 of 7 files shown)


contracts/colony/Colony.sol
```diff
@@ -60,6 +63,9 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
 
     require(sig != APPROVE_SIG, "colony-cannot-call-erc20-approve");
     require(sig != TRANSFER_SIG, "colony-cannot-call-erc20-transfer");
+    require(sig != TRANSFER_FROM_SIG, "colony-cannot-call-erc20-transfer-from");
+    require(sig != BURN_SIG, "colony-cannot-call-burn");
```
kronosapiens, line 67: Shouldn't these messages include the `erc20-` component?

contracts/colony/Colony.sol
```diff
@@ -471,6 +480,17 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     ITokenLocking(tokenLockingAddress).transferStake(_user, _amount, token, _beneficiary);
   }
 
+  function burnTokens(
+    address _token,
+    uint256 _amount
+  ) public stoppable auth {
```
kronosapiens, line 486: Could probably put arguments into a single line.

contracts/colony/Colony.sol
area: The [ERC20 Spec](https://eips.ethereum.org/EIPS/eip-20#methods) does not include the `burn` functions.

kronosapiens (approved):

PR #916, opened 2021-01-04
Have previous version resolvers respond accurately

area:
Previously, resolvers were registered as previous versions, but `.version()` was still the current version number.  This makes them respond accurately to `version`, at least, which means that relevant UI development is possible.

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -18,7 +18,8 @@ module.exports = {
       'patriciaTree',
       'testHelpers/ContractEditing.sol', // only used in setting up colony-network-recovery.js tests, never in production
       'testHelpers/NoLimitSubdomains.sol',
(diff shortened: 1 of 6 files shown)


kronosapiens (approved):

area:
While using this branch, Alicja ran in to an issue where calling `upgrade` wasn't actually possible. The call to `version()` in upgrade is internal, so always returned `5`, even if to the outside world the colony responded to `version()` calls with a different number.

This is effectively changing the contracts for the benefit of development, and while there is a slim justification here (calls to version should always be consistent) it still doesn't sit quite right with me.

I also considered inheriting `Colony` in to the contracts in PreviousVersion, but that would have meant adding `virtual` to `upgrade` and `version` which I thought would be more egregious.

kronosapiens (approved):

PR #917, opened 2021-01-07
Multiple chain support

### LINKED ISSUE #872: Multiple chain support
Just an issue to hold things I encounter as I deploy on xDai that we need proper solutions for going forward. 

* Possibly a better solution to usernames than a bespoke ENS deployment every time?
* Reputation mining rewards need to be handled appropriately
* Elegantly prevent mining of 'CLNY'

Upgrading solidity and using the `chainId` assembly opcode is going to be valuable, I suspect. Would also mean we could get rid of the 'patches' in colonyJS (that are the current workaround for #645)

area:
Closes #872, with clean implementations of the different required functionality across multiple chains.


area:
Marking this as 'ready for review', despite the fact the build fails (only because of coverage). It also can't be merged until #907 is merged, which is now also going to contain removal of the metacolony stipend functionality in the core contracts, which otherwise would also require chain-specific functionality to be added here.

contracts/colony/Colony.sol
```diff
@@ -20,11 +20,12 @@ pragma experimental ABIEncoderV2;
 
 import "./../common/ERC20Extended.sol";
 import "./../common/IEtherRouter.sol";
+import "./../common/GetChainId.sol";
```
kronosapiens, line 23: Can we call this `MultiChain`? It doesn't seem right to name a contract after a function.

contracts/colony/Colony.sol
```diff
@@ -191,6 +192,12 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
     require(msg.sender == colonyNetworkAddress, "colony-access-denied-only-network-allowed");
     // Function only valid on the Meta Colony
     require(address(this) == IColonyNetwork(colonyNetworkAddress).getMetaColony(), "colony-access-denied-only-meta-colony-allowed");
+    // Not callable on Xdai
+    uint256 chainId = getChainId();
+    if (chainId == 100 || chainId == 265669100) {
```
kronosapiens, line 197: I'd rather these numbers be defined as constants somewhere, probably in the `GetChainId` contract.

contracts/colony/Colony.sol
```diff
@@ -191,6 +192,12 @@ contract Colony is ColonyStorage, PatriciaTreeProofs {
...
+      require(false, "colony-network-forbidden-on-xdai");
```
kronosapiens, line 198: This could be done as `require(chainId != 100 etc, "forbidden")`

contracts/colonyNetwork/ColonyNetworkMining.sol
```diff
@@ -173,7 +174,11 @@ contract ColonyNetworkMining is ColonyNetworkStorage {
     }
 
     // II. Disburse reputation and tokens
-    IMetaColony(metaColony).mintTokensForColonyNetwork(realReward);
+    uint256 chainId = getChainId();
+    // On Xdai, we can only use bridged tokens, so no minting
+    if (chainId != 100 && chainId != 265669100) {
```
kronosapiens, line 179: One idea is to include these types of conditionals in the `GetChainId` contract itself, so these contracts can be more semantic, e.g. `if isMainnet()` etc.

contracts/common/GetChainId.sol
```diff
@@ -0,0 +1,30 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.7.3;
+
+
+
+contract GetChainId {
```
kronosapiens, line 22: As per my above comments, I think we can put more useful conditional logic in this contract and keep the rest of the contracts more semantic.

contracts/testHelpers/ChainId.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity 0.7.3;
+
+
+contract ChainId {
```
kronosapiens, line 4: Part of me would like this to inherit from `GetChainId`, as simple as it is.

test-chainid/chainid-dependent-behaviour.js
```diff
@@ -0,0 +1,93 @@
...
+contract("Contract Storage", (accounts) => {
+  const MINER1 = accounts[5];
+
+  let metaColony;
+  let clnyToken;
+  let colonyNetwork;
+  let chainId;
+
+  before(async () => {
+    const cid = await ChainId.new();
```
kronosapiens, line 26: I'm looking slanty-eyed at these variable names. Calling the contract `MultiChain` would help.

test-chainid/chainid-dependent-behaviour.js
```diff
@@ -0,0 +1,93 @@
...
+
+    await colonyNetwork.initialiseReputationMining();
+    await colonyNetwork.startNextCycle();
+  });
+
+  describe("Should behave differently based on the network deployed to", () => {
+    it("should be able to get the domain name", async () => {
+      await metaColony.registerColonyLabel("meta", "", { from: accounts[0] });
+      console.log(chainId);
+      if (chainId === 1 || chainId === 2656691) {
```
kronosapiens, line 46: Same, use constants.

kronosapiens (commented):
Nice PR! I like it.

kronosapiens (dismissed):
Lovely!

contracts/colony/Colony.sol
```diff
@@ -20,10 +20,12 @@ pragma experimental ABIEncoderV2;
 
 import "./../common/ERC20Extended.sol";
 import "./../common/IEtherRouter.sol";
+import "./../common/MultiChain.sol";
+import "./../extensions/ColonyExtension.sol";
```
kronosapiens, line 24: Is this doing anything?

contracts/common/MultiChain.sol
```diff
@@ -0,0 +1,48 @@
...
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.7.3;
+
+
+
+contract MultiChain {
+  function chainId() internal view returns (uint256) {
```
kronosapiens, line 23: Nitpick: can we call this function `getChainId`? Also no reason not to make it public.

contracts/testHelpers/ChainId.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity 0.7.3;
+
+import "../common/MultiChain.sol";
+
+contract ChainId is MultiChain {
```
kronosapiens, line 5: You know, we could probably just drop this helper entirely and use `MultiChain` in the tests directly.

test-chainid/chainid-dependent-behaviour.js
kronosapiens: Yeah, doing:

```
const multiChain = await MultiChain.new()
chainId = await multiChain.getChainId()
```

Would be much better.

contracts/colony/Colony.sol
area: `ColonyExtension` is used in the file... I guess it's able to pick it up from the import in ColonyStorage?

contracts/common/MultiChain.sol
area: It's weird to make this public without exposing it on the resolver though - it's only being made public for the tests, and I have a well documented hatred of changing production code for the benefit of the tests :stuck_out_tongue:

contracts/colony/Colony.sol
area: Yeah, it doesn't need this to compile successfully.

contracts/common/MultiChain.sol
area: Okay I'll make it external and it shouldn't cost more gas as far as I can tell.

contracts/colony/Colony.sol
kronosapiens: I mention it because the PR doesn't introduce it anywhere other than the import, which tells me that any existing usage is already functioning.

contracts/common/MultiChain.sol
kronosapiens: Yeah, I hear you about the resolver. Happy leaving it internal I suppose...

kronosapiens (approved):
Lovely!

PR #919, opened 2021-01-20
Add script to deploy 'true' old versions of the contracts

area:
For some of the work @alicjakujawa has been doing, she needed to be able to easily deploy old versions of the contracts. The kludge that we have in the migrations was inadequate (see #916) because while those versions claimed to be old versions, they were actually version 5 contracts. That meant that events and so forth had an unexpected number of arguments (especially after #911) and so colonyJS would error out, making the development of the frontend impossible in this scenario.

This script checks out old versions as appropriate and deploys them to the existing network deployment found at the address in `etherrouter-address.json` (introduced in [this commit](https://github.com/JoinColony/colonyNetwork/commit/adf8b3e995cb3cbba4631cf09086d49a5cfdb8a8) with little fanfare at the time)).

### DIFF
--- a/scripts/deployUnmockedColonyVersions.js
+++ b/scripts/deployUnmockedColonyVersions.js
@@ -40,7 +40,7 @@ module.exports = async () => {
       res = res.stdout;
     }
     let index = res.indexOf("Colony version 3 set to Resolver");
-    v3ResolverAddress = res.substring(index + 33, index + 33 + 42);
+    v3ResolverAddress = res.substring(index + RESOLVER_LOG_OFFSET, index + RESOLVER_LOG_OFFSET + ADDRESS_LENGTH);
     console.log("v3 address:", v3ResolverAddress);
     await metaColony.addNetworkColonyVersion(3, v3ResolverAddress);
 
--- a/scripts/deployUnmockedColonyVersions.js
+++ b/scripts/deployUnmockedColonyVersions.js
@@ -44,7 +44,7 @@ module.exports = async () => {
     res = await exec("yarn run truffle migrate --reset");
     index = res.indexOf("Colony version 4 set to Resolver");
     v4ResolverAddress = res.substring(index + 33, index + 33 + 42);
-    await metaColony.addNetworkColonyVersion(3, v4ResolverAddress);
+    await metaColony.addNetworkColonyVersion(4, v4ResolverAddress);
   } catch (err) {
     console.log(err);
   }
--- a/scripts/deployUnmockedColonyVersions.js
+++ b/scripts/deployUnmockedColonyVersions.js
@@ -26,6 +26,7 @@ module.exports = async () => {
   const metaColonyAddress = await colonyNetwork.getMetaColony();
   const metaColony = await IMetaColony.at(metaColonyAddress);
   try {
+    await exec("mv ./build ./buildBackup");
     await exec("git checkout ad5569de24567517aa12624e29600c9136fb594d && git submodule update");
 
     // Comment out uneeded parts of file
@@ -66,5 +67,5 @@ module.exports = async () => {
   // put things back how they were.
   await exec(`git checkout -f ${currentHash}`);
   await exec("git submodule update");
-  await exec("yarn run truffle compile");
+  await exec("mv ./buildBackup ./build");
 };
--- a/scripts/deployUnmockedColonyVersions.js
+++ b/scripts/deployUnmockedColonyVersions.js
@@ -17,7 +17,7 @@ const IMetaColony = artifacts.require("./IMetaColony");
 
 module.exports = async () => {
   // While debugging, always checkout currenthash
-  await exec("git checkout 2690fba3d3002fd72912cd74f1fbf4932c734e94 && git submodule update");
+  // await exec("git checkout 2690fba3d3002fd72912cd74f1fbf4932c734e94 && git submodule update");
   let res = await exec("git log -1 --format='%H'");
   const currentHash = res.stdout.trim();
   let v3ResolverAddress;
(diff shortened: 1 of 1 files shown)


scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,72 @@
...
+const cnAddress = require("../etherrouter-address.json").etherRouterAddress; // eslint-disable-line import/no-unresolved
+
+const IColonyNetwork = artifacts.require("./IColonyNetwork");
+const IMetaColony = artifacts.require("./IMetaColony");
+
+module.exports = async () => {
+  // While debugging, always checkout currenthash
+  // await exec("git checkout 2690fba3d3002fd72912cd74f1fbf4932c734e94 && git submodule update");
+  const currentHash = await exec("git log -1 --format='%H'");
+  // const currentHash = res.stdout.trim();
```
kronosapiens, line 22: Delete this?

scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,72 @@
...
+  const metaColonyAddress = await colonyNetwork.getMetaColony();
+  const metaColony = await IMetaColony.at(metaColonyAddress);
+  try {
+    await exec("mv ./build ./buildBackup");
+    await exec("git checkout ad5569de24567517aa12624e29600c9136fb594d && git submodule update");
+
+    // Comment out uneeded parts of file
+    await exec("sed -i'' -e '26,27 s|^|//|' ./migrations/4_setup_colony_version_resolver.js");
+    await exec("sed -i'' -e '31 s|^|//|' ./migrations/4_setup_colony_version_resolver.js");
+    await exec("sed -i'' -e 's|await ContractRecovery.deployed()|await ContractRecovery.new()|' ./migrations/4_setup_colony_version_resolver.js");
```
kronosapiens, line 35: I don't love how brittle this is...

Here's a suggestion: put this logic inside a conditional, and simply `sed` a constant, something along the lines of `const skipForBackVersion = true;` which would then bypass these lines.

scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,72 @@
...
+    await exec("rm -rf ./build");
+    let res = await exec("yarn run truffle migrate --reset -f 4 --to 4");
+    if (res.stdout) {
```
kronosapiens, line 38: A ternary operator would be cleaner here: `res = (res.stdout) ? res.stdout : res;`

scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,72 @@
...
+      // How this response looks changes node 10->12
+      res = res.stdout;
+    }
+    let index = res.indexOf("Colony version 3 set to Resolver");
+    v3ResolverAddress = res.substring(index + 33, index + 33 + 42);
```
kronosapiens, line 43: Can we have these numbers as constants rather than magic numbers?

scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,72 @@
...
+  const currentHash = await exec("git log -1 --format='%H'");
+  // const currentHash = res.stdout.trim();
+  let v3ResolverAddress;
+  let v4ResolverAddress;
+  const colonyNetwork = await IColonyNetwork.at(cnAddress);
+  const metaColonyAddress = await colonyNetwork.getMetaColony();
+  const metaColony = await IMetaColony.at(metaColonyAddress);
+  try {
+    await exec("mv ./build ./buildBackup");
+    await exec("git checkout ad5569de24567517aa12624e29600c9136fb594d && git submodule update");
```
kronosapiens, line 30: Does this version not have a release tag? That's too bad.

scripts/deployUnmockedColonyVersions.js
area: If I understand your suggestion correctly, you're talking about adding that to the migration file? That's not really possible - the file exists in the git history, I can't really go back and add to it without rebasing everything that's been on `master` since then. I would argue it is not brittle for that reason though - we are checking out a specific commit, so this will never change.

scripts/deployUnmockedColonyVersions.js
area: I mean, a ternary operator means it can go on one line, but I think mine is much easier to grok.

scripts/deployUnmockedColonyVersions.js
area: It does, I am just bad :smile:

scripts/deployUnmockedColonyVersions.js
kronosapiens: Ah yes fair, this isn't the `develop` migrations. Carry on 🚀

scripts/deployUnmockedColonyVersions.js
kronosapiens: Ok, I don't feel strongly.

scripts/deployUnmockedColonyVersions.js
```diff
@@ -0,0 +1,73 @@
...
+    await exec("sed -i'' -e '31 s|^|//|' ./migrations/4_setup_colony_version_resolver.js");
+    await exec("sed -i'' -e 's|await ContractRecovery.deployed()|await ContractRecovery.new()|' ./migrations/4_setup_colony_version_resolver.js");
+    await exec("rm -rf ./build");
+    let res = await exec("yarn run truffle migrate --reset -f 4 --to 4");
+    if (res.stdout) {
+      // How this response looks changes node 10->12
+      res = res.stdout;
+    }
+    let index = res.indexOf("Colony version 3 set to Resolver");
+    v3ResolverAddress = res.substring(index + 33, index + 33 + 42);
```
kronosapiens, line 43: The consts!

area: :facepalm:

kronosapiens (approved):

PR #924, opened 2021-02-04
Add function for a colony to unlock its own token, if owner

area:
In order for 'unlock token' to appear as an action in the frontend, there needs to be an event for it to key off of. The token is already deployed, so we can't alter is there, so I've added a function to the colony that calls `unlock` on the token. To work on the frontend we'll have to give the colony `owner`ship of the token, but arguably we should have been doing that anyway.

I had to steal my changes in `test-data-generator` from #907 to write the test, so they're here too.

### DIFF
--- a/test/contracts-network/colony.js
+++ b/test/contracts-network/colony.js
@@ -216,6 +216,19 @@ contract("Colony", (accounts) => {
       expect(rewardPotInfo.associatedTypeId).to.be.zero;
       expect(rewardPotInfo.payoutsWeCannotMake).to.be.zero;
     });
+
+    it("should allow the token to be unlocked by a root user only", async () => {
+      ({ colony, token } = await setupRandomColony(colonyNetwork, true));
+      await token.setOwner(colony.address);
+      let locked = await token.locked();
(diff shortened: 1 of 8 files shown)


test/contracts-network/colony.js
```diff
@@ -216,6 +216,19 @@ contract("Colony", (accounts) => {
...
       expect(rewardPotInfo.payoutsWeCannotMake).to.be.zero;
     });
+
+    it("should allow the token to be unlocked by a root user only", async () => {
+      ({ colony, token } = await setupRandomColony(colonyNetwork, true));
+      await token.setOwner(colony.address);
+      let locked = await token.locked();
+      expect(locked).to.be.equal(true);
+
+      await checkErrorRevert(colony.unlockToken({ from: accounts[1] }), "ds-auth-unauthorized");
```
kronosapiens, line 226: Can we add a `const USER1 = accounts[1]` up at the top?

kronosapiens (commented):
Look good!

area:
Used super-powers to merge.

PR #925, opened 2021-02-12
Do not burn CLNY on Xdai during mining

area:
If we've bridged CLNY, burning the bridged token does nothing to the supply on mainnet, and the bridge will just hold the tokens for eternity (I think?). For now, send the CLNY to the metacolony. We can decide to unbridge and 'truly' burn down the line, if we like.

### DIFF
--- a/contracts/colonyNetwork/ColonyNetworkMining.sol
+++ b/contracts/colonyNetwork/ColonyNetworkMining.sol
@@ -248,7 +248,13 @@ contract ColonyNetworkMining is ColonyNetworkStorage, MultiChain {
   function burnUnneededRewards(uint256 _amount) public stoppable onlyReputationMiningCycle() {
     address clnyToken = IMetaColony(metaColony).getToken();
     ITokenLocking(tokenLocking).withdraw(clnyToken, _amount, true);
-    ERC20Extended(clnyToken).burn(_amount);
+    if (isXdai()){
+      // On Xdai, I'm burning bridged tokens is certainly not what we want. So let's
+      // send them to the metacolony for now.
+      ERC20Extended(clnyToken).transfer(metaColony, _amount);
+    } else {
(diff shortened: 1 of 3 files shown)


area:
Changes to `expectEvent` are to accommodate more than one event of the same name being fired in a transaction. We lose the per-argument expects, but this is more general and should still be reasonably informative with the error that gets thrown for debugging purposes.

kronosapiens:
My main feedback on this is that the amount of pseudo-burned tokens is not stored on-chain anywhere. So the act of "actually burning" the bridged tokens is something that requires off-chain event parsing, etc. My first thought is it could be nice to have a "burnedXDaiTokens" storage variable which tracks this number so we can write a function to bridge/burn those tokens. But maybe that's not really viable since that variable would live on the XDai chain and we'd have to get it over to mainnet somehow anyway, which itself is probably an off-chain process.

area:
But then we also have to decrement or otherwise track that variable when we actually do burn the tokens, which I'm not sure actually saves us a huge amount of bother.

kronosapiens:
True true... I guess this feels a bit like "we'll burn the tokens, we promise" which seems like the sort of thing we want to avoid // defeats the purpose of "burning". I remember in the past we'd talked about burning vs. giving to the metacolony and IIRC the answer was "if we give to the MC it creates an incentive to slash" or something along those lines. I can imagine a future XDai MC saying "wellll, we could burn these tokens or we could just KEEP THEM and use them".

area:
It is better than "We can no longer burns these tokens and they are stuck in the bridge forever".

kronosapiens:
My point is that burning and reclaiming are different actions with different incentive structures. But I agree that this is a good step and so we can go with this for now and course-correct down the road.

kronosapiens (approved):

PR #927, opened 2021-02-22
Fix first issues arising from actual use

area:
Could be a worse crop, all things considered...

1. We were not correctly tracking the recovery role count when using `setUserRoles`. This fixes that. The consequences of this are frustrating, and are going to be dealt with on the frontend to a limited extent. Basically, the frontend will prevent people upgrading unless there is exactly one person with the recovery role. It can tell this from events, and not from `numRecoveryRoles`, which isn't incremented when using `setUserRoles`, which is all the frontend uses. `numRecoveryRoles` will then be correct, and the users can upgrade and re-award the recovery roles as they see fit.
If someone upgrades without doing that, by interacting directly with the contracts, then not even recovery mode can save them, as `numRecoveryRoles` is protected, which is a little awkward. I don't really think there's much to be done there, though.
Only two colonies have been affected by this so far, and this was discovered internally, so hopefully won't affect anyone meaningfully once we're deployed.

2. Upgraded colonies missed a permission being given to them in `finishUpgrade`, just like we'd always feared. This one was tricky to spot, as the function can be called by two different permissions. I discovered this through using the Metacolony. It only affects colonies that have upgraded 4->5, so shouldn't be too big a deal as very few fall in to that category.

3. The designed frontend for recovery mode requires annotations to be allowed in recovery mode. I can't see an issue with this, so I've enabled it there.

I've also bumped the colony version to six.

### DIFF
--- a/contracts/colony/ColonyRoles.sol
+++ b/contracts/colony/ColonyRoles.sol
@@ -104,7 +104,7 @@ contract ColonyRoles is ColonyStorage, ContractRecoveryDataTypes {
         setTo = uint256(roles) % 2 == 1;
 
         ColonyAuthority(address(authority)).setUserRole(_user, _domainId, roleId, setTo);
-        if (roleId == 0){
+        if (roleId == uint8(ColonyRole.Recovery)) {
           if (setTo){
             recoveryRolesCount++;
           } else {
--- a/contracts/colony/ColonyRoles.sol
+++ b/contracts/colony/ColonyRoles.sol
@@ -19,9 +19,10 @@ pragma solidity 0.7.3;
 pragma experimental ABIEncoderV2;
 
 import "./ColonyStorage.sol";
+import "./../common/ContractRecoveryDataTypes.sol";
 
 
-contract ColonyRoles is ColonyStorage {
+contract ColonyRoles is ColonyStorage, ContractRecoveryDataTypes {
 
   function setRootRole(address _user, bool _setTo) public stoppable auth {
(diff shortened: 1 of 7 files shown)


contracts/colony/ColonyRoles.sol
```diff
@@ -103,7 +104,14 @@ contract ColonyRoles is ColonyStorage {
         setTo = uint256(roles) % 2 == 1;
 
         ColonyAuthority(address(authority)).setUserRole(_user, _domainId, roleId, setTo);
-
+        if (roleId == 0){
+          if (setTo){
```
kronosapiens, line 108: I don't think this quite gets us there, actually, since it doesn't check if the user already has the role or not. If I already have the recovery role, and then I use `setUserRoles` to give it to myself again, then this number will be bumped twice. We'd want something like

```
if (setTo && !hasUserRole(user, recovery))
```

area: I disagree. This is inside the `if (changed)` statement, which stops that, no?

area: Either way, testing for this wouldn't go amiss

kronosapiens: Ah right, that bit was clipped out of the Github viewer. I agree then. My other request is that we use `uint8(ColonyRoles.Recovery)` instead of `0` in the external conditional.

kronosapiens (dismissed):
Fabulous

kronosapiens (approved):

kronosapiens (dismissed):

kronosapiens (approved):

PR #929, opened 2021-02-25
Fix slither for extensions

kronosapiens:
Partially solves #319

Completes the slither updates by bringing extension code in line with best-practices.

On-hold until the frontend is ready to handle versioned extensions.

### DIFF
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -68,7 +68,7 @@ contract CoinMachine is ColonyExtension {
 
   /// @notice Returns the version of the extension
   function version() public override pure returns (uint256) {
-    return 1;
+    return 2;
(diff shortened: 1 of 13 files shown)


area:
Raul is working on the dapp/colonyJS supporting upgrades, so I've pointed him at this branch for testing. The extensions that have been changed I've upgraded to v2, and I've added a script to install previous versions for his testing, as well.

I've taken the potentially controversial decision to remove the restriction on adding versions in order from the network contracts. We don't have that restriction for colony versions, so I don't see why we should have it here.

kronosapiens:
@area I agree, I was thinking about making that same change for #931 . The in-order restriction was a nice idea but ultimately is more cumbersome than its worth.

area (approved):

PR #932, opened 2021-03-12
First pass at allowing arbitrary approve transactions

area:
Okay, this is a first pass at this ready for feedback. I'm not completely happy with merging this yet (might be able to extend this idea to other 'forbidden' transactions), but happy enough for it to start being reviewed. 

Basically, this introduces some more internal accounting so that when a colony `approve`s another contract to take its tokens via the very common ERC20 interface, we only allow tokens to be removed from the root ~~address~~ domain. Tokens may be taken, but the internal bookkeeping won't update until another approve transaction to the same address is made, or any user calls the `updateApprovalAmount` function.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -91,7 +91,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
     return res;
   }
 
-  function updateApprovalAmount(address _token, address _spender) public {
+  function updateApprovalAmount(address _token, address _spender) stoppable public {
     updateApprovalAmountInternal(_token, _spender, false);
   }
 
@@ -491,11 +491,11 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
     emit TokenUnlocked();
   }
 
-  function getTokenApproval(address _token, address _spender) public returns (uint256 amount) {
+  function getTokenApproval(address _token, address _spender) public view returns (uint256 amount) {
     return tokenApprovals[_token][_spender];
   }
 
-  function getTotalTokenApproval(address _token) public returns (uint256 amount) {
+  function getTotalTokenApproval(address _token) public view returns (uint256 amount) {
     return tokenApprovalTotals[_token];
   }
 
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -60,10 +60,8 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
 
     address spender;
     if (sig == APPROVE_SIG){
-      address approvalAmount;
       assembly {
         spender := mload(add(_action, 0x24))
-        approvalAmount := mload(add(_action, 0x44))
       }
       updateApprovalAmountInternal(_to, spender, false);
     }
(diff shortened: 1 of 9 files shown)


kronosapiens:
As an initial comment, is there any reason to include this in `makeArbitraryTransaction`,  rather than have a separate `approveErc20` function? That way `makeArbitraryTransaction` can stay restricted and we put all the special bookkeeping logic in the other function.

area:
I hadn't really considered it, but doing it this way makes it much easier for everything that actually wants to make arbitrary transactions - e.g. the dapp interface designed does not distinguish between any arbitrary transactions, and so ideally wouldn't want to distinguish between the functions being called.

kronosapiens:
Sure, I guess more my point is that ERC20 approval is specific functionality that we want to support which requires specific logic, then maybe it should actually be a separate interface, i.e. not a part of the arbitrary transaction flow. Potentially something you do via the colony management screen. I guess it comes down to what is an "arbitrary transaction". Is it anything that calls out to an external address?

kronosapiens:
I haven't worked through it entirely but another benefit of a separate `approveErc20` function would be to dispense with the "internal" approve and have only a single public function, since any special bookkeping you need to do during the approval call can be handled inside of `approveErc20`.

area:
Would you have separate functions for burn, transfer, etc?

kronosapiens:
Hmm... an "ERC20 Utils" interface which lets colonies handle all these functions could make sense separate from "arbitrary transactions". "Arbitrary transactions" is turning into mostly a big conditional for handling a very specific type of transaction...

area:
I guess I consider an arbitrary transaction any transaction the colony sends with a contract outside the network. I agree it's inelegant, but having different interfaces for every external contract we want to support is also not particularly elegant.

kronosapiens:
I don't fully agree, it seems much more like path-dependence than anything else. If the spec had been "add support for common ERC20 transactions" we would almost certainly have written separate functions, but since instead we're coming at it from the frame of "arbitrary transactions with special logic" it seems like too much of a change.

I don't buy the argument that we'll need separate interfaces for every external contract, ERC20s seem like a very well-defined special case, there's still a lot of residual that we can throw into "arbitrary transactions". Put another way, if we anticipate that 80% of "arbitrary transactions" will be ERC20 calls, that seems to support the argument for splitting those out.

I don't feel strongly enough to keep pushing for it though, we can do it this way.

kronosapiens:
I'll just point out that a "switch statement plus argument unpacking" is exactly what functions are for.

area:
> I don't fully agree, it seems much more like path-dependence than anything else. If the spec had been "add support for common ERC20 transactions" we would almost certainly have written separate functions, 

I agree this seems very likely.

> but since instead we're coming at it from the frame of "arbitrary transactions with special logic" it seems like too much of a change.

The real point of comparison should be 'if we had implemented separate functions for ERC20 transactions, and then added arbitrary transactions, what would we have decided to do"? In a real-world scenario, I think we would end up where you describe, but only _because_ of the path dependency and not wanting to remove elements of the interface that already existed. Without that constraint, I'm still not convinced this isn't the best implementation.

I've factored out the pre/post transaction handling that is required for these in to separate functions to trim down the arbitrary transaction function. Do you find this structure more palatable? The arbitrary transaction function is basically the same as it would be if the ERC20 functions were split out entirely (because we would still have to guard against those ERC20 functions being called through this arbitrary transaction function, and so the `if` lines would be the requires that we had in originally). The one exception is the post-approval bookkeeping, but that seems like a small price to pay to keep the interface much simpler.

> I'll just point out that a "switch statement plus argument unpacking" is exactly what functions are for.

Hahaha, this is true! This is definitely on the line. If it were just the contracts, I'd be more easily swayed, but in terms of implementing what has actually been designed, I think this is the best path.

kronosapiens:
ALRIGHT I'm satisfied.

kronosapiens (approved):

PR #937, opened 2021-03-22
Create KycOracle package

kronosapiens:
Introduce a `KycOracle` service which reads data from an off-chain KYC service (such as Synaps) and writes it on-chain, using the `Whitelist` extension introduced in #931  (and which this is rebased off of).

Still a bit of a WIP since we haven't finished hammering out the Synaps-side flow, but this is basically what I'm thinking.

### DIFF
--- a/packages/kyc-oracle/KycOracle.js
+++ b/packages/kyc-oracle/KycOracle.js
@@ -1,0 +1,129 @@
+const axios = require("axios");
+const ethers = require("ethers");
+const express = require("express");
+
+class KycOracle {
+  /**
+   * Constructor for KycOracle
+   * @param {string} privateKey              The private key of the address that has the administration permission.
+   *                                           If used, `userAddress` is not needed and will be derived.
+   * @param {string} userAddress             The address that has the administration permission to edit the whitelist.
+   * @param {string} whitelistAddress        The address the whitelist.
+   * @param {Object} loader                  The loader for loading the contract interfaces. Usually a TruffleLoader.
+   * @param {Object} provider                Ethers provider that allows access to an ethereum node.
+   * @param {Number} [port]                  The port the oracle will serve on
+   */
+  constructor({ privateKey, userAddress, apiKey, loader, provider, port = 3000 }) {
+    if (privateKey) {
+      this.wallet = new ethers.Wallet(privateKey, this.provider);
+      this.userAddress = this.wallet.address;
+    } else {
+      this.wallet = provider.getSigner(userAddress);
+      this.userAddress = userAddress;
+    }
+    console.log("Transactions will be signed from ", this.userAddress);
+
+    this.apiKey = apiKey;
+
+    this.loader = loader;
+    this.provider = provider;
+
+    this.app = express();
+
+    this.app.use(function (req, res, next) {
+      res.header("Access-Control-Allow-Origin", "*");
+      next();
+    });
+
+    this.app.get("", async (req, res) => {
+      return res.status(200).send("KycOracle is running!");
+    });
+
+    // Create a new session
+    this.app.get("/session", async (req, res) => {
+      const { data } = await axios.post(
+        "https://workflow-api.synaps.io/v2/session/init",
+        {}, // No data
+        {
+          headers: {
+            "Api-Key": this.apiKey,
+          },
+        }
+      );
+
+      return res.status(200).send(data);
+    });
+
+    // Query for KYC status and update the whitelist
+    this.app.get("/status/:sessionId", async (req, res) => {
+      const { data } = await axios.get("https://workflow-api.synaps.io/v2/workflow/details", {
+        headers: {
+          "Api-Key": this.apiKey,
+          "Session-Id": req.params.sessionId,
+        },
+      });
+
+      // TODO: Update this logic once the final verification flow is ready
+      const validated = data.identity_step[0].state === "VALIDATED";
+      const userAddress = this.userAddress;
+
+      if (validated) {
+        const alreadyApproved = await this.whitelist.getApproval(userAddress);
+        if(!alreadyApproved) {
+          // TODO: Pull actual user address from Synaps, this is only a placeholder
+          const gasEstimate = await this.whitelist.estimateGas.approveUser(userAddress, true);
+          await this.whitelist.approveUser(userAddress, true, { gasLimit: gasEstimate, gasPrice: this.gasPrice });
+        }
+      }
+
+      return res.status(200).send(data);
+    });
+
+    this.server = this.app.listen(port, () => {
+      console.log(`⭐️ KycOracle running on port ${this.server.address().port}`);
+    });
+  }
+
+  /**
+   * Initialises the Kyc Oracle so that it knows where to find the `Whitelist` contract
+   * @param  {string}  whitelistAddress The address of the current `Whitelist` contract
+   * @return {Promise}
+   */
+  async initialise(whitelistAddress) {
+    const network = await this.provider.getNetwork();
+    this.chainId = network.chainId;
+
+    this.whitelistContractDef = await this.loader.load({ contractName: "Whitelist" }, { abi: true, address: false });
+    this.whitelist = new ethers.Contract(whitelistAddress, this.whitelistContractDef.abi, this.wallet);
+
+    await this.updateGasEstimate("safeLow");
+  }
+
+  /**
+   * Update the gas estimate
+   * @param  {string}  Transaction speed (fastest, fast, safeLow)
+   * @return {Promise}
+   */
+  async updateGasEstimate(type) {
+    if (this.chainId === 100) {
+      this.gasPrice = ethers.utils.hexlify(1000000000);
+      return;
+    }
+
+    try {
+      // Get latest from ethGasStation
+      const { data } = await axios.get("https://ethgasstation.info/json/ethgasAPI.json");
+
+      if (data[type]){
+        this.gasPrice = ethers.utils.hexlify(data[type] / 10 * 1e9);
+      } else {
+        this.gasPrice = ethers.utils.hexlify(20000000000);
+      }
+    } catch (err) {
+      console.log(`Error during gas estimation: ${err}`);
+      this.gasPrice = ethers.utils.hexlify(20000000000);
+    }
+  }
+}
+
+module.exports = KycOracle;
--- a/packages/kyc-oracle/KycOracle.js
+++ b/packages/kyc-oracle/KycOracle.js
@@ -12,9 +12,9 @@ class KycOracle {
   /**
    * Constructor for KycOracle
    * @param {string} privateKey              The private key of the address that has the administration permission.
-   *                                           If used, `adminAddress` is not needed and will be derived.
-   * @param {string} adminAddress             The address that has the administration permission to edit the whitelist.
-   * @param {string} whitelistAddress        The address the whitelist.
+   *                                         If used, `adminAddress` is not needed and will be derived.
+   * @param {string} adminAddress            The address that has the administration permission to edit the whitelist.
+   * @param {string} whitelistAddress        The address of the whitelist.
    * @param {Object} loader                  The loader for loading the contract interfaces. Usually a TruffleLoader.
    * @param {Object} provider                Ethers provider that allows access to an ethereum node.
    * @param {Number} [port]                  The port the oracle will serve on
@@ -183,7 +183,12 @@ class KycOracle {
 
   async createDB() {
     const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
-    await db.run("CREATE TABLE IF NOT EXISTS users ( address text NOT NULL UNIQUE, session_id text NOT NULL )");
+    await db.run(
+      `CREATE TABLE IF NOT EXISTS users (
+        address text NOT NULL UNIQUE,
+        session_id text NOT NULL
+      )`
+    );
     await db.close();
   }
 
@@ -207,7 +212,10 @@ class KycOracle {
     console.log(address, session);
     const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
     await db.run(
-      `INSERT INTO users (address, session_id) VALUES ("${address}", "${session}") ON CONFLICT(address) DO UPDATE SET session_id = "${session}"`
+      `INSERT INTO users (address, session_id)
+       VALUES ("${address}", "${session}")
+       ON CONFLICT(address) DO
+       UPDATE SET session_id = "${session}"`
     );
     await db.close();
   }
--- a/packages/kyc-oracle/KycOracle.js
+++ b/packages/kyc-oracle/KycOracle.js
@@ -93,7 +93,7 @@ class KycOracle {
             }
           );
           sessionId = data.session_id;
-          if (!/^[0-9a-f\-]+$/.test(sessionId)) {
+          if (!/^[0-9a-f-]+$/.test(sessionId)) {
             return res.status(500).send("Synaps session ID obtained was invalid");
           }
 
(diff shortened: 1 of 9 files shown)


packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,96 @@
...
+
+    this.apiKey = apiKey;
+
+    this.loader = loader;
+    this.provider = provider;
+
+    this.app = express();
+
+    this.app.use(function (req, res, next) {
+      res.header("Access-Control-Allow-Origin", "*");
```
area, line 34: Seems like if this is going to be an API exposed to the world, we should do better than this.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,88 @@
+const axios = require("axios");
+const ethers = require("ethers");
+const express = require("express");
+
+const { SYNAPS_API_KEY } = require("./config");
```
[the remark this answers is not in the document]
kronosapiens: I was actually thinking of making it a command-line argument.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Initialises the mining client so that it knows where to find the `ColonyNetwork` contract
+   * @param  {string}  colonyNetworkAddress The address of the current `ColonyNetwork` contract
+   * @return {Promise}
+   */
+  async initialise(whitelistAddress) {
+    const network = await this.provider.getNetwork();
+    this.chainId = network.chainId;
```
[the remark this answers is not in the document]
kronosapiens: I guess we don't really need it, the oracle should  be agnostic to the network, it's just a provider in the end.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    this.app.get("", async (req, res) => {
+      return res.status(200).send("KycOracle is running!");
+    });
+
+    // Create a new session
+    this.app.get("/session", async (req, res) => {
+      const session = await axios.post(
+        "https://workflow-api.synaps.io/v2/session/init",
```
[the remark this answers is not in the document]
kronosapiens: I do think we'll need this service to generate sessions, otherwise we'd have to put the API key in the dapp which would make it public.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,96 @@
...
+
+      // TODO: Update this logic once the final verification flow is ready
+      const validated = status.data.identity_step[0].state === "VALIDATED";
+
+      if (validated) {
+        // TODO: Pull actual user address from Synaps, this is only a placeholder
+        await this.whitelist.approveUser(this.userAddress, true);
+      }
+
+      return res.status(200).send(validated);
```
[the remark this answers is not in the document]
kronosapiens: I had it that way initially, I'm open to whatever return values will be most useful.

scripts/kyc-bootstrap.js
```diff
@@ -0,0 +1,37 @@
+/* globals artifacts */
+const { soliditySha3 } = require("web3-utils");
+
+const EtherRouter = artifacts.require("EtherRouter");
+const IColonyNetwork = artifacts.require("IColonyNetwork");
+const IMetaColony = artifacts.require("IMetaColony");
+const Whitelist = artifacts.require("Whitelist");
+
+const WHITELIST = soliditySha3("Whitelist");
+const API_KEY = "rbKPwh6GvzQ1epdMRknHSDtWOirQaKq1";
```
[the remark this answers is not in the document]
kronosapiens: I'm developing against their sandbox. We can generate new API keys trivially :)

packages/kyc-oracle/KycOracle.js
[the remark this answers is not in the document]
area: That's fine too :smile:

packages/kyc-oracle/KycOracle.js
kronosapiens: I suppose restricting access to the dapp would be appropriate.

area:
* Challenge/response verification to make sure we're talking to someone in control of the address being claimed (i.e. we don't give someone's session to someone pretending to be them, allowing them to see the documents they've submitted).
* Synaps endpoints updated
* SQLite DB to persist Synaps authentication session ids

This is exactly the flow we currently use when logging in to the app, so hopefully is acceptable!

area:
Commands to make testing easier, from a truffle console:

```
const axios = require("axios");

res = await axios.post("http://127.0.0.1:3000/auth/challenge",{address:accounts[3]});

signature = await web3.eth.sign(res.data.challenge, accounts[3])

res = await axios.post("http://127.0.0.1:3000/auth/token", {challenge: res.data.challenge, signature})

sessionid = res.data.sessionId

To check the current state:

res = await axios.get(`http://127.0.0.1:3000/status/${sessionid}`)
```

The flow can be accessed at: `https://verify.synaps.io/?session_id=${sessionid}&service=workflow`

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
+const axios = require("axios");
+const ethers = require("ethers");
+const express = require("express");
+const etherpass = require("etherpass");
+
+const { getChallenge } = etherpass;
+const { verifyEthSignature } = etherpass;
```
kronosapiens, line 7: Could these two be done in one line?

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+    this.provider = provider;
+
+    this.app = express();
+
+    this.app.use(function (req, res, next) {
+      const regex = /(.*colony\.io)/;
+      const origin = req.get("origin");
+      const matches = regex.exec(origin);
+      if (matches) {
+        res.header("Access-Control-Allow-Origin", matches[1]);
```
kronosapiens, line 50: Can we add a comment as to why we use index 1 here?

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+      next();
+    });
+
+    this.app.use(bodyParser.json());
+
+    this.app.get("", async (req, res) => {
+      return res.status(200).send("KycOracle is running!");
+    });
+
+    this.app.post("/auth/challenge", (req, res) => {
```
kronosapiens, line 61: Seems like it should be a `get` but if we're mirroring the app flow then by all means leave it as-is.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+    });
+
+    this.app.post("/auth/token", async (req, res) => {
+      try {
+        if (typeof req.body.challenge !== "string" || typeof req.body.signature !== "string") {
+          throw new Error("Invalid challenge/signature");
+        }
+
+        // This throws if bad signature
+        const address = verifyEthSignature(req.body.challenge, req.body.signature);
```
kronosapiens, line 78: Why not just put this line in the try/catch?

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+      }
+    } catch (err) {
+      console.log(`Error during gas estimation: ${err}`);
+      this.gasPrice = ethers.utils.hexlify(20000000000);
+    }
+  }
+
+  async createDB() {
+    const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
+    await db.run("CREATE TABLE IF NOT EXISTS users ( address text NOT NULL UNIQUE, session_id text NOT NULL  )");
```
kronosapiens, line 186: TWO SPACES AFTER `NOT NULL` ?????

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+    const res = await db.all(
+      `SELECT DISTINCT users.session_id as session_id
+       FROM users
+       WHERE users.address="${address}"`
+    );
+    await db.close();
+    const sessionIds = res.map((x) => x.session_id);
+    console.log(sessionIds);
+    if (sessionIds.length > 1) {
+      throw new Error("More than one matching address");
```
kronosapiens, line 201: I wonder if this is the best way to handle this? Why not return the most recent sessionId?

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+    return sessionIds[0];
+  }
+
+  async setSessionForAddress(address, session) {
+    console.log(address, session);
+    const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
+    await db.run(
+      `INSERT INTO users (address, session_id) VALUES ('${address}', '${session}') ON CONFLICT(address) DO UPDATE SET session_id = '${session}'`
```
kronosapiens, line 210: Stylistically you're using single quotes here and double quotes earlier in the SQL queries.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+    const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
+    const res = await db.all(
+      `SELECT DISTINCT users.address as address
+       FROM users
+       WHERE users.session_id="${sessionId}"`
+    );
+    await db.close();
+    const addresses = res.map((x) => x.address);
+    if (addresses.length > 1) {
+      throw new Error("More than one matching session");
```
kronosapiens, line 225: `More than one matching address`

packages/kyc-oracle/bin/index.js
```diff
@@ -0,0 +1,41 @@
...
+  .option("whitelistAddress", { string: true })
+  .option("apiKey", { string: true });
+
+const KycOracle = require("../KycOracle");
+const TruffleLoader = require("../TruffleLoader").default;
+
+const { adminAddress, privateKey, whitelistAddress, apiKey, network, providerPort, providerAddress, dbPath } = argv;
+const supportedInfuraNetworks = ["mainnet"];
+
+if ((!adminAddress && !privateKey) || !whitelistAddress) {
```
kronosapiens, line 20: Check for `apiKey` here as well.

kronosapiens (commented):
Looks good! I'd like to merge this after #931 if possible, since this PR contains a lot of the same code.

packages/kyc-oracle/KycOracle.js
area: Looking at this again I think we can get away with `matches[0]`.

packages/kyc-oracle/KycOracle.js
area: Yeah, I've just scooped this out of the app / etherpass documentation. Dodging caching given the challenge is only valid for 10 minutes seems like a good reason for `POST` over `GET` here though.

packages/kyc-oracle/KycOracle.js
area: I don't follow? It's already inside one?

packages/kyc-oracle/KycOracle.js
area: I mean, this is the equivalent of an `assert` in Solidity in that this should never happen. The table's primary key is addresses, so there should only ever be one match. If there is more than one, something is *very* wrong and I'd rather fail out entirely.

packages/kyc-oracle/KycOracle.js
kronosapiens: Why check the arguments and throw an exception in the previous lines?

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+const sqlite = require("sqlite");
+const bodyParser = require("body-parser");
+
+class KycOracle {
+  /**
+   * Constructor for KycOracle
+   * @param {string} privateKey              The private key of the address that has the administration permission.
+   *                                           If used, `adminAddress` is not needed and will be derived.
+   * @param {string} adminAddress             The address that has the administration permission to edit the whitelist.
+   * @param {string} whitelistAddress        The address the whitelist.
```
kronosapiens, line 17: "The address of the whitelist"

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+        const challenge = getChallenge(address);
+        return res.json({ challenge });
+      } catch (err) {
+        return res.status(500).send(err);
+      }
+    });
+
+    this.app.post("/auth/token", async (req, res) => {
+      try {
+        if (typeof req.body.challenge !== "string" || typeof req.body.signature !== "string") {
```
kronosapiens, line 73: Not sure if this check is necessary.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+      }
+    } catch (err) {
+      console.log(`Error during gas estimation: ${err}`);
+      this.gasPrice = ethers.utils.hexlify(20000000000);
+    }
+  }
+
+  async createDB() {
+    const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
+    await db.run("CREATE TABLE IF NOT EXISTS users ( address text NOT NULL UNIQUE, session_id text NOT NULL )");
```
kronosapiens, line 186: Maybe put a uniqueness constraint on the pairs of address/session_id?

https://stackoverflow.com/questions/12763726/mysql-make-a-pair-of-values-unique

That would avoid the need for the ` < 1 ` checks later on.

Something like:

```
`CREATE TABLE IF NOT EXISTS users (
    address text NOT NULL UNIQUE,
    session_id text NOT NULL UNIQUE,
    UNIQUE (address, session_id)
)`
```

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,231 @@
...
+      throw new Error("More than one matching address");
+    }
+    return sessionIds[0];
+  }
+
+  async setSessionForAddress(address, session) {
+    console.log(address, session);
+    const db = await sqlite.open({ filename: this.dbPath, driver: sqlite3.Database });
+    await db.run(
+      `INSERT INTO users (address, session_id) VALUES ("${address}", "${session}") ON CONFLICT(address) DO UPDATE SET session_id = "${session}"`
```
kronosapiens, line 210: Can we format the SQL queries nicely (with newlines)?

packages/kyc-oracle/KycOracle.js
area: They have it in [colonyServer](https://github.com/JoinColony/colonyServer/blob/be4db1fd333a7c918ac2a4410914f98860ba7fcf/src/index.ts) where I cribbed this from, so I left it in.

packages/kyc-oracle/KycOracle.js
area: I'm slightly uncomfortable with making `session_id` unique. We definitely hope it is, but we don't know how Synaps generate them.

I don't understand what you're suggesting here regarding the composite key. If address is `unique` (which it already is), by definition the combination of `address` and `session_id` in each row much be unique.

area:
More changes:

* Make `adminAddress` command line parameter be forced to be a string
* Validate the characters in provided Synaps session IDs, both from Synaps and the user, to avoid SQL injections.
* Endpoints that expect a Synaps session ID now expect it to be passed in the header. This is slightly preferred to having them in the URL.

packages/kyc-oracle/KycOracle.js
```diff
@@ -0,0 +1,266 @@
...
+            {
+              headers: {
+                "Api-Key": this.apiKey,
+                alias: address,
+              },
+            }
+          );
+          sessionId = data.session_id;
+          if (!/^[0-9a-f-]+$/.test(sessionId)) {
+            return res.status(500).send("Synaps session ID obtained was invalid");
```
kronosapiens, line 97: Not sure if we need to test data from Synaps but I'll allow it.

kronosapiens (commented):
Looks good to me. Github says this is yours to approve :)

area (approved):

PR #939, opened 2021-03-24
Slight mining improvement and bootstrap correction

area:
Two unrelated changes, but both small so I thought I'd bundle them up.

* Jack reported [a large list of incoming transactions](https://www.notion.so/colony/Metacolony-incoming-fees-are-all-denominated-in-CLNY-and-from-the-same-address-8d27401478b6436785852b33812338e2) in the metacolony. These are from the xDai implementation of the `burnUnneededRewards` function, which even if there are no rewards, went ahead and did a 0-value transfer anyway. We shortcircuit the 0 case now, which is good for gas, and good for the app. To test this, I also refactored the helper functions around testing for the presence / lack of presence of an event in a transaction.
* When we added expenditures and payments, bootstrapping got left behind. This PR also prevents bootstrapping if a payment or an expenditure has been created, in the same way bootstrapping is not allowed if a task has been created.

### DIFF
--- a/contracts/colony/ColonyStorage.sol
+++ b/contracts/colony/ColonyStorage.sol
@@ -198,7 +198,7 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, ColonyNetworkDataTypes
   }
 
   modifier isInBootstrapPhase() {
-    require(taskCount == 0, "colony-not-in-bootstrap-mode");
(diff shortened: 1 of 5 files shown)


kronosapiens (dismissed):

area:
Testing that an event doesn't happen doesn't really fill me with confidence, so I've changed the test so that there are rewards that have to be dealt with.

kronosapiens:
I figured, I just wanted to fix the break and let you come back to it in the morning.

kronosapiens (approved):

PR #942, opened 2021-04-08
Only use required stake in migration

area:
Staking loads of CLNY in the migration during setup was causing issues with the app dev environment when the miner tried to submit more entries but blocktime was fast-forwarded at the same time. Easiest solution was to stake only the minimum amount of stake.

### DIFF
--- a/test/reputation-system/client-sync-functionality.js
+++ b/test/reputation-system/client-sync-functionality.js
@@ -5,7 +5,7 @@ import chai from "chai";
 import bnChai from "bn-chai";
 
 import TruffleLoader from "../../packages/reputation-miner/TruffleLoader";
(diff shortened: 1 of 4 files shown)


test/reputation-system/client-sync-functionality.js
```diff
@@ -59,7 +59,7 @@ process.env.SOLIDITY_COVERAGE
         await reputationMiner1.initialise(colonyNetwork.address);
 
         const lock = await tokenLocking.getUserLock(clnyToken.address, MINER1);
-        expect(lock.balance).to.eq.BN(DEFAULT_STAKE);
+        expect(lock.balance).to.eq.BN(MIN_STAKE);
```
kronosapiens, line 62: Should still be `DEFAULT_STAKE` since you're redefining the constant.

kronosapiens (approved):

PR #943, opened 2021-04-16
Add AmbExecutor.sol

### LINKED ISSUE #940: Prototype an extension for sending arbitrary transactions across the xDai AMB
One potentially useful use-case for Colony is for colonies to make decision on xDai, due to the low cost of gas, and then execute that decision on mainnet. Doing this would require leveraging the arbitrary message bridge, or AMB, which xDai has kindly made available.

We should explore the feasibility of this functionality. It will likely involve an extension which takes in arbitrary transactions (in the format we use for motions, etc) and invokes a function on the xDai AMB contract, to send the transaction to mainnet. One key open question is how gas costs are handled -- while there is a 2m gas limit per arbitrary message, xDai is almost certainly not taking responsibility for eth-side gas costs.

*Update*: After discussing with Auryn (who is developing similar functionality on the Gnosis Safe side), gas costs are paid by whoever executes the tx -- the idea being that validators submit signatures, but the actual execution can be triggered by anyone.

Some resources:

https://docs.tokenbridge.net/eth-xdai-amb-bridge/about-the-eth-xdai-amb

https://docs.tokenbridge.net/amb-bridge/how-to-develop-xchain-apps-by-amb

https://docs.tokenbridge.net/amb-bridge/gas-token-minting

https://github.com/gnosis/amb-module/blob/safebridge/contracts/SafeBridgeModule.sol

kronosapiens:
Closes #940 

## Notes

Encoding the AMB transaction:

`function requireToPassMessage(address _contract, bytes _data, uint256 _gas) external;`

`_contract` is the destination across the bridge.
`_data` is the transaction to execute, including the function signature and all encoded arguments.

On the AmbExecutor side, we have:

`function executeTransaction(address to, bytes memory action)`

Note that this entire function call (including the signature for `executeTransaction`, as well as `to` and `action`, are what we encoded in `_data` previously. Now, `to` is the address we are transacting to in order to perform our desired action, such as manipulating funds.

In other words:

```
_data = encode("executeTransaction", to, action);
```

## Resources

Based off https://github.com/gnosis/amb-module/blob/safebridge/contracts/SafeBridgeModule.sol

See also https://docs.tokenbridge.net/amb-bridge/how-to-develop-xchain-apps-by-amb#receive-a-method-call-from-the-amb-bridge

AMB tx dashboard: https://alm-rinkeby.herokuapp.com/100/0xeca054161cb422c35ded25b71926059162f0236bc35770e068c8ee984439ac78

### DIFF
--- a/contracts/extensions/AmbExecutor.sol
+++ b/contracts/extensions/AmbExecutor.sol
@@ -1,0 +1,122 @@
+/*
+  This file is part of The Colony Network.
+
+  The Colony Network is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  The Colony Network is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.7.3;
+pragma experimental ABIEncoderV2;
+
+import "./ColonyExtension.sol";
+
+// ignore-file-swc-108
+
+// https://docs.tokenbridge.net/amb-bridge/how-to-develop-xchain-apps-by-amb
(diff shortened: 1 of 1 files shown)


kronosapiens:
After more discussion with @jakzilla and @auryn-macmillan , we realized we didn't need this contract after all -- all AMB bridging functionality can be handled via arbitrary transactions or contracts developed elsewhere. Huzzah!

PR #946, opened 2021-04-23
Fix maths bugs in ReputationVoting (and potentially elsewhere)

area:
This PR does two things

1. Fixes a bug where if a motion was escalated, but not staked fully, incorrect amounts of stake were returned to users (as `requiredStake` was used, rather than the amount that was actually staked), resulting in some users not being able to claim their stake.
2. Fixes a bug where if multiple people staked, not everyone would be able to claim their rewards in some circumstances. This was because the WAD-based calculations would round up, so e.g. if everyone staked 1/6th of the required value, their stakes would be rounded to 0.166...67 of the required value, slightly more than they had actually staked. I have taken this round-up functionality of the WAD calculation and dealt with it with extreme prejudice, and it now always rounds down, in line with the rest of solidity maths. This chance has been made on the `dappysys-monolithic` repository. As a result of this, some tests have changed, and not just for reputation voting.

a) Where numbers are different... they are simply different. That said, I can't work out why the numbers that changed in the funding queue tests were the ones that changed, and not some of the other ones.
b) Where `divRound` was used, I've removed it and replaced with `div` where appropriate
c) Where 'raw' calculations have been done, I've had to reorder some elements to more accurately represent the calculation happening on chain. As a happy side effect of this though, I've also managed to remove the `addn` or `subn`s that were floating around some of the tests which is much more satisfying.

We still do not have a solution for cleanup up small amounts of tokens left behind by this process, but I feel like that's a feature for another day.

### DIFF
--- a/contracts/extensions/VotingReputation.sol
+++ b/contracts/extensions/VotingReputation.sol
@@ -756,7 +756,6 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
     uint256 stakeFraction = wdiv(stakes[_motionId][_staker][_vote], totalSideStake);
 
     uint256 realStake = wmul(stakeFraction, motion.stakes[_vote]);
-    uint256 requiredStake = getRequiredStake(_motionId);
 
     uint256 stakerReward;
     uint256 repPenalty;
@@ -764,13 +763,23 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
     // Went to a vote, use vote to determine reward or penalty
     if (add(motion.votes[NAY], motion.votes[YAY]) > 0) {
 
-      uint256 loserStake = sub(requiredStake, motion.paidVoterComp);
+      uint256 loserStake;
+      uint256 winnerStake;
+      if (motion.votes[YAY] > motion.votes[NAY]){
+        loserStake = motion.stakes[NAY];
+        winnerStake = motion.stakes[YAY];
+      } else {
+        loserStake = motion.stakes[YAY];
+        winnerStake = motion.stakes[NAY];
+      }
+
+      loserStake = sub(loserStake, motion.paidVoterComp);
       uint256 totalVotes = add(motion.votes[NAY], motion.votes[YAY]);
       uint256 winFraction = wdiv(motion.votes[_vote], totalVotes);
       uint256 winShare = wmul(winFraction, 2 * WAD); // On a scale of 0-2 WAD
 
       if (winShare > WAD || (winShare == WAD && _vote == NAY)) {
-        stakerReward = wmul(stakeFraction, add(requiredStake, wmul(loserStake, winShare - WAD)));
+        stakerReward = wmul(stakeFraction, add(winnerStake, wmul(loserStake, winShare - WAD)));
       } else {
         stakerReward = wmul(stakeFraction, wmul(loserStake, winShare));
         repPenalty = sub(realStake, stakerReward);
@@ -779,6 +788,7 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
     // Determine rewards based on stakes alone
     } else {
       assert(motion.paidVoterComp == 0);
+      uint256 requiredStake = getRequiredStake(_motionId);
 
       // Your side fully staked, receive 10% (proportional) of loser's stake
       if (
--- a/test/extensions/coin-machine.js
+++ b/test/extensions/coin-machine.js
@@ -456,7 +456,7 @@ contract("Coin Machine", (accounts) => {
       const periodLength = await coinMachine.getPeriodLength();
       const maxPerPeriod = await coinMachine.getMaxPerPeriod();
       const windowSize = await coinMachine.getWindowSize();
-      const alphaAsWad = new BN(2).mul(WAD).divRound(windowSize.addn(1));
+      const alphaAsWad = new BN(2).mul(WAD).div(windowSize.addn(1));
 
       let currentPrice;
       let evolvePrice;
@@ -500,7 +500,7 @@ contract("Coin Machine", (accounts) => {
       await coinMachine.updatePeriod();
 
       const emaIntake = WAD.muln(100).mul(WAD.sub(alphaAsWad)).add(maxPerPeriod.mul(alphaAsWad));
-      const expectedPrice = emaIntake.divRound(WAD.muln(100));
+      const expectedPrice = emaIntake.div(WAD.muln(100));
       currentPrice = await coinMachine.getCurrentPrice();
       expect(currentPrice).to.eq.BN(expectedPrice);
     });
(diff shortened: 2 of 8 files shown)


contracts/extensions/VotingReputation.sol
```diff
@@ -756,21 +756,30 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
...
     uint256 stakerReward;
     uint256 repPenalty;
 
     // Went to a vote, use vote to determine reward or penalty
     if (add(motion.votes[NAY], motion.votes[YAY]) > 0) {
 
-      uint256 loserStake = sub(requiredStake, motion.paidVoterComp);
+      uint256 loserStake;
+      uint256 winnerStake;
+      if (motion.votes[YAY] > motion.votes[NAY]){
```
kronosapiens, line 768: Too much?

```
uint256 loserStake = motion.stakes[(motion.votes[YAY] > motion.votes[NAY]) ? YAY : NAY]; 
uint256 winnerStake = motion.stakes[(motion.votes[YAY] > motion.votes[NAY]) ? NAY : YAY]; 
```

test/extensions/coin-machine.js
```diff
@@ -404,7 +404,7 @@ contract("Coin Machine", (accounts) => {
       const periodLength = await coinMachine.getPeriodLength();
       const maxPerPeriod = await coinMachine.getMaxPerPeriod();
       const windowSize = await coinMachine.getWindowSize();
-      const alphaAsWad = new BN(2).mul(WAD).divRound(windowSize.addn(1));
+      const alphaAsWad = new BN(2).mul(WAD).div(windowSize.addn(1));
```
kronosapiens, line 407: Can you mention in the PR description why we don't need `divRound` anymore?

test/extensions/voting-rep.js
```diff
@@ -1405,11 +1406,15 @@ contract("Voting Reputation", (accounts) => {
       await voting.claimReward(motionId, 1, UINT256_MAX, USER0, YAY);
       await voting.claimReward(motionId, 1, UINT256_MAX, USER1, NAY);
 
+      let votingPayout = await voting.getVoterReward(motionId, new BN(user0Value.slice(2, 66), 16));
```
kronosapiens, line 1409: ~Mm, can we use `user0Influence` here? This slicing is peculiar, I don't like it.~

Just realized we don't introduce those variables until the token/hybrid PR. Can we just include the WAD amounts explicitly? User 0 is WAD, User 1 is WAD.muln(2).

test/extensions/voting-rep.js
```diff
@@ -1455,18 +1460,22 @@ contract("Voting Reputation", (accounts) => {
...
-      const expectedReward1 = REQUIRED_STAKE.add(loserStake.divn(3)).divn(3).muln(2); // stake + ((stake * .8) * (1 - (winPct = 2/3 * 2))
+      const loserStake = REQUIRED_STAKE.sub(votingPayout); // Take out voter comp
+      const expectedReward0 = loserStake.muln(2).divn(3); // (stake * .8) * (winPct = 1/3 * 2)
+      const expectedReward1 = REQUIRED_STAKE.add(loserStake.divn(3)).muln(2).divn(3); // stake + ((stake * .8) * (1 - (winPct = 2/3 * 2))
       const expectedReward2 = REQUIRED_STAKE.add(loserStake.divn(3)).divn(3); // stake + ((stake * .8) * (1 - (winPct = 2/3 * 2))
 
       expect(new BN(user0LockPost.balance).sub(new BN(user0LockPre.balance))).to.eq.BN(expectedReward0);
-      expect(new BN(user1LockPost.balance).sub(new BN(user1LockPre.balance))).to.eq.BN(expectedReward1.addn(1)); // Rounding
-      expect(new BN(user2LockPost.balance).sub(new BN(user2LockPre.balance))).to.eq.BN(expectedReward2.addn(1)); // Rounding
+      expect(new BN(user1LockPost.balance).sub(new BN(user1LockPre.balance))).to.eq.BN(expectedReward1);
```
kronosapiens, line 1477: This is satisfying!

test/extensions/voting-rep.js
```diff
@@ -1504,16 +1513,75 @@ contract("Voting Reputation", (accounts) => {
...
 
-      const loserStake = REQUIRED_STAKE.divn(10).muln(8); // Take out voter comp
-      const expectedReward0 = loserStake.divn(3).muln(2).divn(3).muln(2); // (stake * .8) * (winPct = 1/3 * 2)
-      const expectedReward1 = REQUIRED_STAKE.add(loserStake.divn(3)); // stake + ((stake * .8) * (1 - (winPct = 2/3 * 2))
-      const expectedReward2 = loserStake.divn(3).muln(2).divn(3); // (stake * .8) * (winPct = 1/3 * 2)
+      let votingPayout = await voting.getVoterReward(motionId, new BN(user0Value.slice(2, 66), 16));
+      const voter1reward = await voting.getVoterReward(motionId, new BN(user1Value.slice(2, 66), 16));
+      votingPayout = votingPayout.add(voter1reward);
+
+      const loserStake = REQUIRED_STAKE.sub(votingPayout); // Take out voter comp
```
kronosapiens, line 1520: Can we keep the comments here?

test/extensions/funding-queue.js
```diff
@@ -558,7 +558,7 @@ contract("Funding Queues", (accounts) => {
       // So 1 - (1 - 1/2 * 2/3) = 1/3 (33.3%) of the balance should be transferred
       const balanceAfter = await colony.getFundingPotBalance(1, token.address);
       const amountTransferred = balanceBefore.sub(balanceAfter);
-      const expectedTransferred = new BN("333743300899454444");
+      const expectedTransferred = new BN("333743300899535035");
```
kronosapiens, line 561: I'm curious why the differences here are so larger? (Larger than 1 wei I mean).

kronosapiens (commented):
Can we include a discussion in the PR description of the changes you made to `math.sol`?

area:
> Can we include a discussion in the PR description of the changes you made to math.sol?

Does point 2 in the description not already explain those?

test/extensions/voting-rep.js
```diff
@@ -257,7 +258,7 @@ contract("Voting Reputation", (accounts) => {
       const identifier = await voting.identifier();
       const version = await voting.version();
       expect(identifier).to.equal(VOTING_REPUTATION);
-      expect(version).to.eq.BN(reputationVotingVersion);
+      expect(version).to.eq.BN(2);
```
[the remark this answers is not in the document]
area: 100%, not sure what happened here :confused:

contracts/extensions/VotingReputation.sol
area: Yeah, I think so. It's harder to read, and I think it uses more gas as well.

test/extensions/coin-machine.js
area: Is this not also covered in point 2? I can make it more explicit if you like.

test/extensions/funding-queue.js
area: The values that are calculated using `wmul`/`wdiv` are [raised to a large power](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/extensions/FundingQueue.sol#L360) in the calculation for how much funding to transfer, and so a 1 wei difference in those numbers is going to result in a large discrepancy here.

test/extensions/coin-machine.js
kronosapiens: Mm it's alright.

kronosapiens (approved):

PR #947, opened 2021-04-27
Make getStakerReward behaviour consistent

area:
If no-one stakes on a side, and we attempt to query staker rewards on that side, `getStakerReward` errors out. This is because it divides by zero. This doesn't affect anything on the contracts (if no-one staked, then no-one can claim for that side, so the fact that it would divide by zero is irrelevant), but this proved annoying for the frontend team. Dividing by zero is something that we shouldn't do anyway, so let's just return 0 in this scenario (which will always be correct).


kronosapiens (approved):
LGTM

PR #949, opened 2021-04-29
Domain motions with alternative targets

### LINKED ISSUE #944: Allow votes to arbitrary targets + domains
Currently, all arbitrary targets require root votes. We should consider allowing arbitrary targets for domain votes, in the instance where we want to target an extension for domain-specific functionality (such as voting on a `OneTxPayment` in a subdomain).

kronosapiens:
Closes #944 

- Add `getCapabilityRoles` to the `ColonyExtension` interface.
- Add the ability for `makeDomainMotion` to accept an alternative target.
- Restrict `makeRootMotion` to only accept root-permissioned or non-permissioned functions.
- Update `colonyNetwork.getChildSkillId` to handle `UINT256_MAX` as one would expect.

### DIFF
--- a/contracts/extensions/ColonyExtension.sol
+++ b/contracts/extensions/ColonyExtension.sol
@@ -45,7 +45,9 @@ abstract contract ColonyExtension is DSAuth, DSMath {
   function deprecate(bool _deprecated) public virtual;
   function uninstall() public virtual;
 
-  function getCapabilityRoles(bytes4 _sig) public view virtual returns (bytes32);
+  function getCapabilityRoles(bytes4 _sig) public view virtual returns (bytes32) {
+    return bytes32(0);
+  }
 
   function getDeprecated() public view returns (bool) {
     return deprecated;
--- a/contracts/extensions/ColonyExtension.sol
+++ b/contracts/extensions/ColonyExtension.sol
@@ -45,6 +45,8 @@ abstract contract ColonyExtension is DSAuth, DSMath {
   function deprecate(bool _deprecated) public virtual;
   function uninstall() public virtual;
 
+  function getCapabilityRoles(bytes4 _sig) public view virtual returns (bytes32);
+
   function getDeprecated() public view returns (bool) {
     return deprecated;
   }
--- a/contracts/colonyNetwork/ColonyNetwork.sol
+++ b/contracts/colonyNetwork/ColonyNetwork.sol
@@ -280,9 +280,13 @@ contract ColonyNetwork is ColonyNetworkStorage, MultiChain {
   }
 
   function getChildSkillId(uint _skillId, uint _childSkillIndex) public view returns (uint256) {
-    Skill storage skill = skills[_skillId];
-    require(_childSkillIndex < skill.children.length, "colony-network-out-of-range-child-skill-index");
-    return skill.children[_childSkillIndex];
+    if (_childSkillIndex == UINT256_MAX) {
+      return _skillId;
+    } else {
+      Skill storage skill = skills[_skillId];
+      require(_childSkillIndex < skill.children.length, "colony-network-out-of-range-child-skill-index");
+      return skill.children[_childSkillIndex];
+    }
   }
 
   function deprecateSkill(uint256 _skillId) public stoppable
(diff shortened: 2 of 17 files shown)


area (commented):
Can we get a test demonstrating that the domain motions can't act in inappropriate domains?

area:
1. I cannot work out why the smoke tests need updating here (they started failing after the first commit) and it's giving me the heebie-jeebies. Any ideas?
2. After talking with some of the frontend team, there is an ambiguity here for making motions for actions that require a non-root permission in the root domain. They can be made through both `createRootMotion` and `createDomainMotion`. Do you think we should remove that ambiguity? I've told the fronted team to, if a motion is being created in root, they should use `createRootMotion` as it will be cheaper, gas wise - so should we error if `createDomainMotion` is tried to be used to create a motion in root?

kronosapiens:
Regarding the smoke test, we add a function to all the extensions, which should change the hash of the deployed contracts, no?

Regarding the ambiguity, I would support restricting `createRootMotion` to root functions only, and require `createDomainMotion` for funding, etc motions in the root domain. That seems the most semantic. Going the other way (restricting domain motions in the root domain, for non-root functions) seems a bit more arbitrary.

contracts/extensions/ColonyExtension.sol
```diff
@@ -45,6 +45,8 @@ abstract contract ColonyExtension is DSAuth, DSMath {
   function deprecate(bool _deprecated) public virtual;
   function uninstall() public virtual;
 
+  function getCapabilityRoles(bytes4 _sig) public view virtual returns (bytes32);
```
area, line 48: We could have a default implementation of this function, that returns `bytes32(0)`?

area (commented):
* We need tests for `getCapabilityRoles` on each extension to get the coverage green again
* Update smoke tests

Is the separation of `createRootMotion` and `createDomainMotion` necessary now? We could just have a single `createMotion` function with an `if` statement that, if it's a function that requires a domain proof, does the checks.

area:
Not a problem with alternative targets, but realised while thinking about the permissions we're relying on in combination with talking with the frontend team: The permissions imposed on `moveFundsBetweenPots` are not respected by motions. Effectively, we check that the motion is in a domain that has control over the funds being moved, but not that they have (funding) control over the destination.

kronosapiens:
@area I actually noticed that myself recently. I believe it's a non-issue, since if the condition checked by the voting contract (the domain proof for the first funding pot) is valid, then we can actually guarantee that the second condition (checked by the colony itself) will be valid, since the permission domain will be the same.

area:
I either have not explained myself well, or I am not understanding something. There are three domain proofs involved, and then some checks on those proofs. Let's have domain A, with children B and C, and a motion is being made that moves funds from B to C. The VotingReputation contract has funding permission in A.

Proof 1 (in the action). Proves that a permission in A gives the same permission in domain B
Proof 2 (in the action). Proves that a permission in A gives the same permission in domain C

Proof 3 (in the call to `makeDomainMotion`). This (is meant to) prove that the motion is being made in a domain that either is the domain in which the motion has the effect on, or is a direct parent of the domain in which the motion has an effect on. It works out the domain that Proof 1 points to (by taking the first two arguments in the action), and compares that to the domain that Proof 3 points to, and if they're the same then it allows the motion to be created.

Unfortunately, in the case of `moveFundsBetweenPots`, that is inadequate. The way the permissions work on `moveFundsBetweenPots`, funding permission in B is inadequate - I need funding permission in A. We should therefore (presumably), not allow a motion that only people in B get to vote on to change a decision that needed permission in A to make.

The proposed resolution here is make it so that you only need funding permission in the domain the funds are being moved from, for both motions and also for addresses leveraging the funding permission.

kronosapiens:
@area yes I understand your point now. Consider this tree:

```
    A
  /   \
B    C
      / \
    D  E 
```

In this example, I could make a motion in domain C to move funds from D to B, by creating `moveFundsBetweenPots(A, i, D, j, B)`. The voting extension would check that C is valid to make a vote in D, but would be unaware that B falls outside that purview. Relaxing the requirement on B would eliminate this issue.

Removing this requirement would let people transfer funds "up" the domain tree, which I don't really see as being too much of a problem anyway.

area (approved):
Nothing major I'm objecting to here, so approving and merging to unblock frontend and will do some cleanup in my PR for `moveFundsBetweenPots` changes.

PR #951, opened 2021-05-06
Add setDefaultGlobalClaimDelay for Expenditures

### LINKED ISSUE #948: Add a colony-wide default claim delay for expenditures
Currently, the default `globalClaimDelay` for expenditures is 0. It seems prudent to make this non-zero, and also to let colonies determine this value for themselves (along the lines of setting the reward inverse), via a root function `setExpenditureGlobalClaimDelay` or something along those lines.

kronosapiens:
Closes #948

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "test:contracts:watch": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle watch --network development",
     "test:contracts:e2e": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle test test-system/end-to-end.js --network development",
     "test:contracts:smoke": "truffle compile && truffle test ./test-smoke/colony-storage-consistent.js --network storageSmoke",
-    "test:security:slither": "slither . --solc-disable-warnings --exclude-low --exclude-informational --exclude-optimization --exclude uninitialized-state-variables,uninitialized-local-variables,unused-return --filter-paths 'lib|contracts/testHelpers'",
+    "test:security:slither": "slither . --solc-disable-warnings --exclude-low --exclude-informational --exclude-optimization --exclude uninitialized-state-variables,uninitialized-local-variables --filter-paths 'lib|contracts/testHelpers'",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "test:contracts:watch": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle watch --network development",
     "test:contracts:e2e": "npm run start:blockchain:client & truffle migrate --reset --compile-all && truffle test test-system/end-to-end.js --network development",
     "test:contracts:smoke": "truffle compile && truffle test ./test-smoke/colony-storage-consistent.js --network storageSmoke",
-    "test:security:slither": "slither . --solc-disable-warnings --exclude-low --exclude-informational --exclude-optimization --exclude uninitialized-state-variables,uninitialized-local-variables --filter-paths 'lib|contracts/testHelpers'",
+    "test:security:slither": "slither . --solc-disable-warnings --exclude-low --exclude-informational --exclude-optimization --exclude uninitialized-state-variables,uninitialized-local-variables,unused-return --filter-paths 'lib|contracts/testHelpers'",
```
area, line 39: I'm not wild about doing this, this is a legitimate mistake which it's usefully flagged before. Looking at the output from a failed build, maybe disabling the rule for the function in question would work?

area (commented):
I think the naming you've used is ambiguous - the most natural reading of the function/variable/event names is that you're setting the global claim delay for all existing expenditures. I think e.g. `setDefaultGlobalClaimDelay` would be much clearer.

package.json
kronosapiens: The issue is that there are two exclusions necessary for that line -- something about the try/catch logic throws up `unused return` and `unchecked transfer`. I've tried every way I can think of (two consecutive lines of ignore, including two detectors in one exclude) but slither doesn't seem to be able to handle it.

area (commented):
I think I'm good with this now, but still want to hold off merging.

package.json
area: Didn't I get rid of this at some point and it still worked?

kronosapiens: Mm, I think I remember that. Let's give it a go.

area (approved):

PR #954, opened 2021-05-08
Relax permission requirement on moveFundingBetweenPots

kronosapiens:
Reduce the requirement that the `toPot` is in the subtree in which the caller has the `funding` permission. Now, only the source of funds must be in the permitted subtree. In practice, this means that users can now transfer funds "up" the domain tree, all the way up to the root, regardless of the domain in which they hold the permission. This also means that users can now transfer funds to arbitrarily remote domains, even very specific ones, without having the inherited permission in that domain. The main downside (or is it a feature?) here is that now funds may arrive in a domain without anyone in that domain's inheritance path having had anything to do with it.

### DIFF
--- a/contracts/colony/ColonyAuthority.sol
+++ b/contracts/colony/ColonyAuthority.sol
@@ -110,6 +110,9 @@ contract ColonyAuthority is CommonAuthority {
     addRoleCapability(ROOT_ROLE, "editColony(string)");
     addRoleCapability(ROOT_ROLE, "burnTokens(address,uint256)");
     addRoleCapability(ROOT_ROLE, "unlockToken()");
+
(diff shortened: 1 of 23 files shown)


area:
After thinking about this a bit harder, I think our logic previously was faulty, and this restriction should remain and be treated as a special case for `moveFundsBetweenPots` in the case of the voting contracts.

The flaw is that having funding permission in a domain does _not_ mean that you can pay the tokens out - you need administration permission to do that. So you can award funding permission to someone in a domain and know that the administrators in that domain are still in control of any funds being paid out. But with these changes, if you award funding permission to someone with administration in _any_ other domain then they can pay those tokens out by first transferring them to that domain.

kronosapiens:
Closing due to security concerns.

PR #956, opened 2021-05-11
Permission moveFundsBetweenPots properly for voting

area:
This is my take at fixing up `moveFundsBetweenPots` so that what movement of funds are allowed is consistent regardless of whether the function is being executed directly or a motion has been made.

I tried a couple of other ways, but there just simply isn't enough information provided in the combination of the existing action and the existing motion creation function. You have to prove:

The domain I have the permission in is a parent of the 'fromPot' domain
The domain I have the permission in is a parent of the 'toPot' domain
The domain the vote is taking place in is a parent of the 'fromPot' domain
The domain the vote is taking place in is a parent of the 'toPot' domain

But we only have three proofs, between `createMotion` and the action if we create special functions to extract the second proof from the action. The action would provide the first two, but there's no way to prove both of the second with a single proof.

Stepping back, I realised that we're in this predicament is because there is an ambiguity in the general case for `moveFundsBetweenPots` as to which domain we're acting in. Consider the domain hierarchy:

```
      A 
      |
      B
    /   \
   C     D
```

Let's say I have the permission explicitly in A, and I want to move funds from C to D. What domain am I acting in? We could say it's A by convention, but that would mean it would be impossible to have a vote in B about this, which clearly should be allowed. There's no way to specify which domain we're acting in. So the solution I've implemented here adds a new `moveFundsBetweenPots` that removes this ambiguity. It still includes multiple domain proofs, but those now prove that the domain we're acting in is a parent of C and a parent of D and so don't need to be considered for the creation of the motion, nor do they need to meet the unofficial standard we've adopted (which the proof for 'which domain we're acting in' does, and therefore plays nicely with the motions and disputes).

### DIFF
--- a/contracts/colony/ColonyFunding.sol
+++ b/contracts/colony/ColonyFunding.sol
@@ -245,68 +245,6 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
     moveFundsBetweenPotsFunctionality(_fromPot, _toPot, _amount, _token);
   }
 
-  function moveFundsBetweenPotsFunctionality(
-    uint256 _fromPot,
-    uint256 _toPot,
-    uint256 _amount,
-    address _token
-  )
-  internal {
-    FundingPot storage fromPot = fundingPots[_fromPot];
-    FundingPot storage toPot = fundingPots[_toPot];
-
-    fromPot.balance[_token] = sub(fromPot.balance[_token], _amount);
-    toPot.balance[_token] = add(toPot.balance[_token], _amount);
-
-    if (_fromPot == 1){
-      // If we're moving from the root pot, then check we haven't dropped below what we need
-      // to cover any approvals that we've made.
-      require(fromPot.balance[_token] >= tokenApprovalTotals[_token], "colony-funding-too-many-approvals");
-    }
-
-    // If this pot is associated with a Task or Expenditure, prevent money
-    // being taken from the pot if the remaining balance is less than
-    // the amount needed for payouts, unless the task was cancelled.
-    if (fromPot.associatedType == FundingPotAssociatedType.Task) {
-      require(
-        tasks[fromPot.associatedTypeId].status == TaskStatus.Cancelled ||
-        fromPot.balance[_token] >= fromPot.payouts[_token],
-        "colony-funding-task-bad-state"
-      );
-    }
-    if (fromPot.associatedType == FundingPotAssociatedType.Expenditure) {
-      require(
-        expenditures[fromPot.associatedTypeId].status == ExpenditureStatus.Cancelled ||
-        fromPot.balance[_token] >= fromPot.payouts[_token],
-        "colony-funding-expenditure-bad-state"
-      );
-    }
-
-    if (
-      fromPot.associatedType == FundingPotAssociatedType.Expenditure ||
-      fromPot.associatedType == FundingPotAssociatedType.Payment ||
-      fromPot.associatedType == FundingPotAssociatedType.Task
-    ) {
-      uint256 fromPotPreviousAmount = add(fromPot.balance[_token], _amount);
-      updatePayoutsWeCannotMakeAfterPotChange(_fromPot, _token, fromPotPreviousAmount);
-    }
-
-    if (
-      toPot.associatedType == FundingPotAssociatedType.Expenditure ||
-      toPot.associatedType == FundingPotAssociatedType.Payment ||
-      toPot.associatedType == FundingPotAssociatedType.Task
-    ) {
-      uint256 toPotPreviousAmount = sub(toPot.balance[_token], _amount);
-      updatePayoutsWeCannotMakeAfterPotChange(_toPot, _token, toPotPreviousAmount);
-    }
-
-    if (_toPot == 0 ) {
-      nonRewardPotsTotal[_token] = sub(nonRewardPotsTotal[_token], _amount);
-    }
-
-    emit ColonyFundsMovedBetweenFundingPots(msg.sender, _fromPot, _toPot, _amount, _token);
-  }
-
   function claimColonyFunds(address _token) public stoppable {
     uint toClaim;
     uint feeToPay;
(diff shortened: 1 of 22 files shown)


contracts/colony/ColonyFunding.sol
```diff
@@ -205,6 +205,27 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
...
+    address _token
+  )
+  public
+  stoppable
+  authDomain(_permissionDomainId, _childSkillIndex, _domainId)
+  authDomain(_domainId, _fromChildSkillIndex, getDomainFromFundingPot(_fromPot))
+  authDomain(_domainId, _toChildSkillIndex, getDomainFromFundingPot(_toPot))
+  validFundingTransfer(_fromPot, _toPot)
+  {
+    moveFundsBetweenPotsFunctionality(_fromPot, _toPot, _amount, _token);
```
kronosapiens, line 226: I like this but I would prefer an approach which wrapped the unmodified `moveFundsBetweenPots`, so something along the lines of:

```
 function moveFundsBetweenPots(
    uint256 _permissionDomainId,
    uint256 _childSkillIndex,
    uint256 _domainId,
    uint256 _fromChildSkillIndex,
    uint256 _toChildSkillIndex,
    uint256 _fromPot,
    uint256 _toPot,
    uint256 _amount,
    address _token
  )
  public
  stoppable
  authDomain(_permissionDomainId, _childSkillIndex, _domainId)
{
  moveFundsBetweenPots(_domainId, _fromChildSkillIndex, _toChildSkillIndex, _fromPot, _toPot, _amount, _token);
}
```

That way the original function is left unchanged, and we have a special function for use in voting.

contracts/extensions/VotingReputation.sol
```diff
@@ -232,6 +242,9 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
 
     address target = getTarget(_altTarget);
     bytes4 action = getSig(_action);
+    if (action == bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,address)"))){
+      require(false, "voting-rep-disallowed-function");
```
kronosapiens, line 246: Could be `require(action != ..., "voting...")`

contracts/colony/ColonyFunding.sol
area: If I'm understanding you correctly, don't think what you're proposing is possible, because of the reputation proofs. In the original function, the proofs prove the relationship between a domain where I have the permission, and the domains where we're moving funds from and to. In this new function, the reputation proofs prove the relationship between the domain where I'm saying I'm acting, and the domains we're moving the funds from and to. There is no universal way to convert from one to the other, and so we're not able to wrap them the way you're proposing.

I'm also very much in favour of deprecating that original function (indeed, I already have it marked as such in this PR), and switching to using the new `moveFundsBetweenPots` introduced here.

contracts/extensions/VotingReputation.sol
area: Wow, yeah, not sure what came over me here.

contracts/colony/ColonyFunding.sol
```diff
@@ -205,6 +205,27 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
...
+    uint256 _toChildSkillIndex,
+    uint256 _fromPot,
+    uint256 _toPot,
+    uint256 _amount,
+    address _token
+  )
+  public
+  stoppable
+  authDomain(_permissionDomainId, _childSkillIndex, _domainId)
+  authDomain(_domainId, _fromChildSkillIndex, getDomainFromFundingPot(_fromPot))
```
kronosapiens, line 222: Looking at this again, I'm realizing that I don't see how this works -- if the caller has the permission in `_permissionDomainId`, then how can this call succeed? Even if the index is correct, the caller may not have the permission in `_domainId` and so this check fails.

area: You're absolutely right, the second two `authDomain`s should only be `validateDomainInheritance`, and not the full `authDomain`s

contracts/extensions/VotingReputation.sol
```diff
@@ -314,29 +322,7 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
...
-    bytes memory _action,
-    bytes memory _key,
-    bytes memory _value,
-    uint256 _branchMask,
-    bytes32[] memory _siblings
-  )
-    public
-  {
-    createDomainMotion(_domainId, _childSkillIndex, address(0x0), _action, _key, _value, _branchMask, _siblings);
+    createMotion(_domainId, _childSkillIndex, address(0x0), _action, _key, _value, _branchMask, _siblings);
```
kronosapiens, line 325: Didn't we deprecate domain motions in #949  ?

test/contracts-network/colony-funding.js
```diff
@@ -95,10 +95,29 @@ contract("Colony Funding", (accounts) => {
       expect(pot2Balance).to.eq.BN(51);
     });
 
+    it("when moving tokens between pots, should respect permission inheritance", async () => {
```
kronosapiens, line 98: I wonder why this test didn't catch the `authDomain` issue.

area: It didn't exist :stuck_out_tongue:

contracts/extensions/VotingReputation.sol
area: Yes? I'm not sure what you're saying here :disappointed:

contracts/extensions/VotingReputation.sol
```diff
@@ -957,6 +943,17 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
     return colonyNetwork.getChildSkillId(permissionSkillId, childSkillIndex);
   }
 
+  function getActionPermissionSkillId(bytes memory _action) internal view returns (uint256) {
```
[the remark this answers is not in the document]
area: _Mea cupla_, it's a leftover from my first implementation of a solution here. Will be nixed with great prejudice.

contracts/extensions/VotingReputation.sol
kronosapiens: Oh rather, didn't we add support for `altTargets`? Why is this PR changing the interface?

area: Oh, I see. The version of `createDomainMotion` that took `altTarget` was both added and deprecated in that PR, which shouldn't have happened. I'm removing it so it's never released.

contracts/colony/ColonyFunding.sol
```diff
@@ -220,6 +242,16 @@ contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
   authDomain(_permissionDomainId, _toChildSkillIndex, getDomainFromFundingPot(_toPot))
   validFundingTransfer(_fromPot, _toPot)
   {
+    moveFundsBetweenPotsFunctionality(_fromPot, _toPot, _amount, _token);
+  }
+
+  function moveFundsBetweenPotsFunctionality(
```
kronosapiens, line 248: Can this be moved further down with the rest of the `internal` functions?

contracts/extensions/VotingReputation.sol
```diff
@@ -214,6 +214,17 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
 
   // Public functions (interface)
 
+  bytes32 constant oldMoveFundsSig = bytes4(keccak256("moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,address)"));
```
kronosapiens, line 217: Can we put in `ALL_CAPS` to denote a constant?

kronosapiens (approved):

PR #959, opened 2021-05-19
Change how rewards are calculated for vote revealers

area:
These changes mean that more of the reward pool will be paid out - essentially, the pool is split (notionally) between all those who voted, rather than all of those who could have voted. Note that you are still required to reveal your vote to receive the rewards, so it remains possible for much of the pool to not be paid out if many people who voted do not reveal their vote for whatever reason.

I've also added a new function called `getVoterRewardRange` which can be used to see what the potential payout for a user will be should they choose to vote and then reveal.

### DIFF
--- a/contracts/extensions/VotingReputation.sol
+++ b/contracts/extensions/VotingReputation.sol
@@ -804,6 +804,8 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
   }
 
   /// @notice Get the voter reward
+  /// NB This function will only return a meaningful value if in the reveal state.
+  /// Prior to the reveal state, getVoterRewardRange should be used.
   /// @param _motionId The id of the motion
   /// @param _voterRep The reputation the voter has in the domain
   /// @return The voter reward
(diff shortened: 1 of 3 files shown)


kronosapiens (approved):
Nicely done

contracts/extensions/VotingReputation.sol
```diff
@@ -804,16 +804,47 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
...
   function getVoterReward(uint256 _motionId, uint256 _voterRep) public view returns (uint256) {
     Motion storage motion = motions[_motionId];
-    uint256 fractionUserReputation = wdiv(_voterRep, motion.skillRep);
+    uint256 fractionUserReputation = wdiv(_voterRep, motion.repSubmitted);
     uint256 totalStake = add(motion.stakes[YAY], motion.stakes[NAY]);
     return wmul(wmul(fractionUserReputation, totalStake), voterRewardFraction);
   }
 
+  /// @notice Get the range of potential rewards for a voter on a specific motion, intended to be
+  /// used when the motion is in the reveal state.
```
alicjakujawa, line 820: shouldn't be here in the vote state?

contracts/extensions/VotingReputation.sol
```diff
@@ -804,16 +804,47 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
...
+  /// Once a motion is in the reveal state the reward is known, and getVoterRewardRange should be used.
+  /// @param _motionId The id of the motion
```
alicjakujawa, line 822: in reveal state we should use getVoterReward as reward is known

PR #960, opened 2021-05-20
Per-user Coin Machine limits

### LINKED ISSUE #952: Maximum per-user purchases for Coin Machine
We would like to introduce a parameter governing a maximum amount of tokens an address can purchase, expressed as a percentage of the total tokens sold. A value of 100% is therefore no limit.

kronosapiens:
Closes #952 

- Introduce a new initialisation parameter, `userLimitFraction`, which governs what % of the total sold amount any user can buy. If the total amount of tokens sold is less than `maxPerPeriod`, that value is used.
- Introduce `getUserLimit` which returns the total amount of tokens any given user can purchase, across many periods.
- Replace `getNumAvailable` with `getSellableTokens`, which does not take into account the token balance, only the remaining tokens in the period.
- Introduce `getMaxPurchase` which  returns the amount of tokens a user can purchase in the current period, taking into account all factors (token balance, user limits, and remaining tokens in the period)

### DIFF
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -61,9 +61,8 @@ contract CoinMachine is ColonyExtension {
 
   bool evolvePrice; // If we should evolve the price or not
 
+  uint256 userLimitFraction; // Limit an address can buy of the total amount (in WADs)
   uint256 soldTotal; // Total tokens sold by the coin machine
-  uint256 userLimit; // Limit any address can buy of the total amount (as WAD-denominated fraction)
-
   mapping(address => uint256) soldUser; // Tokens sold to a particular user
 
   // Modifiers
@@ -138,7 +137,7 @@ contract CoinMachine is ColonyExtension {
     uint256 _windowSize,
     uint256 _targetPerPeriod,
     uint256 _maxPerPeriod,
-    uint256 _userLimit,
+    uint256 _userLimitFraction,
     uint256 _startingPrice,
     address _whitelist
   )
@@ -153,8 +152,8 @@ contract CoinMachine is ColonyExtension {
     require(_windowSize <= 511, "coin-machine-window-too-large");
     require(_targetPerPeriod > 0, "coin-machine-target-too-small");
     require(_maxPerPeriod >= _targetPerPeriod, "coin-machine-max-too-small");
-    require(_userLimit <= WAD, "coin-machine-limit-too-large");
-    require(_userLimit > 0, "coin-machine-limit-too-small");
+    require(_userLimitFraction <= WAD, "coin-machine-limit-too-large");
+    require(_userLimitFraction > 0, "coin-machine-limit-too-small");
 
     token = _token;
     purchaseToken = _purchaseToken;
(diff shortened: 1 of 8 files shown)


kronosapiens:
@area we did, but thinking more I see no reason not to leave those two features orthogonal. In practice, a limit without a whitelist would be ineffectual, but I would rather that be a choice that users make.

area:
I would argue a limit without a whitelist is effective at being biased towards those who know how to work around the restriction.

test/extensions/coin-machine.js
```diff
@@ -382,6 +392,56 @@ contract("Coin Machine", (accounts) => {
       expect(balance).to.eq.BN(maxPerPeriod);
     });
 
+    it("cannot buy more than ther user limit allows", async () => {
```
area, line 395: As far as I can tell, this test still passes even if you remove the checks on the limits in the contracts? So it's not really testing what it claims :stuck_out_tongue:

contracts/extensions/CoinMachine.sol
```diff
@@ -341,6 +359,15 @@ contract CoinMachine is ColonyExtension {
     );
   }
 
+  /// @notice Get the maximum amount of tokens a user can purchase
+  function getMaxPurchase(address _user) public view returns (uint256) {
```
area, line 363: This function doesn't work at the start of a new period. Should this function also take in to account the remaining in the window, the token balance, etc? Basically fold in (most of) the `min` calls.

kronosapiens: My preference is to keep this context-free as much as possible -- returning the maximum the user can buy in general, and letting questions of period max, etc be handled elsewhere.

kronosapiens: Also, I don't follow that the function doesn't work at the start of a new period?

area: > Also, I don't follow that the function doesn't work at the start of a new period?
Me neither. I'm going to blame the fumes I was casually inhaling all day.

That said, I do think there should be a separate function that does include all appropriate limits, or at the very least check with the frontend that they're happy doing all the required calculations, and the comment here should make it _very_ clear that the returned value might not, actually, be the amount of tokens you can purchase right now, which is a bit of a surprise for a public function called `getMaxPurchase`.

kronosapiens: Fair, maybe renaming the function `getUserLimit` would be better, and the variable `userLimitFraction`.

contracts/extensions/CoinMachine.sol
```diff
@@ -190,17 +199,23 @@ contract CoinMachine is ColonyExtension {
       "coin-machine-unauthorised"
     );
 
-    uint256 tokenBalance = getTokenBalance();
-    uint256 numTokens = min(min(_numTokens, maxPerPeriod - activeSold), tokenBalance);
+    uint256 maxPurchase = getMaxPurchase(msg.sender);
+    uint256 numTokens = min(maxPurchase, _numTokens);
```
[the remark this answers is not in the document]
area: Or maybe return - such a transaction could be sent in good faith, and a return would stop the events being emitted and use as little gas as possible.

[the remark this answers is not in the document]
kronosapiens: Yes, I agree with the latter thinking. I generally lean to not erroring for zero values, since transactions might be sent in good faith (i.e. the period just sold out).

area (dismissed):

area:
This _can_ be merged once rebased.

area (approved):

area (approved):

PR #962, opened 2021-05-28
Introduce metatransaction support to the contracts

area:
## Description

This PR introduces metatransactions to the majority of the contracts a user will be interacting with - Colony, ColonyNetwork, TokenLocking, extensions. Metatransactions mean that a user can sign a message, which is then broadcast to the network in a transaction paid for by a third party. This means that in terms of onboarding, users do not need tokens to pay for gas. I also believe this implementation brings in [Network Agnostic Transactions](https://docs.biconomy.io/products/enable-gasless-transactions/network-agnostic-transactions) though I have not properly investigated that in any depth.

~~The big exception to this is the token contract, and nor is it possible to deploy a new token contract. Xdai bridged tokens all appear to have [EIP-2612 permit](https://eips.ethereum.org/EIPS/eip-2612) support, so I think a new version of our token contract with that support, that can be deployed via a contract transaction from the network, will solve those issues. Those changes are not in this PR, however.~~

## Implementation

This implementation is based on the `BasicMetatransaction.sol` supplied by [Biconomy](https://github.com/bcnmy/metatx-standard/blob/basic-signature-metatx/src/contracts/BasicMetaTransaction.sol). It could not be a straight import, however, as the implementation assumes that any self-targeting transaction is a metatransaction, which is not the case for our contracts - `finishUpgrade` being a notable example, which specifically _has_ to be implemented as a self-calling transaction to work.

The way the default implementation works is that, when performing a metatransaction, the address the transaction is being conducted on behalf on is appended to `msg.data`, and extracted when `msgSender()` is called and `msg.sender` is `address(this)`. All instances where you would use `msg.sender` are replaced with a call to `msgSender(), which returns `msg.sender` in the case of a normal transaction and the appended address in the case of a metatransaction.

While I believe it is the case that the original implementation would have worked if we had no self-calling transactions that also didn't want to use `msg.sender`, which I think might even be the case in the current codebase, that felt like a restriction that was a) hard to script checks for and b) would be very easy to do down the line unintentionally, so I went for a more robust slight re-implementation. I append a specific `bytes32` as well as the address we are acting on behalf of, and even if `msg.sender` is `address(this)`, if that bytes32 is not there as the penultimate `bytes32` of `msg.data`, `msgSender` returns `msg.sender` as normal.

Note that, in general, you can append whatever you want on to `msg.data`, so if someone were able to engineer a custom `msg.data` and have `msg.sender` be equal to `address(this)` on a contract that supports meta-transactions, they could act as if they were anyone by appending to `msg.sender` appropriately. I don't believe this is currently possible (we disallow arbitrary transactions on a colony targeting itself), but making sure that's the case should be a top concern when reviewing this PR.

Another issue with the implementation was trying to keep the code DRY as much as possible. The default implementation accesses `metatransactionNonces` mapping directly which is sadly not possible for us to keep here if we want to keep the importing pattern, and only one copy of the meta-transaction code, because that mapping is in different storage slots in different contracts. I ended up introducing virtual functions to be overridden in the appropriate contracts to access that mapping. 

## Recovery mode

Another sticking point with metatransactions is their interaction with recovery mode. The way that replays of transactions are prevented is by using nonces (as is traditional), but as-it-was, recovery mode would be able to decrement the nonce and replay the transaction repeatedly - obviously a serious issue for anyone using metatransactions, as such a transaction could be a OneTxPayment or similar. The existing recovery mode had no way of protecting slots in a mapping like this.

This has been mitigated by a new style of protection, though it requires functions that touch those mappings to know that they have to be protected. Essentially, if the slot _s_ is being edited in recovery mode, the value in slot `keccak(s)` is also accessed. If either slot is found to be equal to a magic constant, then the slot is unable to be edited in recovery mode - either the slot is protected, or the slot signifies another slot is protected. In either case, the edit is disallowed. When a variable in a protected mapping is set, `keccak(s)` is also set to the magic constant. This means that the slot is only protected once a metatransaction has been executed, and the nonce goes from 0 to 1, but I can't see a way for that to be abused.

Potentially, all existing slots that are protected via the slots being hard-coded in the contracts could move to this protection, but that's not been done here.


contracts/colony/Colony.sol
```diff
@@ -473,25 +378,39 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
 
   }
 
+  function getMetatransactionNonce(address userAddress) override public view returns (uint256 nonce){
```
kronosapiens, line 381: Let's stick to convention -- argument should be `_user`.

contracts/colony/Colony.sol
```diff
@@ -473,25 +378,39 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
+    return metatransactionNonces[userAddress];
+  }
+
+  function incrementMetatransactionNonce(address user) override internal {
```
kronosapiens, line 385: Argument should be `_user`.

contracts/colony/Colony.sol
```diff
@@ -473,25 +378,39 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
+    // We need to protect the metatransaction nonce slots, otherwise those with recovery
+    // permissions could replay metatransactions, which would be a disaster.
+    // What slot are we setting?
+    // This mapping is in slot 34 (see ColonyStorage.sol);
+    uint256 slot = uint256(keccak256(abi.encode(uint256(user), uint256(34))));
```
kronosapiens, line 390: Let's make this a constant instead of `34`.

contracts/colony/ColonyArbitraryTransaction.sol
```diff
@@ -0,0 +1,136 @@
...
+pragma solidity 0.7.3;
+pragma experimental ABIEncoderV2;
+
+import "./../common/ERC20Extended.sol";
+import "./../common/IEtherRouter.sol";
+import "./../common/MultiChain.sol";
+import "./../tokenLocking/ITokenLocking.sol";
+import "./ColonyStorage.sol";
+
+contract ColonyArbitraryTransaction is ColonyStorage, PatriciaTreeProofs, MultiChain {
```
kronosapiens, line 27: Why `PatriciaTreeProofs` and `MultiChain`?

contracts/colony/ColonyRecovery.sol
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+pragma solidity 0.7.3;
+pragma experimental "ABIEncoderV2";
+
+import "./../common/ContractRecovery.sol";
+import "./ColonyStorage.sol";
+
+
+contract ColonyRecovery is ColonyStorage, ContractRecovery { // ignore-swc-123
```
kronosapiens, line 25: Why?

contracts/colony/ColonyStorage.sol
```diff
@@ -102,6 +102,8 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, ColonyNetworkDataTypes
   // Mapping of token address -> total amount approved
   mapping (address => uint256 ) tokenApprovalTotals; // Storage slot 33
 
+  mapping(address => uint256) metatransactionNonces; // Storage slot 34
```
kronosapiens, line 105: I'd say create the constant here:

```
uint256 constant METATRANSACTION_NONCES_SLOT = 34;
```

contracts/colony/ColonyStorage.sol
```diff
@@ -237,21 +239,21 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, ColonyNetworkDataTypes
     _;
   }
 
-  modifier auth override {
-    require(isAuthorized(msg.sender, 1, msg.sig), "ds-auth-unauthorized");
+  modifier auth virtual override {
```
kronosapiens, line 242: Why?

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -349,6 +350,20 @@ contract ColonyNetwork is ColonyNetworkStorage, MultiChain {
     emit TokenWhitelisted(_token, _status);
   }
 
+  function getMetatransactionNonce(address userAddress) override public view returns (uint256 nonce){
```
kronosapiens, line 353: `_user`

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -349,6 +350,20 @@ contract ColonyNetwork is ColonyNetworkStorage, MultiChain {
...
+    return metatransactionNonces[userAddress];
+  }
+
+  function incrementMetatransactionNonce(address user) override internal {
```
kronosapiens, line 357: `_user`

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -349,6 +350,20 @@ contract ColonyNetwork is ColonyNetworkStorage, MultiChain {
...
+    // We need to protect the metatransaction nonce slots, otherwise those with recovery
+    // permissions could replay metatransactions, which would be a disaster.
+    // What slot are we setting?
+    // This mapping is in slot 41 (see ColonyNetworkStorage.sol);
+    uint256 slot = uint256(keccak256(abi.encode(uint256(user), uint256(41))));
```
kronosapiens, line 362: Same thing about the constant.

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,64 @@
...
+  function getMetatransactionNonce(address userAddress) public view virtual returns (uint256 nonce);
+
+  // NB if implementing this functionality in a contract with recovery mode,
+  // you MUST prevent the metatransaction nonces from being editable with recovery mode.
+  function incrementMetatransactionNonce(address user) internal virtual;
+
+  event MetaTransactionExecuted(address userAddress, address payable relayerAddress, bytes functionSignature);
+
+
+  function getChainID() public pure returns (uint256) {
```
kronosapiens, line 17: Why not inherit from `MultiChain`?

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,64 @@
+pragma solidity 0.7.3;
+
+import "../../lib/dappsys/math.sol";
+import "./MetaTransactionMsgSender.sol";
+
+abstract contract BasicMetaTransaction is DSMath, MetaTransactionMsgSender {
+
+  function getMetatransactionNonce(address userAddress) public view virtual returns (uint256 nonce);
```
kronosapiens, line 8: Same comments on argument naming conventions.

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,64 @@
...
+   * personal_sign method.
+   * @param userAddress Address of user trying to do meta transaction
+   * @param functionSignature Signature of the actual function to be called via meta transaction
+   * @param sigR R part of the signature
+   * @param sigS S part of the signature
+   * @param sigV V part of the signature
+   */
+  // slither-disable-next-line locked-ether
+  function executeMetaTransaction(address userAddress, bytes memory functionSignature,
+      bytes32 sigR, bytes32 sigS, uint8 sigV) public payable returns(bytes memory) {
```
kronosapiens, line 38: Space after `returns`.

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,64 @@
...
+
+      require(verify(userAddress, getMetatransactionNonce(userAddress), getChainID(), functionSignature, sigR, sigS, sigV), "Signer and signature do not match");
```
kronosapiens, line 40: Comment does not follow convention.

contracts/common/CommonStorage.sol
```diff
@@ -18,18 +18,29 @@
...
   uint256 constant UINT256_MAX = 2**256 - 1;
 
   uint256 constant AUTHORITY_SLOT = 0;
   uint256 constant OWNER_SLOT = 1;
   uint256 constant RESOLVER_SLOT = 2;
 
+  bytes32 constant PROTECTED = keccak256("Recovery Mode Protected Slot");
+
+  function protectSlot(uint256 _slot) internal always {
+    uint256 flagSlot = uint256(keccak256(abi.encodePacked("RECOVERY_PROTECTED", _slot)));
```
kronosapiens, line 37: Why not just hash the slot itself?

contracts/common/ContractRecovery.sol
```diff
@@ -26,14 +26,23 @@ import "./IRecovery.sol";
...
     require(_slot != OWNER_SLOT, "colony-common-protected-variable");
     require(_slot != RESOLVER_SLOT, "colony-common-protected-variable");
 
+    bytes32 flag;
+    uint256 flagSlot = uint256(keccak256(abi.encodePacked("RECOVERY_PROTECTED", _slot)));
+    assembly {
+      flag := sload(flagSlot)
+    }
+
+    // uint256 protectFlag = uint256(PROTECTED);
```
kronosapiens, line 43: Delete

contracts/common/MetaTransactionMsgSender.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity 0.7.3;
+
+import "../../lib/dappsys/math.sol";
+
+abstract contract MetaTransactionMsgSender is DSMath {
```
kronosapiens, line 5: Why not include this function in `BasicMetaTransaction`?

contracts/extensions/CoinMachine.sol
```diff
@@ -61,10 +61,19 @@ contract CoinMachine is ColonyExtension {
 
   bool evolvePrice; // If we should evolve the price or not
 
+  mapping(address => uint256) metatransactionNonces;
```
kronosapiens, line 64: Can we not roll these into `ColonyExtension` as a base class? Is it because it'll mess up the storage slots?

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -22,37 +22,46 @@ import "./../../lib/dappsys/math.sol";
...
-      sub(userLocks[_token][msg.sender].balance, _amount) >= totalObligations[msg.sender][_token],
+      sub(userLocks[_token][msgSender()].balance, _amount) >= totalObligations[msgSender()][_token],
       "colony-token-locking-excess-obligation"
     );
     _;
   }
 
   // Public functions
 
+  function getMetatransactionNonce(address userAddress) override public view returns (uint256 nonce){
```
kronosapiens, line 57: Is there no way to inherit these functions from elsewhere? This repetition is a buzzkill!

helpers/test-data-generator.js
```diff
@@ -393,3 +395,28 @@ export async function setupColony(colonyNetwork, tokenAddress) {
   const colony = await IColony.at(colonyAddress);
   return colony;
 }
+
+export async function getMetatransactionParameters(txData, userAddress, targetAddress) {
+  // const txData = await tokenLocking.contract.methods["deposit(address,uint256,bool)"](token.address, usersTokens, true).encodeABI();
```
kronosapiens, line 400: Delete

test/contracts-network/colony-network-recovery.js
```diff
@@ -195,6 +202,45 @@ contract("Colony Network Recovery", (accounts) => {
       await colonyNetwork.exitRecoveryMode();
     });
 
+    it("should not allow editing of a protected variable in a mapping", async () => {
```
kronosapiens, line 205: Nice

contracts/common/CommonStorage.sol
area: That would collide with actual storage in the event that we wanted to protect the length of an array.

contracts/extensions/CoinMachine.sol
area: Yep.

contracts/tokenLocking/TokenLocking.sol
area: The problem is they're all in different storage slots, so they each need their own declaration.

contracts/common/MetaTransactionMsgSender.sol
area: Any contract that inherits `BasicMetaTransaction` needs to define the virtual functions, and `msgSender` is needed basically everywhere, so split up the functionality between `BasicMetaTransaction` (functionality required to execute a metatransaction) and `MetaTransactionMsgSender` (the functionality required to support metatransactions) in order to reduce the repetition of the virtual function declarations.

area:
Contrary to my original PR, this now adds MetaTxToken.sol, which is a token that supports both metatransactions, in line with those introduced to the rest of the contracts, and [EIP-2612 signed permit calls](https://eips.ethereum.org/EIPS/eip-2612).

contracts/colony/ColonyStorage.sol
kronosapiens: Bump

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+  function getMetatransactionNonce(address _user) override public view returns (uint256 nonce){
+    return metatransactionNonces[_user];
+  }
+
+  function incrementMetatransactionNonce(address _user) override internal {
+    metatransactionNonces[_user] = add(metatransactionNonces[_user], 1);
+  }
+
+  event Mint(address indexed guy, uint wad);
+  event Burn(address indexed guy, uint wad);
```
kronosapiens, line 177: I'd put the events first here as well

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+    uint chainId;
+    assembly {
+        chainId := chainid()
+    }
+
+    DOMAIN_SEPARATOR = keccak256(
+        abi.encode(
+            keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
+            keccak256(bytes(name)),
+            keccak256(bytes("1")),
```
kronosapiens, line 201: Why?

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+            chainId,
+            address(this)
+        )
+    );
+  }
+
+  function transferFrom(address src, address dst, uint wad) public
```
kronosapiens, line 208: Shouldn't the base meta contract provide all this?

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+  auth
+  {
+    locked = false;
+  }
+
+  // Pinched from https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol
+  // Which is also licenced under GPL V3
+
+  // keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
+  bytes32 public constant PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
```
kronosapiens, line 252: Why not just set the constant to the hash result directly?

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+
+  function permit(address owner, address spender, uint value, uint deadline, uint8 v, bytes32 r, bytes32 s) external unlocked {
+      require(deadline >= block.timestamp, "colony-token-expired-deadline");
+
+      bytes32 digest = keccak256(
+          abi.encodePacked(
+              "\x19\x01",
```
kronosapiens, line 259: Would love this assigned to a variable

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+              DOMAIN_SEPARATOR,
+              keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, value, metatransactionNonces[owner]++, deadline))
```
kronosapiens, line 261: This `++` usage is slightly contra to our convention but I'll allow it

That said, if we're doing it here I'm not sure why we wouldn't do it for all the metatransaction nonces. There's not really a possibility of overflow...

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+          )
+      );
+      address recoveredAddress = ecrecover(digest, v, r, s);
+      require(recoveredAddress != address(0) && recoveredAddress == owner, "colony-token-invalid-signature");
+      _approvals[owner][spender] = value;
+      emit Approval(owner, spender, value);
```
kronosapiens, line 267: Newline before event

contracts/testHelpers/TestExtensions.sol
```diff
@@ -37,6 +37,16 @@ abstract contract TestExtension is ColonyExtension {
   function uninstall() public override auth {
     selfdestruct(address(uint160(address(colony))));
   }
+
+  mapping(address => uint256) metatransactionNonces;
```
kronosapiens, line 41: If we added this to `ColonyExtension` we could drop this

migrations/8_setup_meta_colony.js
```diff
@@ -62,23 +62,45 @@ module.exports = async function (deployer, network, accounts) {
...
+  const ColonyArbitraryTransaction = artifacts.require("./ColonyArbitraryTransaction");
 
   const colony = await Colony.new();
   const colonyFunding = await ColonyFunding.new();
   const colonyExpenditure = await ColonyExpenditure.new();
   const colonyRoles = await ColonyRoles.new();
   const colonyTask = await ColonyTask.new();
   const colonyPayment = await ColonyPayment.new();
-  const contractRecovery = await ContractRecovery.deployed();
+  const contractRecovery = await ContractRecovery.new();
```
kronosapiens, line 73: Why this change?

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+chai.use(bnChai(web3.utils.BN));
+
+const EtherRouter = artifacts.require("EtherRouter");
+const IColony = artifacts.require("IColony");
+const IColonyNetwork = artifacts.require("IColonyNetwork");
+const MetaTxToken = artifacts.require("MetaTxToken");
+
+const ADDRESS_ZERO = ethers.constants.AddressZero;
+
+contract("MetaTx token", (addresses) => {
```
kronosapiens, line 18: Why two words?

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+  let metatxToken;
+  let colonyNetwork;
+
+  before(async () => {
+    const etherRouter = await EtherRouter.deployed();
+    colonyNetwork = await IColonyNetwork.at(etherRouter.address);
+  });
+
+  beforeEach(async () => {
+    metatxToken = await MetaTxToken.new("Test", "TEST", 18);
```
kronosapiens, line 29: Stylistically, shouldn't these variables be named `metaTxToken`?

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+    const { logs } = await colonyNetwork.createColony(metatxToken.address, 0, "", "");
+    const { colonyAddress } = logs.filter((x) => x.event === "ColonyAdded")[0].args;
```
kronosapiens, line 31: This should be a helper function frankly

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+    colony = await IColony.at(colonyAddress);
+    await colony.setRewardInverse(100);
+  });
+  describe("when using the contract directly", () => {
```
kronosapiens, line 35: Newline above?

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+    describe("when working with MetaTxToken, should behave like normal token", () => {
+      beforeEach("mint 1500000 tokens", async () => {
+        await metatxToken.unlock({ from: addresses[0] });
```
kronosapiens, line 38: Can we add `USER0 = addresses[0]` at the start of the file? Also why not `accounts` like we do elsewhere?

test/contracts-network/metatx-token.js
```diff
@@ -0,0 +1,725 @@
...
+        await checkErrorRevert(metatxToken.mint(addresses[0], 1500000, { from: addresses[2] }), "ds-auth-unauthorized");
+        const totalSupply = await metatxToken.totalSupply();
+        expect(totalSupply).to.be.zero;
+      });
+
+      it("should be able to burn others' tokens when they have approved them to do so", async () => {
+        await metatxToken.mint(addresses[1], 1500000, { from: addresses[0] });
+        await metatxToken.methods["approve(address,uint256)"](addresses[2], 500000, { from: addresses[1] });
+        // await metatxToken.approve(addresses[2], 500000, { FROM: addresses[1] });
+        // await metatxToken.burn(addresses[1], 500000, { from: addresses[2] });
```
kronosapiens, line 224: Delete

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,56 @@
...
+
+  // NB if implementing this functionality in a contract with recovery mode,
+  // you MUST prevent the metatransaction nonces from being editable with recovery mode.
+  function incrementMetatransactionNonce(address _user) internal virtual;
+
+  event MetaTransactionExecuted(address user, address payable relayerAddress, bytes functionSignature);
+
+  /**
+   * Main function to be called when user wants to execute meta transaction.
+   * The actual function to be called should be passed as param with name functionSignature
```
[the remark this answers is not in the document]
area: You're right, but this was lifted straight from the [canonical implementation](https://github.com/bcnmy/metatx-standard/blob/basic-signature-metatx/src/contracts/BasicMetaTransaction.sol). I can change it.

contracts/common/CommonStorage.sol
```diff
@@ -18,18 +18,29 @@
...
+    uint256 protectFlag = uint256(PROTECTED);
```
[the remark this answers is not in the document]
area: While we save a type coercion here, we then get two extra coercions in `ContractRecovery.sol` for the recovery event that's emitted (which takes `bytes32`s).

contracts/extensions/TokenSupplier.sol
```diff
@@ -39,6 +39,14 @@ contract TokenSupplier is ColonyExtension {
   uint256 tokenIssuanceRate;
   uint256 lastIssue;
   uint256 lastRateUpdate;
+  mapping(address => uint256) metatransactionNonces;
```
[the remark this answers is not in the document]
area: Existing extensions would have to stop inheriting from that contract.

contracts/metatxToken/MetaTxToken.sol
area: This is the version field in the [EIP-712](https://eips.ethereum.org/EIPS/eip-712) definition, which is used in the [PermittableTokenImplementation](https://blockscout.com/xdai/mainnet/address/0xf8D1677c8a0c961938bf2f9aDc3F3CFDA759A9d9/contracts) that tokens bridged on to xdai use. I'm mimicking that particular implementation so that the same Permit flow can be used in the front end for both types of tokens.

contracts/metatxToken/MetaTxToken.sol
area: We want it to be `locked`, so we reimplement it with that modifier and call `super`. This is copied straight from the ColonyToken repository.

contracts/metatxToken/MetaTxToken.sol
area: This came from the existing implementation, so I have no skin in this particular game, but I believe this is cheaper gas-wise.

contracts/metatxToken/MetaTxToken.sol
area: Why?

contracts/metatxToken/MetaTxToken.sol
area: Fair. Again, this came from the existing implementation. For what it's worth, I quite like the ++ pattern here (in that it guarantees a nonce can only be used once at a very low level), but yes, we could use it elsewhere.

migrations/8_setup_meta_colony.js
area: This got changed to `ColonyRecovery.new()` originally with my first crack, but when you pointed out we didn't need separate recovery contracts with the implementation, I changed all the `ColonyRecovery`s back to `ContractRecovery`, but obviously missed the `deployed/new` subtlety. Will be changed back.

contracts/colonyNetwork/ColonyNetworkAuction.sol
```diff
@@ -127,12 +129,20 @@ contract DutchAuction is DSMath, MultiChain {
...
     clnyToken = ERC20Extended(_clnyToken);
     token = ERC20Extended(_token);
   }
 
+  function getMetatransactionNonce(address userAddress) override public view returns (uint256 nonce){
+    return metatransactionNonces[userAddress];
+  }
+
+  function incrementMetatransactionNonce(address user) override internal {
+    metatransactionNonces[user] = add(metatransactionNonces[user], 1);
```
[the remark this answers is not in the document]
area: Yeah, it doesn't have a recovery mode.

contracts/metatxToken/MetaTxToken.sol
```diff
@@ -0,0 +1,269 @@
...
+            return false;
+        } else {
+            return authority.canCall(src, address(this), sig);
+        }
+    }
+}
+
+/// base.sol -- basic ERC20 implementation
+
+// Copyright (C) 2015, 2016, 2017  DappHub, LLC
```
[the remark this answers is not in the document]
area: I've split them up to keep the original notices in tact.

contracts/extensions/TokenSupplier.sol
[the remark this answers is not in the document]
kronosapiens: I was thinking about it this morning and I think that's a fair trade -- rename the current `ColonyExtension` to `ColonyExtensionNoMeta` and use that for existing extensions, and then have the new `ColonyExtension` moving forward. Assuming there are more extensions in our future it will be well worth it.

contracts/metatxToken/MetaTxToken.sol
kronosapiens: To add some context as to why these particular characters.

contracts/metatxToken/MetaTxToken.sol
kronosapiens: I'd be happy seeing it in all the metatransaction nonce updates.

contracts/extensions/TokenSupplier.sol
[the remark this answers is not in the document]
area: Okay. And you think that should be in this PR?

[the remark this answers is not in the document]
kronosapiens: Mm I could go either way. I could see it in this PR but also happy to put together a quickie PR after this is merged.

area:
This PR now includes a service that can be used to broadcast (and pay for) meta-transactions we want to. Those transactions are:

* Any transaction to a 'core' contract - TokenLocking, ColonyNetwork, any Colony, any Extension
* Any transaction sending or giving control of tokens to one of those contracts.

kronosapiens:
Can we make this broadcaster a separate PR? It's a lot of code to introduce into an already large feature.

area:
This PR now includes a function on ColonyNetwork that allows a token to be deployed via metatransaction.

contracts/extensions/CoinMachine.sol
```diff
@@ -65,10 +65,19 @@ contract CoinMachine is ColonyExtension {
   uint256 soldTotal; // Total tokens sold by the coin machine
   mapping(address => uint256) soldUser; // Tokens sold to a particular user
 
+  mapping(address => uint256) metatransactionNonces;
+  function getMetatransactionNonce(address userAddress) override public view returns (uint256 nonce){
+    return metatransactionNonces[userAddress];
+  }
+
+  function incrementMetatransactionNonce(address user) override internal {
+    metatransactionNonces[user]++;
```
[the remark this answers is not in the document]
area: I feel like you're thin-end-of-the-wedging me here. [You originally](https://github.com/JoinColony/colonyNetwork/pull/962#discussion_r661865542) only suggested expanding it to the extensions.

I only _like_ it in the place it was originally, where we wish to increment and use the new value simultaneously. For a pure increment, I think `+= 1` and/or `add(... , 1)` are both better.

contracts/colony/ColonyArbitraryTransaction.sol
```diff
@@ -0,0 +1,165 @@
...
+          success = false;
+        }
+      }
+      require(success, "colony-arbitrary-transaction-failed");
+    }
+    return true;
+  }
+
+  function makeSingleArbitraryTransaction(address _to, bytes memory _action)
+  external stoppable self
```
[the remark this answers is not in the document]
area: Try / catch only works with external calls, so in order to accommodate the `_strict` paramter in `makeArbitraryTransactions` it has to be an external call.

contracts/colony/ColonyStorage.sol
area: I definitely did deal with this... I assume I lost it in a rebase somewhere.

contracts/extensions/CoinMachine.sol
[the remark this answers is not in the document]
kronosapiens: Ok, I'll drop it then.

kronosapiens (dismissed):
Nicely done

contracts/common/BasicMetaTransaction.sol
```diff
@@ -0,0 +1,54 @@
...
+      bytes32 _sigR, bytes32 _sigS, uint8 _sigV) public payable returns (bytes memory) {
+
+      require(verify(_user, getMetatransactionNonce(_user), getChainId(), _payload, _sigR, _sigS, _sigV), "metatransaction-signer-signature-mismatch");
+      incrementMetatransactionNonce(_user);
+
+      // Append _user at the end to extract it from calling context
+      (bool success, bytes memory returnData) = address(this).call(abi.encodePacked(_payload, METATRANSACTION_FLAG, _user));
+      require(success, "colony-metatx-function-call-unsuccessful");
+
+      emit MetaTransactionExecuted(_user, msg.sender, _payload);
```
area, line 37: Should this be msgSender()? What happens if I make a metatransaction to make a metatransaction...?

contracts/colony/ColonyStorage.sol
```diff
@@ -234,19 +237,19 @@ contract ColonyStorage is CommonStorage, ColonyDataTypes, ColonyNetworkDataTypes
...
-    require(address(this) == msg.sender, "colony-not-self");
+    require(address(this) == msgSender(), "colony-not-self");
     _;
   }
 
   modifier onlyExtension() {
-    // Ensure msg.sender is a contract
-    require(isContract(msg.sender), "colony-sender-must-be-contract");
+    // Ensure msgSender() is a contract
+    require(isContract(msgSender()), "colony-sender-must-be-contract");
```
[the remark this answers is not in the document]
kronosapiens: Good point, by my understanding metatransactions can only replace "initial" transactions, signed by externally-owned accounts.

You're suggesting some kind of `assert` here, e.g. `assert(msg.sender = msgSender()`? I'm not opposed, I would put that in a separate line above the `require`.

contracts/common/BasicMetaTransaction.sol
kronosapiens: How would that work? I make a transaction, and then I wrap that transaction in the `executeMetaTransaction` call and then send that out as a metaTransaction? I don't see any reason to disallow it...?

area: Sure, but I'm saying that in that scenario, this `msg.sender` would be wrong?

kronosapiens: Sure, I see that. The "inner" metatransaction should be credited to the signer, not to the person who paid (who would get credit for the "outer" metatransaction call).

Say A signs the Tx, and then B signs the Metatransaction, which C executes, resulting in a second call to execute (where B should be credited as the sender). In that case, `msg.sender` would return C, which is incorrect. So, let's use `msgSender()`.

kronosapiens: Although... wouldn't `msg.sender` in the inner metatransaction return `address(this)`? Since the metatransaction contract is executing against itself?

contracts/colony/ColonyStorage.sol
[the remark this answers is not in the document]
area: > Good point, by my understanding metatransactions can only replace "initial" transactions, signed by externally-owned accounts.

Yep, I believe this to be true.

> You're suggesting some kind of assert here, e.g. assert(msg.sender = msgSender()? I'm not opposed, I would put that in a separate line above the require.

Yeah, exactly.

(Found this un-sent in a browser window, clearly got distracted yesterday)

contracts/common/BasicMetaTransaction.sol
area: It would, yes, which is wrong.

area:
`deployToken` has been renamed to make updates to colonyJS easier. They already had an override with that name which they attached to the contract object they create. The new name of `deployTokenViaNetwork` looks somewhat redundant on the contracts, but will make sense further up the stack (i.e. in colonyJS).

contracts/common/TokenAuthority.sol
```diff
@@ -0,0 +1,59 @@
...
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.7.3;
+
+import "../../lib/dappsys/auth.sol";
+
+
+contract TokenAuthority is DSAuthority {
```
kronosapiens, line 23: Why not use the existing contract in `lib`? I don't see any `msgSender()` calls here.

contracts/colony/ColonyStorage.sol
```diff
@@ -243,6 +243,7 @@ contract ColonyStorage is ColonyDataTypes, ColonyNetworkDataTypes, DSMath, Commo
 
   modifier onlyOwnExtension() {
     require(isOwnExtension(msgSender()), "colony-must-be-own-extension");
+    assert(msgSender() == msg.sender);
```
kronosapiens, line 246: I would put this first but that's just me.

kronosapiens (dismissed):

contracts/common/TokenAuthority.sol
area: The solidity version.

contracts/colony/ColonyStorage.sol
area: If it's first, it's hittable, and therefore not appropriate for an assert.

kronosapiens (approved):

PR #965, opened 2021-06-07
A number (currently four) of smallish changes and improvements

### LINKED ISSUE #933: Multi-transaction Motions
Can we motion on actions which involve multiple transactions in sequence? Let's find out.

### LINKED ISSUE #957: Should emit obligation events in TokenLocking
Currently, we do not emit events for obligations which is going to get us in hot water with the frontend eventually.

area:
## TokenLocking events

We were missing events for some function calls in TokenLocking. This adds them. Closes #957.

## VotingReputation reputation proof.

We were not checking that, in the case that the action domain and the voting domain were the same, that the supplied ChildSkillId was UINT256Max, which we do elsewhere. The only practical effect I can see right now was that motions-to-create-motions would be wrong, and that's a very niche case, but better to nip this one in the bud before it trips us up down the line.

## Multiple arbitrary transactions

Probably the biggest thing in this branch, but a pretty straightforward addition, as we've discussed - deprecating the old arbitrary transaction function and replacing it with something identical that loops over two arrays containing the target and the transaction data. Closes #933. The test is pretty basic - on reflection, perhaps it should be mint-and-claim, which was where this missing functionality was spotted during QA.

## Test extension management via vote

Couldn't see a reason why this wouldn't work, but was specifically queried by the frontend as they will be implementing it eventually, so added a test for a motion to add an extension. There are not currently corresponding tests for initialising / deprecating / uninstalling... should there be?

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -57,7 +57,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
   returns (bool)
   {
     require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
-    for (uint256 i = 0; i < _targets.length; i+= 1){
+    for (uint256 i; i < _targets.length; i+= 1){
       require(
         makeArbitraryTransactionFunctionality(_targets[i], _actions[i]),
         "colony-arbitrary-transaction-failed"
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -49,25 +49,34 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
   public stoppable auth
   returns (bool)
   {
-    return makeArbitraryTransactionFunctionality(_to, _action);
+    return this.makeSingleArbitraryTransaction(_to, _action);
   }
 
-  function makeArbitraryTransactions(address[] memory _targets, bytes[] memory _actions)
+  function makeArbitraryTransactions(address[] memory _targets, bytes[] memory _actions, bool _strict)
   public stoppable auth
   returns (bool)
   {
     require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
-    for (uint256 i; i < _targets.length; i+= 1){
-      require(
-        makeArbitraryTransactionFunctionality(_targets[i], _actions[i]),
-        "colony-arbitrary-transaction-failed"
-      );
+    for (uint256 i; i < _targets.length; i += 1){
+      bool success = true;
+      try this.makeSingleArbitraryTransaction(_targets[i], _actions[i]) returns (bool ret){
+        if (_strict){
+          success = ret;
+        }
+      } catch {
+        // We failed in a require, which is only okay if we're not in strict mode
+        if (_strict){
+          success = false;
+        }
+      }
+      require(success, "colony-arbitrary-transaction-failed");
     }
     return true;
   }
 
-  function makeArbitraryTransactionFunctionality(address _to, bytes memory _action)
-  internal
+  function makeSingleArbitraryTransaction(address _to, bytes memory _action)
+  external
+  self
   returns (bool)
   {
     // Prevent transactions to network contracts
@@ -486,7 +495,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
     ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
     bytes4 sig;
 
-    sig = bytes4(keccak256("makeArbitraryTransactions(address[],bytes[])"));
+    sig = bytes4(keccak256("makeArbitraryTransactions(address[],bytes[],bool)"));
     colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
   }
 
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -59,6 +59,7 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
     require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
     for (uint256 i; i < _targets.length; i += 1){
       bool success = true;
+      // slither-disable-next-line unused-return
       try this.makeSingleArbitraryTransaction(_targets[i], _actions[i]) returns (bool ret){
         if (_strict){
           success = ret;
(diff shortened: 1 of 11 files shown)


contracts/colony/Colony.sol
```diff
@@ -48,6 +48,27 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
+  }
+
+  function makeArbitraryTransactions(address[] memory _targets, bytes[] memory _actions)
+  public stoppable auth
+  returns (bool)
+  {
+    require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
+    for (uint256 i = 0; i < _targets.length; i+= 1){
+      require(
+        makeArbitraryTransactionFunctionality(_targets[i], _actions[i]),
```
kronosapiens, line 62: Why not just call `makeArbitraryTransaction`? Internal calls are authorized by default.

contracts/colony/Colony.sol
```diff
@@ -48,6 +48,27 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
+  {
+    return makeArbitraryTransactionFunctionality(_to, _action);
+  }
+
+  function makeArbitraryTransactions(address[] memory _targets, bytes[] memory _actions)
+  public stoppable auth
+  returns (bool)
+  {
+    require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
+    for (uint256 i = 0; i < _targets.length; i+= 1){
```
kronosapiens, line 60: No need to initialise `= 0` here.

contracts/tokenLocking/TokenLocking.sol
```diff
@@ -157,6 +157,7 @@ contract TokenLocking is TokenLockingStorage, DSMath { // ignore-swc-123
 
   function approveStake(address _user, uint256 _amount, address _token) public calledByColonyOrNetwork() {
     approvals[_user][_token][msg.sender] = add(approvals[_user][_token][msg.sender], _amount);
+    emit UserTokenApproved(_token, _user, msg.sender, _amount);
```
kronosapiens, line 160: Can we keep to the convention of having a newline before event emissions?

contracts/tokenLocking/TokenLockingDataTypes.sol
```diff
@@ -27,6 +27,10 @@ interface TokenLockingDataTypes {
   event UserTokenClaimed(address token, address user, uint256 amount);
   event UserTokenTransferred(address token, address user, address recipient, uint256 amount);
   event UserTokenWithdrawn(address token, address user, uint256 amount);
+  event UserTokenObligated(address token, address user, address obligatedBy, uint256 amount);
+  event UserTokenDeobligated(address token, address user, address obligatedBy, uint256 amount);
+  event UserTokenApproved(address token, address user, address approvedBy, uint256 amount);
+  event StakeTransferred(address token, address by, address from, address to, uint256 amount);
```
kronosapiens, line 33: I wonder if some of these values should be indexed?

contracts/colony/Colony.sol
kronosapiens: Ah, you want to deprecate the single version.

contracts/tokenLocking/TokenLockingDataTypes.sol
area: We only currently have the `TokenLocked` as an event with an indexed parameter. I don't recall a specific discussion, but I assume there was some logic here around these underlying events being triggered by colony / colonyNetwork, which would be emitting relevant events with indexed parameters?

kronosapiens: Ok, as long as we're happy about it.

contracts/colony/Colony.sol
kronosapiens: My only request is that we rename the internal function `makeArbitraryTransaction` when the time comes.

contracts/colony/Colony.sol
```diff
@@ -48,6 +48,27 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
+    return makeArbitraryTransactionFunctionality(_to, _action);
+  }
+
+  function makeArbitraryTransactions(address[] memory _targets, bytes[] memory _actions)
+  public stoppable auth
+  returns (bool)
+  {
+    require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
+    for (uint256 i; i < _targets.length; i+= 1){
+      require(
```
kronosapiens, line 61: Should these be wrapped in a `require`? In other uses, we allow transactions to fail and return a `false` bool. At minimum, let's consider adding a `strict` flag which throws an error if any tx fails, otherwise let's return a bool, making it a user decision.

area: I can get on board with a flag.

contracts/colony/Colony.sol
```diff
@@ -49,25 +49,34 @@ contract Colony is ColonyStorage, PatriciaTreeProofs, MultiChain {
...
   {
     require(_targets.length == _actions.length, "colony-targets-and-actions-length-mismatch");
-    for (uint256 i; i < _targets.length; i+= 1){
-      require(
-        makeArbitraryTransactionFunctionality(_targets[i], _actions[i]),
-        "colony-arbitrary-transaction-failed"
-      );
+    for (uint256 i; i < _targets.length; i += 1){
+      bool success = true;
+      try this.makeSingleArbitraryTransaction(_targets[i], _actions[i]) returns (bool ret){
```
kronosapiens, line 62: I would have done this differently:

```
bool success = true;
for (uint256 i; i < _targets.length; i += 1) {
   success = success && this.makeSingleArbitraryTransaction(_targets[i], _actions[i]);
   require(success || !_strict, "colony-arbitrary-transaction-failed);
}
return success;

area: I believe we have to use try and an external call here because of the `require`s in `makeSingleArbitraryTransaction` - if one of those fail, with your pattern, the whole transaction fails even if `_strict` is false.

kronosapiens: Also why `this`? We had it as an internal call earlier it it was all groovy.

area: > I don't follow -- makeSingleArbitraryTransaction returns a boolean, doesn't it?

If a require fails, it does not return a boolean and instead the whole transactions fails.

> It doesn't have an internal require.

The first three lines in the function are `require`s, at least? There are then also `require`s in the `preparation` functions.

> Also why this? We had it as an internal call earlier it it was all groovy.

`try` only works on external calls

kronosapiens: Oh, right -- I was thinking specifically of the `executeCall` but there are a lot of validation checks which could throw. All good!

kronosapiens (approved):

PR #966, opened 2021-06-08
Increase flexability miner shows towards mining cycle duration

area:
The 24 hour duration was hard coded in a fair number of places. This PR removes that dependency, and means that, going forward, changing the mining cycle duration will just involve replacing the `miningCycleResolver` address on the network contract, with no miner changes required.

This PR also drops us down to a 1 hour duration on the contracts side. Deployments wise, I will deploy a new version of the miner, which should work with the old cycle, and then deploy the new resolver to switch us to the one hour period once we see that's working.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -1079,7 +1079,7 @@ class ReputationMiner {
 
     this.miningCycleDuration = await repCycle.getMiningWindowDuration();
     this.miningCycleDuration = this.miningCycleDuration.toNumber();
-    this.constant = ethers.constants.MaxUint256.sub(1).div(this.miningCycleDuration);
+    this.constant = ethers.constants.MaxUint256.div(this.miningCycleDuration);
 
     if (calculatedPeriodLength !== this.miningCycleDuration) {
       this._adapter.log(
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -130,6 +130,10 @@ class ReputationMiner {
     this.justificationHashes = {};
     this.reverseReputationHashLookup = {};
     const repCycle = await this.getActiveRepCycle(blockNumber)
+    // Update fractions
+    const decayFraction = await repCycle.getDecayConstant({ blockTag: blockNumber });
+    this.decayNumerator = decayFraction.numerator;
+    this.decayDenominator = decayFraction.denominator;
 
     // Do updates
     this.nReputationsBeforeLatestLog = ethers.BigNumber.from(this.nReputations.toString());
@@ -241,12 +245,9 @@ class ReputationMiner {
     if (updateNumber.lt(this.nReputationsBeforeLatestLog)) {
       const key = await Object.keys(this.reputations)[updateNumber];
       const reputation = ethers.BigNumber.from(`0x${this.reputations[key].slice(2, 66)}`);
-      // These are the numerator and the denominator of the fraction we wish to reduce the reputation by. It
-      // is very slightly less than one (0.5 ** (1/2160) for a 1-hr mining cycle, 0.5 ** (1/90) for a 24-hr cycle).
-      // Disabling prettier on the next line so we can have these two values aligned so it's easy to see
-      // the fraction will be slightly less than one.
-      const numerator = ethers.BigNumber.from("992327946262944");
-      const denominator = ethers.BigNumber.from("1000000000000000");
+
+      const numerator = ethers.BigNumber.from(this.decayNumerator);
+      const denominator = ethers.BigNumber.from(this.decayDenominator);
 
       const newReputation = reputation.mul(numerator).div(denominator);
       const reputationChange = newReputation.sub(reputation);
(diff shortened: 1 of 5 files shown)


packages/reputation-miner/ReputationMiner.js
```diff
@@ -13,9 +13,9 @@ const sqlite = require("sqlite");
 // do in the future.
 // const accountAddress = "0xbb46703786c2049d4d6dd43f5b4edf52a20fefe4";
 const secretKey = "0xe5c050bb6bfdd9c29397b8fe6ed59ad2f7df83d6fd213b473f84b489205d9fc7";
-const miningCycleDuration = ethers.BigNumber.from(60).mul(60).mul(24); // 24 hours
 const minStake = ethers.BigNumber.from(10).pow(18).mul(2000);
-const constant = ethers.BigNumber.from(2).pow(256).sub(1).div(miningCycleDuration);
+
+const DAY_IN_SECONDS = 3600 * 24;
```
kronosapiens, line 18: I'd prefer `60 * 60 * 24` for consistency

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1060,6 +1062,34 @@ class ReputationMiner {
...
+    let count = -1;
+    while (value.gt(ethers.BigNumber.from("500000000000000000"))){
+      value = value.mul(numerator).div(denominator);
+      count += 1;
+    }
+    const calculatedPeriodLength = Math.floor(90 * DAY_IN_SECONDS / count);
+
+    this.miningCycleDuration = await repCycle.getMiningWindowDuration();
+    this.miningCycleDuration = this.miningCycleDuration.toNumber();
+    this.constant = ethers.BigNumber.from(2).pow(256).sub(1).div(this.miningCycleDuration);
```
kronosapiens, line 1082: Doesn't ethers provide a UINT256_MAX constant out of the box?

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -288,13 +287,20 @@ class ReputationMinerClient {
...
           clearTimeout(this.confirmTimeoutCheck);
         }
-        // If we don't see this next cycle completed in the next day and ten minutes, then report it
-        this.confirmTimeoutCheck = setTimeout(this.reportConfirmTimeout.bind(this), (DAY_IN_SECONDS + 10 * MINUTE_IN_SECONDS) * 1000);
+        await this._miner.updatePeriodLength(repCycle);
+
+        // If we don't see this next cycle completed in the period length and ten minutes, then report it
+        this.confirmTimeoutCheck = setTimeout(
+          this.reportConfirmTimeout.bind(this),
+          (this._miner.getMiningCycleDuration() + 10 * MINUTE_IN_SECONDS) * 1000
```
kronosapiens, line 302: For full correctness, shouldn't the open timestamp be used here also?

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1060,6 +1062,34 @@ class ReputationMiner {
...
+
+    if (calculatedPeriodLength !== this.miningCycleDuration) {
```
kronosapiens, line 1084: Ok, so we use the constants for the canonical window cycle, but there's still the explicit value in the contracts, which we use to throw warnings? Remind me why we don't just use the explicit value?

area: I mostly just want something to be doing a sanity check on these values (which we don't currently do).

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1079,7 +1079,7 @@ class ReputationMiner {
 
     this.miningCycleDuration = await repCycle.getMiningWindowDuration();
     this.miningCycleDuration = this.miningCycleDuration.toNumber();
-    this.constant = ethers.BigNumber.from(2).pow(256).sub(1).div(this.miningCycleDuration);
+    this.constant = ethers.constants.MaxUint256.sub(1).div(this.miningCycleDuration);
```
kronosapiens, line 1082: I believe `MaxUint256` includes the `sub(1)`

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -296,10 +296,12 @@ class ReputationMinerClient {
...
         await this._miner.updatePeriodLength(repCycle);
 
-        // If we don't see this next cycle completed in the period length and ten minutes, then report it
+        // If we don't see this next cycle completed at an appropriate time, then report it
+
+        const openTimestamp = await repCycle.getReputationMiningWindowOpenTimestamp();
         this.confirmTimeoutCheck = setTimeout(
           this.reportConfirmTimeout.bind(this),
-          (this._miner.getMiningCycleDuration() + 10 * MINUTE_IN_SECONDS) * 1000
+          (this._miner.getMiningCycleDuration() + 10 * MINUTE_IN_SECONDS - (Date.now() / 1000 - openTimestamp)) * 1000
```
kronosapiens, line 304: If we have duplicate logic in two places I'm wondering if it could be turned into a single function called twice?

packages/reputation-miner/ReputationMiner.js
area: That's what I get for trying to rush this through :smile:

kronosapiens (approved):
Nice!

PR #972, opened 2021-07-06
Bumps / cleanups / updates for next release

area:
As of the last release we adopted a convention that we bump everything as soon as a release is made (as this make life easiest for the front-end team). So here it is! I *think* I got everything...


kronosapiens (approved):

kronosapiens:
Is this going in `develop` or `e-lwss`?

area:
So I realised that what we discussed actually wouldn't work. So I think the branch with new features needs to be `next`, or something generic - as if we needed a 'patch' release, that would need to end up being `elwss`. So this is going in to `develop`.

PR #982, opened 2021-08-16
Add ColonyExtensionMeta

kronosapiens:


kronosapiens:
Noting that we would like to add some (let's say 32) storage slot placeholders in case we need to add new storage to this base class down the road.

PR #985, opened 2021-09-03
Many miner and oracle improvments

area:
There's a lot of medium sized things in this PR, but they're all around the reputation miner/oracle.

### More performant version of SQLite

Migrated to [Better SQLite](https://github.com/JoshuaWise/better-sqlite3) which does a much better job of showing off why it's good on the linked GitHub page than I do.

### More performant use of SQLite

We now use only one connection to the database which is left open, rather than opening and closing the file for every request, and also use prepared queries for improved performance. By not repeatedly opening and closing the file we remove the possibility of collisions, which we were starting to see as more people use the app.

### Cache relevant responses

Requested reputation proofs corresponding to a particular user / reputation / state combination are now cached for an hour. This duration was chosen to match the mining cycle length - the majority of requests are for a particular reputation in the current state, which only lasts for an hour, and so caching for longer seems only very occasionally relevant (e.g. caching a reputation proof involved in a vote would be useful).

### New oracle endpoint

I've added an oracle endpoint to return all reputations that a user has in a colony. It's at `:reputationRootHash/:colonyAddress/all/:userAddress`, intended to align with the endpoint used for a reputation proof for a specific reputation. It is required to implement this GUI element in a sensible way (i.e. without querying the oracle for every reputation a user could possibly have):

FireShot_Capture_431_-_Colony_Simplified_Figma_-_www figma com

### Improve oracle and miner reliability with unreliable endpoints

The ability to use multiple endpoints through [Ether's Fallback Provider](https://docs.ethers.io/v5/api/providers/other/#FallbackProvider). We're using a wrapped version of it, that in the event of a failed request will retry it up to ten times (the default of the `exponential-backoff` [npm package](https://www.npmjs.com/package/exponential-backoff).

This version of the miner is current running on QA. There may be further tweaks/improvements/adjustments as I see it fail in new and more interesting ways, but there's enough here to start getting feedback on. By creating a branch/PR it also means the frontend team can use it, and continue development using the new oracle endpoint.


packages/reputation-miner/ReputationMiner.js
```diff
@@ -1313,67 +1346,59 @@ class ReputationMiner {
...
-    );
-    await db.close();
+    const res = await this.queries.getAddressesWithReputation.all(reputationRootHash, colonyAddress.toLowerCase(), skillId);
     const addresses = res.map(x => x.user_address)
     return addresses;
   }
 
+  async getReputationsForAddress(reputationRootHash, colonyAddress, userAddress) {
+    const res = await this.queries.getReputationsForAddress.all(reputationRootHash, colonyAddress.toLowerCase(), userAddress.toLowerCase());
+    return res.map(function(x){ return {
```
kronosapiens, line 1359: Newline before `return`

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1313,67 +1346,59 @@ class ReputationMiner {
...
 
     // Do we have to do a database upgrade, from when we renamed n_nodes to n_leaves?
-    const nNodesColumn = await db.all("SELECT * FROM PRAGMA_TABLE_INFO('reputation_states') WHERE name='n_nodes';")
+    const nNodesColumn = await this.db.prepare("SELECT * FROM PRAGMA_TABLE_INFO('reputation_states') WHERE name='n_nodes';").all()
     if (nNodesColumn.length === 1) {
-      await db.run("ALTER TABLE 'reputation_states' RENAME COLUMN n_nodes to n_leaves");
-      const check1 = await db.all("SELECT * FROM PRAGMA_TABLE_INFO('reputation_states') where name='n_nodes'")
-      const check2 = await db.all("SELECT * FROM PRAGMA_TABLE_INFO('reputation_states') where name='n_leaves'")
+      await this.db.prepare("ALTER TABLE 'reputation_states' RENAME COLUMN n_nodes to n_leaves").run();
+      const check1 = await this.db.prepare("SELECT * FROM PRAGMA_TABLE_INFO('reputation_states') where name='n_nodes'").all()
```
kronosapiens, line 1385: What's the difference between the `run` and `all` functionality? Is the latter a select?

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1313,67 +1346,59 @@ class ReputationMiner {
...
   async resetDB() {
-    const db = await sqlite.open({filename: this.dbPath, driver: sqlite3.Database});
-    await db.run(`DROP TABLE IF EXISTS users`);
-    await db.run(`DROP TABLE IF EXISTS colonies`);
-    await db.run(`DROP TABLE IF EXISTS skills`);
-    await db.run(`DROP TABLE IF EXISTS reputations`);
-    await db.run(`DROP TABLE IF EXISTS reputation_states`);
-    await db.close();
+    // Again, not regularly used, so not preparing and saving the statements for reuse.
+    await this.db.prepare(`DROP TABLE IF EXISTS users`).run();
```
kronosapiens, line 1397: How is this preparing and saving if we don't save the statements to some constant?

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -125,8 +129,25 @@ class ReputationMinerClient {
         }
       });
 
+      // Query all reputation for a single user in a colony
+      this._app.get("/:rootHash/:colonyAddress/all/:userAddress", cache('1 hour'), async (req, res) => {
```
kronosapiens, line 133: I'd have a mild preference for `/:rootHash/:colonyAddress/:userAddress/all` if you're amenable to that

packages/reputation-miner/bin/index.js
```diff
@@ -49,22 +65,24 @@ if (network) {
...
-
-  if (!rpcEndpoint) {
-      rpcEndpoint = `http://${localProviderAddress || "localhost"}:${localPort || "8545"}`;
-  }
-
+} else if (providerAddress.length === 0){
+  const rpcEndpoint = `${localProviderAddress || "http://localhost"}:${localPort || "8545"}`;
   provider = new ethers.providers.JsonRpcProvider(rpcEndpoint);
+} else {
+  console.log('here')
```
kronosapiens, line 72: ??

test/reputation-system/client-calculations.js
```diff
@@ -74,8 +74,8 @@ process.env.SOLIDITY_COVERAGE
       });
 
       beforeEach(async () => {
-        await goodClient.resetDB();
         await goodClient.initialise(colonyNetwork.address);
+        await goodClient.resetDB();
```
kronosapiens, line 78: Why?

packages/reputation-miner/ReputationMiner.js
area: There are... a lot of places where we don't do this? Is this a thing we do in javascript?

packages/reputation-miner/ReputationMiner.js
area: `run` executes the statement, but doesn't expect data back from the statement. `all` returns all rows in a `SELECT` statement, and `get` would return the first row only.

packages/reputation-miner/ReputationMiner.js
area: Perhaps a misreading of the comment, which explicitly says we are _not_ preparing and saving the statement?

packages/reputation-miner/bin/index.js
area: Debugging leftover!

test/reputation-system/client-calculations.js
area: Think this got caught in the crossfire of having to change the order of `initlialise` and `resetDB` elsewhere.

packages/reputation-miner/ReputationMiner.js
kronosapiens: Ok, never mind then.

packages/reputation-miner/ReputationMiner.js
kronosapiens: Yup, correct.

packages/reputation-miner/bin/index.js
```diff
@@ -35,25 +35,49 @@ const {
...
+    super(url);
+    this.adapter = adapterObject;
+  }
 
-if ((!minerAddress && !privateKey) || !colonyNetworkAddress || !syncFrom) {
-  console.log("❗️ You have to specify all of ( --minerAddress or --privateKey ) and --colonyNetworkAddress and --syncFrom on the command line!");
-  process.exit();
-}
+  static attemptCheck(err, attemptNumber){
+    if (attemptNumber === 11){
```
kronosapiens, line 45: Why this value?

packages/reputation-miner/bin/index.js
```diff
@@ -50,14 +50,16 @@ class RetryProvider extends ethers.providers.JsonRpcProvider {
   }
 
   getNetwork(){
-    return backoff(() => super.getNetwork(), {retry: this.attemptCheck});
+    const that = this;
```
kronosapiens, line 53: Why?

packages/reputation-miner/bin/index.js
area: The backoff module tries ten times. If it's going to fail out entirely, my understanding is that this function will be called with `attemptNumber` equal to eleven and I want to see what went so wrong in the logs. If it's retrying but still might succeed, no need to log.

packages/reputation-miner/bin/index.js
area: I saw an error around `this.attemptCheck` and correctly using `.bind` is a big blind spot for me, so reverted to the `that=this` pattern out of safety. I think the actual issue here though is that I declared the function as static for some reason (even though it uses `this`), and so won't be available on the instances, just on the class itself.

kronosapiens:
Take this off draft whenever you're ready and I'll give it another read-through.

packages/reputation-miner/bin/index.js
kronosapiens: Can we indicate that with some constant somewhere? E.g. `BACKOFF_ATTEMPTS + 1`

packages/reputation-miner/bin/index.js
kronosapiens: JS scoping is not my strong suit either so I'm going to assume you did an excellent job.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -1161,13 +1222,31 @@ class ReputationMiner {
...
     let applyLogs = false;
 
+    // Run through events backwards find the most recent one that we know...
+    let syncFromIndex = 0;
+    for (let i = events.length - 1 ; i >= 0 ; i -= 1){
+      const event = events[i];
+      const hash = event.data.slice(0, 66);
+      const nLeaves = ethers.BigNumber.from(`0x${event.data.slice(66, 130)}`);
+      // Do we have such a state?
+      const res = await this.db.prepare(`SELECT COUNT ( * ) AS "n" FROM reputation_states WHERE root_hash='${hash}' AND n_leaves='${nLeaves}'`).get()
```
kronosapiens, line 1232: I feel like this could be `prepare`d outside of the loop?

kronosapiens (approved):

kronosapiens (dismissed):

kronosapiens (approved):

kronosapiens (dismissed):

PR #987, opened 2021-09-10
Add events for reputation award/smite

area:
Actions need events for the frontend to work properly, so we need to add them to these functions which the frontend team is developing against.

We believe that using the same event for all of these is fine, context (i.e. the skill id and the amount being positive/negative) should allow them to be distinguished.

Not wild about `customReputationUpdate` as a name, open to suggestions.


kronosapiens (approved):
I could go with `ArbitraryReputationUpdate` but it's your call.

kronosapiens (approved):

PR #988, opened 2021-09-10
Cannot confirm failing motion before a week passed

area:
While we've been messing around with the gas estimation / buffer process on the frontend, it actually doesn't solve an underlying problem. A bad actor could, in theory, finalize a motion with exactly the right amount of gas that means the motion fails its execution but the finalize transaction succeeds. 

I believe this is the 'path of least resistance' to fix the issue. The frontend will require no changes, and a 'proper' solution can be incorporated down the road when the frontend team have bandwidth to change the flow in the frontend (e.g. why can't users claim their tokens once its finalizable? With this solution, in the event of a motion that's corresponding transaction is now failing, users will have to wait a week to be able to reclaim)

### DIFF
--- a/contracts/extensions/VotingReputation.sol
+++ b/contracts/extensions/VotingReputation.sol
@@ -605,10 +605,30 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
 
     if (canExecute) {
       executed = executeCall(_motionId, motion.action);
-      require(executed || sub(block.timestamp, motion.events[REVEAL_END]) > 7 * 24 * 3600, "voting-execution-failed-not-one-week");
+      require(executed || failingExecutionAllowed(_motionId), "voting-execution-failed-not-one-week");
     }
 
     emit MotionFinalized(_motionId, motion.action, executed);
+  }
+
+
+  /// @notice Return whether a motion, assuming it's in the finalizable state,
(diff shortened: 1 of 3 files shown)


contracts/extensions/VotingReputation.sol
```diff
@@ -605,11 +605,32 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs {
...
     emit MotionFinalized(_motionId, motion.action, executed);
   }
 
+
+  /// @notice Return whether a motion, assuming it's in the finalizable state,
+  // is allowed to finalize without the call executing successfully.
+  /// @param _motionId The id of the motion
+  /// @dev We are only expecting this to be called from finalize motion in the contracts.
+  /// It is marked as public only so that the frontend can use it.
+  function failingExecutionAllowed(uint256 _motionId) public returns (bool) {
```
kronosapiens, line 620: Can we mark this as `view`?

kronosapiens (approved):

contracts/extensions/VotingReputation.sol
area: Oh, yeah, this definitely should be `view` otherwise it's useless for the frontend 😄

kronosapiens (approved):

PR #992, opened 2021-09-22
Change how extensions receive tokens and reputation

### LINKED ISSUE #991: Prevent extensions being awarded reputation, taking network fee
During CoinMachine QA'ing, it's been brought to everyone's attnetion that when funding the coinmachine with the native tokens of a colony, the coinmachine is awarded both reputation and the network fee is paid.

This is undesirable. Iff a payout is being made to a colony's own extension, the network fee is not taken and the reputation should not be awarded. 

When Colony A sends tokens to colony B's extension, the network fee should be taken, but the reputation should not be awarded.

area:
Closes #991 (when it's done).

Currently, this just has some failing tests!

### DIFF
--- a/contracts/colony/ColonyFunding.sol
+++ b/contracts/colony/ColonyFunding.sol
@@ -23,13 +23,13 @@ import "./ColonyStorage.sol";
 
 
 contract ColonyFunding is ColonyStorage, PatriciaTreeProofs { // ignore-swc-123
-  function lockToken() public stoppable onlyExtension returns (uint256) {
+  function lockToken() public stoppable onlyOwnExtension returns (uint256) {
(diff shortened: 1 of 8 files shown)


kronosapiens:
Currently, I do not believe we can support the spec of "extensions for other colonies", as we cannot determine from an extension address which colony it is installed in.

Once #971 is merged, however, we can support this, as we can do a lookup from extension address to colony. I would suggest we merge the feature as-is and update it once #971 is complete.

kronosapiens (dismissed):

area:
Using superpowers to merge this. If you have a problem with my last commit @kronosapiens then bring it up with me, but in another PR :stuck_out_tongue:

PR #996, opened 2021-10-21
Improve oracle performance

area:
The real world really sucks, huh?

So even with the `/noProof/` endpoint, performance in production was poor. So I went down a profiling hole.

A big, easy win was adding relevant indexes to the DB. Without, even for the `noProof` endpoint, all of the relevant tables had to be examined. With them, the daemon [can be much more efficient](https://www.sqlitetutorial.net/sqlite-index/).

While I was in there, I took a look at the bottleneck for the endpoint that returns proofs, and it's 100% `soliditySha3`.

screenshot-2021-10-21_17-10-14

That's going to be much harder to win... it's not clear that a JS binding to a native implementation would be faster, as the data is small and there are overheads. If you dig all the way down, the implementation is [this one](https://github.com/paulmillr/noble-hashes), which claims to be excellent. 

As more and more reputations are added to the state, the performance of that endpoint is going to get worse and worse. It's okay-ish for now, but it's already close to unacceptable and, if we do nothing, will eventually get there. Making a PR for this quick win, though, seemed worthwhile.

I also activated [WAL mode](https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/performance.md) and added caching to another relevant endpoint, but neither of those helped noticeably when I was testing (and I wouldn't have expected the latter to, just something I noticed while I was looking).

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -1392,7 +1392,8 @@ class ReputationMiner {
   async getAddressesWithReputation(reputationRootHash, colonyAddress, skillId) {
     const res = await this.queries.getAddressesWithReputation.all(reputationRootHash, colonyAddress.toLowerCase(), skillId);
     const addresses = res.map(x => x.user_address)
-    return addresses;
+    const reputations = res.map(x => new BN(x.value.slice(2, 66), 16).toString())
+    return { addresses, reputations };
   }
 
   async getReputationsForAddress(reputationRootHash, colonyAddress, userAddress) {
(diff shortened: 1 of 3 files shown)


area:
So it's our [Patricia Tree](https://github.com/JoinColony/colonyNetwork/blob/develop/packages/reputation-miner/patricia-base.js) implementation, which makes sense - the whole reputation tree gets rebuilt, which involves a lot of hashing things together (I think O(N^2) hashes, where N is the number of reputations in the state).

Part of my profiling was to check where/for how long the event loop was being blocked, and this hashing wasn't a culprit - it's just the proof requests that take a long(er) time, they don't prevent other requests being serviced. The biggest blocker seemed to be some network request, but I didn't figure out where from exactly - it was on the order of 200 milliseconds, and the requests with proofs take several seconds, so didn't appear to be the dominant factor.

PR #998, opened 2021-11-01
Fix version check script, bump needed versions

area:
The version check script for extensions was wrong - when #962 was rebased, should have thrown a big wobbly. This fixes it, and increments all the versions that should have been at the time.

### DIFF
--- a/test/contracts-network/colony-arbitrary-transactions.js
+++ b/test/contracts-network/colony-arbitrary-transactions.js
@@ -207,14 +207,14 @@ contract("Colony Arbitrary Transactions", (accounts) => {
   it("should not be able to make arbitrary transactions to the colony's own extensions", async () => {
     const COIN_MACHINE = soliditySha3("CoinMachine");
 
-    let coinMachineVersion = 0;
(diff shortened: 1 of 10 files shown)


test/contracts-network/colony-arbitrary-transactions.js
```diff
@@ -206,7 +206,15 @@ contract("Colony Arbitrary Transactions", (accounts) => {
 
   it("should not be able to make arbitrary transactions to the colony's own extensions", async () => {
     const COIN_MACHINE = soliditySha3("CoinMachine");
-    await colony.installExtension(COIN_MACHINE, 2);
+
+    let coinMachineVersion = 0;
+    let coinMachineResolverAddress = ethers.constants.AddressZero;
+
+    while (coinMachineResolverAddress === ethers.constants.AddressZero) {
```
kronosapiens, line 213: I don't love this loop as a way to pull the latest version, but so it goes.

kronosapiens (approved):

kronosapiens (approved):
Nice :tada:

PR #1005, opened 2021-12-07
Change how CoinMachine limits are enforced

area:
In light of [colonyDapp#2995](https://github.com/JoinColony/colonyDapp/issues/2995), the behaviour we understood was desired to be implemented was not actually desired.

The user limit is now enforced against the total number of tokens already sold plus those that are still to be sold that have already been given to coinmachine. This means that a user can, if the max sold per period allows, buy their entire allowance in one transaction at the start of the sale.

This means that, if a sale is terminated early, some users may have more than their allowance. This has been deemed acceptable.

### DIFF
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -361,7 +361,7 @@ contract CoinMachine is ColonyExtension {
     return
       (userLimitFraction == WAD || whitelist == address(0x0)) ?
       UINT256_MAX :
-      sub(wmul(add(ERC20(token).balanceOf(address(this)), soldTotal), userLimitFraction), soldUser[_user])
+      sub(wmul(add(getTokenBalance(), soldTotal), userLimitFraction), soldUser[_user])
     ;
   }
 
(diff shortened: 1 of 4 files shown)


contracts/extensions/CoinMachine.sol
```diff
@@ -358,11 +358,11 @@ contract CoinMachine is ColonyExtension {
...
   function getUserLimit(address _user) public view returns (uint256) {
-    // ((max(soldTotal, targetPerPeriod) * userLimitFraction) - soldUser) / (1 - userLimitFraction)
-    return (userLimitFraction == WAD || whitelist == address(0x0)) ? UINT256_MAX : wdiv(
-      sub(wmul(max(soldTotal, targetPerPeriod), userLimitFraction), soldUser[_user]),
-      sub(WAD, userLimitFraction)
-    );
+    return
+      (userLimitFraction == WAD || whitelist == address(0x0)) ?
+      UINT256_MAX :
+      sub(wmul(add(ERC20(token).balanceOf(address(this)), soldTotal), userLimitFraction), soldUser[_user])
```
kronosapiens, line 364: Can we use `getTokenBalance()` instead of making the direct call? For consistency mostly.

kronosapiens (approved):

PR #1011, opened 2022-01-12
Delegated mining

### LINKED ISSUE #1002: Delegated mining
Was chatting with Alex about security for private keys in miners, and we thought it would be a great idea to let users delegate their tokens to a different address, allowing that address to mine on their behalf. This way, the miner's private key can be compromised without putting the user's tokens at risk.

Brief discussion suggested such a thing could be accomplished with two mapping on the colony network. The first, from `user => miner` to ensure that users cant delegate twice, and the second, from `miner => user` lets the mining cycle determine what stake a miner can claim for mining purposes.

area:
Closes #1002 

Draft status, as not ready to merge, but after feedback.

Implements what Daniel and I vaguely talked about a while ago. Basically, everything stays the same, but instead of using `msg.sender` during mining, we instead call a function that either returns `msg.sender` (i.e. the fallback case should be the existing functionality) or, if a delegation has been awarded to `msg.sender`, returns the address they are allowed to mine on behalf of.

### DIFF
--- a/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
+++ b/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
@@ -32,7 +32,7 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
   // in reputationMiningCycleRespond. If you change one, you should change the other.
   uint256 constant MINING_WINDOW_SIZE = 60 * 60 * 1; // 1 hour
 
-  function getMinerAddress() internal view returns (address) {
+  function getMinerAddressIfStaked() internal view returns (address) {
     // Is msg.sender a miner themselves? See if they have stake.
     uint256 lockBalance = ITokenLocking(tokenLockingAddress).getObligation(msg.sender, clnyTokenAddress, colonyNetworkAddress);
     if (lockBalance > 0) {
@@ -150,7 +150,7 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
       return false;
     }
 
-    address minerAddress = getMinerAddress();
+    address minerAddress = getMinerAddressIfStaked();
 
     uint256 windowOpenFor = block.timestamp - _responseWindowOpened;
 
--- a/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
+++ b/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
@@ -31,6 +31,17 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
   // in reputationMiningCycleRespond. If you change one, you should change the other.
   uint256 constant MINING_WINDOW_SIZE = 60 * 60 * 1; // 1 hour
 
+  function getMinerAddress() internal view returns(address){
+    // Is msg.sender a delegated miner?
+    // See if this address is claiming to be a delegate
+    address delegator = IColonyNetwork(colonyNetworkAddress).getMiningDelegator(msg.sender);
+    if (delegator != address(0x00)){
+      return delegator;
+    }
+    // Otherwise, return msg.sender
+    return msg.sender;
+  }
+
   function expectedBranchMask(uint256 _nLeaves, uint256 _leaf) public pure returns (uint256) {
     // Gets the expected branchmask for a patricia tree which has nLeaves, with keys from 0 to nLeaves -1
     // i.e. the tree is 'full' - there are no missing leaves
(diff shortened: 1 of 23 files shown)


area:
So I've added restrictions that mean that anyone doing any sort of mining must either have tokens staked, or be a delegate miner for another address. If they have both, the former takes priority. This is to avoid the possibility that existed in the original implementation of 'stealing' miners by naming them as your delegate. In this implementation, you still can, but it has no effect (while they have tokens staked). Previously, `confirmNewHash` could be called by anyone, but now needs to be called by someone with tokens staked.

There were a good number of knock-on effects in the tests, unfortunately, but I think broadly speaking this is better. I don't think it presents a risk of getting in to issues either.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -31,6 +32,20 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
   // in reputationMiningCycleRespond. If you change one, you should change the other.
   uint256 constant MINING_WINDOW_SIZE = 60 * 60 * 1; // 1 hour
 
+  function getMinerAddress() internal view returns(address){
+    // Is msg.sender a miner themselves? See if they have stake.
+    uint256 lockBalance = ITokenLocking(tokenLockingAddress).getObligation(msg.sender, clnyTokenAddress, colonyNetworkAddress);
+    if (lockBalance > 0){
```
kronosapiens, line 38: Space before curly brace

kronosapiens:
What was the reason for restricting `confirmNewHash` to staked miners?

area:
> What was the reason for restricting `confirmNewHash` to staked miners?

So it originally came up because we call `getMinerAddress` in `responsePossible`, and the changes I made to `getMinerAddress` to deal with the griefing angle. I could have solved it by only calling `getMinerAddress` if we're inside the `SUBMITTER_ONLY_WINDOW_DURATION`, but realised it makes the mining cycle as a whole slightly more resilient. Picture a scenario where only one person has tokens staked.

1. This person unstakes their tokens
2. Someone calls `confirmNewHash`
3. The mining cycle is now broken. No-one has had tokens staked since before this cycle started, and so cannot submit the next hash.

By requiring this function call to be made by someone with tokens staked, we guarantee at least one person is able to continue the mining cycle. If the only person with tokens staked is malicious, then they're still able to break things (by unstaking after calling `confirmNewHash`), but we are now at least not able to get in to such a situation by everyone behaving non-maliciously, which was the case before (both steps 1 and 2 could be done by well-behaving participants).

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -31,6 +32,20 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
   // in reputationMiningCycleRespond. If you change one, you should change the other.
   uint256 constant MINING_WINDOW_SIZE = 60 * 60 * 1; // 1 hour
 
+  function getMinerAddress() internal view returns (address) {
```
kronosapiens, line 35: I would prefer this function return `msg.sender` in the event of no delegator (seems like the "right" behavior, since this isn't the place to assert whether the miner is staked), but I don't feel strongly enough to hold up the review.

packages/reputation-miner/ReputationMiner.js
```diff
@@ -106,6 +102,18 @@ class ReputationMiner {
...
+    // has been used for realWallet and is where transactions will be signed from may or may not be acting
+    // on behalf of another address via delegate mining. So check which address we're actually mining for.
+    const signingAddress = await this.realWallet.getAddress();
+    const delegator = await this.colonyNetwork.getMiningDelegator(signingAddress);
+    if (delegator === ethers.constants.AddressZero){
+      this.minerAddress = signingAddress;
+    } else {
+      this.minerAddress = delegator;
+    }
+    console.log("Mining on behalf of ", this.minerAddress)
```
kronosapiens, line 116: Nitpick but this style of string construction automatically inserts a space between arguments. I would slightly prefer the string template approach, but not really a big deal.

test/reputation-system/happy-paths.js
```diff
@@ -951,12 +951,59 @@ contract("Reputation Mining - happy paths", (accounts) => {
...
+      const delegatedClient = new ReputationMinerTestWrapper({ loader, realProviderPort, useJsTree, minerAddress: WORKER });
+      await colonyNetwork.setMiningDelegate(WORKER, true, { from: MINER1 });
+      await delegatedClient.initialise(colonyNetwork.address);
+      await advanceMiningCycleNoContest({ colonyNetwork, test: this });
+
+      const addr = await colonyNetwork.getReputationMiningCycle(true);
+      let repCycle = await IReputationMiningCycle.at(addr);
+      await forwardTime(MINING_CYCLE_DURATION + SUBMITTER_ONLY_WINDOW + 1, this);
+      await repCycle.submitRootHash("0x00", 0, "0x00", 10, { from: WORKER });
+      await repCycle.confirmNewHash(0, { from: MINER1 });
```
kronosapiens, line 974: Shouldn't this be `WORKER`?

kronosapiens (approved):

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
area: The implication here is that we'd have a `checkStaked` function, or similar, which would then be called after this function, and does exactly the same thing (including multiple calls to other contracts), but returns a boolean instead?

Would a name change placate you?

test/reputation-system/happy-paths.js
area: This is just setting up the test. It's the subsequent `delegatedClient` calls that test the functionality.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
kronosapiens: Sure, `getMinerAddressIfStaked` would be better.

PR #1013, opened 2022-01-24
Reputation miner updates

kronosapiens:


packages/reputation-miner/ReputationMiner.js
```diff
@@ -734,7 +734,7 @@ class ReputationMiner {
 
   async getEntryIndex(startIndex = 1) {
     // Get how much we've staked, and thefore how many entries we have
-    const [stakeAmount] = await this.colonyNetwork.getMiningStake(this.minerAddress);
+    const stakeAmount = await this.getMiningStake();
```
area, line 737: How does this work? `getMiningStake` does not return a single value?

truffle.js
```diff
@@ -41,17 +41,23 @@ module.exports = {
...
+      network_id: "1",
+    },
     goerli: {
       provider: () => {
-        return new HDWalletProvider("replace-with-private-key-when-using", "https://goerli.infura.io/v3/e21146aa267845a2b7b4da025178196d");
+        return new HDWalletProvider("private-key", "https://goerli.infura.io/v3/infura-key");
       },
       network_id: "5",
     },
-    mainnet: {
```
area, line 50: I still want mainnet in this file.

packages/reputation-miner/ReputationMiner.js
kronosapiens: Mm good point, should be `minerStake.amount`

truffle.js
kronosapiens: I moved it up the file, so the networks are ordered by id.

PR #1014, opened 2022-01-26
Miner client updates as real world imperfect

area:
Adds:

* Some extra logging
* Appropriately waits for transactions during disputes
* More timeout resilience
* Copes with blocks that do not advance the dispute process

Should incorporate #1013 or vice-versa.


kronosapiens (approved):
LGTM

PR #1015, opened 2022-02-02
Miner (hopefully) faster sync, faster startup if submission made

area:
Two (mostly) unrelated improvements:

* When syncing past cycles, no need to make the justification tree - they're historical, the time for disputes has passed, so no need to prepare for how we might respond to a dispute. This could be better; some of the lifting to work out keys/values is still done, it's just that the tree that the values are added to performs no-ops and returns nothing, rather than actually building the tree, and so will return quicker than generating the tree and associated proofs. I did is this way because it was easier than picking apart all the bits that generate the justification tree, which are a bit scattered.

* Fast-resume if we've made a submission to the cycle. When we generate the next reputation state and save it to the DB, I also save the data required to recreate the justification tree to a json file (not the database, as this is only needed temporarily and does not need to be queried by the oracle). If we've submitted and the miner restarts, it can see that it submitted, and load that state out of the DB and the Justification Tree out of the temporary file, rather than loading the previous state and (re-)applying the reputation update log. This should be a lot quicker, which is good if a restart happens during a dispute, for example.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -1442,7 +1442,7 @@ class ReputationMiner {
     }
 
     const currentJRH = await this.justificationTree.getRootHash();
-    if (currentJRH !== justificationRootHash) {
+    if (justificationRootHash && currentJRH !== justificationRootHash) {
       console.log("WARNING: The supplied JRH failed to be recreated successfully. Are you sure it was saved?");
     }
   }
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -7,6 +7,9 @@ import Database from "better-sqlite3";
 import PatriciaTreeNoHash from "./patriciaNoHashKey";
 import PatriciaTree from "./patricia";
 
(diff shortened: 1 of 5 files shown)


packages/reputation-miner/ReputationMiner.js
```diff
@@ -192,25 +197,35 @@ class ReputationMiner {
...
-      this.justificationTree = new PatriciaTreeNoHash();
-    } else {
-      const contractFactory = new ethers.ContractFactory(
-        this.patriciaTreeNoHashContractDef.abi,
-        this.patriciaTreeNoHashContractDef.bytecode,
-        this.ganacheWallet
-      );
+  async addLogContentsToReputationTree(blockNumber = "latest", buildJustificationTree = true) {
+    if (buildJustificationTree){
+      if (this.useJsTree) {
```
kronosapiens, line 206: Would be nice to see this pulled out into a separate function, `createJustificationTree("type")` or something which returns either a JS tree, a contract, or the no-op tree.

kronosapiens (approved):

kronosapiens (approved):

PR #1020, opened 2022-02-14
Merge master back in to develop

area:
Need to do this again, so that all the delegated mining stuff in the client lines up with other improvements that, as yet, are only in the client on the `develop` branch.

### DIFF
--- a/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
+++ b/contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
@@ -21,6 +21,7 @@ pragma experimental "ABIEncoderV2";
 import "./../../lib/dappsys/math.sol";
 import "./../patriciaTree/PatriciaTreeProofs.sol";
 import "./../colonyNetwork/IColonyNetwork.sol";
+import "./../tokenLocking/ITokenLocking.sol";
 import "./ReputationMiningCycleStorage.sol";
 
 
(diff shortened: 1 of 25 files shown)


area:
Note that this wasn't a straight merge, there were conflicts and I also tried to catch where new code needed to be changed (mostly around `msg.sender` / `msgSender()`), so I am asking for a proper review here, not a rubber-stamping as it might appear at first glance.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -31,6 +32,20 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
...
+  function getMinerAddressIfStaked() internal view returns (address) {
+    // Is msg.sender a miner themselves? See if they have stake.
+    uint256 stakedForMining = IColonyNetwork(colonyNetworkAddress).getMiningStake(msg.sender).amount;
+    if (stakedForMining > 0) {
+      // If so, they we don't let them mine on someone else's behalf
+      return msg.sender;
+    }
+
+    // Return any delegator they are acting on behalf of
+    address delegator = IColonyNetwork(colonyNetworkAddress).getMiningDelegator(msg.sender);
```
kronosapiens, line 44: Should this be `msgSender()`? Also elsewhere in the function.

area: The reputation mining contract does not support metatransactions as it's not a contract that a beginning user will be interacting with as they struggle to get xDAI to pay for gas (which is the reason for metatransactions, see #962)

PR #1021, opened 2022-02-14
Fix further wrinkles in the mining client

area:
~_NB based on #1020 so cannot be merged until that is. Can be reviewed, however_~

A variety of fixes here:

* The message to refill an address with Ether was not taking in to account delegated mining.
* The mysterious-errors-overnight error, I think, is because by default Nethermind refuses transactions with a gas price lower than 1.1Gwei; I think this is a consequence of EIP1559, but not 100% sure. When gas is higher, it's not seen, but I _think_ that overnight, gas drops below 1.1Gwei according to the Blockscout oracle, which we then use and our node rejects. Didn't want to overly tinker with the node to change it, given it's in production, and accommodating the defaults seems reasonable, anyway.
* If we had been delayed replying to a challenge, we'd invalidate our opponent before confirming our JRH, which could result in us both getting invalidated. We now always respond to a challenge if it's appropriate to do so, and only once it's not possible is invalidating our opponents considered.
* If a challenge needed to be navigated, and we'd loaded the state from disk, `this.nReputationsBeforeLatestLog` was left uninitialized which causes the challenge navigation process to fail.

### DIFF
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -91,7 +91,7 @@ contract CoinMachine is ColonyExtension, BasicMetaTransaction {
 
   /// @notice Returns the version of the extension
   function version() public override pure returns (uint256) {
-    return 4;
+    return 5;
   }
 
   /// @notice Configures the extension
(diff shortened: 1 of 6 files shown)


kronosapiens (commented):
Looks good to me

area:
Using superpowers rather than waiting for formal approval so I can get #1023 merged.

PR #1022, opened 2022-02-15
Two new events to support frontend

kronosapiens:
Updating CoinMachine's `TokensBought` event to emit the token address, and the `TokenUnlocked` event to include the agent. Also bumping CoinMachine's version to 5.

### DIFF
--- a/contracts/extensions/CoinMachine.sol
+++ b/contracts/extensions/CoinMachine.sol
@@ -30,7 +30,7 @@ contract CoinMachine is ColonyExtension, BasicMetaTransaction {
 
   // Events
 
-  event TokensBought(address indexed buyer, address indexed token, uint256 numTokens, uint256 totalCost);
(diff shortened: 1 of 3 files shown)


contracts/extensions/CoinMachine.sol
```diff
@@ -30,7 +30,7 @@ contract CoinMachine is ColonyExtension, BasicMetaTransaction {
 
   // Events
 
-  event TokensBought(address buyer, uint256 numTokens, uint256 totalCost);
+  event TokensBought(address indexed buyer, address indexed token, uint256 numTokens, uint256 totalCost);
```
area, line 33: I can understand indexing `buyer`, but I'm struggling to come up with a justification for indexing `token` here?

kronosapiens: Fair point, given that there is only one value this argument can take. I can remove.

area:
I think we should still merge this?

area (approved):

PR #1030, opened 2022-04-08
Adjust mining window behaviour

area:
This PR adjusts how the submission window and the dispute resolution window work, and who is allowed to submit when.

Before this PR, the submission window lasts an hour, and at the start of the hour no-one with stakes can submit, and at the end of the hour, everyone with stakes can submit. However, we want to ensure that even an unlucky good actor has an opportunity to submit before the window ends, and so this PR intends to change the logic to everyone being able to submit after 50 minutes. This ten minutes is captured in the newly introduced `ALL_ENTRIES_ALLOWED_END_OF_WINDOW` constant in the contracts.

Similarly, during the dispute process, before this PR each stage lasts ten minutes, with no-one able to defend at the start and everyone able to defend at the end. I have changed this to a twenty minute window, where from minute 10 everyone is able to defend. This has the same effect - every staker will have at least ten minutes to submit a defence for a submission. The same logic applies to all stages in the dispute - JRH confirmation, binary search, the transaction demonstrating a correct calculation, and the invalidation of submissions.

In both cases, some responses are still allowed after the ten minute free-for-all window - but crucially the 'next step' also starts becoming valid, whatever that might be.

I've also taken the opportunity to clean up some magic constant literals that were still floating about, as well as slightly improving coverage.


contracts/reputationMiningCycle/ReputationMiningCycle.sol
```diff
@@ -69,7 +69,7 @@ contract ReputationMiningCycle is ReputationMiningCycleCommon {
     }
   }
 
-  uint256 constant X = UINT256_MAX / MINING_WINDOW_SIZE;
+  uint256 constant X = UINT256_MAX / (MINING_WINDOW_SIZE - ALL_ENTRIES_ALLOWED_END_OF_WINDOW);
```
kronosapiens, line 72: I would prefer this expressed as a fraction and calculated using `wmul`, which avoids the possible (albeit unlikely) case of the all-entry window being larger than the mining window.

contracts/reputationMiningCycle/ReputationMiningCycle.sol
```diff
@@ -284,11 +288,11 @@ contract ReputationMiningCycle is ReputationMiningCycleCommon {
       require(disputeRounds[_round][opponentIdx].challengeStepCompleted >= disputeRounds[_round][_idx].challengeStepCompleted, "colony-reputation-mining-less-challenge-rounds-completed");
 
       // Require that it has failed a challenge (i.e. failed to respond in time)
-      require(add(disputeRounds[_round][_idx].lastResponseTimestamp, 600) <= block.timestamp, "colony-reputation-mining-not-timed-out"); // Timeout is ten minutes here.
+      require(add(disputeRounds[_round][_idx].lastResponseTimestamp, CHALLENGE_RESPONSE_WINDOW_DURATION) <= block.timestamp, "colony-reputation-mining-not-timed-out"); // Timeout is ten minutes here.
```
kronosapiens, line 291: Lovely, thank you

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
```diff
@@ -142,19 +143,20 @@ contract ReputationMiningCycleCommon is ReputationMiningCycleStorage, PatriciaTr
   }
 
   uint256 constant UINT256_MAX = 2**256 - 1;
-  uint256 constant SUBMITTER_ONLY_WINDOW_DURATION = 60 * 10;
-  uint256 constant Y = UINT256_MAX / SUBMITTER_ONLY_WINDOW_DURATION;
+  uint256 constant CHALLENGE_RESPONSE_WINDOW_DURATION = 60 * 20;
```
kronosapiens, line 146: Am I correct in thinking that this change doubles the amount of time a dispute will take? Is that acceptable in terms of colony performance (e.g. not having reputation be frozen for a long time during a dispute)?

helpers/constants.js
```diff
@@ -59,8 +59,8 @@ const FINALIZED_TASK_STATE = 2;
 const SECONDS_PER_DAY = 86400;
 
 const MINING_CYCLE_DURATION = 60 * 60 * 1; // 1 hour
-const MINING_CYCLE_TIMEOUT = 60 * 10; // Ten minutes
-const SUBMITTER_ONLY_WINDOW = 60 * 10; // Ten minutes
+const DISPUTE_DEFENCE_WINDOW = 60 * 20; // Twenty minutes
```
kronosapiens, line 62: This is the same variable as `CHALLENGE_RESPONSE_WINDOW_DURATION`, yes? Why different names?

packages/reputation-miner/ReputationMinerClient.js
```diff
@@ -19,6 +19,8 @@ const disputeStages = {
  CONFIRM_NEW_HASH: 5
 }
 
+const DISPUTE_DEFENCE_WINDOW = 1200;
```
kronosapiens, line 22: Can we express this as `60 * 20`?

contracts/reputationMiningCycle/ReputationMiningCycle.sol
area: The scenario you're worried about here is us changing these constants incorrectly down the line?

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
kronosapiens: Is there some middle ground here, like 6 + 6?

contracts/reputationMiningCycle/ReputationMiningCycle.sol
area: For what it's worth, a wad would also have a huge amount of invalid space in its range (anything bigger than 1). 

I don't think I can do what you're proposing without using storage slots and initialising them on deployment, or recalculating every time we use functions that need this constant, as we can only use solidity built-ins for defining constants.

contracts/reputationMiningCycle/ReputationMiningCycleCommon.sol
area: Security, not speed, is the priority here. If I had my way, all these windows would be a lot longer than they are.

contracts/reputationMiningCycle/ReputationMiningCycle.sol
kronosapiens: Alright, so be it.

kronosapiens (dismissed):

PR #1037, opened 2022-04-27
Add dev script to easily start reputation oracle

chmanie:
I added this script to make it easier to start the reputation oracle for development purposes.

It gets the ganache accounts from the RPC node and the `etherrouter-address` from the corresponding json file.


area (approved):

area:
Added a commit that is mostly ESLint fixes, but also moves the `.on('error')` from `res` to `req` which seems like what is necessary from [the docs](https://nodejs.dev/learn/making-http-requests-with-nodejs).

area (approved):

Issue #1034, opened 2022-04-27
Remove dependency on Python 2.7.x


chmanie:
## Summary
tl;dr: 
- Python 2.x is EOL
- `truffle-security` depends on the `sleep` module which is old and required Python 2.x to build
- `truffle-security` itself is deprecated and no longer maintained

## Steps to Reproduce (for bugs)
Clone the repo and run `yarn`

## Expected Behavior
Packages are installed properly when Python 2.x is not available

## Current Behaviour
An error is thrown when building `sleep`:
```
colonyNetwork/node_modules/truffle-security/node_modules/sleep: Command failed.
Exit code: 1
Command: node-gyp rebuild
Arguments: 
Directory: /home/chris/dev/colony/colonyNetwork/node_modules/truffle-security/node_modules/sleep
Output:
gyp info it worked if it ends with ok
gyp info using node-gyp@3.8.0
gyp info using node@12.16.3 | linux | x64
gyp ERR! configure error 
gyp ERR! stack Error: Command failed: /usr/bin/python -c import sys; print "%s.%s.%s" % sys.version_info[:3];
gyp ERR! stack   File "<string>", line 1
gyp ERR! stack     import sys; print "%s.%s.%s" % sys.version_info[:3];
gyp ERR! stack                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gyp ERR! stack SyntaxError: Missing parentheses in call to 'print'. Did you mean print(...)?
gyp ERR! stack 
gyp ERR! stack     at ChildProcess.exithandler (child_process.js:303:12)
gyp ERR! stack     at ChildProcess.emit (events.js:310:20)
gyp ERR! stack     at maybeClose (internal/child_process.js:1021:16)
gyp ERR! stack     at Socket.<anonymous> (internal/child_process.js:443:11)
gyp ERR! stack     at Socket.emit (events.js:310:20)
gyp ERR! stack     at Pipe.<anonymous> (net.js:672:12)
gyp ERR! System Linux 5.15.35-1-lts
gyp ERR! command "node/v12.16.3/bin/node" "colonyNetwork/node_modules/.bin/node-gyp" "rebuild"
gyp ERR! cwd colonyNetwork/node_modules/truffle-security/node_modules/sleep
```

## Possible Solution
Remove `truffle-security` and use [mythx-cli](https://github.com/dmuhs/mythx-cli) instead (as proposed by the developers).

## Context
[...]

PR #1043, opened 2022-05-02
Remove babel dependency

chmanie:
This PR removes all `babel` related packages and adjusts all the import/export statements to be commonJS compatible,


chmanie:
I hereby claim that this is working now.

area:
Closes #1034

area:
# High level changes

The original purpose of this PR, before I thought "oh, wouldn't it be a good idea if...", was to remove babel, which Chris did successfully. This mostly required changing imports to requires, and (I think) still represent the bulk of the changes in this PR.

Then I got my hands on it, and also decided this PR shouuld upgrades the node version (as Chris has pointed out v12 is going EOL very soon)... which in turn meant that truffle, ganache and solidity versions all also needed updating. Note that `ganache-cli` and `ganache-core` were originally separate packages, but have now both been replaced by `ganache`. `ganache-cli` is still installed as (a link to) the executable, but we should migrate to just using `ganache`.

## Coverage changes

The big change that `solidity-coverage` introduced with this new version is that modifiers on functions are now tracked independently and count as branch points unless specified otherwise. So for functions with modifiers, you need to show it both passing through the modifier and reverting. For modifiers that don't behave in that way (like 'always', for us, which is essentially just syntatical sugar for some of our automated checks), they can be added to the whitelist.

As a result of this, a lot more tests have been added to this PR.

## Contract changes

* Version bookkeeping - bumping versions, removing upgrades
* Removal of useless modifiers. Some read-only functions had recovery-mode modifiers, and I don't see the point in them? It could only ever make recovery mode harder to use.
* Reordering of some modifiers. This partly was to make tests easier to write, but also allows the calls in question to fail quickly (i.e. without spending as much gas) if easy-to-check conditions are not met.
* Removal of unused modifiers and functions. `expenditureLocked` was an unused modifier, and `setReputationRootHash` has an overload that is not used. The latter is not used by the dapp at all, so I'm viewing this as a taster of how colonyJS etc deals with the removal of a function, as it's one that doesn't (shouldn't...) matter.

## Test changes

As well as adding a boatload of tests, some tests have been changed. The new version of ganache seems like it is more sensitive to function calls that should be `await`ed not being so, so some we've had for a while (unknowingly) were finally thrown up. This includes the Migrations files, where calls to the `deployer` weren't awaited.

test/contracts-network/colony-payment.js
```diff
@@ -253,6 +253,11 @@ contract("Colony Payment", (accounts) => {
       await checkErrorRevert(colony.finalizePayment(1, UINT256_MAX, paymentId, { from: accounts[10] }), "ds-auth-unauthorized");
     });
 
+    it("should not allow a finazlied payment to be finalized again", async () => {
```
kronosapiens, line 256: "finazlied"

kronosapiens (approved):
Nicely done

kronosapiens (dismissed):

area:
I ended up running this one down via `truffle debug` which _now works on our codebase_ :tada: 

If a function that calls `issueTokens` (so either `issueTokens` directly or `setTokenIssuanceRate`), and is `estimateGas`'d when no tokens would be minted, we get some amount of gas returned. If the transaction is executed with that amount of gas, but the block timestamp has advanced (by one second) to a point when some amount of tokens would be minted when the transaction is mined, it tries to mint tokens, doesn't have enough gas, and reverts.

This wasn't happening much, about 1-in-20 runs, but enough to annoy me...

chmanie:
This is amazing work! Cheers Alex!

PR #1054, opened 2022-06-20
Replace DEPRECATED with @deprecated

### LINKED ISSUE #1035: Replace `DEPRECATED` with `@deprecated` in the dev docs
## Summary
In order to make the docs look nicer in the auto-generated markdown docs, I recommend to use the typedoc compatible `@deprecated` tag instead of just the uppercase `DEPRECATED`. This will look much nicer in the docs.

image

vs

image

Please also quickly check the other available [tag references](https://tsdoc.org/pages/tags/typeparam/) and coordinate with @chmanie to see whether using other ones would also make sense and improve readability.

chmanie:
Closes #1035 

I also took the liberty to fix some of the NatSpec warnings we got when running build:docs.

### DIFF
--- a/contracts/colony/IColony.sol
+++ b/contracts/colony/IColony.sol
@@ -61,8 +61,8 @@ interface IColony is ColonyDataTypes, IRecovery, IBasicMetaTransaction {
   /// @return tokenAddress Address of the token contract
   function getToken() external view returns (address tokenAddress);
 
+  /// @deprecated
   /// @notice Execute arbitrary transaction on behalf of the Colony
-  /// DEPRECATED
   /// @param _to Contract to receive the function call (cannot be this contract, network or token locking)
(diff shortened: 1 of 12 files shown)


chmanie:
Ah, I see, okay. `@deprecated` is not a natspec tag. Yeah, fine. I'll work around that.

area:
What does that mean for this PR?

chmanie:
That we likely have to use

```
/// @notice @deprecated
```

instead. Sorry I didn't get to adjusting that.

area:
So the issue here is that you've changed VotingReputation. There are no functional changes, but the script isn't bright enough to realise that and thinks it should have its version bumped.

I'm not really sure what to do about this, I'm going to be fairly resistant to changing that script.

area (approved):

PR #1055, opened 2022-06-20
Switch to npm

### LINKED ISSUE #1050: Readme changes
- Node version needs to be 16.15.0, probably other versions (yarn, solc) need bumps as well.
- The delegate proxy link does not work.

There might be more, I'll add it as I find it.

chmanie:
`yarn` was great at a time when npm was lacking the speed, the consistency and the features it provided. These days are over now as npm has caught up. I propose to switch this project (back?) to npm and with that improve compatibility (node version usually determines npm version) and remove dependencies while at the same time create consistency with all other Colony projects.

I used https://github.com/imsnif/synp to convert the `yarn.lock` to a `package-lock.json`. The process is like so:

1) Freshly clone the repository with the newest `yarn.lock` file
2) Run `yarn --frozen-lockfile`
3) Remove any existing `package-lock.json`
4) Run `npx synp --source-file yarn.lock --with-workspace`
5) Run `npm install`
6) After that you should end up with a valid `package-lock.json`

We also move from Meta to Microsoft :grimacing: 

Changes for the run-scripts:

* To run executables from the `.bin` folder, in the `package.json` one does not need to use `yarn` or `yarn run`, so I removed those
* `yarn EXECUTABLE` in other places was replaced with `npx EXECUTABLE` (the npm canonical way to execute the `.bin` executables)
* `yarn (run) SCRIPT` in the `package.json` was replaced with `npm run SCRIPT`

It seems to be going through fine. We still have to test/adjust the following:

* Does the reputation miner still work?
* We need to adjust various docker images I assume, especially rebuild the reputation miner's
* We probably also need to adjust things on some production servers?

**Note:** as this PR is based on top of #1052 I suggest to merge that PR first and have the CircleCI checks run on this one

This PR closes #1050.


rdig:
Can we please be kept in the look as when this makes it's way into `develop` as it will affect the dapp's `provision` script

chmanie:
It works! All the tests are passing and I also tested the reputation miner locally.

chmanie:
Rebased and working again! I applied `npm audit fix` which should have resolved all the security issues

kronosapiens (approved):

kronosapiens:
@rdig heads up this has been merged to develop

PR #1061, opened 2022-06-24
Move checkReputation to ColonyExtension

kronosapiens:
Checking reputation proofs is generally useful functionality. Rather than have extensions re-implement, let's provide a common function.


kronosapiens:
@area after a long saga the builds are green again, I think this is clear to merge without too much trouble.

PR #1065, opened 2022-07-08
Fix CoinMachine token accounting bug

kronosapiens:
Porting the fix for https://github.com/JoinColony/coinMachine/issues/81 into this repo

### DIFF
--- a/test/extensions/coin-machine.js
+++ b/test/extensions/coin-machine.js
@@ -615,6 +615,34 @@ contract("Coin Machine", (accounts) => {
       expect(currentPrice).to.eq.BN(expectedPrice);
     });
 
+    it("can correctly account for sold tokens between periods when adding a balance", async () => {
+      await colony.uninstallExtension(COIN_MACHINE, { from: USER0 });
+      await colony.installExtension(COIN_MACHINE, version, { from: USER0 });
+      const coinMachineAddress = await colonyNetwork.getExtensionInstallation(COIN_MACHINE, colony.address);
+      coinMachine = await CoinMachine.at(coinMachineAddress);
+
+      await purchaseToken.mint(USER0, WAD.muln(10), { from: USER0 });
+      await purchaseToken.approve(coinMachine.address, WAD.muln(10), { from: USER0 });
+
+      // Iniitalise with a zero balance, then add tokens
+      await coinMachine.initialise(token.address, purchaseToken.address, 60 * 60, 10, WAD, WAD, WAD, WAD, ADDRESS_ZERO);
+      await token.mint(coinMachine.address, WAD.muln(200));
+
+      const periodLength = await coinMachine.getPeriodLength();
+
+      let activeSold;
+
+      await coinMachine.buyTokens(WAD, { from: USER0 });
+      activeSold = await coinMachine.getActiveSold();
+      expect(activeSold).to.eq.BN(WAD);
+
+      await forwardTime(periodLength.toNumber(), this);
+
+      await coinMachine.buyTokens(WAD, { from: USER0 });
+      activeSold = await coinMachine.getActiveSold();
+      expect(activeSold).to.eq.BN(WAD);
+    });
+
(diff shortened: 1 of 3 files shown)


test/extensions/coin-machine.js
```diff
@@ -537,7 +537,7 @@ contract("Coin Machine", (accounts) => {
       expect(currentPrice).to.eq.BN(WAD);
 
       // Now we reload the token balance (and advance a period)
-      // Price doesn't adjust in the period you update
+      // Price adjusts in the period you update
```
area, line 540: Is this a change in behaviour that we're happy with? Was a requirement along these lines how the issue crept in in the first place - a guard against an unintended drop in price (because of selling a number of tokens below the target, but only because we ran out)?

test/extensions/coin-machine.js
```diff
@@ -602,20 +595,41 @@ contract("Coin Machine", (accounts) => {
...
+      expect(activeSold).to.eq.BN(WAD);
+
+      await forwardTime(periodLength.toNumber(), this);
+
+      await coinMachine.buyTokens(WAD, { from: USER0 });
+      activeSold = await coinMachine.getActiveSold();
+      expect(activeSold).to.eq.BN(WAD);
+    });
+
+    it.skip("it monotonically adjusts prices according to demand", async () => {
```
area, line 632: :raised_eyebrow:

kronosapiens: That test keeps timing out on my machine... 😭

test/extensions/coin-machine.js
area: > I would default to not changing existing behaviour wherever possible

While this is true in general, I also I _really_ don't want to have to update [the logic in the dapp](https://github.com/JoinColony/colonyDapp/blob/3699bb1424c2eeb88c0319f5fa2c512df1019de1/src/data/resolvers/coinMachine.ts).

arrenv: It seems more fair in general for both token seller and token buyer to stop price evolution as soon as buys are not able to happen.

Regarding having to update the resolver logic for the DApp, this is likely going to need to be revised anyway, to move it over to the new CDApp architecture which will also be used for CoinMachine.

area: I had a conversation with Arren on Discord to explain the issue - let me know if you feel like I've misrepresented something (as I said to you on Discord, I didn't follow the whole of your message):

**Alex**
> I think the way to think about it is (though at this point, I'm not 100% certain)
>
> Period n: We sell out of tokens, evolution stops, purchasing is not possible
>_some time later_
>Period n+x: We add more tokens, purchasing is possible again
>Period n+x+1: Sale continues
>
>Does the price update in period n+x, based on tokens sold in period n before we sold out, or does the price only update in period n+x+1 based on tokens sold in period n+x
>
>(for reference, the bug being fixed is that it would only update in period n+x+1, but do so based on tokens sold in period n and period n+x, which combined could only sell the amount expected to sell in one period)

**Arren**
>So, does the batch time start from 0 when the tokens are added, or could this be half way through a batch period?
>
>If it sold above the threshold in n, then it makes sense that n+x would increase in price.
However, if it sold below the threshold in n due to selling out of tokens, then it doesn't make sense that it would decrease the price.
>
>So, I think the middle ground for this is that it increases in n+x+1

**Alex**
>>So, does the batch time start from 0 when the tokens are added, or could this be half way through a batch period?

>It could be halfway through for both the original and new implementations
>>So, I think the middle ground for this is that it increases in n+x+1

>Or decreases, based only on what was sold in n+x?

**Arren**
>Yes, sorry, exactly

test/extensions/coin-machine.js
area: Can we make the loop size tied to whether we're running on CI or something then (via env variables)? The test is the way it is because we thought it was more likely to fail with rounding errors etc the longer a sale went, and we shouldn't compromise it because of local issues.

area (approved):

PR #1071, opened 2022-07-22
Add support for decisions that take no action

### LINKED ISSUE #1068: Create Simple Decisions extension
See spec here:

- https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1018%3A54646
- https://www.notion.so/colony/Simple-Decisions-2d869923870640a5889f9f24b183c65b

area:
Closes #1068.

Not happy with the ternary, would very much like a function with a more understandable if statement, but can't come up with a semantic name for that function beyond `finalizableOrFinalized`. Please suggest something!

Should the 'special action' be longer than just a signature? Obviously it's all meaningless, but...

### DIFF
--- a/contracts/extensions/VotingReputation.sol
+++ b/contracts/extensions/VotingReputation.sol
@@ -799,10 +799,10 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs, BasicMetaTrans
         return MotionState.Staking;
       // If not, did the YAY side stake?
       } else if (motion.stakes[YAY] == requiredStake) {
-        return getSig(motion.action) == NO_ACTION ? MotionState.Finalized : MotionState.Finalizable;
+        return finalizableOrFinalized(motion.action);
       // If not, was there a prior vote we can fall back on?
       } else if (add(motion.votes[NAY], motion.votes[YAY]) > 0) {
-        return getSig(motion.action) == NO_ACTION ? MotionState.Finalized : MotionState.Finalizable;
+        return finalizableOrFinalized(motion.action);
       // Otherwise, the motion failed
       } else {
         return MotionState.Failed;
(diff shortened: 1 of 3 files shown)


kronosapiens:
My first comment here is you bought up the idea of having this be a separate extension, so colonies can choose to install these individually. Is this just a prototype implementation for feedback, or do you want to keep the functionality bundled?

area:
See my comment on #1068

contracts/extensions/VotingReputation.sol
```diff
@@ -797,10 +799,10 @@ contract VotingReputation is ColonyExtension, PatriciaTreeProofs, BasicMetaTrans
         return MotionState.Staking;
       // If not, did the YAY side stake?
       } else if (motion.stakes[YAY] == requiredStake) {
-        return MotionState.Finalizable;
+        return getSig(motion.action) == NO_ACTION ? MotionState.Finalized : MotionState.Finalizable;
```
kronosapiens, line 802: Stylistically I would do the following:

```
MotionState finalizableState = getSig(motion.action) == NO_ACTION : MotionState.Finalized : MotionState.Finalizable;
```

And then return `finalizableState` as needed. I think `finalizableState` is an appropriate name, since it's `Finalizable` normally but `Finalized` under special circumstances.

kronosapiens: Another option is to simply `return` in `finalizeMotion` after setting `motion.finalized` to true if the action is the null action, although I suppose you'd want to avoid having users make that transaction.

test/extensions/voting-rep.js
```diff
@@ -1525,6 +1525,22 @@ contract("Voting Reputation", (accounts) => {
...
+      motionId = await voting.getMotionCount();
+
+      await voting.stakeMotion(motionId, 1, UINT256_MAX, YAY, REQUIRED_STAKE, user0Key, user0Value, user0Mask, user0Siblings, { from: USER0 });
+
+      await forwardTime(STAKE_PERIOD, this);
+
+      await checkErrorRevert(voting.finalizeMotion(motionId), "voting-rep-motion-not-finalizable");
+
+      const motionState = await voting.getMotionState(motionId);
+      expect(motionState).to.eq.BN(EXECUTED);
```
kronosapiens, line 1542: Just noting how in the tests we call these `EXECUTED` while in the contracts they're `Finalized`, but perhaps not a big deal.

kronosapiens (approved):
I'd actually be inclined to call the function `finalizable()` because in ~most cases it returns the `Finalized` state, and so would make `getMotionState` read a bit better. The intended state in those branches is `Finalizable` except for this new exception, so Id' say we could justify tucking that exception away in the `finalizable` function for people who are curious. But I'll leave that up to you.

PR #1072, opened 2022-07-22
Make the metatransaction broadcaster more robust

area:
Based on #1070 (as this branch is being used by the frontend team), so no merging for now.

The main issue with the metatransaction broadcaster in the real world is to do with nonces for the account doing the broadcasting. With the current (committed, though not deployed) implementation, two requests for a metatransaction at the same time would end up with the same nonce, and one wouldn't be accepted. This is obviously bad, so I started using the experimental [NonceManager](https://docs.ethers.io/v5/api/experimental/#experimental-noncemanager) from Ethers, which gives us finer control over nonces. Basically, it signs transactions with sequential nonces as requests come in, regardless of what the RPC node it is connected to tells it.

There are issues with this - if the nonce gets out-of-sync, for whatever reason, things start going wrong. So I've added logic to the NonceManager around the nonce getting out of sync to try and allow it to recover. It also keeps track of transactions its signed, and if it seems like the node it's connected to has forgotten about them, will re-send them (with the same nonce). If it somehow gets ahead of itself, it should also go back and 'fill in' missing nonces.

There is also some generic logic around resetting the transaction count if an error message is seen with the word 'nonce' in it. We also apply some belt-and-braces logic by adding a queue, so only one broadcast is serviced at a time. It's possible that this will slow down sending metatransactions, but until we get complaints about it, it's something else that should help nonces going wrong.

This PR also adds support for the EIP2612 transactions we use when interacting with (bridged) tokens.

We also add some logic around allowing `finalizeMotion` where we can - basically, if the motion is for something that we'd allow, we will pay for it.

`setAuthority` on a token was a tricky one. As it takes an address, I didn't want to allow all `setAuthority` calls, as in principle that argument could be used to pass anything, and we don't want to pay for everything that a user could do on GnosisChain. I ended up assuming that the user will always be going through flows in the frontend, and so we should be able to check the passed address is an authority that was just deployed by the network.

Finally, `setOwner` has similar logic - when going through flows in the frontend, we only ever call `setOwner` with the parameter of a colony, so that can be easily allowed (or not).

### DIFF
--- a/packages/metatransaction-broadcaster/Dockerfile
+++ b/packages/metatransaction-broadcaster/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16-bullseye
+FROM node:14-bullseye
 COPY ./packages ./packages
 COPY ./package.json ./
 COPY ./yarn.lock ./
--- a/packages/metatransaction-broadcaster/Dockerfile
+++ b/packages/metatransaction-broadcaster/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:16-bullseye
+COPY ./packages ./packages
+COPY ./package.json ./
+COPY ./yarn.lock ./
+COPY ./build ./build
+RUN yarn
+EXPOSE 3000
+CMD node $NODE_ARGS packages/metatransaction-broadcaster/bin/index.js --colonyNetworkAddress $COLONYNETWORK_ADDRESS --privateKey $PRIVATE_KEY --gasLimit $GASLIMIT $ARGS
--- a/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
+++ b/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
@@ -219,12 +219,13 @@ class MetatransactionBroadcaster {
         if (motionTarget === "0x0000000000000000000000000000000000000000") {
           motionTarget = await possibleVotingRep.getColony();
         }
-        // Is the motion doing something we'd allow? Duplicated logic from main function
-
-        const addressValid = await this.isAddressValid(motionTarget);
-        // It's possible it's not a valid address, but could be a valid transaction with a token.
-        const validTokenTransaction = await this.isTokenTransactionValid(motionTarget, motion.action, userAddress);
-        const allowedColonyFamilyTransaction = await this.isColonyFamilyTransactionAllowed(motionTarget, motion.action);
+        // Is the motion doing something we'd allow?
+        const { addressValid, validTokenTransaction, allowedColonyFamilyTransaction } = await this.doValidTransactionChecks(
+          motionTarget,
+          motion.action,
+          userAddress
+        );
+        console.log(addressValid, validTokenTransaction, allowedColonyFamilyTransaction);
         if (!(addressValid && allowedColonyFamilyTransaction) && !validTokenTransaction) {
           return false;
         }
@@ -298,10 +299,12 @@ class MetatransactionBroadcaster {
         }
       } else if (!spender && target) {
         // Then it's a 'normal' metatransaction.
-        const addressValid = await this.isAddressValid(target);
-        // It's possible it's not a valid address, but could be a valid transaction with a token.
-        const validTokenTransaction = await this.isTokenTransactionValid(target, payload, userAddress);
-        const allowedColonyFamilyTransaction = await this.isColonyFamilyTransactionAllowed(target, payload, userAddress);
+        const { addressValid, validTokenTransaction, allowedColonyFamilyTransaction } = await this.doValidTransactionChecks(
+          target,
+          payload,
+          userAddress
+        );
+
         if (!(addressValid && allowedColonyFamilyTransaction) && !validTokenTransaction) {
           const data = {};
           if (!addressValid) {
@@ -356,6 +359,14 @@ class MetatransactionBroadcaster {
         message: err,
       });
     }
+  }
+
+  async doValidTransactionChecks(target, payload, userAddress) {
+    const addressValid = await this.isAddressValid(target);
+    // It's possible it's not a valid address, but could be a valid transaction with a token.
+    const validTokenTransaction = await this.isTokenTransactionValid(target, payload, userAddress);
+    const allowedColonyFamilyTransaction = await this.isColonyFamilyTransactionAllowed(target, payload, userAddress);
+    return { addressValid, validTokenTransaction, allowedColonyFamilyTransaction };
   }
 
   async processMetatransaction(req, res) {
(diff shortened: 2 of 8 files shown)


packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -215,25 +176,50 @@ class MetatransactionBroadcaster {
         valid = await this.isAddressValid(tx.args[1]);
       } else if (tx.signature === "approve(address,uint256)") {
         valid = await this.isAddressValid(tx.args[0]);
+      } else if (tx.signature === "setAuthority(address)") {
+        // Get the most recent metatx this user sent on colonyNetwork
```
kronosapiens, line 180: This should be put into it's own function to be consistent with the style of the other branches

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -244,13 +230,172 @@ class MetatransactionBroadcaster {
...
+        // Get the motion
+        let motionTarget = motion.altTarget;
+        if (motionTarget === "0x0000000000000000000000000000000000000000") {
+          motionTarget = await possibleVotingRep.getColony();
+        }
+        // Is the motion doing something we'd allow? Duplicated logic from main function
+
+        const addressValid = await this.isAddressValid(motionTarget);
+        // It's possible it's not a valid address, but could be a valid transaction with a token.
+        const validTokenTransaction = await this.isTokenTransactionValid(motionTarget, motion.action, userAddress);
```
kronosapiens, line 244: This looks similar to other code snippets elsewhere. Perhaps a function which takes the arguments, performs all these checks, and returns a single boolean?

test/extensions/voting-rep.js
```diff
@@ -1525,6 +1530,85 @@ contract("Voting Reputation", (accounts) => {
       const pastVote = await voting.getExpenditurePastVote(slotHash);
       expect(pastVote).to.eq.BN(REQUIRED_STAKE);
     });
+
+    it("transactions that try to execute an allowed method on Reputation Voting extension are accepted by the MTX broadcaster", async function () {
```
kronosapiens, line 1534: I'm thinking whether it would be better to keep all the MTX tests inside a single test file which loads in all the extensions etc vs having to put a few similar tests in every test file.

test/extensions/voting-rep.js
```diff
@@ -1525,6 +1530,85 @@ contract("Voting Reputation", (accounts) => {
...
+      const action = await encodeTxData(colony, "setExpenditureState", [1, UINT256_MAX, expenditureId, 25, [true], ["0x0"], WAD32]);
+
+      await voting.createMotion(1, UINT256_MAX, ADDRESS_ZERO, action, domain1Key, domain1Value, domain1Mask, domain1Siblings);
+      motionId = await voting.getMotionCount();
+
+      await voting.stakeMotion(motionId, 1, UINT256_MAX, YAY, REQUIRED_STAKE, user0Key, user0Value, user0Mask, user0Siblings, { from: USER0 });
+
+      await forwardTime(STAKE_PERIOD, this);
+
+      const txData = await voting.contract.methods.finalizeMotion(1).encodeABI();
```
kronosapiens, line 1547: `motionId` please

test/extensions/voting-rep.js
```diff
@@ -1525,6 +1530,85 @@ contract("Voting Reputation", (accounts) => {
...
+      motionId = await voting.getMotionCount();
+
+      await voting.stakeMotion(motionId, 1, UINT256_MAX, YAY, REQUIRED_STAKE, user0Key, user0Value, user0Mask, user0Siblings, { from: USER0 });
+
+      await forwardTime(STAKE_PERIOD, this);
+
+      const txData = await voting.contract.methods.finalizeMotion(motionId.toString()).encodeABI();
+
+      const realProviderPort = process.env.SOLIDITY_COVERAGE ? 8555 : 8545;
+      const provider = new ethers.providers.JsonRpcProvider(`http://127.0.0.1:${realProviderPort}`);
```
kronosapiens, line 1588: One benefit of having all the MTX tests in a single file is that this setup can be tucked into a `before` or `beforeEach` block

test/extensions/voting-rep.js
area: These tests require a reputation state, so I opted to have these tests here, rather than copy the `before` that's in this file in to the metatransaction file (which is much larger than the equivalent `before`s I copied in to here).

packages/metatransaction-broadcaster/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:16-bullseye
```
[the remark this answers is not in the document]
area: Fair, I guess we should take all the dockerfiles back to 14 for consistency...

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
area: The problem is that the other place that this logic is used, it uses the details of what specific checks failed to change the message returned, so returning a single boolean isn't adequate.

kronosapiens: Hmm. Could we return `(result, err)` then? Are the error messages consistent (or could they be made consistent)? Alternatively we could just return `err === undefined` if the checks pass.

test/extensions/voting-rep.js
kronosapiens: Can we put all the MTX tests in a single block so we can have a `beforeEach` for this setup?

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -219,12 +219,13 @@ class MetatransactionBroadcaster {
...
-        // It's possible it's not a valid address, but could be a valid transaction with a token.
-        const validTokenTransaction = await this.isTokenTransactionValid(motionTarget, motion.action, userAddress);
-        const allowedColonyFamilyTransaction = await this.isColonyFamilyTransactionAllowed(motionTarget, motion.action);
+        // Is the motion doing something we'd allow?
+        const { addressValid, validTokenTransaction, allowedColonyFamilyTransaction } = await this.doValidTransactionChecks(
+          motionTarget,
+          motion.action,
+          userAddress
+        );
+        console.log(addressValid, validTokenTransaction, allowedColonyFamilyTransaction);
```
kronosapiens, line 228: Remove this?

kronosapiens (approved):

PR #1073, opened 2022-07-25
Improve noProof oracle endpoint and fast-failure of proof endpoint

area:
When I made the `noProof` endpoint, I removed the logic around building the tree, as it wasn't needed. But I could have also removed the bit where it loads all reputations in the state from the database! So I've done that. Similarly, the proof endpoint will now not load all reputations just to check if the requested reputation does not exist.

### DIFF
--- a/packages/reputation-miner/ReputationMiner.js
+++ b/packages/reputation-miner/ReputationMiner.js
@@ -201,6 +201,7 @@ class ReputationMiner {
     );
 
     this.queries.getReputationStateCount = this.db.prepare(`SELECT COUNT ( * ) AS "n" FROM reputation_states WHERE root_hash=? AND n_leaves=?`);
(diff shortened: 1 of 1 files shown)


arrenv:
Nice one, this should clean up the browser console a little :)

packages/reputation-miner/ReputationMiner.js
```diff
@@ -901,8 +902,8 @@ class ReputationMiner {
 
     // Not in the accepted or next tree, so let's look at the DB
 
-    const allReputations = await this.queries.getAllReputationsInHash.all(rootHash);
-    if (allReputations.length === 0) {
+    const res = await this.queries.getReputationHashCount.get(rootHash);
```
kronosapiens, line 905: Efficient!

kronosapiens (approved):

kronosapiens (approved):

PR #1074, opened 2022-07-29
Make metatransaction broadcaster estimateGas call stricter

area:
[Some spotty tests](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/4687/workflows/40ffdd4e-f7c0-4622-b067-f542948f33c3/jobs/25046) came down to some behaviour here, where the oracle took a long time to responsd, and on thinking about it it did raise a potential issue with the broadcaster, so have decided to change its behaviour slightly (already cleared with the frontend). By including the gas limit with `estimateGas`, we return as quickly as possible. We lose information about whether the transaction fails or is too expensive, but I think that's fine to prevent people deliberately trying metatransactions that use a lot of gas to slow down / DoS the oracle (which is still possible, just harder).

### DIFF
--- a/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
+++ b/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
@@ -46,20 +46,12 @@ class MetatransactionBroadcaster {
         let gasEstimate;
 
         try {
-          gasEstimate = await contract.estimateGas.executeMetaTransaction(userAddress, payload, r, s, v, { gasPrice: this.gasPrice });
-          if (gasEstimate > gasLimit) {
-            return res.status(400).send({
-              status: "fail",
-              data: {
-                payload: "Transaction too expensive and will not be broadcast",
-              },
-            });
-          }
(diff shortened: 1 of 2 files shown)


packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -46,20 +46,12 @@ class MetatransactionBroadcaster {
...
-              },
-            });
-          }
+          gasEstimate = await contract.estimateGas.executeMetaTransaction(userAddress, payload, r, s, v, { gasPrice: this.gasPrice, gasLimit });
         } catch (err) {
           return res.status(400).send({
             status: "fail",
             data: {
-              payload: "Transaction reverts and will not be broadcast",
+              payload: "Transaction reverts and will not be broadcast. It either fails outright, or uses too much gas.",
```
kronosapiens, line 54: I'm inclined to suggest we save all error messages to variables and use the variables here, but it's little more than a vague notion so I'll just let you think about it.

kronosapiens (approved):

PR #1075, opened 2022-07-29
Remove ganache-core

area:
When we did #1052, we upgraded to the new `ganache`, but we clung on to `ganache-core` for the smoke tests. This was because the [internals of the provider were no longer available](https://github.com/trufflesuite/ganache/issues/3076), but the latest release of ganache has provided a mechanism (`eth_getProof`) to get the state hash again. 

Unfortunately this does mean that the state hashes change for this upgrade (see the issue in the ganache repo for the details), but should behave as before going forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,6 @@
   ],
   "dependencies": {
     "@umaprotocol/truffle-ledger-provider": "^1.0.5",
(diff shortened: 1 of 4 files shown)


kronosapiens (approved):
Out of curiosity, why do the state hashes change?

PR #1079, opened 2022-08-10
Sync miner code with docs

kronosapiens:
Some minor tweaks to sync the code with the docs, plus a bit more logging

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -91,7 +91,9 @@ module.exports = {
       network_id: "*",
     },
     xdai: {
-      url: "https://xdai-archive.blockscout.com/",
+      provider: () => {
+        return new HDWalletProvider("replace-with-private-key-when-using", "https://xdai-archive.blockscout.com/",);
+      },
       gasPrice: 2000000000,
       network_id: 100,
     },
--- a/truffle.js
+++ b/truffle.js
@@ -91,9 +91,7 @@ module.exports = {
       network_id: "*",
     },
     xdai: {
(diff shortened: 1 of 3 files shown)


area:
Okay, so, explaining what I've done here:

In order to improve the performance of the `latestState` endpoint, I had to change the type of `reputations.reputation_rowid` to INTEGER from TEXT. That had the unintended knock-on effect of changing the order reputations were returned in from `getAllReputationsInHash`. So, to fix that, I added an ORDER BY to that query, which orders by the unique ID that every reputation has in the state, which determines, among other things, the order that decays should be applied when evolving the state. This was the order that they were returned in previously (which was the order they appeared in the table in the database), but now the correct ordering is always forced.

As part of my rooting around, I discovered that `rowid`s can change under [certain circumstances](https://sqlite.org/lang_vacuum.html) which, given we are using them, would be a _bad thing_. By explicitly declaring a column in the relevant tables called `rowid` which is `INTEGER PRIMARY KEY`, we make sure that these IDs can never change. By naming these columns `rowid`, we didn't have to change associated queries.

truffle.js
```diff
@@ -91,7 +91,9 @@ module.exports = {
       network_id: "*",
     },
     xdai: {
-      url: "https://xdai-archive.blockscout.com/",
```
area, line 94: I actually quite value having `--network xdai` as something that works in the default repo if I have to check something on the contract level.

area (approved):

PR #1081, opened 2022-08-19
First pass at storage layout fixes

area:
Don't merge yet, I want to figure out the git issues, and am not sure if this is the right base. But currently based against (a branch that is a copy of) flwss for reviewing purposes.


kronosapiens (commented):
Looks good to me

kronosapiens (approved):

PR #1086, opened 2022-09-01
Introduce ReputationBootstrapper

### LINKED ISSUE #1069: Support Reputation bootstrapping via email
As discussed today in Product Sync, we would like to allow colony owners to invite users via email, and give them reputation.

After working through a few ideas, we seem to have settled on the following design:

1. A colony admin inputs a list of user emails and the amount of reputation to bootstrap.
2. A `Bootstrapping` extension receives arrays of `sha256(pin)` and reputation amounts.
3. Users receive emails containing their pin, and given a link to the onboarding process.
4. After onboarding, the users can submit their email and pin to claim the reputation.

Incorporating pins makes the process secure -- the admin uploads the hashes, which cannot be reverse-engineered without knowing the email addresses and the pins. As long as the email accounts are secure, then only the users should be able to claim their tokens.

Pin generation should take place in the client, which then sends the email and configures the extension. The pins should never have to be stored or transmitted publicly.

kronosapiens:
Closes #1069 

Add `ReputationBootstrapper`, which allows for setting & claiming of reputation+token grants.

- Grants can be set in bulk
- The extension must be "locked" before grants can be claimed (similar to expenditures)
- A user must "commit" the secret and wait one hour before claiming
- If the `giveTokens` flag is set to `true`, tokens will be given along with reputation
- The reputation grant begins decaying when it is set, at the same rate as reputation in general
- Reputation decay is implemented in `O(logn)` time, in the number of hours of decay


contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,137 @@
...
+    selfdestruct(address(uint160(address(colony))));
+  }
+
+  // Public
+
+  function setGrants(bytes32[] memory _hashedSecrets, uint256[] memory _amounts) public onlyRoot {
+    require(_hashedSecrets.length == _amounts.length, "reputation-bootsrapper-invalid-arguments");
+
+    for (uint256 i; i < _hashedSecrets.length; i++) {
+      require(_amounts[i] <= INT256_MAX, "repuatation-bootsrapper-invalid-amount");
```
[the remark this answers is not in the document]
kronosapiens: I forgot it was that!

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,137 @@
...
+    bytes32 hashedPin = keccak256(abi.encodePacked(_secret));
+    uint256 grantAmount = grants[hashedPin].amount;
+    uint256 grantTimestamp = grants[hashedPin].timestamp;
+
+    require(grantAmount > 0, "reputation-bootstrapper-nothing-to-claim");
+
+    delete grants[hashedPin];
+
+    uint256 tokenAmount = min(ERC20(token).balanceOf(address(this)), grantAmount);
+    require(tokenAmount >= uint256(grantAmount) || tokenAmount <= 0, "reputation-bootstrapper-insufficient-tokens");
```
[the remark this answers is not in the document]
kronosapiens: It's because of the "optional" token behavior. Basically, users can either give reputation OR reputation and tokens. I chose to implement it implicitly, e.g. without having to initialise with a flag, by simply checking to see if the contract held tokens. If the contract has tokens, send them. If the contract has only enough tokens to partially fill the grant, throw an error. I suppose this was my attempt to avoid introducing a flag, but that would have allowed for a more clear implementation.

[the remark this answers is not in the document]
area: Ah, I see... I'm not wild about this approach towards that functionality. In particular, the scenario where I've done one batch of bootstrapping and set up a second, if I add the grants before adding the tokens, enthusiastic users won't get the tokens they're expecting.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -102,6 +99,10 @@ contract ReputationBootstrapper is ColonyExtensionMeta {
 
   // Public
 
+  function setGiveTokens(bool _giveTokens) public onlyRoot {
```
area, line 102: Should this be changeable during the process?

test/extensions/reputation-bootstrapper.js
```diff
@@ -165,14 +166,6 @@ contract("Reputation Bootstrapper", (accounts) => {
       await checkErrorRevert(reputationBootstrapper.claimGrant(PIN1, { from: USER1 }), "reputation-bootstrapper-nothing-to-claim");
     });
 
-    it("cannot claim reputation amounts and tokens if the token amount only partially covers the balance", async () => {
```
area: Shouldn't this still be the case?

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    for (uint256 i; i < _hashedSecrets.length; i++) {
+      require(_amounts[i] <= INT128_MAX, "reputation-bootstrapper-invalid-amount");
+      grants[_hashedSecrets[i]] = Grant(_amounts[i], block.timestamp);
+
+      emit GrantSet(_hashedSecrets[i], _amounts[i]);
+    }
+  }
+
+  function claimGrant(uint256 _secret) public {
```
area, line 117: Unfortunately, I think this type of claim process is fundamentally flawed. If I see a pending transaction for `claimGrant`, I can front-run the transaction with the secret from my own account with a higher gas price, and 'snipe' it.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,148 @@
...
+    bytes32 hashedSecret = keccak256(abi.encodePacked(_secret));
+    uint256 grantAmount = grants[hashedSecret].amount;
+    uint256 tokenAmount = grants[hashedSecret].amount;
+    uint256 grantTimestamp = grants[hashedSecret].timestamp;
+
+    require(grantAmount > 0, "reputation-bootstrapper-nothing-to-claim");
+
+    delete grants[hashedSecret];
+
+    // First, decay by any full days
```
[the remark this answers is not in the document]
kronosapiens: Ok, I see what you're getting at, if we cut the time in half and double the decay factor we'll be able to apply only a handful of calculations rather than all of them. So for a 50 hour decay we'd end up applying essentially `(k**2)(k**16)(k**32)` which is equivalent to `(k**50)`.

contracts/extensions/ReputationBootstrapper.sol
kronosapiens: It's a good point, I worry about going down a recursive sort of rabbit hole. Do we add a "locked" boolean, which prevents more claims from being set? Can the "locked" boolean itself be set back to false? If so, then an admin could go back and disable tokens anyway. We could have the "locked" boolean be one-way, which also checks to make sure there's sufficient tokens available. But someone could always deprecated & uninstall the extension and reclaim the funds, unless we say you cannot delete the extension until all the funds are claimed, but what if then someone loses the email and never comes to claim? Are funds locked permanently?

I would lean in favor of leaving it open, since ultimately a colony is choosing to give tokens here. I think we can assume that they will continue to make that choice. Put another way, I don't think this is such an adversarial setting.

contracts/extensions/ReputationBootstrapper.sol
kronosapiens: I see that. One solution which comes to mind would be to require claiming addresses to have a `joincolony.eth` ENS name before the claim. That would require users to go through the onboarding flow before presenting their secret, and would help alleviate this risk (but not remove it entirely). What do you think of that? I suppose miners could go ahead and create their ENS names before proceeding to run this attack.

Another idea would be to require a two-part "handshake" where users present the secret and has their address recorded, but ultimately the admin has to go back and hit the button to actually transfer the funds. That would be a product decision though, so if you think it's promising we should rope in the rest of the crew.

test/extensions/reputation-bootstrapper.js
kronosapiens: Yes, I deleted it because we'd only be checking the underlying underflow math for token transfers, which I didn't think was particularly interesting, but I can put it back.

area: That's ultimately the error that throws, but we're interested in checking the logic that if that does end up throwing, then we don't also award reputation.

contracts/extensions/ReputationBootstrapper.sol
area: As-ever, when it's a question of the actual requirements (rather than the specific implementation), we need to go back to product (-:

contracts/extensions/ReputationBootstrapper.sol
area: > I suppose miners [sic] could go ahead and create their ENS names before proceeding to run this attack.

Indeed.

>Another thing we could say is that this contract is knowingly vulnerable to this attack, and so should be used only for small amounts of bootstrapping funds. I expect you won't like that solution, but we can put it on the table.

I do indeed not like that solution!

> ultimately the admin has to go back and hit the button to actually transfer the funds.

There seems very little benefit to using this extension, in that case?

I think the only possible on-chain solution is a commit-and-reveal on the user's side as well, with a minimum timestamp between those two transactions to prevent front-running of both transactions.

I do think product needs to be involved to sign off on any of this though.

kronosapiens: @arrenv what are your thoughts on this? Alex is saying that the design as-is is vulnerable to miner front-running. One solution would be to require two transactions, the first essentially "registering" the address and the intended hashed secret, and the second providing the secret itself, checking that the caller is the initial address. By introducing a delay (on the order of ~30 seconds or so I would suppose, maybe less) then a miner could not front-run. Would this two-transactions-plus-delay flow be acceptable?

contracts/extensions/ReputationBootstrapper.sol
kronosapiens: @arrenv can you weigh in here? Essentially the question is whether the flag to give tokens or not should be locked, or whether we can leave it an open setting which can (in theory) be changed by admins at any point. Locking the flag introduces other considerations, as I've touched on in my earlier comment.

contracts/extensions/ReputationBootstrapper.sol
area: I would want it to be at least an hour

kronosapiens: That seems like a very long amount of time and meaningfully affects the UX. I would expect an experience along the lines of a token bridge (e.g. two transactions, one on each chain, 15 or so blocks apart) would be more palatable to users.

area: How long have you had to wait for a transaction to be mined in the past? If the timescale is smaller than that, even a commit-and-reveal can be front-run.

kronosapiens: When bridging from xdai to mainnet it's quick, but that's because we're waiting on the xdai blocks which are snappy. If we're waiting on mainnet then I suppose it would be longer, although post-merge maybe not so much. In any case, what do you think of phrasing the delay in terms of block confirmations and not an absolute amount of time?

kronosapiens:
@arrenv can you weigh in on the above issues?

contracts/extensions/ReputationBootstrapper.sol
arrenv: From what I'm understanding, the question is whether an admin should be able to toggle the bootstrapping claims. This does not seem very useful unless values could be changed.

It does seem like a good idea to be able to cancel a bootstrapping if a mistake is made, this seems like it could be handled by deprecating the extension.

kronosapiens: I think a good solution here would be to add a one-way "locked" switch, similar to what we do with expenditures. While in a "draft" state, the parameters can all be changed. Once "locked", the parameters are set and people can begin claiming.

The extension could then be deprecated only once the token balance was down to zero (as then everyone would have claimed, or if the only grants are reputation, there's no cost to removing them).

Would that be acceptable? It gives admins the flexibility to configure the contract, but then gives some guarantees to end users once they start claiming.

arrenv: @kronosapiens Nice, yeah that sounds acceptable to me.

contracts/extensions/ReputationBootstrapper.sol
arrenv: Having the two transactions sounds acceptable to me, then using a block confirmation as a delay. As it would then be compatible on different chains. In the UI we can then use an approx. time for making the claim based on the chains block rate.

area: Why is block confirmations better than an absolute amount of time?

kronosapiens: I would say because block confirmations get more directly at the issue we are about, which is preventing miners from mining transactions to benefit themselves. The more blocks are confirmed, the harder it becomes for a miner to selfishly mine. Further, block confirmations allow us to move more quickly on chains with short block times, while still providing security on slower chains, waiting "an hour" if need be. If we waited a fixed amount of time, that could either be way too long (on fast chains), or potentially still not long enough (if for whatever reason transactions take a very long time to mine).

area: > more directly at the issue we are about, which is preventing miners from mining transactions to benefit themselves

This isn't really the issue, nor is miners rewriting history as you seem to be implying. The issue is:

* A user commits `keccak(walletAddress, secret)` (transaction 1)
* That user then reveals `secret` via another transaction (transaction 2), claiming their reputation 
* A malicious user sees transaction2 in the pending transaction pool, and commits `keccak(walletAddress2, secret` and then reveals `secret`, all before transaction 2 is mined (because they're paying very high gas fees).

The 'honest' transactions are likely to take a little while to be mined, as they are almost certainly metatransactions that we're paying for, and so going to be at the low end of the fee scale. The likely time for those transactions to be mined has to be _significantly_ lower than the minimum allowed time between a commit and reveal.

kronosapiens: I'm following. The issue at hand is less the amount of blocks since the initial tx was confirmed and more about making sure that a slower tx isn't scooped up by a faster one.

Sounds like the timeline you have in mind is:

```
t0         [... hour... ]          t1                                     [...hour...]              t2 
secret committed           secret revealed
                                         malicious user commits secret                   malicious user reveals secret
```

You are concerned with making sure that the honest transaction at t1 is mined before the dishonest one at t2.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -137,11 +145,27 @@ contract ReputationBootstrapper is ColonyExtensionMeta {
     }
   }
 
+  /// @notice Commit the secret, beginning the security delay window
+  /// @param _committedSecret A sha256 hash of (userAddress, secret)
+  function commitSecret(bytes32 _committedSecret) public locked {
+    committedSecrets[_committedSecret] = block.timestamp;
```
area, line 151: Unfortunately an attacker can call `commitSecret` with the committed (good) secret after the user has broadcast the `claimGrant` transaction and prevent them from claiming indefinitely, allowing the attacker to claim at their leisure. This mapping needs to be based on `msgSender()` to prevent that, I think? I guess to future-proof ourselves in case a user needs to go through this process multiple times, it can be based on `keccak(msgSender(), committedSecret` to allow them to do so in parallel.

kronosapiens: Gotcha. I had initially thinking about including `msgSender` but didn't as the hash includes `msgSender`, but I see your point about the griefing. In what case do we need to future-proof, assuming each secret can be claimed only once?

kronosapiens:
@arrenv as @area mentioned in the team meeting, we'd like to hop on a call with you to iron out some edge-case behaviors around paying tokens, specifically around balancing security / efficiency (more bookkeeping vs lower gas costs, etc).

arrenv:
> @arrenv as @area mentioned in the team meeting, we'd like to hop on a call with you to iron out some edge-case behaviors around paying tokens, specifically around balancing security / efficiency (more bookkeeping vs lower gas costs, etc).

Of course, I have a call this afternoon for CDApp, and tomorrow is product call, would Thursday work?

kronosapiens:
Following on the product sync call:

- Fund contract, add grants, claim in an ongoing way
- Deprecation means all grants cannot be claimed
- Each grant gets a "payTokens" flag, used in a mapping to keep track
- Can update the values for a grant
- We are fine with a one-hour hardcoded security delay

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,194 @@
...
+
+      emit GrantSet(_paid[i], _hashedSecrets[i], _amounts[i]);
+    }
+  }
+
+  /// @notice Commit the secret, beginning the security delay window
+  /// @param _committedSecret A sha256 hash of (userAddress, secret)
+  function commitSecret(bytes32 _committedSecret) public notDeprecated {
+    bytes32 addressHash = keccak256(abi.encodePacked(msgSender(), _committedSecret));
+    committedSecrets[addressHash][_committedSecret] = block.timestamp;
```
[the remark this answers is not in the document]
kronosapiens: If we only key on address, then we would preclude someone getting a grant in two domains (part of the "future proofing" you brought up earlier).

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,194 @@
...
+        grantAmount = grantAmount * adjustedNumerator / decayDenominator;
+      }
+      // slither-disable-next-line divide-before-multiply
+      adjustedNumerator = adjustedNumerator * adjustedNumerator / decayDenominator;
+      decayEpochs >>= 1;
+    }
+
+    colony.emitDomainReputationReward(1, msgSender(), int256(grantAmount));
+
+    if (tokenAmount > 0) {
```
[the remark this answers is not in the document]
kronosapiens: `grantAmount` is decayed earlier in the function and we've already deleted the grant as part of the `check-effects-interaction` pattern, so I'm not sure that would work

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,194 @@
...
+
+    require(
+      commitTimestamp > 0 && commitTimestamp + SECURITY_DELAY <= block.timestamp,
+      "reputation-bootstrapper-commit-window-unelapsed"
+    );
+
+    bytes32 hashedSecret = keccak256(abi.encodePacked(_secret));
+    uint256 grantAmount = grants[_paid][hashedSecret].amount;
+    uint256 tokenAmount = _paid ? grants[_paid][hashedSecret].amount : 0;
+    uint256 grantTimestamp = grants[_paid][hashedSecret].timestamp + SECURITY_DELAY; // Don't decay during delay window
```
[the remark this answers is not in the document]
area: I mean... so what? If everyone is penalized, no-one is penalized. If anything, I would say this gives an advantage to people who are bootstrapped though the extension over those who are awarded reputation directly, as their reputation has decayed for an hour less.

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
area: Could we then not do this transfer before the grant is decayed? So still after the grant has been deleted, and before the decay and emission of reputation?

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
area: Neither of my suggestions key only on address. They both involved the address and the committed secret once each, just in two different ways. The current implementation depends on the address once and the committed secret twice, which seems unnecessary to me.

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
kronosapiens: Not following your logic -- are people awarded reputation directly decayed by two hours?

I don't feel strongly, if anything I would've expected you to come down on the side of precision, so I can leave this out.

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
kronosapiens: We can move the code, but wouldn't putting the boolean in the event obscure the actual amount of tokens received?

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
kronosapiens: Yup, I'm following.

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
area: If one person is awarded 1000 reputation directly, and another person is awarded 1000 reputation via bootstrapping at the same time (and goes through the claim process immediately), a day later the first person has had their reputation decay for 24 hours, and the second person has had their reputation decay for 23 hours, and so has more influence than the first, which (seems to me) was quite clearly not the intention of the admin.

I think maybe we're just not agreeing on when the 'clock' should start, but in my opinion it's pretty clear that it should be when it's put anywhere in to the system. Otherwise, it becomes very hard for a user to calibrate the reputation they're awarding correctly if they're using the two different processes.

contracts/extensions/ReputationBootstrapper.sol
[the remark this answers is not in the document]
area: I only really introduced the idea of the boolean to let us get rid of the `tokenAmount` variable (gas optimisation hat on), but if we can get rid of that variable and still emit the amount, we can as well.

It didn't really bother me as there will be a `Transfer` event in the transaction if someone cares about the amount.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,191 @@
...
+  // Data structures
+
+  struct Grant {
+    uint256 amount;
+    uint256 timestamp;
+  }
+
+  // Storage
+
+  address public token;
```
area, line 50: Is there a reason we've gone with public variables rather than our own defined getters? I like it, but just wondering.

kronosapiens: Good question. I was going to say "because it's not an EtherRouter" but then I would be thinking of the CoinMachine contract, in which we made them public for that reason. Perhaps because there's no interface contract? I can't fully remember, and going over other extensions I see that it isn't the style we use elsewhere.

kronosapiens: In the spirit of consistency I'll make them private and add getters.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,194 @@
...
+    require(_paid.length == _hashedSecrets.length, "reputation-bootstrapper-invalid-arguments");
+    require(_hashedSecrets.length == _amounts.length, "reputation-bootstrapper-invalid-arguments");
+    uint256 balance = ERC20(token).balanceOf(address(this));
+
+    for (uint256 i; i < _hashedSecrets.length; i++) {
+      require(_amounts[i] <= INT128_MAX, "reputation-bootstrapper-invalid-amount");
+
+      if (_paid[i]) {
+        uint256 priorGrant = grants[_paid[i]][_hashedSecrets[i]].amount;
+        if (priorGrant < _amounts[i]) {
```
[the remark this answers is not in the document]
kronosapiens: For some reason, this optimization seems to be totally dropped by the compiler. I tried running it several times and each time I got the same results for both:

Naive:
Min: 106086
Max: 174898
Max: 129023

Optimized:
Min: 106203
Max: 174603
Avg: 129003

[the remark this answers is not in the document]
kronosapiens: So all else being (literally) equal I would go with the naive version for readability.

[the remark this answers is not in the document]
area: How many grants were you setting at once?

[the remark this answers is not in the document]
kronosapiens: That was for two. I've re-run it with five and have the following:

Naive:
Min: 297112
Max: 319872
Avg: 304699

Optimized:
Min: 297229
Max: 318959
Avg: 304472

Looks like it made a difference there but only very slightly

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,219 @@
...
+  /// @param _committedSecret A sha256 hash of (userAddress, secret)
+  function commitSecret(bytes32 _committedSecret) public notDeprecated {
+    bytes32 addressHash = keccak256(abi.encodePacked(msgSender(), _committedSecret));
+    committedSecrets[addressHash] = block.timestamp;
+  }
+
+  /// @notice Claim the grant, after committing the secret and having the security delay elapse
+  /// @param _secret The secret corresponding to a reputation grant
+  function claimGrant(bool _paid, uint256 _secret) public notDeprecated {
+    bytes32 committedSecret = keccak256(abi.encodePacked(msgSender(), _secret));
```
area, line 146: Is this not still an unnecessary amount of hashing? This could just be `keccak256(abi.encodePacked(_secret))`, no?

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    colony.emitDomainReputationReward(1, msgSender(), int256(grantAmount));
+
+    emit GrantClaimed(msgSender(), grantAmount, _paid);
+  }
+
+// View
+
+  function getToken() public view returns (address) {
```
area, line 192: These could all be external, I believe.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,219 @@
...
+        grantAmount = grantAmount * adjustedNumerator / decayDenominator;
+      }
+      // slither-disable-next-line divide-before-multiply
+      adjustedNumerator = adjustedNumerator * adjustedNumerator / decayDenominator;
+      decayEpochs >>= 1;
+    }
+
+    colony.emitDomainReputationReward(1, msgSender(), int256(grantAmount));
+
+    emit GrantClaimed(msgSender(), grantAmount, _paid);
```
area, line 187: Have we (or product?) given any amount of consideration as to which `grantAmount` should actually be emitted here (i.e. the decayed amount, or the original)?

contracts/extensions/ReputationBootstrapper.sol
kronosapiens: I don't think so, because otherwise someone could supply a hash on behalf of another user. Although I suppose there's nothing wrong with that because it is only to the user's advantage (e.g. they can claim earlier than they otherwise would have)...

contracts/extensions/ReputationBootstrapper.sol
kronosapiens: Not to my knowledge.

@arrenv, do you have any preference? When a user claims a grant, should we emit the initial amount, the decayed amount actually claimed, or both?

arrenv: It makes the most sense that this should emit the decayed amount, as they never receive the initial amount and thus it would be inaccurate.

contracts/extensions/ReputationBootstrapper.sol
```diff
@@ -0,0 +1,219 @@
...
+  /// @notice Commit the secret, beginning the security delay window
+  /// @param _committedSecret A sha256 hash of (userAddress, secret)
+  function commitSecret(bytes32 _committedSecret) public notDeprecated {
+    bytes32 addressHash = keccak256(abi.encodePacked(msgSender(), _committedSecret));
+    committedSecrets[addressHash] = block.timestamp;
+  }
+
+  /// @notice Claim the grant, after committing the secret and having the security delay elapse
+  /// @param _secret The secret corresponding to a reputation grant
+  function claimGrant(bool _paid, uint256 _secret) public notDeprecated {
```
area, line 145: I guess it doesn't actually matter, but it seems weird to me that `_secret` is a `uint256` rather than a `bytes32`.

kronosapiens: That's a legacy from when it was a numeric pin. I can change it, but I actually kind of prefer it as a number.

area: Why's that?

kronosapiens: I guess I'm assuming that generating random numbers vs random hashes will be more straightforward client-side, and seeing a number in an email is a bit more familiar than seeing a hash. Information-wise they are equivalent.

kronosapiens: @arrenv do you have a preference as to whether the secrets use to claim reputation are numbers or hashes?

arrenv: My preference would be numbers, as that is more consistent experience for users. And I agree, hashes are more of denotation of addresses.

area (approved):

PR #1090, opened 2022-09-23
Enable 'No action' (simple decicion) motions in subdomains

area:
The original implementation in #1071 failed to consider we would want to make such NO_ACTION motions in subdomains. In `createMotion`, `ColonyRoles(target).getCapabilityRoles(action) | ROOT_ROLES == ROOT_ROLES` would evaluate to true, as the colony does not know a function with the signature `0x12345678`, and so would return `0x0` ffrom `getCapabilityRoles`.

As a result, attempting to create such a motion in anything other than the root domain would fail.

This PR fixes the issue by adding a special case on creation for such a motion, and adds a test demonstrating the issues is resolved. I have asserted that such a motion has an 'effect' on the domain the vote is taking place in.


test/extensions/voting-rep.js
```diff
@@ -1627,6 +1636,29 @@ contract("Voting Reputation", (accounts) => {
...
     });
 
+    it("motions with the special NO_ACTION signature can be created in subdomains", async function () {
+      const action = "0x12345678";
+
+      const user0Key2 = makeReputationKey(colony.address, domain2.skillId, USER0);
+      const user0Value2 = makeReputationValue(WAD.divn(3), 9);
+      const [user0Mask2, user0Siblings2] = await reputationTree.getProof(user0Key2);
+
+      console.log(2, UINT256_MAX, ADDRESS_ZERO, action, domain2Key, domain2Value, domain2Mask, domain2Siblings);
```
kronosapiens, line 1646: Remove this

contracts/testHelpers/VotingReputationMisalignedFixed.sol
```diff
@@ -0,0 +1,1021 @@
+/*
```
kronosapiens, line 1: This looks like changes related to a separate issue -- does this belong in this PR?

area: These changes are to fix the tests around the storage misalignment issue, which are necessary because the version for VotingReputation is bumped here.

These could be in another PR, or indeed removed entirely (do we need tests showing that upgrade works, now it's in the past?)

kronosapiens: I would be happy to remove this then, it's a lot of code that serves minimal purpose.

kronosapiens (dismissed):

area:
Used superpowers as Daniel is off today but had approved this (modulo changes to the 'skip coverage' commit, but I assume that he's fine with that given we discussed the intention there).

Issue #1058, opened 2022-06-23
Investigate colony creation bundling


kronosapiens:
- Stage 1
        - Bundle 1:
            - Create Colony
            - Create Token
            - Assign Token
            - Mint Token
            - Claim Tokens
            - OneTxPayment Extension
            - Expenditures Extension
        - Bundle 2:
            - Create Colony
            - Assign Token
            - OneTxPayment Extension
            - Expenditures Extension
    - Stage 2
        - Bundle 3:
            - Multisig only
        - Bundle 4:
            - Governance only
        - Bundle 5:
            - Multisig + Governance
    - Stage 3
        - Bundle 6:
            - Airdropping tokens or just reputation

arrenv:
@kronosapiens, I came across these as well, which may be more versatile options if feasible.

- [https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol](https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol)
- [https://github.com/safe-global/safe-contracts/blob/da66b45ec87d2fb6da7dfd837b29eacdb9a604c5/contracts/libraries/MultiSend.sol](https://github.com/safe-global/safe-contracts/blob/da66b45ec87d2fb6da7dfd837b29eacdb9a604c5/contracts/libraries/MultiSend.sol)

area:
Assuming it works with Etherrouters (which I believe it should), a Multicall-esque implementation should be our preference here.

area:
The `feat/multicall` branch now exists with the first tentative steps for this.

PR #1091, opened 2022-09-23
Add multicall support to contracts

area:
Closes https://github.com/JoinColony/colonyNetwork/issues/1058

This PR adds Multicall support to colony contracts - currently just to Colony and ColonyNetwork, but will be added to extensions as well. Looking for a review, but not a merge, at this point.

The multicall functionality allows a single transaction to call multiple functions on the same contract, assuming that all the parameters are know at the time of the call. So, for example, this wouldn't be appropriate for creating an expenditure, and then editing it, because the id of the expenditure is not known at the time of the transaction. 

The multicall implementation is from [Uniswap](https://github.com/Uniswap/v3-periphery), though I've changed it. I've removed payable, to stop us [tripping ourselves up in the future](https://github.com/Uniswap/v3-periphery/issues/52).

It also required some changing to allow compatability with metatransactions, which with the default implementation wouldn't work as expected. That's where I'm most interested in getting feedback - as with anything related to metatransactions, that needs to be considered very carefully and has to be right otherwise there will be a _lot_ of problems.

Some changes were also required on the metatransaction broadcaster, as we don't want calls that we otherwise wouldn't pay for being allowed just because they're wrapped inside a multicall (i.e. we don't want to just pay for all multicalls).


contracts/common/Multicall.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Mostly taken from Uniswap/v3-periphery repository, with payable removed and solidity version
+// adjustments
+pragma solidity 0.7.3;
+pragma experimental "ABIEncoderV2";
+import "./MetaTransactionMsgSender.sol";
+
+/// @title Multicall
+/// @notice Enables calling multiple methods in a single call to the contract
```
kronosapiens, line 9: Not consistent with our style to have comments above contracts...

contracts/common/Multicall.sol
```diff
@@ -0,0 +1,38 @@
...
+abstract contract Multicall is MetaTransactionMsgSender {
+
+
+    function multicall(bytes[] calldata data) public returns (bytes[] memory results) {
+        // First off, is this a metatransaction?
+        address sender = msgSender();
+        bytes memory affix;
```
kronosapiens, line 16: Would prefer this as a ternary, but that's just a personal style preference:

```
bytes memory affix = (msg.sender == address(this) && msgSender() != address(this)
    ? abi.encodePacked(METATRANSACTION_FLAG, msgSender())
    : "0x0"
```

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -234,6 +234,25 @@ class MetatransactionBroadcaster {
       // Not a voting rep related transaction (we recognise)
     }
 
+    const multicallDef = await this.loader.load({ contractName: "Multicall" }, { abi: true, address: false });
+    const possibleMulticall = new ethers.Contract(target, multicallDef.abi, this.wallet);
+
+    try {
+      const tx = possibleMulticall.interface.parseTransaction({ data: txData });
+      if (tx.signature === "multicall(bytes[])") {
+        // We check for each multicall whether it's doing something we'd allow
```
kronosapiens, line 243: Setting `const actions = tx.args[0]` would make the `for` loop a bit more readable, and you could use `actions[i]` pretty clearly.

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -234,6 +234,25 @@ class MetatransactionBroadcaster {
...
+        for (let i = 0; i < tx.args[0].length; i += 1) {
+          const action = tx.args[0][i];
+          const valid = await this.isColonyFamilyTransactionAllowed(target, action, userAddress);
+          if (!valid) {
+            return false;
+          }
+        }
+      }
+    } catch (err) {
```
kronosapiens, line 252: Under what circumstance would we `err` here? E.g. why is this wrapped in try/catch?

test/contracts-network/colony-network-extensions.js
```diff
@@ -163,6 +163,36 @@ contract("Colony Network Extensions", (accounts) => {
...
+    it("allows a root user to install multiple extensions in a single transaction via multicall", async () => {
+      const installExtension1Data = await colony.contract.methods.installExtension(TEST_EXTENSION, 2).encodeABI();
+      const installExtension2Data = await colony.contract.methods.installExtension(TEST_VOTING_TOKEN, 1).encodeABI();
+      await checkErrorRevert(colony.multicall([installExtension1Data, installExtension2Data], { from: USER }), "ds-auth-unauthorized");
+
+      let extension1Address = await colonyNetwork.getExtensionInstallation(TEST_EXTENSION, colony.address);
+      let extension2Address = await colonyNetwork.getExtensionInstallation(TEST_VOTING_TOKEN, colony.address);
+
+      expect(extension1Address).to.equal(ethers.constants.AddressZero);
+      expect(extension2Address).to.equal(ethers.constants.AddressZero);
```
kronosapiens, line 175: I don't know if it's necessary to test the zero address here since we've already established the transaction failed, it just makes the tests longer and harder to parse.

test/packages/metaTransactionBroadcaster.js
```diff
@@ -489,6 +490,123 @@ contract("Metatransaction broadcaster", (accounts) => {
...
+      expect(roles).to.equal(ethers.utils.hexZeroPad(ethers.BigNumber.from(expectedRoles).toHexString(), 32));
+    });
+
+    it("a multicall transaction that calls something invalid is rejected", async function () {
+      const txData1 = await colony.contract.methods.setArchitectureRole(1, UINT256_MAX, USER1, 1, true).encodeABI();
+
+      const colonyAsEtherRouter = await EtherRouter.at(colony.address);
+      const resolverAddress = await colonyAsEtherRouter.resolver();
+
+      const txData2 = await colony.contract.methods
```
kronosapiens, line 546: Can you add a comment clarifying this tx? It's a lot to parse.

contracts/common/Multicall.sol
area: I'm not exactly sure what you're after here?

contracts/common/Multicall.sol
area: I know you like ternaries more than I do, but even so... do you really think yours is easier to reason about/understand than mine?

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
area: When it's not a multicall transaction? I think I'm not understanding the intent of the question :sweat_smile:

test/contracts-network/colony-network-extensions.js
area: Have we established that the transaction failed? What if, to pick a totally random example, someone refactored `checkErrorRevert` incorrectly :laughing:

contracts/common/Multicall.sol
kronosapiens: I'm saying it's stylistically inconsistent to have natspec comments above contract definitions

contracts/common/Multicall.sol
kronosapiens: I think it's a bit more succinct, but as I said, it's a style choice I leave to you

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
kronosapiens: Curious which line would throw the exception. Is it this line `const tx = possibleMulticall.interface.parseTransaction({ data: txData });`?

test/contracts-network/colony-network-extensions.js
kronosapiens: I suppose so, especially in light of that. It's never wrong to be rigorous.

contracts/common/Multicall.sol
area: Ah, I had parsed your original comment as "Not consistent with [our style to have comments above contracts...]" and was confused :sweat_smile:

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
area: Yep - if it's not a multicall transaction, then that interface cannot parse the transaction data and throws an error.

contracts/common/Multicall.sol
area: Would you like me to remove them from other places they are present as well (i.e. IMulticall, IRecovery, ContractRecovery)?

kronosapiens: I suppose that is what I'm saying, although now I feel like a nitpick. Do you have strong feelings about the stylistic consistency?

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
kronosapiens: In that case, can we limit the try/catch to just that line? Wrapping the entire section in a try/catch suggests that other parts of the statement could throw.

contracts/common/Multicall.sol
area: If we were code golfing, I'd be all for it. But we're trying to write an easily-maintainable (and therefore easily understandable) codebase.

packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
area: We don't want any code in the try to run if that line errors though. Are you looking for something like:
```
try {
   statementThatMightThrow
} catch (err) {
   error = err
}
if (!error) {
   rest of Code
}
```

I'll also note that this exact pattern is used twice more in the file; am I sorting those out too?

contracts/common/Multicall.sol
area: Nope, I just copied a random contract that happened to already have them.

kronosapiens: Ok. I'm not going to ask you to go around editing unrelated files, so let's leave it at this for now.

kronosapiens:
@area is this still a draft?

area:
I still don't want this merged quite yet, but good for reviewing.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -93,6 +94,43 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
     return colonyAddress;
   }
 
+  function createColonyForFrontend(
```
kronosapiens, line 97: Why not have two functions, one which accepts a token and one which deploys? The frontend could choose which to call.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -93,6 +94,43 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
+    address _tokenAddress,
+    string memory _name,
+    string memory _symbol,
+    uint8 _decimals,
```
kronosapiens, line 101: IMO we should drop this argument, I expect users choose something other than 18 extremely rarely.

contracts/common/Multicall.sol
```diff
@@ -0,0 +1,34 @@
...
+        if (msg.sender == address(this) && sender != address(this)){
+            // If it's a metatransaction, we re-append the metatransaction identifier to each call we make
+            affix = abi.encodePacked(METATRANSACTION_FLAG, sender);
+        }
+
+        results = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
```
kronosapiens, line 19: Minor point but elsewhere we don't explicitly initialise the values in for loops, e.g. it's just `uint256 i; ...`

test/packages/metaTransactionBroadcaster.js
```diff
@@ -489,6 +490,125 @@ contract("Metatransaction broadcaster", (accounts) => {
       expect(allowed).to.eq.BN(0);
     });
 
+    it("a valid transaction that uses multicall is broadcast and mined", async function () {
+      await colony.contract.methods.setArchitectureRole(1, UINT256_MAX, USER1, 1, true).send({ from: USER0 });
```
kronosapiens, line 494: Isn't this redundant?

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: That's a totally fine opinion to hold, but this isn't the place to make that decision (which needs to start at the top with product, and work down from there)

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: Mostly I don't want to encourage the idea that to add a new flow we just add a new function. I dislike that we're putting this function in in the first place.

kronosapiens (dismissed):
🚀

kronosapiens (approved):

PR #1092, opened 2022-09-28
Use self-consistent coverage configurations for separate runs

area:
We've seen [this issue](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/4931/workflows/7cd8361c-b366-4c33-ac6f-5bf6b809a315/jobs/27125) in a [lot of builds](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/4921/workflows/2487d6f6-097f-434e-8eba-4f62c470529d/jobs/26996) lately. It was intermittent, though, and so tough to pin down. ~~I was also unable to reproduce locally. It ended up being due to [this issue](https://github.com/gotwarlost/istanbul/pull/940) now that we're on Node 14 (and I'm attributing the inablility to reproduce locally due to `node`/`npm` version weirdness I've got going on locally).~~

~~I ended up forking `istanbul-combine` to use a fixed `istanbul` under the hood, which seems to solve the issue.~~

~~Unfortunately, the resulting `package-lock.json` [threw an error](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/4937/workflows/16500303-2a9a-4ff1-8f24-86282d9be670/jobs/27135) when using `npm ci` (though not `npm i`, weirdly). Updating `npm` for the `check-coverage` stage of the build fixes that.~~

### DIFF
--- a/.solcover.extensions.js
+++ b/.solcover.extensions.js
@@ -1,33 +1,5 @@
-const { execSync } = require("child_process");
-const log = console.log;
+const config = require("./.solcover.js")
 
-// Copies pre-built token artifacts to .coverage_artifacts/contracts
-function provisionTokenContracts(config){
-  let output;
-  const provisionColonyToken = `bash ./scripts/provision-token-contracts.sh`;
+config.istanbulFolder = "./coverage-extensions"
 
-  log('Provisioning ColonyToken contracts...')
-  output = execSync(provisionColonyToken);
-  log(output.toString())
-}
-
-module.exports = {
-    skipFiles: [
-      'Migrations.sol',
-      'common/EtherRouter.sol',
-      'patriciaTree',
-      'testHelpers',
-      'ens'
-    ],
-    providerOptions: {
-      port: 8555,
-      network_id: 1999,
-      account_keys_path: "./ganache-accounts.json",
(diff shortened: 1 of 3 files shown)


chmanie:
I'm taking a look at this. The complete overhaul of the package-lock.json seems a bit suspicious.

area:
Taking another swing at this. Diving down in to the guts, it turns out there was a race condition, but the warning being shown was unrelated. I do still think this was triggered somehow by Node 14, but at this point I just want out rather than understanding why :stuck_out_tongue: 

Because some of our coverage instrumented the `always` modifier, and some didn't, we ended up with coverage runs for the same file (e.g. `Colony.sol`) that had different numbers of branch points. If one with fewer branchpoints was the first to be merged, then when the time came to merge the one with more branchpoints there was an error silently thrown in `istanbul`.

To fix this, I've inhereted the 'base' config where possible for `solidity-coverage` across all our runs, and changed the configuration where appropriate (primarily, the output directory).

We still have the warning about the circular dependency due to node 14, but that doesn't seem to be affecting anything with this change, so happy to leave it as-is.

kronosapiens (approved):

PR #1094, opened 2022-10-11
Fix incorrect refactor of checkErrorRevert

area:
Looks like we got away with this one - nothing was overlooked in this repository, except one test where we [decided to remove the functionality it tested](https://github.com/JoinColony/colonyNetwork/pull/1045#discussion_r922224613)


kronosapiens (approved):
😅

PR #1096, opened 2022-10-18
Fix streaming payment extension

area:
There's a bug in Streaming Payments that my multicall work uncovered. This PR fixes it.

EDIT: I think that function really should be expecting the expenditure id, but it's much more gas efficient like this.


kronosapiens (approved):
Yup, that definitely should've been `domainFundingPotId`. Thanks for catching that. I think it's fine to pass in `domainFundingPotId` to `getAmountClaimable` also, since ultimately we care about the funding pot in that function, not the streaming payment which points to it.

PR #1112, opened 2023-01-18
Fix MTX broadcaster intermittent bug; pay for bridge transactions

### LINKED ISSUE #1111: Whitelist AMB Contracts in MetaTx for Safe Control
## Description
In order to be able to provide a smoother experience for the Safe Control feature. We would like to try whitelisting the AMB contracts in MetaTx for Safe Control arbitrary transactions.

The contract addresses for each of the current 2 AMB contracts can be found here - https://github.com/JoinColony/colonyDapp/blob/67f63c98b51bfd8d38e16712866fbbf4ea46b133/src/modules/constants.ts#L295

area:
Closes #1111 . We want people using the safe-control feature to not be obliged to pay for the bridging transaction on the Gnosis Chain side, so have whitelisted the two (currently known / cared about) bridge contracts, and allowed calls to the arbitrary transactions functions on a colony to call exactly `requireToPassMessage` on those contracts and still be eligible for using our metatransaction broadcaster.

Also fixes an untracked bug where, if the gas-price oracle used returned certain values, a non-integer would be passed to hexlify which would error. That should now no longer be possible, regardless of what value is returned by the oracle.

Based on #1113 so no merging until that is merged.

### DIFF
--- a/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
+++ b/packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
@@ -6,6 +6,9 @@ const queue = require("express-queue");
 const NonceManager = require("./ExtendedNonceManager");
 const { colonyIOCors, ConsoleAdapter, updateGasEstimate } = require("../package-utils");
 
+const ETHEREUM_BRIDGE_ADDRESS = "0x75Df5AF045d91108662D8080fD1FEFAd6aA0bb59";
+const BINANCE_BRIDGE_ADDRESS = "0x162E898bD0aacB578C8D5F8d6ca588c13d2A383F";
+
 class MetatransactionBroadcaster {
   /**
    * Constructor for MetatransactionBroadcaster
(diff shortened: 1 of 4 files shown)


packages/metatransaction-broadcaster/MetatransactionBroadcaster.js
```diff
@@ -194,14 +197,35 @@ class MetatransactionBroadcaster {
...
+        }
+        if (addresses.length !== calls.length) {
+          // If the arrays don't line up, return false.
+          return false;
+        }
+        for (let i = 0; i < addresses.length; i += 1) {
+          if (
+            !MetatransactionBroadcaster.isBridgeAddress(addresses[i]) || // It's not going to a bridge address
+            // Or it is, but not calling the function we allow
+            calls[i].slice(0, 10) !== ethers.utils.keccak256(ethers.utils.toUtf8Bytes("requireToPassMessage(address,bytes,uint256)")).slice(0, 10)
```
kronosapiens, line 224: Can this be a constant?

```
ethers.utils.keccak256(ethers.utils.toUtf8Bytes("requireToPassMessage(address,bytes,uint256)"))
```

kronosapiens (dismissed):

kronosapiens:
@area I modified the last commit (it was a small change; I figured it would be more expedient) -- it's approved, take a look and merge if you're happy wit hit.

kronosapiens (approved):

PR #1113, opened 2023-01-19
Slither updates

area:
Fixes issues the newest version of slither raises. This includes reentrancy risks inside loops, variables that could be immutable but weren't, and what I think I agree was an unnecessary external call to `version` in the `upgrade` function. Not convinced I've fixed these in the best possible ways, so open to suggestions.

### DIFF
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -109,8 +109,11 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
       require(uint256(_amounts[i]) <= fundingPots[1].balance[token], "colony-bootstrap-not-enough-tokens");
       fundingPots[1].balance[token] = sub(fundingPots[1].balance[token], uint256(_amounts[i]));
       nonRewardPotsTotal[token] = sub(nonRewardPotsTotal[token], uint256(_amounts[i]));
+    }
 
-      assert(ERC20Extended(token).transfer(_users[i], uint256(_amounts[i])));
+    // After doing all the local storage changes, then do all the external calls
+    for (uint i = 0; i < _users.length; i++) {
+      require(ERC20Extended(token).transfer(_users[i], uint256(_amounts[i])), "colony-bootstrap-token-transfer-failed");
       IColonyNetwork(colonyNetworkAddress).appendReputationUpdateLog(_users[i], _amounts[i], domains[1].skillId);
     }
 
--- a/contracts/colony/Colony.sol
+++ b/contracts/colony/Colony.sol
@@ -29,7 +29,7 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
 
   // This function, exactly as defined, is used in build scripts. Take care when updating.
   // Version number should be upped with every change in Colony or its dependency contracts or libraries.
-  function version() public pure returns (uint256 colonyVersion) { return 11; }
+  function version() public pure returns (uint256 colonyVersion) { return 12; }
 
   function getColonyNetwork() public view returns (address) {
     return colonyNetworkAddress;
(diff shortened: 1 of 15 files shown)


contracts/colony/Colony.sol
```diff
@@ -109,8 +109,11 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
       require(uint256(_amounts[i]) <= fundingPots[1].balance[token], "colony-bootstrap-not-enough-tokens");
       fundingPots[1].balance[token] = sub(fundingPots[1].balance[token], uint256(_amounts[i]));
       nonRewardPotsTotal[token] = sub(nonRewardPotsTotal[token], uint256(_amounts[i]));
+    }
 
-      assert(ERC20Extended(token).transfer(_users[i], uint256(_amounts[i])));
+    // After doing all the local storage changes, then do all the external calls
+    for (uint i = 0; i < _users.length; i++) {
```
kronosapiens, line 115: Is `i` being declared twice here? Stylistically I would use `j` to indicate a second loop but that's just me.

contracts/colony/ColonyAuthority.sol
```diff
@@ -28,6 +28,8 @@ contract ColonyAuthority is CommonAuthority {
   uint8 constant ARCHITECTURE_ROLE = uint8(ColonyDataTypes.ColonyRole.Architecture);
   uint8 constant ROOT_ROLE = uint8(ColonyDataTypes.ColonyRole.Root);
 
+  // colony is used in the constructor by calls to addRoleCapability, despite what slither thinks
```
kronosapiens, line 31: I don't see where...

kronosapiens (commented):
Looks reasonable to me, will take another look on Monday

contracts/colony/Colony.sol
kronosapiens: Oh, can you declare the same identifier twice in different loops? I assumed they were in the same scope, guess not.

area: Yep, they are [different scopes](https://docs.soliditylang.org/en/v0.8.17/control-structures.html#scoping-and-declarations):

> As an exception to this rule, variables declared in the initialization part of a for-loop are only visible until the end of the for-loop.

contracts/colony/ColonyAuthority.sol
area: https://github.com/JoinColony/colonyNetwork/blob/b7201a0b0051ed9a985a0a15e73779b9ced5d780/contracts/colony/ColonyAuthority.sol#L136 - this call in `addRoleCapability` reads the `colony` storage variable.

contracts/colony/Colony.sol
kronosapiens, line 115: Can we make it `uint256` for consistency with the rest of the codebase?

contracts/colony/ColonyRoles.sol
```diff
@@ -104,20 +104,25 @@ contract ColonyRoles is ColonyStorage, ContractRecoveryDataTypes {
...
-          emit RecoveryRoleSet(_user, setTo);
-        }
         emit ColonyRoleSet(msgSender(), _user, _domainId, roleId, setTo);
 
       }
       roles >>= 1;
       rolesChanged >>= 1;
     }
+
+    // Update the storage slot tracking number of recovery roles after all the external calls are complete
```
kronosapiens, line 114: Shouldn't this be before the external calls?

area: Yup, I really messed up there :disappointed:

contracts/colony/ColonyAuthority.sol
area: Prior to [solidity 0.8.8](https://github.com/ethereum/solidity/releases/tag/v0.8.8) immutable variables could not be read in the constructor. One to be picked up in #1053

kronosapiens: Ah, fascinating.

kronosapiens (approved):
Fab

PR #1131, opened 2023-04-10
Introduce StagedPayment

### LINKED ISSUE #1129: Add support for staged payments to StakedExpenditures
Add ability to create a "staged payment", where the owner can "release" payments by setting claim delays to zero (using the arbitration permission)

kronosapiens:
Closes #1129 

Implements Staged Payments:

- Add a `_stagedPayment` boolean to the existing `makeExpenditureWithStake` function
- Add a `releaseStagedPayment` to let creator set slot claim delays to 0 once finalized

Otherwise, staged payments are behaviorally identical to advanced expenditures.

### DIFF
--- a/contracts/extensions/StakedExpenditure.sol
+++ b/contracts/extensions/StakedExpenditure.sol
@@ -29,7 +29,8 @@ contract StakedExpenditure is ColonyExtensionMeta {
 
   // Events
 
-  event ExpenditureMadeViaStake(address indexed creator, uint256 expenditureId, uint256 stake);
+  event ExpenditureMadeViaStake(address indexed creator, uint256 expenditureId, uint256 stake, bool stagedPayment);
+  event StagedPaymentReleased(uint256 indexed expenditureId, uint256 slot);
   event ExpenditureCancelled(uint256 expenditureId);
   event StakeReclaimed(uint256 expenditureId);
   event StakeFractionSet(uint256 stakeFraction);
@@ -38,13 +39,13 @@ contract StakedExpenditure is ColonyExtensionMeta {
 
   struct Stake {
     address creator;
+    bool stagedPayment;
     uint256 amount;
   }
 
   // Storage
 
   uint256 stakeFraction;
-
   mapping (uint256 => Stake) stakes;
 
   // Modifiers
@@ -109,6 +110,7 @@ contract StakedExpenditure is ColonyExtensionMeta {
   /// @param _value Reputation value indicating the total reputation in _domainId
   /// @param _branchMask The branchmask of the proof
   /// @param _siblings The siblings of the proof
+  /// @param _stagedPayment Whether the expenditure is a staged payment
   function makeExpenditureWithStake(
     uint256 _permissionDomainId,
     uint256 _childSkillIndex,
@@ -116,7 +118,8 @@ contract StakedExpenditure is ColonyExtensionMeta {
     bytes memory _key,
     bytes memory _value,
     uint256 _branchMask,
-    bytes32[] memory _siblings
+    bytes32[] memory _siblings,
+    bool _stagedPayment
   )
     public
     notDeprecated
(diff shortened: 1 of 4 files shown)


contracts/extensions/StakedExpenditure.sol
```diff
@@ -129,17 +132,44 @@ contract StakedExpenditure is ColonyExtensionMeta {
...
+    require(
+      colony.getExpenditure(_expenditureId).status == ColonyDataTypes.ExpenditureStatus.Finalized,
+      "staked-expenditure-not-finalized"
+    );
+
+    bool[] memory mask = new bool[](2); mask[0] = false; mask[1] = true;
+    bytes32[] memory keys = new bytes32[](2); keys[0] = bytes32(0); keys[1] = bytes32(uint256(1));
+    colony.setExpenditureState(_permissionDomainId, _childSkillIndex, _expenditureId, 26, mask, keys, bytes32(0));
+
+    emit StagedPaymentReleased(_expenditureId, _slot);
```
area, line 165: Again, one for @arrenv : Is there a situation where you wouldn't want to release and pay? In which case, we could/should call 'claim' here.

contracts/extensions/StakedExpenditure.sol
```diff
@@ -129,17 +132,44 @@ contract StakedExpenditure is ColonyExtensionMeta {
...
   }
 
   /// @notice Reclaims the stake if the expenditure is finalized or cancelled
   /// @param _expenditureId The id of the expenditure
   function reclaimStake(uint256 _expenditureId) public {
     Stake storage stake = stakes[_expenditureId];
-    require(stake.creator != address(0x0), "staked-expenditure-nothing-to-claim");
+    require(stake.amount > 0, "staked-expenditure-nothing-to-claim");
```
area, line 172: More instructive, certainly - was there another reason for this change?

contracts/extensions/StakedExpenditure.sol
```diff
@@ -109,14 +110,16 @@ contract StakedExpenditure is ColonyExtensionMeta {
...
   function makeExpenditureWithStake(
     uint256 _permissionDomainId,
     uint256 _childSkillIndex,
     uint256 _domainId,
     bytes memory _key,
     bytes memory _value,
     uint256 _branchMask,
-    bytes32[] memory _siblings
+    bytes32[] memory _siblings,
+    bool _stagedPayment
```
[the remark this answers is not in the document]
arrenv: To clarify, the idea here is that you can toggle between this being a staged payment or not.

I'm assuming this has been included because it was desired that you could add or remove staged payments while in draft and removing some so that there is only one would result in it not being needed as a staged payment?

Otherwise, no, being able to change an expenditure type from say split to staged or vice versa while it draft does not need to be possible.

contracts/extensions/StakedExpenditure.sol
arrenv: Yes, if a milestone is never delivered, then it should be possible to not have to release and pay that payment.

contracts/extensions/StakedExpenditure.sol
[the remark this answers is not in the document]
area: > Otherwise, no, being able to change an expenditure type from say split to staged or vice versa while it draft does not need to be possible.

Grand, then the current functionality is okay.

contracts/extensions/StakedExpenditure.sol
arrenv: So, should staged payments have claim delays?

At this stage, it is not required. As from what I understand this would also make it more complicated given that staged payments are essentially really high claim delays.

contracts/extensions/StakedExpenditure.sol
```diff
@@ -129,17 +132,44 @@ contract StakedExpenditure is ColonyExtensionMeta {
...
+  {
+    require(stakes[_expenditureId].stagedPayment, "staked-expenditure-not-staged-payment");
+    require(stakes[_expenditureId].creator == msgSender(), "staked-expenditure-not-creator");
+    require(
+      colony.getExpenditure(_expenditureId).status == ColonyDataTypes.ExpenditureStatus.Finalized,
+      "staked-expenditure-not-finalized"
+    );
+
+    bool[] memory mask = new bool[](2); mask[0] = false; mask[1] = true;
+    bytes32[] memory keys = new bytes32[](2); keys[0] = bytes32(0); keys[1] = bytes32(uint256(1));
```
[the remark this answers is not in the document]
kronosapiens: They are functionally linked, essentially a constant operation, and I don't think it should take three lines to initialise an array.

contracts/extensions/StakedExpenditure.sol
```diff
@@ -38,13 +39,13 @@ contract StakedExpenditure is ColonyExtensionMeta {
 
   struct Stake {
     address creator;
+    bool stagedPayment;
```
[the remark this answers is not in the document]
kronosapiens: I thought of you

contracts/extensions/StakedExpenditure.sol
kronosapiens: I think what Alex is saying is that we should claim the payment (e.g. send funds to the user) in the same transaction in which we release the claim delay. Otherwise we would force the user to come back to claim the payment, which seems unecessary.

contracts/extensions/StakedExpenditure.sol
kronosapiens: It caught my eye and I realized that the previous approach was suboptimal. I've clearly matured in the months since that was originally written.

contracts/extensions/StakedExpenditure.sol
kronosapiens: To do these we would need to pass an array of any tokens being claimed at that stage. I've gone ahead and added that because I assume that will be preferable.

contracts/extensions/StakedExpenditure.sol
```diff
@@ -129,17 +132,50 @@ contract StakedExpenditure is ColonyExtensionMeta {
...
+      colony.getExpenditure(_expenditureId).status == ColonyDataTypes.ExpenditureStatus.Finalized,
+      "staked-expenditure-not-finalized"
+    );
+
+    bool[] memory mask = new bool[](2); mask[0] = false; mask[1] = true;
+    bytes32[] memory keys = new bytes32[](2); keys[0] = bytes32(0); keys[1] = bytes32(uint256(1));
+    colony.setExpenditureState(_permissionDomainId, _childSkillIndex, _expenditureId, 26, mask, keys, bytes32(0));
+
+    for (uint256 i; i < _tokens.length; i++) {
+      colony.claimExpenditurePayout(_expenditureId, _slot, _tokens[i]);
```
area, line 168: Can we get a test demonstrating that it is possible to finalize, and not claim, in this scheme? (This will become important once, for example, you finish the changes to voting and there is an ongoing vote when this function is called)

On reflection, I also wonder if setting to zero is as good as we can do? Could we set it to the delay such that it can only be called `now`, once all pending objections are resolved?

area: I have added such a test. 

I have also grown less interested in the second point, as well, as it's verging on inter-extension awareness/interoperability, so until it's an issue, I'm inclined to leave it as-is.

kronosapiens:
After a call with @area and @arrenv, we want to split the stage payment functionality into its own extension. Any expenditure owner will be able to indicate that an expenditure should be treated as a staged payment (while in draft state) and then release the payments once finalized.

PR #1132, opened 2023-04-17
Multi-chain reputation

area:
## Goal

There is a desire for Colony to be deployed to multiple chains. However, the idea of requiring reputation mining to be deployed to multiple chains at once is unpalatable for a variety of reasons - needing the token bridged to all relevant chains, needing to stake separately on all chains, having to monitor all chains to name a few. 

We therefore aim for a simpler extension of reputation mining; mining remains on a single chain, but where necessary (e.g. adding new log entries to the mining log), we use bridges to send transactions from other chains to the chain where mining is taking place.

## Implementation
The main change to existing functions is the function that adds a reputation update to the log. On the mining chain, it behaves as it currently does, and on any other chain it sends a transaction to the bridge. That transaction, once taken across the bridge (by whatever mechanism the bridge uses) will add the log entry to the log. Similarly, on skill creation, if not on the mining chain, a bridging transaction is sent in addition to all the usual behaviour, which when bridged to the mining chain registers the skill as created, and therefore eligible for reputation to be earned in.

The support for bridging transactions is intended to be agnostic to the actual details of the bridges themselves, as we certainly cannot guarantee that all bridges will follow the interface that our mock bridge does. When adding details of bridges on either the mining chain or another chain, we provide the address of the bridge, the chainId of the chain on the other side of the bridge, and some combination of pairs of bytes representing the three types of bridge transactions that there are: 

* Bridging a reputation update (intended to go from non-mining chain to mining chain)
* Bridging skill creation (intended to go from non-mining chain to mining chain)
* Bridging a new reputation state (intended to go from mining chain to non-mining chain)

When constructing a bridging transaction, the contracts construct the transaction that needs to be called on the other side of the bridge, and then concatenates the 'before' and 'after' bytes arrays that were provided when registering the bridge. Not all of these arrays will be provided for any given bridge registration, as each colonyNetwork instance will only call a bridge in at most two of these three ways, depending on whether it is on the mining chain or not. A functional deployment will require all of these to be set between the two colony network deployments on the two chains.

Where appropriate, functions have been marked with a new `onlyMiningChain` modifier to try and make sure they're not accidentally called. Functions only intended to be called by bridges check that they are coming from a known bridge.

We expect only one bridge to be registered on any non-mining chain - the address of a bridge back to the mining chain; this address is stored in the `miningBridgeAddress` storage slot. This slot is not intended to be used on the mining chain, where many bridges will be registered (i.e. all the bridges to all the non-mining chains).

In principle, I believe this scheme also allows the possibility of 'middlemen' - if a bridge does not exist between the mining chain and the desired foreign chain, then a bridge can be registered for the desired chain but is, actually, a bridge to an intermediary chain. The transactions constructed could be such that they call another bridge on that chain directly, which is a bridge to a desired foreign chain. Obviously, gas fees, and having to make sure such transactions are executed all the way along, but probably lower effort to write a service that does this rather than deploying and maintaining our own bridges for any and all chain combinations.

## Assumptions

Note that inherently, here, we trust bridges to be honest. If a bridge is compromised, and starts sending 'bridged' transactions arbitrarily:

* Arbitrary reputation values will be able to be added/removed to colonies on the chain corresponding to the compromised bridge.
* The skill trees on the two chains will get out of sync

Notably, the issues (should be...) confined to colonies on the chain corresponding to the compromised bridge. In the case that a 'bad' reputation state is sent to the colonyNetwork on a chain with a compromised bridge, the 'right' state can be emitted again and bridged.

## Design Decisions

`skillId`s on non-mining chains start at a value of `chainId << 128`. This makes skill ids unique across chains - a requirement for reputation mining. I believe this also means there is no transitional period required - once deployed, and bridged skills and transactions start appearing, they should just be incorporated in to the log. I am a little wary of the fact that `chainId` is [by definition](https://eips.ethereum.org/EIPS/eip-1344) a `uint256` whereas we're essentially truncating it in to a `uint128` here. I am therefore very tempted to hash the `chainId` and use (part of) the resulting hash as the first 'half' of the `skillId`. This is better than hashing both the skill count and the `chainId` together, as `skillId`s remain monotonically increasing per-chain allowing checks for whether a skill exists and not requiring a storage slot to be set per-skill even if it's not otherwise necessary.

Note that everything to do with global skills is limited to the mining chain only, and is ultimately planned to be deprecated there too - we do not need universal global skills, and so there is no requirement to bridge those that already exist from the mining chain, nor a requirement for their creation to be supported on other chains.

The intention is for the system to be relatively resilient to bridges behaving in unexpected ways. For example, a bridge might not bridge a transaction, or might bridge transactions out of order. Skills must be added in order (on a per-chain basis), which could cause issues. As a result, if a skill is attempted to be added out-of-order on the mining chain, it's stored and can be added later. Once all previous skills have been added, either via re-bridging or via this 'pending' mechanism, the skill can be added as normal with an appropriate transaction on the mining chain.

The current reputation state can be bridged to any chain at any time on demand. It is not automatically bridged on an update for gas reasons; if such behaviour is desired, it should be implemented via a standalone service.

Currently, if an entry in the reputation log is bridged for a skill that does not currently exist, it is dropped. On reflection, I think a better behaviour would be a similar scheme for skill addition, where such an emission would be stored and could be added to the log once the skill has been created, with the value decayed by an appropriate amount based on the time between the original emission and the addition to the log.

In general, all these failure cases are open to debate and should arguably go up-stream to a conversation involving some or all of Jack / Aron / Daniel / Arren


kronosapiens:
>> This is better than hashing both the skill count and the `chainId` together, as `chainId`s remain monotonically increasing per-chain allowing checks for whether a skill exists and not requiring a storage slot to be set per-skill even if it's not otherwise necessary.

Assuming you mean `skillId`s are monotonically increasing?

area:
I do, updated.

package.json
```diff
@@ -122,9 +123,9 @@
     "rimraf": "^3.0.2",
     "shortid": "^2.2.16",
     "solhint": "^3.3.7",
-    "solidity-coverage": "^0.8.0-beta.1",
+    "solidity-coverage": "0.8.0-beta.1",
     "solidity-steamroller": "^1.1.0",
-    "truffle": "^5.5.30",
+    "truffle": "5.5.30",
```
kronosapiens, line 128: Why the pins?

parity-genesis.template.json
```diff
@@ -8,7 +8,7 @@
         "accountStartNonce": "0x0",
         "maximumExtraDataSize": "0x20",
         "minGasLimit": "0x1388",
-        "networkID" : "0x7CE",
+        "networkID" : "0xFD5C9EC",
```
kronosapiens, line 11: Can we add a comment indicating which chain this is?

truffle.js
```diff
@@ -99,7 +99,7 @@ module.exports = {
       network_id: "*",
     },
     xdai: {
-      url: "https://xdai-archive.blockscout.com/",
+      url: "https://xdai.colony.io/rpc/",
```
kronosapiens, line 102: Are we running our own node now?

test/reputation-system/root-hash-submissions.js
```diff
@@ -216,6 +221,10 @@ contract("Reputation mining - root hash submissions", (accounts) => {
       const miningSkillId = 3;
 
       await metaColony.setReputationMiningCycleReward(WAD.muln(10));
+
+      // Need tokens to pay out rewards
+      await giveUserCLNYTokens(colonyNetwork, colonyNetwork.address, WAD.muln(100));
```
kronosapiens, line 226: Why is this needed now?

test-smoke/colony-storage-consistent.js
```diff
@@ -149,11 +149,12 @@ contract("Contract Storage", (accounts) => {
       console.log("miningCycleStateHash:", miningCycleStateHash);
       console.log("tokenLockingStateHash:", tokenLockingStateHash);
 
-      expect(colonyNetworkStateHash).to.equal("0x711cd2107e1466eee7e531b4abee44853d8ac1a0708f31004a615f57f3de819c");
-      expect(colonyStateHash).to.equal("0x54a0edcb2097270bd95d610dc827869cc827241d131461f58788f7c3257ca151");
-      expect(metaColonyStateHash).to.equal("0x15fab25907cfb6baedeaf1fdabd68678d37584a1817a08dfe77db60db378a508");
-      expect(miningCycleStateHash).to.equal("0x632d459a2197708bd2dbde87e8275c47dddcdf16d59e3efd21dcef9acb2a7366");
-      expect(tokenLockingStateHash).to.equal("0x30fbcbfbe589329fe20288101faabe1f60a4610ae0c0effb15526c6b390a8e07");
+      // TODO: Did we expect these to change?
```
kronosapiens, line 152: Seems like functions were added to everything except Token Locking... that's the one I would check

test/cross-chain/cross-chain.js
```diff
@@ -39,47 +58,153 @@ contract("Cross-chain", () => {
...
+    const foreignNetworkId = await ethersForeignSigner.provider.send("net_version", []);
+    foreignChainId = await ethersForeignSigner.provider.send("eth_chainId", []);
+
     let etherRouterInfo;
     // 0x539 is the chain id used by truffle by default (regardless of networkid), and if
     // we see it in our tests that's the coverage chain, which builds the contract artifacts
     // in to a different location. If we see another chain id, we assume it's non-coverage
     // truffle and look for the build artifacts in the normal place.
-    if (homeChainId.toString() === "0x539") {
+    if (process.env.SOLIDITY_COVERAGE && process.env.TRUFFLE_FOREIGN === "false") {
```
kronosapiens, line 110: I would do a nested conditional here, a bit more clear IMO

test/cross-chain/cross-chain.js
```diff
@@ -39,47 +58,153 @@ contract("Cross-chain", () => {
...
+        // eslint-disable-next-line max-len
+      )}000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000f42400000000000000000000000000000000000000000000000000000000000000044`,
+      "0x00000000000000000000000000000000000000000000000000000000" // root hash after
+    );
+    await tx.wait();
+
+    // Bridge over skills that have been created on the foreign chain
+
+    const count = await foreignColonyNetwork.getSkillCount();
+    for (let i = ethers.BigNumber.from(foreignChainId).mul(ethers.BigNumber.from(2).pow(128)).add(1); i <= count; i = i.add(1)) {
```
kronosapiens, line 174: I would initialise `i` in a separate line here

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -111,6 +133,10 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
     colonyVersionResolver[_version] = _resolver;
     currentColonyVersion = _version;
 
+    if (!isMiningChain()){
+      skillCount = getChainId() << 128;
```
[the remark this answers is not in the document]
area: Yeah, I ended up doing it this way so that, whenever we're dealing with a `skillId`, we know that `skillId` is unique and doesn't need any further modifications (consider the frontend, too, having to work out whether it needs to change anything about a `skillId` before querying e.g. the reputation endpoint).

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -102,6 +102,28 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
   }
 
+  function setBridgeData(address bridgeAddress, bytes memory updateLogBefore, bytes memory updateLogAfter, uint256 gas, uint256 chainId, bytes memory skillCreationBefore, bytes memory skillCreationAfter, bytes memory setReputationRootHashBefore, bytes memory setReputationRootHashAfter) public
+  always
+  {
+    require(msgSender() == metaColony, "colony-network-not-metacolony");
+    if (!isMiningChain()) {
+      miningBridgeAddress = bridgeAddress;
+    }
+    bridgeData[bridgeAddress] = Bridge(updateLogBefore, updateLogAfter, gas, chainId, skillCreationBefore, skillCreationAfter, setReputationRootHashBefore, setReputationRootHashAfter);
```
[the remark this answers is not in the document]
area: The main reason is to allow lookups for incoming transactions to bridge-only endpoints using `msg.sender`. If we didn't store based on the bridge address, we'd have to include the `chainId` in every function call, to then look up and check it matched the expected bridge address.

With this scheme, there's also the potential for multiple bridges on the mining-chain side can point at the same destination chain. I'm... not sure what the benefits of that are though :sweat_smile:

contracts/testHelpers/BridgeMock.sol
```diff
@@ -28,15 +28,26 @@ contract BridgeMock {
...
-    assembly {
-              // call contract at address a with input mem[in…(in+insize))
-              //   providing g gas and v wei and output area mem[out…(out+outsize))
-              //   returning 0 on error (eg. out of gas) and 1 on success
-
-              // call(g,     a,  v,     in,              insize,      out, outsize)
-      success := call(_gasLimit, _target, 0, add(_data, 0x20), mload(_data), 0, 0)
-    }
+
+      (bool success, bytes memory returndata) = address(_target).call{gas:_gasLimit}(_data);
```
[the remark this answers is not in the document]
area: It's not new solidity functionality, I just wanted access to the `returndata` in the event of an error. I don't think changing how a test helper / mock works is worthy of a separate PR?

parity-genesis.template.json
area: I don't believe so, as it's a JSON file.

truffle.js
area: We have been for quite some time. `xdai-archive.blockscout.com` is also now dead.

test/cross-chain/cross-chain.js
area: I don't think I understand the suggestion here. The best I think I can do with a nested conditional is:

```
    if (process.env.SOLIDITY_COVERAGE) {
      if (process.env.TRUFFLE_FOREIGN === "false"){
        etherRouterInfo = JSON.parse(fs.readFileSync("./build-coverage/contracts/EtherRouter.json"));
      } else {
        etherRouterInfo = JSON.parse(fs.readFileSync("./build/contracts/EtherRouter.json"));
      }
    } else {
      etherRouterInfo = JSON.parse(fs.readFileSync("./build/contracts/EtherRouter.json"));
    }
```

Which seems a lot worse to me?

package.json
area: Because unfortunately it's only a very special version of truffle / solidity-coverage that work.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -167,12 +220,57 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
-    emit SkillAdded(skillCount, _parentSkillId);
-    return skillCount;
+  function addSkillFromBridge(uint256 _parentSkillId, uint256 _skillId) public always onlyMiningChain() {
+    // Require is a known bridge
+    Bridge storage bridge = bridgeData[msgSender()];
+    require(bridge.chainId != 0, "colony-network-not-known-bridge");
+
+    // Check skill count - if not next, then store for later.
+    if (networkSkillCounts[bridge.chainId] + 1 == _skillId){
+      if (_parentSkillId > bridge.chainId << 128){
```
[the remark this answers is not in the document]
area: I _think_ this is a remnant of when I was trying to deal with global skills and can be removed...

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -208,7 +306,33 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
   calledByColony
   returns (uint256)
   {
-    return skillCount++;
+    skillCount += 1;
+    if (!isMiningChain()){
+      bridgeSkill(skillCount);
```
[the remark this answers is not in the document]
area: I'd like to keep the semantics - changing the function name to `bridgeSkillIfNotMiningChain` or similar?

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -167,12 +220,57 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+    Bridge storage bridge = bridgeData[msgSender()];
+    require(bridge.chainId != 0, "colony-network-not-known-bridge");
+
+    // Check skill count - if not next, then store for later.
+    if (networkSkillCounts[bridge.chainId] + 1 == _skillId){
+      if (_parentSkillId > bridge.chainId << 128){
+        addSkillToChainTree(_parentSkillId, _skillId);
+      }
+      networkSkillCounts[bridge.chainId] += 1;
+      emit SkillAdded(_skillId, _parentSkillId);
```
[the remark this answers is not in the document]
area: Another hangover from trying to incorporate global skills, which wouldn't have gone through the `addSkillToChainTree` path. But now they're not here, again, I see no reason why this can't be done.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -208,7 +306,33 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
   calledByColony
   returns (uint256)
   {
-    return skillCount++;
+    skillCount += 1;
```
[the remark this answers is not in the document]
area: I don't believe so. This is only called through `initialiseColony`, which is only called on colony creation through ColonyNetwork's `deployColony` function, which is the functional core of the many `createColony` overloads and deprecated variants. But these all require a colony version to be deployed, which has to be done via `initialise` on `colonyNetwork` (note that a metacolony can't be deployed to use `addColonyVersion` until `initialise` is called either). Given `initialise` on `colonyNetwork` sets the skill count, the check shouldn't be necessary. Unless a) I've gotten any of the above wrong (quite possible) or b) you're just coming at this from a safety perspective?

test/cross-chain/cross-chain.js
```diff
@@ -148,4 +373,597 @@ contract("Cross-chain", () => {
...
+
+      // const t = homeColonyNetwork.interface.encodeFunctionData("bridgeSetReputationRootHash", [
+      //   "0xb8b89e7cf61d1d39d09e98c0ccbb489561e5e1173445a6b34e469f362ebdb221",
+      //   "0xb8b89e7cf61d1d39d09e98c0ccbb489561e5e1173445a6b34e469f362ebdb221"
+      // ]);
+      // console.log(t);
+      // const txDataToBeSentToAMB = homeBridge.interface.encodeFunctionData("requireToPassMessage", [foreignColonyNetwork.address, t, 1000000]);
+      // console.log(txDataToBeSentToAMB);
+
+      // process.exit(1)
```
kronosapiens, line 601: Delete all this

kronosapiens (commented):
Second review

test/cross-chain/cross-chain.js
area: Bridging tests are run twice with coverage:

* Once with coverage tracking the chain where reputation mining is occuring (corresponding to `TRUFFLE_FOREIGN=false`)
* Once with coverage tracking the chain where reputation mining is not occurring (corresponding to `TRUFFLE_FOREIGN=true`)

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -102,6 +102,30 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
     emit ColonyVersionAdded(_version, _resolver);
   }
 
+  function setBridgeData(address bridgeAddress, bytes memory updateLogBefore, bytes memory updateLogAfter, uint256 gas, uint256 chainId, bytes memory skillCreationBefore, bytes memory skillCreationAfter, bytes memory setReputationRootHashBefore, bytes memory setReputationRootHashAfter) public
+  always
+  calledByMetaColony
+  {
+    if (!isMiningChain()) {
+      miningBridgeAddress = bridgeAddress;
+      require(isMiningChainId(chainId), "colony-network-can-only-set-mining-chain-bridge");
```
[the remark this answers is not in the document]
area: I mean, they're similar, in that they both care about what the mining chain's `chainId` is, but one is concerned with the chain we're currently on, and one is concerned about the bridge data that is being set. Can you give me a code snippet to show me what you mean here, because I don't think I follow?

test/cross-chain/cross-chain.js
```diff
@@ -88,13 +233,99 @@ contract("Cross-chain", () => {
...
   });
 
+  describe("administrating cross-network bridges", async () => {
+    it("bridge data can be queried", async () => {
+      const bridgeData = await homeColonyNetwork.getBridgeData(homeBridge.address);
+
+      expect(bridgeData.gas.toNumber()).to.equal(1000000);
+      expect(ethers.BigNumber.from(bridgeData.chainId).toHexString()).to.equal(ethers.BigNumber.from(foreignChainId).toHexString());
+      expect(bridgeData.setReputationRootHashBefore.toLowerCase()).to.equal(
+        `0xdc8601b3000000000000000000000000${foreignColonyNetwork.address.slice(
```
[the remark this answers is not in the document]
area: I've repurposed the code blocks that originally generated them (that you flagged to be deleted) to generate them 'on the fly' at the start of the tests, primarily as documentation as to how these are generated.

contracts/colony/Colony.sol
```diff
@@ -315,14 +323,8 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
     ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
     bytes4 sig;
 
-    sig = bytes4(keccak256("makeArbitraryTransactions(address[],bytes[],bool)"));
-    colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
-
-    sig = bytes4(keccak256("setDefaultGlobalClaimDelay(uint256)"));
+    sig = bytes4(keccak256("setBridgeData(address,bytes,bytes,uint256,uint256,bytes,bytes,bytes,bytes)"));
```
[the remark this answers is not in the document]
area: Yeah, I agree with this

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -208,7 +309,105 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+  }
+
+  function addBridgedReputationUpdate(uint256 _chainId, address _colony) public stoppable onlyMiningChain {
+    uint256 nextUpdateNumber = reputationUpdateCount[_chainId][_colony] + 1;
+    // Bridged update must exist
+    require(pendingReputationUpdates[_chainId][_colony][nextUpdateNumber].colony != address(0x00), "colony-network-next-update-does-not-exist");
+    // It should be the next one
+    assert(pendingReputationUpdates[_chainId][_colony][nextUpdateNumber - 1].colony == address(0x00));
+
+    PendingReputationUpdate storage pendingUpdate = pendingReputationUpdates[_chainId][_colony][nextUpdateNumber];
```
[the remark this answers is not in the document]
area: I found this improvement very pleasing.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -208,7 +309,105 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+      bridgeData[miningBridgeAddress].updateLogAfter
+    );
+
+    delete pendingReputationUpdates[getChainId()][_colony][_updateNumber];
+
+    (bool success, ) = miningBridgeAddress.call(payload);
+    require(success, "colony-network-bridging-tx-unsuccessful");
+  }
+
+  function addBridgedReputationUpdate(uint256 _chainId, address _colony) public stoppable onlyMiningChain {
```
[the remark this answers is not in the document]
area: The reorganisation I agree with, the issue I see with a separate contract is that the non-bridging functions need to be able to call the "bridging" equivalents, and in some cases vice-versa, and so if they're in a separate contract that requires an external call (and more gas).

test/reputation-system/root-hash-submissions.js
area: The chainId we're testing on is now not one where we can mint tokens on payouts.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -102,6 +102,28 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
     emit ColonyVersionAdded(_version, _resolver);
   }
 
+  function setBridgeData(address bridgeAddress, bytes memory updateLogBefore, bytes memory updateLogAfter, uint256 gas, uint256 chainId, bytes memory skillCreationBefore, bytes memory skillCreationAfter, bytes memory setReputationRootHashBefore, bytes memory setReputationRootHashAfter) public
+  always
+  {
+    require(msgSender() == metaColony, "colony-network-not-metacolony");
+    if (!isMiningChain()) {
+      miningBridgeAddress = bridgeAddress;
```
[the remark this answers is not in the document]
area: See [this comment](https://github.com/JoinColony/colonyNetwork/pull/1132#discussion_r1169304700).

contracts/colonyNetwork/ColonyNetwork.sol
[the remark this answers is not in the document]
area: `initialiseColony` is in the Colony contract, and this function is in the `ColonyNetwork` contract?

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -208,7 +309,105 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+    uint256 nextUpdateNumber = reputationUpdateCount[_chainId][_colony] + 1;
+    // Bridged update must exist
+    require(pendingReputationUpdates[_chainId][_colony][nextUpdateNumber].colony != address(0x00), "colony-network-next-update-does-not-exist");
+    // It should be the next one
+    assert(pendingReputationUpdates[_chainId][_colony][nextUpdateNumber - 1].colony == address(0x00));
+
+    PendingReputationUpdate storage pendingUpdate = pendingReputationUpdates[_chainId][_colony][nextUpdateNumber];
+
+    // Skill creation must have been bridged
+    require(networkSkillCounts[pendingUpdate.skillId >> 128] >= pendingUpdate.skillId, "colony-network-invalid-skill-id");
```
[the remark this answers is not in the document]
area: Would they? Don't you still have to know whether to call the helper functions or not?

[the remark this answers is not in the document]
kronosapiens: Adding helpers would clarify the semantics of the operations, which would help communicate the intention

contracts/colonyNetwork/ColonyNetwork.sol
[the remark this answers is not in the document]
kronosapiens: I was able to pull out the skills-related functionality into `ColonyNetworkSkills` without introducing any external calls.

kronosapiens:
@area I've addressed all my own comments. There are a few failing tests but they were failing on your last commit so I'm assuming you haven't gotten to them yet. Once you've gotten everything to your liking and taken the PR off draft I'll give it one more review and then I think we should be ready to merge.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -453,36 +435,33 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+        ),
         bridgeData[miningBridgeAddress].updateLogAfter
       );
+
       (bool success, ) = miningBridgeAddress.call(payload);
       if (!success || !isContract(miningBridgeAddress)) {
         // Store to resend later
-        pendingReputationUpdates[getChainId()][msgSender()][reputationUpdateCount[getChainId()][msgSender()]] = PendingReputationUpdate(_user, _amount, _skillId, msgSender(), block.timestamp);
+        PendingReputationUpdate memory pendingReputationUpdate = PendingReputationUpdate(_user, _amount, _skillId, msgSender(), block.timestamp);
+        pendingReputationUpdates[getChainId()][msgSender()][reputationUpdateCount[getChainId()][msgSender()]] = pendingReputationUpdate;
```
area: Interested in the gas cost here. The intention was to directly allocate to storage, but if under-the-hood that uses memory as well, then I agree this is better to read.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -545,12 +524,24 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
     }
   }
 
-  function isContract(address addr) internal returns (bool) {
-    uint256 size;
-    assembly { size := extcodesize(addr) }
-    return size > 0;
+  function toRootSkillId(uint256 _chainId) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
area: `<adjective>skillIdTo<adjective>skillId`, or something?

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -545,12 +524,24 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+    return _chainId << 128;
+  }
+
+  function toChainId(uint256 _skillId) internal pure returns (uint256) {
```
area: Similarly, `<adjective>skillIdToChainId`?

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -148,16 +148,16 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
-    // Initialise the domain tree with defaults by just incrementing the skillCount
-    skillCount += 1;
-    IColonyNetwork(address(this)).bridgeSkillIfNotMiningChain(skillCount);
     colonyCount += 1;
     colonies[colonyCount] = address(colony);
     _isColony[address(colony)] = true;
 
+    // Initialise the domain tree with defaults by just incrementing the skillCount
+    skillCount += 1;
+    IColonyNetwork(address(this)).bridgeSkillIfNotMiningChain(skillCount);
```
area: Hm, is this fixing something?

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -0,0 +1,460 @@
+/*
```
area, line 1: The only thing giving me pause about this is the hectic rebase with #1140 I'm worried it's going to cause. Could we re-order (but stay in the same file), and then do this in a separate PR?

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -277,7 +263,28 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
...
   }
 
-  function addBridgedReputationUpdate(uint256 _chainId, address _colony) public stoppable onlyMiningChain {
+  function addPendingSkill(address _bridgeAddress, uint256 _skillId)
+    public
+    always
+    onlyMiningChain
+  {
+    uint256 bridgeChainId = bridgeData[_bridgeAddress].chainId;
+    require(bridgeChainId != 0, "colony-network-not-known-bridge");
```
area, line 272: Should we check skillId/bridgeChainId consistency here as well?

test/cross-chain/cross-chain.js
area: I don't think you've understood what's happening under coverage, here:

When these tests are run under coverage, there are two chains running. One is being run under coverage, and one is being run not under coverage. In the case of the former, the deployment artifacts are found in `./build-coverage/...`, and in the case of the latter the deployment artifacts are found in `./build/...`. That means under coverage, _both_ need to be loaded, but the question is which artifact corresponds to the mining-chain and which to the non-mining-chain, which `TRUFFLE_FOREIGN` captures.

I'll also note there's the case where neither run under coverage, which also needs to be accounted for in any solution here.

contracts/colonyNetwork/ColonyNetwork.sol
kronosapiens: I am not sure the difference, I was aiming for readability.

contracts/colonyNetwork/ColonyNetwork.sol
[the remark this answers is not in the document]
kronosapiens: I think it's fine, I think it's easy enough to understand, especially when seeing the argument name.

contracts/colonyNetwork/ColonyNetworkSkills.sol
kronosapiens: I hear that... I don't know if there's a way around a tricky rebase, since even if we keep the re-order in one file the line numbers will still be off. I'm happy to go through and do it, though.

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -107,7 +126,7 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
     emit BridgeDataSet(_bridgeAddress);
   }
 
-  function bridgeSkillIfNotMiningChain(uint256 _skillId) public stoppable skillExists(_skillId) {
+  function bridgeSkill(uint256 _skillId) public stoppable skillExists(_skillId) {
```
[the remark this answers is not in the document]
kronosapiens: Fair enough. If I had to choose I would go back to putting the logic in the calling function, but I really think it's fine to leave as-is. People reading the function can see that the exception is made for the mining chain.

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -227,37 +234,16 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
...
     address _colony,
     address _user,
     int256 _amount,
     uint256 _skillId,
     uint256 _updateNumber
   )
     public
-    onlyMiningChain
     stoppable
+    onlyMiningChain
```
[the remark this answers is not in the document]
kronosapiens: I wanted to make the modifier ordering consistent, in general we put the `stoppable`, etc modifiers before more specific ones

contracts/colonyNetwork/ColonyNetworkSkills.sol
kronosapiens: Sure! Potentially worth creating a modifier which does all the checks, since we do them elsewhere.

contracts/colonyNetwork/ColonyNetworkSkills.sol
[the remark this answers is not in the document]
area: I'm just talking about the name of the function :smile:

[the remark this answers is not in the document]
kronosapiens: I know. I'd prefer `bridgeSkill` over `bridgeSkillIfNotMiningChain`. If that means wrapping the call in

```
if (!miningChain()) { bridgeSkill(skilId) };
```

Then that's fine. I would say in that case add a `require(!miningChain(), "not-mining-chain")` in the function body.

contracts/colonyNetwork/ColonyNetwork.sol
area: 40 gas more expensive to do it via memory.

kronosapiens: Ok, I'd vote to keep the more readable implementation then.

contracts/colonyNetwork/ColonyNetwork.sol
kronosapiens: I think it's fine as-is but if you feel strongly feel free to change it

contracts/colonyNetwork/ColonyNetworkSkills.sol
area: Yeah, you're right there in that the big line number changes probably still makes it bad. I guess we just cross that bridge when we come to it.

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -146,7 +151,9 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
     // This succeeds if not set, but we don't want to block e.g. domain creation if that's the situation we're in,
     // and we can re-call this function to bridge later if necessary.
     (bool success, ) = miningBridgeAddress.call(payload);
-    require(success, "colony-network-unable-to-bridge-skill-creation");
+    if (!success) {
+      emit SkillCreationStored(_skillId);
```
kronosapiens, line 155: Is anything actually stored here?

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -210,27 +219,35 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
 
     (bool success, ) = miningBridgeAddress.call(payload);
     require(success, "colony-network-bridging-tx-unsuccessful");
+
+    emit ReputationUpdateSentToBridge(_colony, _updateNumber);
   }
 
   // Bridging (receiving)
 
+  modifier skillAndBridgeConsistent(address bridgeAddress, uint256 _skillId) {
```
kronosapiens, line 228: I might've called this `validBridgeTx`

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -210,27 +219,35 @@ contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall {
...
 
     // Check skill count - if not next, then store for later.
     if (networkSkillCounts[bridgeChainId] + 1 == _skillId){
       addSkillToChainTree(_parentSkillId, _skillId);
       networkSkillCounts[bridgeChainId] += 1;
+      emit SkillAddedFromBridge(_skillId);
     } else if (networkSkillCounts[bridgeChainId] < _skillId){
       pendingSkillAdditions[bridgeChainId][_skillId] = _parentSkillId;
-      // TODO: Event?
+      emit SkillStoredFromBridge(_skillId);
```
kronosapiens, line 250: Stylistically we put newlines before events

contracts/colonyNetwork/ColonyNetworkStorage.sol
```diff
@@ -106,6 +107,26 @@ contract ColonyNetworkStorage is ColonyNetworkDataTypes, DSMath, CommonStorage {
...
+  mapping(address => Bridge) bridgeData; // Storage slot 44
+
+  // A mapping that maps network id -> skill count
+  mapping(uint256 => uint256) networkSkillCounts; // Storage slot 45
+  // A mapping that stores pending bridged skill additions that have been bridged out-of-order
+  // networkId -> skillCount -> parentSkillId
+  mapping(uint256 => mapping(uint256 => uint256)) pendingSkillAdditions; // Storage slot 46
+
+  // A mapping that stores the latest reputation update received from a colony on a particular chain
+  // networkID -> colonyAddress -> updateCount
```
kronosapiens, line 120: `networkId`

contracts/colonyNetwork/ColonyNetworkStorage.sol
```diff
@@ -127,15 +148,23 @@ contract ColonyNetworkStorage is ColonyNetworkDataTypes, DSMath, CommonStorage {
...
     } else {
       require(_isColony[msgSender()] || msgSender() == address(this), "colony-caller-must-be-colony");
     }
     _;
   }
 
-  modifier skillExists(uint skillId) {
+  modifier skillExists(uint256 skillId) {
     require(skillCount >= skillId, "colony-invalid-skill-id");
+    require(isMiningChain() || (skillId >> 128) == getChainId() , "colony-invalid-skill-id");
```
kronosapiens, line 161: Could use `toChainId` here, which would entail moving the function down into the "internal" section

migrations/8_setup_meta_colony.js
```diff
@@ -43,16 +45,23 @@ module.exports = async function (deployer, network, accounts) {
...
-  const tokenLocking = await ITokenLocking.at(tokenLockingAddress);
-  await tokenLocking.methods["deposit(address,uint256,bool)"](clnyToken.address, DEFAULT_STAKE, true, { from: MAIN_ACCOUNT });
-  await colonyNetwork.stakeForMining(DEFAULT_STAKE, { from: MAIN_ACCOUNT });
-  await metaColony.addGlobalSkill();
+  // Check chain id
+  // If not a mining chain, then skip setting up mining
+  const multichain = await MultiChain.new();
+  const chainId = await multichain.getChainId();
+
+  if (chainId.toString() === "265669100" || chainId.toString() === "100") {
```
kronosapiens, line 53: Can we reference constants here?

test/cross-chain/cross-chain.js
```diff
@@ -88,13 +242,98 @@ contract("Cross-chain", () => {
...
+          2
+          // eslint-disable-next-line max-len
+        )}000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000f42400000000000000000000000000000000000000000000000000000000000000044`,
+        "0x00000000000000000000000000000000000000000000000000000000" // root hash after
+      );
+      await tx.wait();
+
+      const countAfter = await homeColonyNetwork.getBridgedSkillCounts(foreignChainId);
+      expect(countAfter).to.not.equal(0);
```
kronosapiens, line 332: Could do `be.zero` here

test/cross-chain/cross-chain.js
```diff
@@ -148,4 +380,585 @@ contract("Cross-chain", () => {
...
+      expect(homeColonyRootLocalSkill.nChildren.toString()).to.equal(foreignColonyRootLocalSkill.nChildren.toString());
+
+      let zeroSkill = await foreignColonyNetwork.getSkill(ethers.BigNumber.from(foreignChainId).mul(ethers.BigNumber.from(2).pow(128)));
+      expect(zeroSkill.nChildren.toNumber()).to.equal(0);
+
+      zeroSkill = await homeColonyNetwork.getSkill(ethers.BigNumber.from(foreignChainId).mul(ethers.BigNumber.from(2).pow(128)));
+      expect(zeroSkill.nChildren.toNumber()).to.equal(0);
+
+      zeroSkill = await homeColonyNetwork.getSkill(0);
+      expect(zeroSkill.nChildren.toNumber()).to.equal(0);
```
kronosapiens, line 594: Is this a meaningful test?

kronosapiens (commented):
Can we take this off draft now?

contracts/colonyNetwork/ColonyNetworkSkills.sol
area: The name of the event was to be consistent with other events related to bridging where things are explicitly stored in the event of failure.

Reattempts are handled by calling this function again, as the comments indicate.

contracts/colonyNetwork/ColonyNetworkSkills.sol
area: I thought along those lines too originally, but in one case, it's not a function that's called from a bridge, it's a function that's called when something's been bridged out of order.

test/cross-chain/cross-chain.js
area: Yes, because until [these changes](https://github.com/JoinColony/colonyNetwork/commit/8e117b1157badc11e5de62a608adea0281a238c7) it wasn't true.

test/cross-chain/cross-chain.js
area: Ethers BNs don't play nicely with that construction.

kronosapiens (dismissed):
Great work! Looks good to me

kronosapiens (dismissed):

kronosapiens (approved):

kronosapiens (approved):

area:
This is now based on top of #1156 so that needs to be merged before this.

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -0,0 +1,495 @@
...
+        _updateNumber
+      ),
+      bridgeData[miningBridgeAddress].updateLogAfter
+    );
+
+    delete pendingReputationUpdates[getChainId()][_colony][_updateNumber];
+
+    // slither-disable-next-line unchecked-lowlevel
+    (bool success, ) = miningBridgeAddress.call(payload);
+    if (!success || !isContract(miningBridgeAddress)) {
```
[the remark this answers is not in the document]
area: If it's not a contract, a .call will (I believe) silently succeed, which we don't want here (it would result in the reputation update being lost).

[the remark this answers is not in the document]
kronosapiens: I'd say this is a behavior which calls for a comment

area:
The reworked version of this ended up in #1216 which was included with the merge of #1237.

PR #1140, opened 2023-05-26
Reputation improvements

### LINKED ISSUE #1124: Improving reputation
After a year and change of usage, it seems like reputation as currently envisioned is too restrictive. A number of suggestions have been made about how to make reputation more flexible. After a call with Jack, Raul, and Arren, here is the proposal to move forward:

- [ ] Introduce the ability to set a "reputation rate" for any token, not just the native token, e.g. payments in DAI could give reputation at a rate of 2:1, while payments in CLNY would give reputation at a rate of 1:1. The default setting would be native token at 1:1, and all other tokens at 1:0.

- [ ] Introduce a per-domain scalar which scales the reputation earned in that domain. If a domain scalar is not set, the domain inherits the scalar from the parent.

- [ ] In extraordinary circumstances, a colony should be able to edit the reputation earned by a payment to an arbitrary amount. It may be possible to implement this using the `payoutScalar`, or may require creating an additional motion.

area:
Closes #1124

# Per-token reputation rate

In order to maintain some of the behaviour around payments (i.e. reputation being emitted on finalization, not claim), after a discussion with Arren, we came to the conclusion that we would limit the number of tokens that, if paid out from a colony, would result in reputation being awarded to the recipient. We landed on ten, which seems like plenty. In principle, this could be increased in the future, but only with the spectre of gas limits hanging over our heads. 

The relevant tokens are stored in a linked list, and tokens must be inserted in order of ascending address in order to ensure there are not duplicate entries - indeed, when any editing of the list is taking place, the previous token must be provided. 

The root permission is required to do this, for fairly obvious reasons. 

An event is emitted when a token rate is set.

# Per-domain reputation rate

Sadly, the same spectre of gas limits hangs over us with respect to per-domain reputation rate, given how it is intended to work (after a discussion with Arren, as this was not clear in the original issue). With multiple domains in a hierarchy, (e.g. A -> B -> C -> D), then the per-domain rate for D is equal to a factor set for D, times the factor (if any) set for C, and so on up the tree. It seems as if that the creation limit (because we note children / parents on skill creation) dwarfs this factor, but something to keep in mind.

Note that strictly, I've implemented this as a _per skill_ reputation scaling, but only provide a mechanism for the skills of domains to be set. This gives us some latitude for the future to add per-skill scaling, but ultimately was driven by the requirement to walk up the skill tree, where it isn't possible to walk up (on-chain) the domain tree.

Again, the root permission is required to set the per-domain reputation rate. This is because of a scenario where the scale factor in a domain is set to 0, any non-root permission being allowed to set it to non-zero allows unexpected earning of reputation in root.

For both per-domain scaling, and per-token scaling, the scale factor is applied before being emitted to the reputation update log. No changes will be required once #1132 is merged as a result, and the reputation miners also need do nothing to accommodate these changes.

An event is emitted when a domain reputation rate is set.

# Updates to payout scalar

Encompassing two bullet points in the requirements, permissions around `payoutScalar` have been changed - ultimately, those who were able to adjust it before can now adjust it between -1 and 0 (previously -1 and 1). This removes the ability to give bonus reputation using the `payoutScalar`. Root permission is now required to set the scalar above 0 (and can be set to an arbitrarily high value), which can be used in extraordinary circumstances to edit the reputation earned by a payment.

# Per-colony decay rate

Fairly self-explanatory. I've limited the precision to the same amount of precision we have currently for the default value. Colonies can return to following our default decay rate even if they've set their own previously.

This feature will require a little additional work to sit alongside #1132, in that the function call that sets the decay rate will need to be bridged, but otherwise decay-rate calculations will occur as previously. Updates to the decay rate require one reputation cycle to complete, which is tracked by storing the address of the currently active cycle and not accessing the storage with a simple getter, but via `getColonyReputationDecayRate`.

The changes to the reputation miner code is such that it should be able to be deployed before the contract changes to allow a seamless transition.

An event is emitted when a colony's decay rate is set.

# Other changes

I've made a few other changes on my way through this work:

* `require(false, "...")`  replaced with `revert(...)`
* Change the solhint script, as IColony.sol is now too large to be piped in to solhint.
* I've changed the `bn2bytes32` function to work as I expected, at least, with negative values. 

# Notes 

I'm not 100% that I've covered all overflow guards for the scaling, so a special eye there would be appreciated.


kronosapiens (commented):
High level comments:

- Colony-specific decay rates I'm broadly happy with
- Ditto changes to payout scalar
- I think that the introduction of a linked list and capping the number of tokens earning reputation simply to support the feature of "give rep on finalization" represents an injudicious complexity/feature tradeoff
- I think that the introduction of domain scaling before the introduction of nested domains is over-engineering, and that feature should wait until we have actual colonies using nested domains

contracts/colony/ColonyStorage.sol
```diff
@@ -358,4 +362,32 @@ contract ColonyStorage is ColonyDataTypes, ColonyNetworkDataTypes, DSMath, Commo
...
   }
+
+  function getTokenScaledReputation(int256 _amount, address _token) internal view returns (int256) {
+    uint256 scaleFactor = tokenReputationRates[token]; // NB This is a WAD
+    if (scaleFactor == 0) { return 0; }
+
+    // Check if too large for scaling
+    int256 amount;
+    int256 absAmount;
+    if (_amount == type(int256).min){
```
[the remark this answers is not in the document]
area: They are not the same number :)

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -269,6 +297,83 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+        if (s.reputationScalingFactor == 0){
+          // If scaling is in effect and is 0, we can short circuit - regardless of the rest of the tree
+          // the scaling factor will be 0
+          return 0;
+        } else {
+          factor = wmul(factor, s.reputationScalingFactor);
+        }
+      }
+    }
+    return factor;
```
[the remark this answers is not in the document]
area: I mean, we _do_ have nested domains, it's just that they're limited to one level. It is a feature that has been requested by users, hence the desire to implement it.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -269,6 +297,83 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+  }
+
+  function setColonyReputationDecayRate(uint256 _numerator, uint256 _denominator) public calledByColony stoppable {
+    require(_numerator < 10**15, "colony-network-decay-numerator-too-big");
```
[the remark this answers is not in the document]
area: To allow the same level of precision as we currently have, and not have to worry about extra overflows if someone tried to set to greater precision.

test/contracts-network/colony-permissions.js
```diff
@@ -408,6 +408,134 @@ contract("ColonyPermissions", (accounts) => {
...
+    it("should be able to apply reputation earned scaling to 150 layers of domains", async () => {
+      await removeSubdomainLimit(colonyNetwork);
+      await colony.addDomain(1, UINT256_MAX, 1);
+      let domainCount = await colony.getDomainCount();
+      domainCount = domainCount.toNumber();
+
+      const limit = domainCount + 150;
+
+      // Limit currently appears to be 160
+      for (let i = domainCount - 2; i < limit - 2; i += 1) {
```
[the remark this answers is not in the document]
area: ¯\_(ツ)_/¯

test/contracts-network/colony.js
```diff
@@ -450,4 +458,224 @@ contract("Colony", (accounts) => {
...
+      let res = await colonyNetwork.getColonyReputationDecayRate(colony.address);
+      expect(res.numerator).to.eq.BN(1);
+      expect(res.denominator).to.eq.BN(2);
+
+      // Get default decay rate from mining cycle
+      const activeReputationMiningCycleAddress = await colonyNetwork.getReputationMiningCycle(true);
+      const activeReputationMiningCycle = await IReputationMiningCycle.at(activeReputationMiningCycleAddress);
+      const defaultDecay = await activeReputationMiningCycle.getDecayConstant();
+
+      await colony.setReputationDecayRate(0, 0); // Special call to reset to follow default rate
```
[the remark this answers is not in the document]
area: That would set it to the current decay rate, and not to the default, which is 'follow the network decay rate, whatever that is'.

test/contracts-network/colony.js
```diff
@@ -450,4 +458,224 @@ contract("Colony", (accounts) => {
...
+      const domain = await colony.getDomain(1);
+      let skill = await colonyNetwork.getSkill(domain.skillId);
+      expect(skill.reputationScalingFactor).to.be.eq.BN(WAD.divn(2));
+      expect(skill.earnedReputationScaling).to.be.true;
+
+      await colony.setDomainReputationScaling(1, false, 0);
+
+      skill = await colonyNetwork.getSkill(domain.skillId);
+      expect(skill.reputationScalingFactor).to.be.eq.BN(0);
+      expect(skill.earnedReputationScaling).to.be.false;
```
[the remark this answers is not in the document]
area: It is a requirement that the domain scaling factor can be set to 0.

test/contracts-network/colony.js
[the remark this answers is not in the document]
kronosapiens: Sure, but couldn't the front-end query `getDecayConstant` from the active mining cycle and pass in the value? That's what happens in the contracts when you pass in 0, and would let us remove some logic from the contract.

test/contracts-network/colony.js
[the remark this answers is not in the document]
kronosapiens: Ok, in that case what do you think of storing the scalar as `1 - value`? Essentially saying that a value of 0 means a scalar of 1, a value of 0 means a scalar of 0, and everything in between.

Setting the scalar would look like:

```
setScalar(uint scalar) {
 storedScalar = 1 - scalar;
}

getScalar() {
 return 1 - storedScalar
}
```

kronosapiens:
@jakzilla @arrenv @area @rdig 

Elaborating on my point at the end of the call, the [current implementation](https://github.com/JoinColony/colonyNetwork/pull/1140/files#diff-ed860c78d6b32d0c8d4f8c141a5bb30ea6ba2631e4a3f194048fd9c540f548b7R340) of `setTokenReputationRate` is 29 lines. Of those, 28 lines are used to implement a special data structure, whose only purpose is to allow reputation to be given out when a payment is finalized (as opposed to when a payment is claimed). Further, this data structure increases the amount of storage required to implement the token-specific reputation feature by 3x, as well as capping the number of tokens which can earn reputation in a colony to 10. If we remove this functionality, `setTokenReputationRate` can be reduced to one line, the storage requirement would be reduced by 2/3, and there would be no cap on the number of tokens which earn reputation.

The initial decision to give reputation when a payment is finalized (as opposed to when it is claimed) was (I believe) made in response to the argument that no one would "claim" a reputation penalty. While this is probably true, our claim functions have been made public, meaning that anyone can claim a payment on another's behalf. This means that if a payment would result in a negative penalty, and the recipient has no incentive to claim that penalty, that someone else is able to "claim" the penalty on their behalf.

One could argue that this second "claim" function would result in more transactions, and more headache for users. However, given that we have now added support for multicall, it would be straightforward to include any reputation penalty "claims" as part of a single transaction, meaning that these implementation decisions would be ultimately invisible to the user.

I am proposing that we drop the more-complex implementation of this functionality, given that we can achieve the desired product goal in an alternative way, which would result in less overall complexity across the contracts and dapp. The addition of a multicall as part of finalization would require the dapp to make a few more queries on finalization, but I believe this would still be a net simplification over the current implementation.

arrenv:
To make it clear, this is what we lose from the simplified version:
- There would be no reputation penalty on claiming (although, offset by public claiming function and multicall)

This is what we gain:
- More then 10 tokens that can earn reputation
- Simpler overall code

If this cover the differences, then I would opt for the simplified version as you have described.

test/contracts-network/colony.js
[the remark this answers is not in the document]
area: I don't think you're understanding what I am saying. What you are proposing would set the decay constant to whatever the current value is - permanently. The functionality when you pass in 0 is that the decay rate follows the network decay rate. If the network decay rate changes, so does the colony decay rate.

contracts/colony/ColonyFunding.sol
```diff
@@ -209,25 +209,27 @@ contract ColonyFunding is ColonyStorage { // ignore-swc-123
...
     delete expenditureSlotPayouts[_id][_slot][_token];
 
-    int256 payoutModifier = imin(imax(slot.payoutModifier, MIN_PAYOUT_MODIFIER), MAX_PAYOUT_MODIFIER);
+    int256 payoutModifier = imax(slot.payoutModifier, MIN_PAYOUT_MODIFIER);
     uint256 payoutScalar = uint256(payoutModifier + int256(WAD));
 
     uint256 repPayout = wmul(initialPayout, payoutScalar);
     uint256 tokenPayout = min(initialPayout, repPayout);
-    uint256 tokenSurplus = initialPayout - tokenPayout;
+    // uint256 tokenSurplus = initialPayout - tokenPayout;
```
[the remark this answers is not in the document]
kronosapiens: Maybe pedantic, but I would do this instead:

```
uint256 payoutScalar = uint256(imax(slot.payoutModifier, MIN_PAYOUT_MODIFIER) + int256(WAD));
```

As `payoutModifier` is never used except as intermediate storage, and `tokenSurplus` gets used twice

test/contracts-network/colony.js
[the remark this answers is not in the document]
kronosapiens: Ok, I see the distinction now

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -269,6 +297,83 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
+    require(_factor <= WAD, "colony-network-invalid-reputation-scale-factor");
+    Domain memory d = IColony(msgSender()).getDomain(_domainId);
+    uint256 skillId = d.skillId;
+    skills[skillId].earnedReputationScaling = _enabled;
+    skills[skillId].reputationScalingFactor = _factor;
+  }
+
+  function getSkillReputationScaling(uint256 _skillId) public view returns (uint256) {
+    uint256 factor;
+    Skill storage s = skills[_skillId];
```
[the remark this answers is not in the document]
area: The natural name of `skill` is very close to lots of other names in this function, which I wanted to distinguish it from.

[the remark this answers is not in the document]
kronosapiens: Fair enough, and I've also seen us do this elsewhere

area:
Need to be able to set scale factor for reputation mining.

contracts/common/ScaleReputation.sol
```diff
@@ -0,0 +1,54 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pragma solidity 0.8.20; // ignore-swc-103
+import "../../lib/dappsys/math.sol";
+
+contract ScaleReputation is DSMath {
```
kronosapiens, line 21: Does this really need to be its own contract?

area: Where else would you suggest it goes? In general, this should probably be more common than it currently is in the codebase. Longer term, I want to do a pass-through to make sure there isn't unused code being deployed with contracts due to too-generous inclusion of rarely used code (in terms of number of contracts that inherit the code vs use it).

kronosapiens: I agree that an audit is overdue. I would say though that a separate contract should contain at least 5 related functions. I would be in favor of pulling out other reputation-related functionality into `ColonyNetworkReputatation`

area: It's shared between Colony and ColonyNetwork, though.

test/contracts-network/colony.js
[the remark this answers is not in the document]
area: I've called this the 'complement', but have not used additional functions to set/load it.

contracts/common/ScaleReputation.sol
kronosapiens: Ok fair. Hopefully we can place more functionality here and rename it something more general in the future, `CommonUtils` or something along those lines.

contracts/colony/ColonyDomains.sol
```diff
@@ -124,6 +127,13 @@ contract ColonyDomains is ColonyStorage {
     return domainCount;
   }
 
+  function setDomainReputationScaling(uint256 _domainId, uint256 _factor) public stoppable auth {
```
area, line 130: We have a big problem with this implementation, as recovery mode can be used to set a domain's skill to an arbitrary skill id, allow colonies to set each-other's scaling, which is obviously unacceptable.

contracts/common/ScaleReputation.sol
area: I'm not convinced that's a good solution; we'd then (presumably) be including code that isn't necessary in contracts? (i.e. a contract that only used this function would still have the code for all the others included).

contracts/colony/ColonyExpenditure.sol
```diff
@@ -316,10 +315,11 @@ contract ColonyExpenditure is ColonyStorage {
 
       // Validate payout modifier
       if (offset == 2) {
-        if (!ColonyAuthority(address(authority)).hasUserRole(msgSender(), 1, uint8(ColonyDataTypes.ColonyRole.Root))){
-          require(int256(uint256(_value)) <= 0, "colony-expenditure-bad-payout-modifier");
-        }
-        require(int256(uint256(_value)) >= MIN_PAYOUT_MODIFIER, "colony-expenditure-bad-payout-modifier");
+        require(
+          (int256(uint256(_value)) <= 0 || IColony(address(this)).hasUserRole(msgSender(), 1, ColonyRole.Root)) &&
```
area, line 319: You've changed from querying the Authority to querying the Colony here. Is that deliberate?

area: I'm also not really a fan of this refactoring in general, I think this is a lot harder to follow.

contracts/common/ScaleReputation.sol
```diff
@@ -20,35 +20,26 @@ import "../../lib/dappsys/math.sol";
...
+    internal
+    pure
+    returns (int256 scaledReputation)
+  {
+    if (reputationAmount == 0 || scaleFactor == 0) { return 0; }
+
+    int256 sgnAmount = (reputationAmount >= 0) ? int256(1) : -1;
+    int256 absAmount = (reputationAmount == type(int256).min)
+      ? type(int256).max // Off by one, but best we can do - probably gets capped anyway
+      : (reputationAmount >= 0) ? reputationAmount : -reputationAmount;
```
area: I'm sorry, but _nested_ ternary statements is where I draw the line. Kudos for discovering a bug in that `solidity-coverage` can't lint this correctly, but the fact that bug seemingly has gone unnoticed until now is probably meaningful in of itself.

contracts/common/ScaleReputation.sol
kronosapiens: Hmm. Sounds like you're angling for more of a "mixin" approach, where we can pull a lot of small contracts in as needed. I guess that approach is defensible from a contract size perspective. I might want to see another directory, `common/mixins/` or something like that so they don't end up cluttering the directory structure.

contracts/colony/ColonyDomains.sol
kronosapiens: I don't love storing the scaling factor on the colony, since it involves making external calls as we walk up the skills tree. Hard to defend from a gas perspective. Can't we add the domain skill as a protected slot in recovery mode? We already have infrastructure for that.

contracts/colony/ColonyExpenditure.sol
kronosapiens: It was deliberate, since we already have a `hasUserRole` in the colony interface. It seemed like an abstraction violation to go to the authority here.

In general, I don't like putting `require` statements inside of conditionals, since that can more easily be expressed as a two-condition `require`. If nothing else is in the branch except a `require`, I would say just put the condition inside the `require`.

contracts/common/ScaleReputation.sol
kronosapiens: Alright, I'll yield on this one, despite how incredible my refactor looked.

contracts/colony/ColonyDomains.sol
area: No, because it's not currently a protected slot (so a colony could set it, then upgrade)

contracts/colony/ColonyExpenditure.sol
area: > It was deliberate, since we already have a hasUserRole in the colony interface. It seemed like an abstraction violation to go to the authority here.

You're right, it's just gassier. I guess it depends which 'incorrect' side we want to err on.

> In general, I don't like putting require statements inside of conditionals, since that can more easily be expressed as a two-condition require. If nothing else is in the branch except a require, I would say just put the condition inside the require.

I fear that by 'more easily', you mean 'in fewer characters', which isn't the goal when we're not code-golfing. All other things being equal, we should go for the easiest to understand with all the nested bracket stacked up in the require, I really don't think it is.

I actually don't think either of us are right here, and it should be expressed with a conditional but should be how you've done it in the require (i.e. checking `value` first, and only doing the external call if necessary, which is much cheaper).

contracts/common/ScaleReputation.sol
area: Again, is 'looked' code for 'fewest characters'? :stuck_out_tongue:

kronosapiens: Ultimately I think we're jointly optimizing two variables: easy of interpreting a block of code, and ease of interpreting a function. Having concise bits of code makes it easier to parse a function or contract, even if a single section of code takes a bit longer to parse (arguable). In my mind, conditionals suggest branching behavior, so using them solely to implement require logic makes the function as a whole more difficult to parse. Ultimately, this is simple logic for setting a single variable, so I think that a condensed version leads to a cleaner overall contract, even if the statement itself requires a few more seconds to parse.

contracts/colony/ColonyExpenditure.sol
kronosapiens: True, it would be better if `hasUserRole` was in `ColonyStorage`, since I think this is the first time we're using the function for an internal call. I could accept a call directly to the authority on the grounds that `hasUserRole` was meant for external calls only and so this usage doesn't represent an abstraction violation.

kronosapiens: Regarding the `require`, I made my point in an earlier comment. When scanning the contract, having multiple conditions inside of a `require` tells the reader that all of this logic ultimately is used to validate the caller. The fact that the contents of the `require` are slightly more obtuse is a trade-off, but one which I argue makes the overall contract easier to understand. If the reader needs to understand the specifics of the `require`, they can then make the extra effort to parse the contents. But if not, then the more dense version makes the contract as a whole easier to interpret.

contracts/colony/ColonyDomains.sol
kronosapiens: Mm, that's frustrating. Your optimization is better, but I do worry about the gas costs of earning reputation being significantly increased. Doesn't seem like there's a better way though.

Is there any possibility of protecting the slot in the next version, and then updating the functionality at some point down the road, once all colonies have upgraded? It does seem like that would be worth it long-term, given how often this function will be called. Definitely a high-value optimization.

area: If it helps, the limit still seems to be about 150 depth which was the limit when the storage was on colonyNetwork in the original implementation.

> Is there any possibility of protecting the slot in the next version, and then updating the functionality at some point down the road, once all colonies have upgraded? It does seem like that would be worth it long-term, given how often this function will be called. Definitely a high-value optimization.

In principle, yes (if we also prevented deployment of earlier versions, which we currently allow), but I'm fairly confident in saying that we would never get to a point where all colonies had upgraded.

kronosapiens: So it goes, I guess.

area: I mean, that's going to be true with any implementation of this feature, as defined, as wherever the data is stored, you have to work your way up the tree to calculate the scaling factor. The question is by how much. This implementation is slightly worse than the original one, but really not by much.

I haven't worked it out, but it's possible that given how memory costs work (i.e. quadratically, if you start requesting a lot), it might be more gas efficient to 'page' the allParents function and call it a couple of times, but it doesn't look like that's needed for now.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -187,20 +220,25 @@ contract ColonyNetwork is ColonyDataTypes, BasicMetaTransaction, ColonyNetworkSt
...
+    emit ColonyReputationDecayRateToChange(_chainId, _colony, activeReputationMiningCycle, _numerator, _denominator);
   }
 
-  function getColonyReputationDecayRate(address _colony) public view returns (uint256, uint256) {
+  function getColonyReputationDecayRate(uint256 _chainId, address _colony) public view returns (uint256, uint256) {
     uint256 numerator;
     uint256 denominator;
 
-    if (activeReputationMiningCycle != colonyDecayRates[_colony].afterMiningCycle) {
+    uint256 chainId = _chainId;
```
kronosapiens, line 230: Could have done this as `uint256 chainId = isMiningChainId(_chainId) ? 0 : _chainId`

test/contracts-network/colony.js
```diff
@@ -450,4 +458,162 @@ contract("Colony", (accounts) => {
...
+
+  describe("when setting the token reputation scaling factor", async () => {
+    it("can read the reputation rate for a token", async () => {
+      const rate = await colony.getTokenReputationRate(token.address);
+      expect(rate).to.eq.BN(WAD);
+    });
+
+    it("can set the reputation rate for more than ten tokens", async () => {
+      let i = ethers.BigNumber.from(1);
+      while (i < 11) {
```
kronosapiens, line 581: Why isn't this a `for` loop?

test/contracts-network/common-scale-reputation.js
```diff
@@ -0,0 +1,61 @@
...
+      expect(scaled).to.eq.BN(INT128_MIN);
+    });
+
+    it("deal with calculations that exceed our reputation cap during calculation, but not once calculation is complete", async () => {
+      const scaled = await scaleReputationTest.scaleReputationPublic(INT128_MIN.addn(10), WAD.subn(1));
+      expect(scaled).to.be.gt.BN(INT128_MIN);
+    });
+
+    it("deal with calculations that exceed our reputation cap during calculation, but not once calculation is complete", async () => {
+      const scaled = await scaleReputationTest.scaleReputationPublic(INT128_MAX.subn(10), WAD.subn(1));
```
kronosapiens, line 57: If these two tests have the same description, I would include both test calls inside the same test.

kronosapiens (commented):
Final pass, mostly cosmetic

contracts/colony/ColonyDomains.sol
```diff
@@ -124,6 +127,15 @@ contract ColonyDomains is ColonyStorage {
...
+    require(domainExists(_domainId), "colony-domain-does-not-exist");
+    require(_factor <= WAD, "colony-network-invalid-reputation-scale-factor");
+
+    skillReputationRateComplements[domains[_domainId].skillId] = WAD - _factor;
```
kronosapiens, line 134: `skillReputationScalingComplements`?

contracts/colony/IColony.sol
```diff
@@ -1109,4 +1109,30 @@ interface IColony is ColonyDataTypes, IRecovery, IBasicMetaTransaction, IMultica
...
+  /// @notice Call to get the reputation scaling applied to payouts made in a particular token
+  /// @param token The token we wish to query
+  /// @return scaleFactor The amount of scaling to applied as a WAD.
+  function getTokenReputationScaling(address token) external view returns (uint256 scaleFactor);
+
+  /// @notice Get the reputation scaling applied to reputation earned in a skill in this colony.
+  /// @dev To look up the scaling in a domain, look up the skill corresponding to that domain
+  /// @param skillId The skill to get the value of scaling in
+  /// @return scaleFactor Returns the scale factor applied to reputation earned in this skill, as a WAD.
+  function getSkillReputationScaling(uint256 skillId) external view returns (uint256 scaleFactor);
```
kronosapiens, line 1137: Is there an argument for making this `getDomainReputationScaling` to mirror `setDomainReputationScaling`?

contracts/colonyNetwork/ColonyNetworkSkills.sol
```diff
@@ -0,0 +1,499 @@
...
+pragma solidity 0.8.20;
+pragma experimental "ABIEncoderV2";
+
+import "./../reputationMiningCycle/IReputationMiningCycle.sol";
+import "./../common/Multicall.sol";
+import "./ColonyNetworkStorage.sol";
+import "./../colony/ColonyDataTypes.sol";
+
+
+contract ColonyNetworkSkills is ColonyNetworkStorage, Multicall, ColonyDataTypes {
```
kronosapiens, line 27: Is `ColonyDataTypes` needed here as well?

contracts/colony/IColony.sol
area: This was mostly me keeping options open for the future (as this strictly implements skill-scaling, not domain-scaling, under the hood) but does also let the reputation reward scaling be queried. I've added a test showing that.

PR #1142, opened 2023-06-16
Reorganize ColonyAuthority contract

kronosapiens:
Currently it is difficult for product folks to keep track of the permissions associated with various roles. A reorganization of the `ColonyAuthority` contract will make it easier to understand different roles and how they relate.


kronosapiens:
Noting that @area would like to see indications of which version including which permission, likely as a comment

kronosapiens:
Closing this because it's stale / unnecessary

PR #1144, opened 2023-07-06
Multisig Permissions Extension

area:
This PR implements a 'multisig permissions' extension. The underlying intention here is to make it so that powerful permissions (e.g. root) don't have to be under the control of a single EOA. So an action that the multisig should take can be suggested, and then if enough people (with the 'right' permissions) sign off on it, then it can be executed.

### Specification

Should mirror the colony-level ('core' being the nomenclature I've adopted for those) permissions exactly - so being able to award permissions in specific domains, as well as being able to leverage permissions in child domains of the domains where the permissions have been explicitly awarded.

In terms of the 'threshold' for execution, the ideal functionality was some fraction of people with the relevant permissions. However, I couldn't establish a way to realistically do this, ~~and so went with an 'acceptable' threshold for 'votes' on an action before it could be executed. If a reviewer comes up with a way to implement fraction-based rather than absolute-based approvals, I'd be interested to hear it.~~

Permissions work as such:

* If a fixed threshold for the relevant domain is set, that number is used
* Otherwise, if a fixed global threshold is set, that number is used
* Otherwise, `n/2 + 1` approvals are required, where `n` is the number of people who _explicitly_ have the relevant permission in the domain in question. People with the permission inherited can approve, but do not count towards the calculation of the threshold.

Less prose-like requirements:
* Removal of a permission doesn't remove approvals already made, and you can remove your approvals even if the permission that originally allowed you to make an approval is removed.
* Anyone can execute an action if it has enough approvals.
* Overapprovals are allowed
* To suggest an action, you require ~the permissions~ at least one of the permissions that would be needed to approve it, and you automatically approve it on creation.
* Multiple transactions can be suggested in a single action (either natively, or via multicall), but they must all require the same permissions.
* If an action fails on execution, the transaction reverts if it's been less than a week since it was eligible for execution. Otherwise, the action fails, though the transaction succeeds overall, and the motion will not be able to be retried.
* Core permissions can award multisig permissions they could award core versions of, and vice-versa (if the threshold is reached when going via multisig)

### Implementation notes

There is some arguably surprising behaviour around what happens if the threshold is changed and motions have already had approvals made, but I've tried to make it as straightforward as possible; if an approval/unapproval causes a motion to cross the threshold, the timestamp is updated. If the timestamp is 0, but we're able to `execute`, the call of `execute` will update the timestamp to `now`, and do nothing else. A second call to execute will actually execute the motion. I'm very open to suggestions here, as I'm not totally happy with where I've landed.

Where multiple permissions are required for a function call (currently the case for OneTxPayment, and nothing else), each permission has a separate threshold, and both must be met for execution to be allowed.


arrenv:
> If an action fails on execution, the transaction reverts if it's been less than a week since it was eligible for execution. Otherwise, the transaction succeeds and the motion will not be able to be retried.

Should this be "Otherwise, the transaction fails and the motion will not be able to be retried."

arrenv:
> There is some arguably surprising behaviour around what happens if the threshold is changed and motions have already had approvals made, but I've tried to make it as straightforward as possible; if an approval/unapproval causes a motion to cross the threshold, the timestamp is updated. If the timestamp is 0, but we're able to execute, the call of execute will update the timestamp to now, and do nothing else. A second call to execute will actually execute the motion. I'm very open to suggestions here, as I'm not totally happy with where I've landed.

What I understand from this is that we have these scenarios:
 - Motion HAS passed the threshold and CAN be finalized, a new EOA is awarded the same permissions and now the threshold changes and now the motion CANNOT be finalized until another approval is given.
 - Motion HAS NOT passed the threshold, an EOA has their permissions revoked and now the threshold changes and now the motion CAN be finalized.

However, you have a timestamp to account for this, but that requires an additional transaction each time? Can they be bundled?

Am I right in this understanding?

area:
>> If an action fails on execution, the transaction reverts if it's been less than a week since it was eligible for execution. Otherwise, the transaction succeeds and the motion will not be able to be retried.

> Should this be "Otherwise, the transaction fails and the motion will not be able to be retried."?

It should be "Otherwise, the action fails, though the transaction succeeds overall, and the motion will not be able to be retried".

area:
> What I understand from this is that we have these scenarios:
> 
>     * Motion HAS passed the threshold and CAN be finalized, a new EOA is awarded the same permissions and now the threshold changes and now the motion CANNOT be finalized until another approval is given.
> 
>     * Motion HAS NOT passed the threshold, an EOA has their permissions revoked and now the threshold changes and now the motion CAN be finalized.
> 
> 
> However, you have a timestamp to account for this, but that requires an additional transaction each time? Can they be bundled?
> 
> Am I right in this understanding?

First off, the number of addresses with a given permission is independent of the threshold, which is just a fixed number. If that many people or more have approved something, it can be called.

In the first scenario, let's say 2 people have approved a motion (which was sufficient for execution, and set the timestamp) and the threshold is now 3. Attempts to execute will fail (as we do not meet the threshold). If a third approves, the timestamp will be updated, and the motion will have a week as expected before failed executions are allowed. No additional transactions are required.

In the second scenario, let's say 2 people had approved a motion, with a threshold of 3. The threshold is now moved to 2, and so the motion can be executed, but the timestamp hasn't been set. The first call of `execute` will set the timestamp and do nothing else. A second call of execute will actually execute the motion (or try to, if it fails). These can be bundled if the motion is going to execute successfully, but if the motion fails and the transaction reverts, if you have bundled them then the timestamp will also not be set, and so the 'clock' for allowing a failing execution will not have started.

I'll also note that in normal activity (i.e. the threshold not being changed), only one `execute` call is ever required.

arrenv:
> First off, the number of addresses with a given permission is independent of the threshold, which is just a fixed number. If that many people or more have approved something, it can be called.

> In the first scenario, let's say 2 people have approved a motion (which was sufficient for execution, and set the timestamp) and the threshold is now 3. Attempts to execute will fail (as we do not meet the threshold). If a third approves, the timestamp will be updated, and the motion will have a week as expected before failed executions are allowed. No additional transactions are required.
> 
> In the second scenario, let's say 2 people had approved a motion, with a threshold of 3. The threshold is now moved to 2, and so the motion can be executed, but the timestamp hasn't been set. The first call of `execute` will set the timestamp and do nothing else. A second call of execute will actually execute the motion (or try to, if it fails). These can be bundled if the motion is going to execute successfully, but if the motion fails and the transaction reverts, if you have bundled them then the timestamp will also not be set, and so the 'clock' for allowing a failing execution will not have started.
> 
> I'll also note that in normal activity (i.e. the threshold not being changed), only one `execute` call is ever required.

arrenv:
> First off, the number of addresses with a given permission is independent of the threshold, which is just a fixed number. If that many people or more have approved something, it can be called.

This is on me, as we spoke about this but I misunderstood. I had understood our two options were a configurable fixed threshold which was difficult and the other was a broad approach using the same formula as exiting recovery `numRequired = totalAuthorized / 2 + 1`. Although, I distinctly remember you saying it needed to be fixed and I seemed to have perceived that as meaning the opposite. So, now I have questions:

 - How is this set? Is it per Colony or Colony Network wide?
 - Is it configurable on extension enabling?
 - Is it changeable after it has been set?
 - Is it configurable per permission type and per team?
 - If it is set to a value higher then the number of permission holders, does that mean Motions can never be approved?

 
> In the second scenario, let's say 2 people had approved a motion, with a threshold of 3. The threshold is now moved to 2, and so the motion can be executed, but the timestamp hasn't been set. The first call of `execute` will set the timestamp and do nothing else. A second call of execute will actually execute the motion (or try to, if it fails). These can be bundled if the motion is going to execute successfully, but if the motion fails and the transaction reverts, if you have bundled them then the timestamp will also not be set, and so the 'clock' for allowing a failing execution will not have started.

So, basically by bundling them you can get around the 1 week rule, in a sense. So, they need to be two transactions so that the rule stays in place.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,457 @@
...
+      (userPermissions & motion.overallPermissions) > 0 ||
+      // Are trying to remove an approval
+      _approved == false
+    , "colony-multisig-no-permissions");
+
+    // Update appropriately
+    if (motionApprovals[_motionId][msgSender()] != _approved){
+      motionApprovals[_motionId][msgSender()] = _approved;
+      // Then they changed their approval
+      if (_approved) {
```
kronosapiens, line 209: Could do a ternary here

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,457 @@
...
+        // They are now approving it
+        motion.approvalCount += 1;
+      } else {
+        motion.approvalCount -= 1;
+      }
+      emit ApprovalChanged(msgSender(), _motionId, _approved);
```
kronosapiens, line 215: Newline

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    // If now over the threshold (for the first time), start the 'timeout' clock
+    if (motion.approvalCount == threshold && _approved == true) {
+      motion.overallApprovalTimestamp = block.timestamp;
+    }
+
+    // If we lose our approvals before execution, un-set the approval timestamp
+    if (motion.approvalCount == threshold - 1 && _approved == false) {
+      motion.overallApprovalTimestamp = 0;
```
kronosapiens, line 225: I would argue that `delete` would be more semantic here

kronosapiens (commented):
Initial skim. My first reaction is that I don't particularly like so much of the permissions logic being simply duplicated here, but will need to think more about the intended functionality before making an argument in any direction.

area:
>Relative value dependent on the number of _explicit_ permission holders of the team, _not inherited_

Just noting this explicitly/precisely for posterity

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,457 @@
...
+  /// @notice Called when deprecating (or undeprecating) the extension
+  /// @param _deprecated Indicates whether the extension should be deprecated or undeprecated
+  function deprecate(bool _deprecated) public override auth {
+    deprecated = _deprecated;
+  }
+
+  function getCapabilityRoles(bytes4 _sig) public view override returns (bytes32) {
+    if (_sig == SETUSERROLES) {
+      return (
+        bytes32(uint256(1)) << uint8(ColonyDataTypes.ColonyRole.Architecture) |
```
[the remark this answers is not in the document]
area: Taken a slightly different, but still constant based, approach.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+    changeApproval(_permissionDomainId, _childSkillIndex, motionCount, true);
+  }
+
+  function changeApproval(uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _motionId, bool _approved) public notExecuted(_motionId) {
+    // Get permissions in _permissionDomainId for user
+    bytes32 userPermissions = getUserRoles(msgSender(), _permissionDomainId);
+
+    // Work out which domain they're using their permissions in, via child skill index
+    uint256 permissionSkillId = colony.getDomain(_permissionDomainId).skillId;
+    uint256 userActingInDomainWithSkillId = colonyNetwork.getChildSkillId(permissionSkillId, _childSkillIndex);
```
kronosapiens, line 198: I would put these three statements in a separate function, since it's re-used. It would keep this and the previous functions more compact as well:

```
function validateMotionDomain(uint256 _permissionDomainId, uint256 _childSkillIndex, uint256 _motionId) {
    uint256 permissionSkillId = colony.getDomain(_permissionDomainId).skillId;
    uint256 userActingInDomainWithSkillId = colonyNetwork.getChildSkillId(permissionSkillId, _childSkillIndex);
    require(userActingInDomainWithSkillId == motions[_motionId].domainSkillId, "colony-multisig-not-same-domain");
}
```
    
```
 function validateUserPermissions(uint256 _permissionDomainId, uint256 _motionId) {
    bytes32 userPermissions = getUserRoles(msgSender(), _permissionDomainId);
    require(userPermissions & motions[_motionId].requiredPermissions > 0, "colony-multisig-no-permissions");
}
```

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+    // If they're trying to approve, needs some overlap of permissions
+    if (_approved){
+      require(userPermissions & motion.requiredPermissions > 0, "colony-multisig-no-permissions");
+    }
+
+    uint8 roleIndex = 0;
+    bool newlyApproved = false;
+    bool anyBelowThreshold = false;
+    while (uint256(motion.requiredPermissions) >= (1 << roleIndex)){
+      if ((motion.requiredPermissions & bytes32(1<<roleIndex)) > 0){
```
kronosapiens, line 213: Spacing

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+
+        // The user either needs to have the permission, or are removing an approval
+        // If not, move on to next permissions
+        if (!(uint256(userPermissions) & (1 << roleIndex) != 0 || !_approved)) {
+          roleIndex += 1;
+          continue;
+        }
+
+        // Update appropriately
+        if (motionApprovals[_motionId][msgSender()][ColonyRole(roleIndex)] != _approved){
```
kronosapiens, line 224: Wondering if we could store all approvals in a single bytes32, rather than many different booleans. Seems like it would be cheaper, gas-wise, when a user might approve with many permissions.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+  }
+
+  function execute(uint256 _motionId) public notExecuted(_motionId) {
+    Motion storage motion = motions[_motionId];
+
+    // TODO: Check per-permission
+    bytes32 motionPermissions = motion.requiredPermissions;
+    uint8 roleIndex = 0;
+    while (motionPermissions > 0){
+      if (uint256(motionPermissions)%2 == 1){
```
kronosapiens, line 270: Spacing

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+  function getDomainSkillRoleCounts(uint256 _domainSkillId, uint8 role) public view returns (uint256) {
+    return domainSkillRoleCounts[_domainSkillId][role];
+  }
+
+  function getDomainSkillThreshold(uint256 _domainSkillId) public view returns (uint256) {
+    return domainSkillThreshold[_domainSkillId];
+  }
+
+  function setDomainSkillThreshold(uint256 _domainSkillId, uint256 _threshold) public returns (uint256) {
+    require(colony.hasUserRole(msgSender(), 1, ColonyDataTypes.ColonyRole.Root), "multisig-permissions-not-core-root");
```
kronosapiens, line 326: I would turn this into a modifier, `onlyRoot`, since it's used elsewhere.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+    domainSkillThreshold[_domainSkillId] = _threshold;
+  }
+
+  function getDomainSkillRoleThreshold(uint256 _domainSkillId, ColonyRole _role) public view returns (uint256) {
+    // If domain skill threshold explicitly set, use that value
```
kronosapiens, line 331: Is there a reason this isn't structured as an if/else?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,541 @@
...
+  }
+
+
+  function setUserRoles(
+    uint256 _permissionDomainId,
+    uint256 _childSkillIndex,
+    address _user,
+    uint256 _domainId,
+    bytes32 _roles
+  ) public //authDomain(_permissionDomainId, _childSkillIndex, _domainId) archSubdomain(_permissionDomainId, _domainId)
```
kronosapiens, line 367: Delete comment

kronosapiens (commented):
Two high-level thoughts:

- There are a lot of comments here. I'm not sure if you intend to leave them there or if they're artifacts of your thinking through the design as you went, but either way I don't think they need to stay in the final version

- There's some code overlap between this and  #1135. We've discussed introducing "mixins" containing reusable functions and inheriting from them. Would you want try doing that here? I'm thinking `getActionDomainSkillId` and `extractCallData` specifically.

contracts/extensions/MultisigPermissions.sol
area: Only the [training I've had from ESLint](https://eslint.org/docs/latest/rules/no-else-return).

contracts/extensions/MultisigPermissions.sol
area: Yeah, this seems like a great idea :+1:

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,552 @@
...
+    address _user,
+    uint256 _domainId,
+    bytes32 _roles
+  ) public
+  {
+    // Validate the permissions proof
+    uint256 colonyDomainCount = colony.getDomainCount();
+    require(_permissionDomainId > 0 && _permissionDomainId <= colonyDomainCount, "multisig-domain-does-not-exist");
+    require(_domainId > 0 && _domainId <= colonyDomainCount, "multisig-domain-does-not-exist");
+    require(validateDomainInheritance(_permissionDomainId, _childSkillIndex, _domainId), "multisig-invalid-domain-inheritance");
```
kronosapiens, line 359: I feel like we could put the first two requires inside of `validateDomainInheritence`

kronosapiens (commented):
Reiterating that I'd like to see a mixin architecture for some of the functionality used in the multisig, since a lot of those functions are pulled from the voting contracts.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,552 @@
...
+    }
+  }
+
+  function execute(uint256 _motionId) public notExecuted(_motionId) {
+    Motion storage motion = motions[_motionId];
+
+    // TODO: Check per-permission
+    bytes32 motionPermissions = motion.requiredPermissions;
+    uint8 roleIndex = 0;
+    while (motionPermissions > 0){
```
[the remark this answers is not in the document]
area: The reason I did it this way was to be able to short-circuit out of the loop early - there's no need to check every permission here. I could change `changeApproval`'s implementation to look more like this one, I suppose?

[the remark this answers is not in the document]
area: Or I guess the while condition could become something like `(while motionPermissions > (1<<roleIndex)`?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,552 @@
...
+  function createMotion(uint256 _permissionDomainId, uint256 _childSkillIndex, address[] memory _targets, bytes[] memory _data) notDeprecated public {
+    require(_targets.length == _data.length, "colony-multisig-invalid-motion");
+    require(_targets.length >= 1, "colony-multisig-invalid-motion");
+
+    motionCount += 1;
+    Motion storage motion = motions[motionCount];
+    motion.targets = _targets;
+    motion.data = _data;
+
+    for (uint256 i = 0; i < motion.data.length; i += 1) {
```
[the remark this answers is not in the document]
area: I think this is required if we're allowing an arbitrary number of targets and actions.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,552 @@
...
+    globalThreshold = _globalThreshold;
+
+    emit ExtensionInitialised();
+  }
+
+  function getGlobalThreshold() public view returns (uint256) {
+    return globalThreshold;
+  }
+
+  function createMotion(uint256 _permissionDomainId, uint256 _childSkillIndex, address[] memory _targets, bytes[] memory _data) notDeprecated public {
```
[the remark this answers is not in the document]
area: I think this is required for a lot of the functionality that is in mind. To give an example relating to something we're in the process of adding: if I want to edit an expenditure and also make it a staged expenditure, then I would need this. I think it is more likely that this will be added in a future version of motions, rather than being removed here.

area:
I'm not sure I've ever had someone say 'there are too many comments'. They certainly originated from when I was thinking through the implementation, but what's your reasoning here for removing them? Do you not see them as helpful?

kronosapiens:
@area ultimately the comments read to me like they were written to describe the logical flow of the contract to help you implement, rather than comments designed to add useful clarifying information.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -64,27 +63,24 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, ExtractCal
 
   // Storage
 
-  IColonyNetwork colonyNetwork;
-
-  uint256 globalThreshold;
+  // User Address => Domain Skill Id => Roles
```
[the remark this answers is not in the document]
area: I'm also curious about the logic to the reordering of the storage.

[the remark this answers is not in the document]
kronosapiens: I think you're right, it's probably the domainId

There wasn't much logic, it was marked `internal` and so I figured I'd put it up top in a privileged position, but I'd be happy moving it

contracts/extensions/MultisigPermissions.sol
```diff
@@ -315,82 +332,61 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, ExtractCal
...
 
-    // Allow this function to be called if the caller:
+    // Core root or architect in a parent domain
     require(
-    // Has core root permissions OR
-      colony.hasUserRole(msgSender(), 1, ColonyDataTypes.ColonyRole.Root) ||
-    // Has core architecture, if we're using that permission in a child domain of where we have it
-      (
-        // Core architecture check
+      colony.hasUserRole(msgSender(), 1, ColonyDataTypes.ColonyRole.Root) || (
```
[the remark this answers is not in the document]
kronosapiens: I figured if I was going through it I might as well try and optimize line lengths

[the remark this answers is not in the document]
area: #1165 will take care of the details of this.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -315,82 +332,61 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, ExtractCal
...
+    // This is not strictly necessary, since these roles are never used in subdomains
+    require(_roles & ROOT_ROLES == 0 || _domainId == 1, "multisig-bad-domain-for-role");
+
+    require(
+      validateDomainInheritance(_permissionDomainId, _childSkillIndex, _domainId),
+      "multisig-invalid-domain-inheritance"
+    );
 
-    // Allow this function to be called if the caller:
+    // Core root or architect in a parent domain
```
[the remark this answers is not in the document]
area: I have restored these comments, as we discussed. 

You also removed a require, just after here, which I have also restored. I would be interested in knowing why.

contracts/extensions/MultisigPermissions.sol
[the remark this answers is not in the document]
area: I would appreciate you checking rather than just trusting me :sweat_smile:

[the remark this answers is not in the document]
kronosapiens: Going off of the implementation in `DomainRoles` it's the `domainId`

contracts/extensions/MultisigPermissions.sol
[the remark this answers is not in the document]
kronosapiens: Sounds good. I don't entirely recall why I removed the require, except perhaps on the grounds of it not being necessary and not being tested. Happy to leave it in.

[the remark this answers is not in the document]
kronosapiens: I remember why -- the same require occurs earlier in the function: https://github.com/JoinColony/colonyNetwork/pull/1144/files#diff-0cddbe45e5c02aa4032aab39009138118f1209c4e4a6f7b195c0fd564a3facfcR423

kronosapiens (approved):

kronosapiens (dismissed):

kronosapiens (approved):

arrenv:
> Need the ability to 'reject' a transaction, not just not-approve it.

The spec has been updated here - https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189?pvs=4

Additions:
- User story
   - As a User, I want to be able to cancel a proposed action as the owner, and collectively if not the owner. This prevents the action from being able to be indefinitely open if not initially approved.

- Acceptance criteria:
   - Cancelling multi-sig actions
      - It should be possible for the owner of the multi-sig action to be able to cancel their own multi-sig action.
      - It should be possible for other permission holders to collectively cancel a multi-sig action.
      
Design for the cancellation/rejection process are here - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=6177%3A354823&mode=design&t=hoAwbRfZm8kN9beP-1

kronosapiens:
@area @arrenv can this be implemented through some type of expiration window? e.g. after 7 days if the motion is not approved, it is automatically cancelled?

arrenv:
> @area @arrenv can this be implemented through some type of expiration window? e.g. after 7 days if the motion is not approved, it is automatically cancelled?

As the only alternative option, or as an additional option?

Would it work in a way where if not enough people sign in the time window, then it auto cancels?

kronosapiens:
I would suggest doing it as the only option, where if it doesn't pass in the window it automatically fails. I suspect that will be much easier to implement (and use) than a parallel downvote system, but I'll let @area weigh in

arrenv:
> I would suggest doing it as the only option, where if it doesn't pass in the window it automatically fails. I suspect that will be much easier to implement (and use) than a parallel downvote system, but I'll let @area weigh in

What about the owner being able to cancel their own created action, would that still work with this approach? As it is something Motions is getting as well.

area:
Personally, I don't think that the creator of a multisig action should have any special permissions compared to anyone else with the equivalent permissions on a multisig. That is mostly a personal preference rather than anything technical though, and is indeed easy to do.

I agree a timeout is easier to implement, though it doesn't solve the primary issue around why rejections are required, it only ameliorates it. If something is proposed, partially approved, forgotten about, and then the thresholds change, a motion might silently become executable that does something that is no longer (or was never, given it was never sufficiently approved) wanted. Limiting that to seven days makes it less likely, but does not remove the issue.

I have put support for rejections that I believe meets the supplied additional spec in [another branch](https://github.com/JoinColony/colonyNetwork/compare/feat/multisig-permissions...feat/multisig-permissions-rejections) while we continue this discussion.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -81,10 +86,14 @@ contract MultisigPermissions is
 
   uint256 motionCount;
   mapping(uint256 => Motion) motions;
-  // Motion Id => User => Approvals
-  mapping(uint256 => mapping(address => bytes32)) motionApprovals;
-  // Motion Id => Permission => Approval Count
-  mapping(uint256 => mapping(ColonyRole => uint256)) motionRoleApprovalCount;
+
+  uint256 constant APPROVAL = 1;
+  uint256 constant REJECTION = 2;
```
kronosapiens, line 92: Broadly, I think this should be implemented using an enum:

```
enum Vote { None, Approval, Rejection }
```

And then the rest of the functionality should orient around `Vote` values.

Specifically, I think the approach of having separate `changeApproval` and `changeRejection` functions is unecessary. A single `changeVote` function should suffice, and would allow you to avoid passing in two variables (`_type` and `_setTo`), which I interpret as an artifact of trying to attach rejection functionality on top of an existing implementation. If we're going to support approvals and rejections then we should write the extension as though that had been a goal from the start.

I expect you could achieve the same goals of being able to handle vote "changes" by checking new `Vote` values against existing `Vote` values, if any, and making `_setTo` an internal value (along the lines of `voteChanged`) and using that down-stream.

area: I've added the struct, and adjusted the interface to be more in line with what you suggest here, though the underlying functionality remains much the same (which, if I understand your suggestion here, I don't think there's much I can do about).

The main difference in functionality between my original implementation and this one is that you cannot simultaneously reject and approve a motion, which does make sense (though is gassier).

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,673 @@
...
+    }
+
+    if (globalThreshold > 0) {
+      return globalThreshold;
+    }
+
+    return (domainSkillRoleCounts[_domainSkillId][uint8(_role)] / 2) + 1;
+  }
+
+  function getActionSummary(
```
kronosapiens, line 461: Can we adapt the implementation to use the shared `GetActionSummary` mixin? Seems wasteful to have a slightly different reimplementation here.

contracts/extensions/votingReputation/VotingReputationStaking.sol
```diff
@@ -49,7 +49,10 @@ contract VotingReputationStaking is VotingReputationStorage {
     // For v9 expenditure motions, only allow counterstaking unless escalated
     if (
       _motionId <= motionCountV10 &&
-      isExpenditureSig(getActionSummary(motion.action, motion.altTarget).sig)
+      isExpenditureSig(
+        getActionSummary(address(colonyNetwork), address(colony), motion.action, motion.altTarget)
```
kronosapiens, line 53: I'm noticing that every invocation casts to address, and then re-casts inside of `getActionSummary`. What do you think about passing in `type IColonyNetwork`, etc directly?

test/extensions/voting-rep.js
```diff
@@ -1754,28 +1754,28 @@ contract("Voting Reputation", (accounts) => {
       let summary;
 
       // No action
-      summary = await voting.getActionSummary(NO_ACTION, ADDRESS_ZERO);
+      summary = await voting.getActionSummary(colonyNetwork.address, colony.address, NO_ACTION, ADDRESS_ZERO);
```
kronosapiens, line 1757: Ditto here. Can we pass in `colonyNetwork` etc directly?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,673 @@
...
+      _vote == Vote.Rejection
+    );
+  }
+
+  function changeVoteFunctionality(
+    uint256 _permissionDomainId,
+    uint256 _childSkillIndex,
+    uint256 _motionId,
+    Vote _vote,
+    bool _setTo
```
kronosapiens, line 265: Is there any way to dispense with this argument? What information does it provide above-and-beyond `_vote`? I see it's used to call `changeVoteFunctionality` twice, but I can't imagine this is the best implementation.

area: Maybe I'm not understanding what you're looking for here. Your top-level suggestion was 

> A single changeVote function should suffice, and would allow you to avoid passing in two variables (_type and _setTo)

In order to provide functionality where users can remove approvals or rejections, the implication of that suggestion (I believe) is that if I call `changeVote(REJECTION)`, it touches the storage for both rejections (to record that I am now voting for a rejection) and approvals (to record that, if I was voting for approval, I am no longer).

That common functionality is factored out in to `changeVoteFunctionality`. It could all be one function, but there would be a lot of duplication, which I assume you're against as well.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,640 @@
...
+
+  // Internal functions
+
+  function getSig(bytes memory motion) internal pure returns (bytes4 sig) {
+    assembly {
+      sig := mload(add(motion, 0x20))
+    }
+  }
+
+  function validateDomainInheritance(
```
kronosapiens, line 560: A lot of external calls here. Might be worth making `validateDomainInheritence` public on the colony itself.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,640 @@
...
+      return _childSkillIndex == UINT256_MAX;
+    } else {
+      Domain memory domain = colony.getDomain(_permissionDomainId);
+      Domain memory childDomain = colony.getDomain(_domainId);
+      uint256 childSkillId = colonyNetwork.getChildSkillId(domain.skillId, _childSkillIndex);
+      return childSkillId == childDomain.skillId;
+    }
+  }
+
+  function validateMotionDomain(
```
kronosapiens, line 582: This could be replaced with a variation on `validateDomainInheritance`

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,640 @@
...
+    uint256 _permissionDomainId,
+    uint256 _childSkillIndex,
+    uint256 _motionId
+  ) internal view {
+    uint256 permissionSkillId = colony.getDomain(_permissionDomainId).skillId;
+    uint256 userActingInDomainWithSkillId = colonyNetwork.getChildSkillId(
```
kronosapiens, line 588: What happens with `_childSkillIndex = UINT256_MAX` here?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    if (globalThreshold > 0) {
+      return globalThreshold;
+    }
+
+    return (domainSkillRoleCounts[_domainSkillId][_role] / 2) + 1;
+  }
+
+  function getActionSummary(
```
kronosapiens, line 431: Can we adapt the implementation to use the shared GetActionSummary mixin? Seems wasteful to have a slightly different reimplementation here.

contracts/extensions/MultisigPermissions.sol
area: We hit [the branch](https://github.com/JoinColony/colonyNetwork/blob/485a950b81101253fd5614598e5a0bc1e6842d6b/contracts/colonyNetwork/ColonyNetwork.sol#L165) for that special case.

(I assume this comment is about whether it works at all, rather than whether it should be short-circuited here).

contracts/extensions/MultisigPermissions.sol
area: I've done what I can the generalise this. I've ended up making the mixin `getSingleActionSummary`, as ReputationVoting and multisig permissions have different requirements, and therefore different implementations past that point:

* Reputation Voting treats the `NO_ACTION`, `OLD_MOVE_FUNDS` signatures differently
* So long as the same domain / expenditure is being targeted, any combination of required permissions is allowed. For Multisig, functions must all require the same permissions (though I'm considering relaxing this requirement - I suspect that it's actually too restrictive for what product would like)

Certainly not sold on this division of functionality, so I would certainly consider it up for further discussion, but I think it's the best we can do with these sets of requirements.

contracts/extensions/MultisigPermissions.sol
area: Are you suggesting using validateDomainInheritance? Because I don't think we can. We have the `motionId`, and therefore the `domainSkillId` - which is not the domain. To do this, we would have to store `domainSkillId` instead.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -439,10 +439,6 @@ contract MultisigPermissions is
     return domainSkillRoleCounts[_domainSkillId][role];
   }
 
-  function getDomainSkillThreshold(uint256 _domainSkillId) public view returns (uint256) {
```
area: Hopefully I'm missing something, but in combination with the commit message, it looks like you deleted this function in order to improve coverage, instead of writing a test for it?

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -187,11 +187,6 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
     return changed;
   }
 
-  /// @notice @deprecated
-  function deprecateSkill(uint256 _skillId) public stoppable {
```
area, line 191: Why is this happening here, when presumably it should be happening in my `maint/deprecation-removal` branch?

contracts/extensions/MultisigPermissions.sol
kronosapiens: The function isn't used anywhere, and the functionality is duplicated in `getDomainSkillRoleThreshold`, so I assumed it was an artifact of an earlier implementation.

contracts/colonyNetwork/ColonyNetwork.sol
kronosapiens: I was looking for opportunities to improve coverage, and removing a function which was deprecated ages ago and is no longer testable seemed like a good place to start. I can move it into the other branch, but this branch wont be mergable until that one is.

contracts/extensions/MultisigPermissions.sol
area: By 'previous version', do you mean removing the mixin entirely, as it would only be used in one place?

I think there are benefits to this refactorisation though - specifically having a place where edge cases like `setUserRole` can live. What is it that you don't like about the reduced functionality in the mixin?

contracts/extensions/MultisigPermissions.sol
area: I mean, it's different. If you get a number back from `getDomainSkillRoleThreshold`, you need access to this function to work out _why_ the number you got back is why it is (i.e. is it set for the this domain specifically vs globally vs based on the number of people with the relevant role in the domain).

That said, I could certainly see this function not being necessary if the block ingestor handles the relevant event where this value is emitted and stores things appropriately for the frontend to access. Generally speaking though, I am uncomfortable with the idea of having storage without corresponding getters.

kronosapiens: Fair enough

contracts/common/GetActionSummary.sol
```diff
@@ -147,15 +147,17 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
...
-        //   the entire multicall is such and we break
-        return actionSummary;
+        //   the entire multicall is such
+        totalSummary.sig = actionSummary.sig;
       } else if (isExpenditureSig(actionSummary.sig)) {
         // If it is an expenditure action, we record the expenditure ids
         //  and ensure it is consistent throughout the multicall.
         //  If not, we return UINT256_MAX which represents an invalid multicall
-        totalSummary.sig = actionSummary.sig;
+        if (totalSummary.sig != NO_ACTION && totalSummary.sig != OLD_MOVE_FUNDS) {
```
kronosapiens, line 158: This could also use `!isSpecialFunction`

area:
The [spec](https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189#fec3699a5fd241a2925bd2809494b157) is now much more comprehensive and we should make sure we still meet it. Of note:

> - Giving permissions requires Admin level or higher tiered permissions (`Architectur` or `Root` custom and contract level permissions). This should work with both the core permissions and the multi-sig permissions. With multi-sig still requiring consensus via the multi-sig approval mechanism.

area:
Should auto-cancel if threshold is met

area:
Should have the same 'default address' functionality with 0x00000...0000 as motions.

area:
Comment to please Chewie, because I haven't taken in to account adding/removing tags, which I probably should :sweat_smile:

contracts/extensions/votingReputation/IVotingReputation.sol
```diff
@@ -281,10 +281,14 @@ interface IVotingReputation is IColonyExtension, VotingReputationDataTypes {
   ) external view returns (uint256 _rewardMin, uint256 _rewardMax);
 
   /// @notice Return a summary of the multicall action
+  /// @param _colonyNetwork The address of the colony network
+  /// @param _colony The address of the colony this extension is installed in
   /// @param _action The id of the motion
   /// @param _altTarget The address of the altTarget, or 0x0 if none exists
   /// @return _summary A summary of the multicall
   function getActionSummary(
```
area, line 289: I feel like we actually shouldn't change this interface - inviting ourselves to trust user input about what `colonyNetwork` is seems like it will set us up for disaster down the line.

kronosapiens: I think that's a good point. Are `colony` and `colonyNetwork` not stored on this extension already?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+import { GetActionSummary, ActionSummary } from "./../common/GetActionSummary.sol";
+
+// ignore-file-swc-108
+
+contract MultisigPermissions is
+  ColonyExtensionMeta,
+  ColonyDataTypes,
+  ExtractCallData,
+  GetActionDomainSkillId,
+  GetActionSummary
```
kronosapiens, line 36: Wonder if its possible to bundle both `ExtractCallData` and `GetActionDomainSkillId` into `GetActionSummary`, since that's what we actually use here.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+      require(_targets[i] != address(colony), "multisig-passed-target-cannot-be-base-colony");
+
+      ActionSummary memory actionSummary = getActionSummary(
+        address(colonyNetwork),
+        address(colony),
+        _data[i],
+        _targets[i]
+      );
+
+      require(actionSummary.domainSkillId < type(uint256).max, "colony-multisig-invalid-motion");
```
kronosapiens, line 183: I assume we made an explicit choice to handle these errors in the caller, but I'm wondering if that was the right choice / if we can move this back into `getActionSummary`.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+        require(
+          motion.domainSkillId == actionSummary.domainSkillId &&
+            motion.requiredPermissions == actionSummary.requiredPermissions,
+          "colony-multisig-invalid-motion"
+        );
+      }
+    }
+
+    validateMotionDomain(_permissionDomainId, _childSkillIndex, motionCount);
+    validateUserPermissions(_permissionDomainId, motionCount);
```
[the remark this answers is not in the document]
kronosapiens: Given that the only vote here is made through `changeVoteFunctionality`, I would assume not.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+    validateMotionDomain(_permissionDomainId, _childSkillIndex, _motionId);
+    changeVoteFunctionality(_permissionDomainId, _motionId, Vote.Approve, _vote == Vote.Approve);
+    changeVoteFunctionality(_permissionDomainId, _motionId, Vote.Reject, _vote == Vote.Reject);
+
+    if (_vote == Vote.Reject && checkThreshold(_motionId, Vote.Reject)) {
+      cancelFunctionality(_motionId);
+    }
+  }
+
+  function changeVoteFunctionality(
```
kronosapiens, line 226: Stylistic comment, but I wouldn't be opposed to the convention of prepending underscores to internal implementation functions, i.e. `_changeVote()` in lieu of `changeVoteFunctionality`. But up to you.

I also think we should move these private functions towards the bottom.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+            emit RejectionChanged(msgSender(), _motionId, roleIndex, _setTo);
+          }
+        }
+
+        if (_vote == Vote.Approve) {
+          uint256 threshold = getDomainSkillRoleThreshold(motion.domainSkillId, roleIndex);
+          if (motionVoteCount[_motionId][_vote][roleIndex] < threshold) {
+            anyBelowThreshold = true;
+          } else if (motionVoteCount[_motionId][_vote][roleIndex] == threshold && _setTo) {
+            newlyAtThreshold = true;
```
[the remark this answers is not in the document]
kronosapiens: Looks as though a user could delay failing execution indefinitely by continuing to refresh this timestamp.

Something like

```
motion.overallApprovalTimestamp  = (motion.overallApprovalTimestamp) ? motion.overallApprovalTimestamp : block.timestamp
```

seems like it would prevent that. I would have preferred using `min` or `max` but those wouldn't work here, since our goal is the intermediate value between `0` and `block.timestamp`.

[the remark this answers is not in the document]
kronosapiens: Also, perhaps we could rename that variable to `atThreshold` since the `newly` seems to be a bit wobbly :)

Or else -- can we make duplicate votes a no-op?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,551 @@
...
+        block.timestamp > motion.creationTimestamp + 7 days,
+      "colony-multisig-not-enough-rejections"
+    );
+
+    cancelFunctionality(_motionId);
+  }
+
+  function cancelFunctionality(uint256 _motionId) private {
+    Motion storage motion = motions[_motionId];
+    motion.rejected = true;
```
kronosapiens, line 308: This could just be `motions[_motionId].rejected = true` since we don't re-use the variable.

contracts/extensions/MultisigPermissions.sol
kronosapiens: I suspect it's because we wanted to reserve the right not to error and to return instead, but I can't remember exactly.

contracts/extensions/votingReputation/IVotingReputation.sol
kronosapiens: Re-reading the changes, I don't think this is a risk -- neither value is user-supplied.

contracts/extensions/MultisigPermissions.sol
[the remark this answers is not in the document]
area: I think with either solution it can't be prevented, as someone could withdraw their vote and then re-approve, which would (I would argue correctly) update the timestamp. Ultimately I don't think there's an issue here, but making duplicate votes a no-op in this context is straightforward so we probably should.

contracts/extensions/votingReputation/IVotingReputation.sol
area: I agree for now it's fine, but it's down-the-line I'm worrying about, where we see this interface and use it without thinking. I'll just make a public wrapper for the mixin, which will still take the parameters.

contracts/extensions/MultisigPermissions.sol
area: Agree on the movement. I'm not 100% sold on the renaming; we have had slip-ups in terms of variables before that only differed by a prepended `_`. I think it's less likely to go unnoticed in this instance, though, as the result would probably be an infinite loop. I'll sit on it for now and see if I get more comfortable with the idea.

contracts/extensions/MultisigPermissions.sol
area: Music to my ears! I've given it a go, but a fair few tests had to change (and I think this revealed some tests were incorrect, but would like input on that front). I've put these changes in [their own commit](https://github.com/JoinColony/colonyNetwork/pull/1144/commits/0b350709fe9da6c1b8a15890498e0a37d1de550e) so they're easier to look at standalone.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -400,14 +400,14 @@ contract MultisigPermissions is
...
           }
-        }
 
-        if (_vote == Vote.Approve) {
-          uint256 threshold = getDomainSkillRoleThreshold(motion.domainSkillId, roleIndex);
-          if (motionVoteCount[_motionId][_vote][roleIndex] < threshold) {
-            anyBelowThreshold = true;
-          } else if (motionVoteCount[_motionId][_vote][roleIndex] == threshold && _setTo) {
-            newlyAtThreshold = true;
+          if (_vote == Vote.Approve) {
```
kronosapiens, line 403: This could be moved up into the previous block.

contracts/extensions/MultisigPermissions.sol
area: Yeah, these imports are totally unnecessary, so removed.

contracts/extensions/MultisigPermissions.sol
area: Done, which I believe also means we can keep the name of `newlyAtThreshold`?

test/extensions/voting-rep.js
```diff
@@ -1889,42 +1893,51 @@ contract("Voting Reputation", (accounts) => {
...
-      multicall = await encodeTxData(colony, "multicall", [[action3, action8]]);
-      summary = await voting.getActionSummary(multicall, ADDRESS_ZERO);
-      expect(summary.sig).to.equal(soliditySha3("addDomain(uint256,uint256,uint256)").slice(0, 10));
-      expect(summary.expenditureId).to.be.zero;
-      expect(summary.domainSkillId).to.eq.BN(domain1.skillId);
+      // multicall = await encodeTxData(colony, "multicall", [[action3, action8]]);
+      // summary = await voting.getActionSummary(multicall, ADDRESS_ZERO);
+      // expect(summary.sig).to.equal(soliditySha3("addDomain(uint256,uint256,uint256)").slice(0, 10));
+      // expect(summary.expenditureId).to.be.zero;
+      // expect(summary.domainSkillId).to.eq.BN(domain1.skillId);
```
kronosapiens, line 1907: Should we delete this commented-out test content?

kronosapiens (dismissed):

area:
Just a reminder to be careful when doing refactoring or copying and pasting (or maybe using Copilot?) - I've found spots where tests you've added or edited have either been [meaningfully changed](https://github.com/JoinColony/colonyNetwork/commit/2ba1d97c0ff469e8aa20e58ee754295a15686e5c#diff-21b9ca4704e7ec0ae6d3a6b7979bcbcb742a74c85e88307d3c9a651660294d55L319-R315) or [don't test what they claim](https://github.com/JoinColony/colonyNetwork/pull/1144/commits/8078944b4968fcfc3fa736b0d9e757c5284d079e#diff-21b9ca4704e7ec0ae6d3a6b7979bcbcb742a74c85e88307d3c9a651660294d55R882-R884)

I'm a little confused by what coverage is telling me with the current state of the PR. There is a branch uncovered that I would have otherwise thought should be being hit. Do you have any idea what's going on there?

kronosapiens:
@area do you mean the test [here](https://github.com/JoinColony/colonyNetwork/pull/1144/files#diff-21b9ca4704e7ec0ae6d3a6b7979bcbcb742a74c85e88307d3c9a651660294d55R346)? At first glance I agree with you... it seems strange.

area:
I think I've decided that this is an issue with `solidity-coverage`. Applying this diff:

```
diff --git a/contracts/extensions/MultisigPermissions.sol b/contracts/extensions/MultisigPermissions.sol
index 1b5350a3..4837281d 100644
--- a/contracts/extensions/MultisigPermissions.sol
+++ b/contracts/extensions/MultisigPermissions.sol
@@ -311,13 +311,13 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
     // Allow this function to be called if the caller:
     require(
       // Has core root permissions OR
-      colony.hasUserRole(msgSender(), 1, ColonyDataTypes.ColonyRole.Root) ||
+      !(!colony.hasUserRole(msgSender(), 1, ColonyDataTypes.ColonyRole.Root) &&
         // Has core architecture, if we're using that permission in a child domain
-        (colony.hasUserRole(
+        !(colony.hasUserRole(
           msgSender(),
           _permissionDomainId,
           ColonyDataTypes.ColonyRole.Architecture
-        ) && (_permissionDomainId != _domainId)),
+        ) && (_permissionDomainId != _domainId))),
       "multisig-caller-not-correct-permissions"
     );
 
```

with no other changes results in that branch being marked as covered (and covered both ways, as expected). I _believe_ that `!(!x && !(y&&z))` is logically equivalent to `x || (y && z)` and I have applied that transformation correctly to the code. If you agree, and it behaves the same for you if you run with the above diff (`git apply file.diff`) then I'm happy for this to be merged.

area:
Got it. This one was a subtle one. 

"Rewrite entire test suite in Typescript" moves up a rung on my priorities 😅

kronosapiens:
FWIW codebase migrations to TS is the sort of thing I'd expect LLMs to be able to start handling soon.

kronosapiens (dismissed):

test/extensions/voting-rep.js
```diff
@@ -1788,7 +1789,7 @@ contract("Voting Reputation", (accounts) => {
       summary = await voting.getActionSummary(NO_ACTION, ADDRESS_ZERO);
       expect(summary.sig).to.equal(NO_ACTION);
       expect(summary.expenditureId).to.be.zero;
-      expect(summary.domainSkillId).to.be.zero;
+      // expect(summary.domainSkillId).to.be.zero; // We leave this undefined
```
area, line 1792: By 'undefined' here do you mean it could have an arbitrary value depending on what other actions are in a possible multicall? It certainly does have a value here and could be tested even if that is the case (as the setup is deterministic).

Somewhat relatedly, I'm a little worried about the change in behaviour in `getActionSummary`, where previously if we hit `NO_ACTION` or `OLD_MOVE_FUNDS` we overwrote any previous skill or domain ids we had seen, but now we just set the signature and continue to iterate. I've not been able to come up with a scenario where something breaks in VotingReputation as a result, but am not at the top of my game today and would appreciate you taking a look as well if you haven't already.

kronosapiens: What I mean by "undefined" is that the spec does not specify a result. So while in this (deterministic) call the result is known, with different parameters the result may be different. So rather than test this particular result, which I suggest could be misleading, I would rather leave the result un-tested.

kronosapiens: Regarding the change in `getActionSummary`, why not simply preserve the existing behavior? It's may be a bit of wasted gas, but not the end of the world. Alternatively, we could take my argument of "undefined results" and extend it here.

area:
I've done a bit of a larger refactor than I intended with `getActionSummary`. This stemmed from the fact that I was uncomfortable that `NO_ACTION` had special meaning in Multisig, when that shouldn't have been the case. By returning early again, I think we've been able to remove some `if` statements (as they shouldn't be able to be hit), and I also realised that the refactor to revert in the helper meant some requires could be removed from VotingReputation. I'd appreciate a careful review here - it's not too imposing, but a misplaced `!` would really ruin our day :smile:

My second commit here is to fix an issue that Kerry spotted, which was that the architecture role could be used to create and vote on motions to award permissions in the domain where the architecture role was held (when it should only be allowed in subdomains of where the permission is held, mirroring the 'core' permission behaviour). That's a much cleaner set of changes!

contracts/common/GetActionSummary.sol
```diff
@@ -147,17 +144,16 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
...
 
-      if (isSpecialFunction(actionSummary.sig)) {
-        // If any of the actions are NO_ACTION or OLD_MOVE_FUNDS,
+      if (sigInArray(actionSummary.sig, _dominantSigs)) {
+        // If any of the actions are classed as 'dominant',
         //   the entire multicall is such
-        totalSummary.sig = actionSummary.sig;
+        return actionSummary;
+      } else if (sigInArray(actionSummary.sig, _forbiddenSigs)) {
+        revert("colony-action-summary-forbidden-sig");
```
kronosapiens, line 152: From the perspective of logical flow, I would flip these around such that the forbidden check comes first.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -503,6 +503,15 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
 
   function validateUserPermissions(uint256 _permissionDomainId, uint256 _motionId) internal view {
     bytes32 userPermissions = getUserRoles(msgSender(), _permissionDomainId);
+    Motion storage motion = motions[_motionId];
+
+    if ((motion.requiredPermissions & bytes32(1 << uint256(ColonyRole.Architecture))) != 0) {
+      require(
+        colony.getDomain(_permissionDomainId).skillId != motion.domainSkillId,
```
kronosapiens, line 510: I think we could do `skillId < motion.domainSkillId`, since domainIds are monotonically increasing. Do you agree?

contracts/common/GetActionSummary.sol
```diff
@@ -147,17 +144,16 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
         totalSummary.domainSkillId = actionSummary.domainSkillId;
```
kronosapiens, line 144: I believe this block could be re-written as:

```
require(
  totalSummary.domainSkillId == 0 || totalSummary.domainSkillId == actionSummary.domainSkillId,
  "colony-action-summary-inconsistent-domain-skill-id"
)

totalSummary.domainSkillId = actionSummary.domainSkillId;
```

I might even move it towards the bottom next to the `requiredPermissions` check.

contracts/common/GetActionSummary.sol
```diff
@@ -147,17 +144,16 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
...
       }
 
-      if (isSpecialFunction(actionSummary.sig)) {
-        // If any of the actions are NO_ACTION or OLD_MOVE_FUNDS,
+      if (sigInArray(actionSummary.sig, _dominantSigs)) {
```
kronosapiens, line 147: I would call this function `includes` but that's just a stylistic preference

contracts/common/GetActionSummary.sol
```diff
@@ -170,10 +166,8 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
         }
       } else {
         // If no expenditure signatures have been seen, we record the latest signature
-        // unless we're already flagged as a NO_ACTION or OLD_MOVE_FUNDS
         // Also, we aggregate the permissions as we go
-
-        if (!isExpenditureSig(totalSummary.sig) && !isSpecialFunction(totalSummary.sig)) {
+        if (!isExpenditureSig(totalSummary.sig)) {
```
kronosapiens, line 170: Is this signature check necessary? Haven't we already established that this is false?

contracts/common/GetActionSummary.sol
```diff
@@ -170,10 +166,8 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
...
           totalSummary.sig = actionSummary.sig;
         }
       }
```
kronosapiens, line 173: I believe the following block could also be re-written:

```
require(
  totalSummary.requiredPermissions == 0 || totalSummary.requiredPermissions == actionSummary.requiredPermissions,
  "colony-action-summary-inconsistent-permissions"
)

totalSummary.requiredPermissions = actionSummary.requiredPermissions;
```

contracts/common/GetActionSummary.sol
area: Moving it to the required permissions check is 100% better, totally agree with that. I'm always a little hesitant about refactors like the one you're proposing here, because I remain convinced that the `if` syntax is more understandable (mostly because, when understanding a block like this, one naturally parses it in to English using the word "if").

That said, I have done it :smile:

contracts/common/GetActionSummary.sol
area: I've renamed it `includes`, and refactored to use [`using...for`](https://docs.soliditylang.org/en/v0.8.26/contracts.html#using-for) syntax so it reads cleanly.

contracts/common/GetActionSummary.sol
area: I don't believe so? We've established that `actionSummary.sig` is not an `ExpenditureSig`, but not `totalSummary.sig`.

contracts/common/GetActionSummary.sol
area: I've also refactored another similar section you didn't highlight for consistency.

contracts/common/GetActionSummary.sol
```diff
@@ -0,0 +1,218 @@
...
+        totalSummary.sig = actionSummary.sig;
+
+        require(
+          totalSummary.expenditureId == 0 ||
+            totalSummary.expenditureId == actionSummary.expenditureId,
+          "colony-action-summary-inconsistent-expenditure-id"
+        );
+
+        totalSummary.expenditureId = actionSummary.expenditureId;
+      } else {
```
kronosapiens, line 156: This could also be re-written as:

```
} else if (!isExpenditureSig(totalSummary.sig)) {
```

kronosapiens (dismissed):

contracts/extensions/MultisigPermissions.sol
```diff
@@ -503,6 +503,15 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
 
   function validateUserPermissions(uint256 _permissionDomainId, uint256 _motionId) internal view {
     bytes32 userPermissions = getUserRoles(msgSender(), _permissionDomainId);
+    Motion storage motion = motions[_motionId];
+
+    if ((motion.requiredPermissions & bytes32(1 << uint256(ColonyRole.Architecture))) != 0) {
```
kronosapiens, line 508: Regarding the failing test, my first thought would be to expand this require to explicitly check the motion signature against the signature which use the `archSubdomain` modifier. That could be done here, or through the "dominant signatures" concept you've introduced elsewhere, or potentially both places.

area: It's possible the solution I've implemented here is too cute, but see what you think.

The reason that `archSubdomain` exists at all is because we allow `root` and `architecture` to call those (permission-setting) functions in different circumstances, but that means that `requiredPermissions` is `1 << ColonyRole.Root & 1 << ColonyRole.Architecture` for those functions, and just `1 << ColonyRole.Architecture` for the other functions (those that affect domains / domain structure). That means this check can be keyed off of `requiredPermissions` by being more precise.

contracts/common/GetActionSummary.sol
```diff
@@ -140,12 +142,15 @@ contract GetActionSummary is ExtractCallData, GetActionDomainSkillId {
       } else if (_dominantSigs.includes(actionSummary.sig)) {
         // If any of the actions are classed as 'dominant',
         //   the entire multicall is such
-        return actionSummary;
+        totalSummary.sig = actionSummary.sig;
```
kronosapiens, line 145: Why add the tracking boolean vs returning directly?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -508,7 +538,10 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
     bytes32 userPermissions = getUserRoles(msgSender(), _permissionDomainId);
     Motion storage motion = motions[_motionId];
 
-    if ((motion.requiredPermissions & ROOT_AND_ARCHITECTURE) == ROOT_AND_ARCHITECTURE) {
+    if (
+      getDominantSigs().includes(motion.sig) &&
```
kronosapiens, line 542: Can we not reference `dominantSigs` here?

area: I'm not sure what you're asking? It's not declared in this scope, and we only need it the once here.

contracts/common/GetActionSummary.sol
area: To ensure that invalid multicalls are still caught. I thought the 'disguising' test I added demonstrated that, but I ended up testing slightly different logic there; I've added a second disguising test that, if you return fails.

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,590 @@
...
+
+  /// @notice Configures the extension
+  /// @param _colony The colony in which the extension holds permissions
+  function install(address _colony) public override auth {
+    super.install(_colony);
+    colonyNetwork = IColonyNetwork(colony.getColonyNetwork());
+  }
+
+  function getCapabilityRoles(bytes4 _sig) public pure override returns (bytes32) {
+    if (_sig == SET_USER_ROLES) {
```
kronosapiens, line 116: Does this need updating to account for the other role-change functions?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -0,0 +1,590 @@
...
+    uint256 _motionId
+  ) internal view {
+    uint256 permissionSkillId = colony.getDomain(_permissionDomainId).skillId;
+    uint256 domainSkillId = colonyNetwork.getChildSkillId(permissionSkillId, _childSkillIndex);
+
+    // slither-disable-next-line incorrect-equality
+    require(domainSkillId == motions[_motionId].domainSkillId, "multisig-not-same-domain");
+  }
+
+  bytes32 constant ROOT_AND_ARCHITECTURE =
```
kronosapiens, line 534: This should probably be up top next to the other permission masks.

area: It's not actually used any more, so I've deleted it.

contracts/common/GetActionSummary.sol
area: I actually forgot to push this test yesterday, so you can take a look at it now.

contracts/extensions/MultisigPermissions.sol
area: No, because they're not functions on this contract.

kronosapiens (dismissed):

contracts/extensions/MultisigPermissions.sol
```diff
@@ -314,10 +315,25 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
     return domainSkillThreshold[_domainSkillId];
   }
 
+  function getMotionVoteThreshold(uint256 _motionId, uint8 _role) public view returns (uint256) {
+    Motion storage motion = motions[_motionId];
+    require(bytes32(1 << _role) & motion.requiredPermissions > 0, "multisig-role-not-required");
```
kronosapiens, line 320: Surprised to see this as a `require` vs simply a return -- don't we iterate through all roles elsewhere in the contract? Those don't return unnecessary errors?

contracts/extensions/MultisigPermissions.sol
```diff
@@ -314,10 +315,25 @@ contract MultisigPermissions is ColonyExtensionMeta, ColonyDataTypes, GetActionS
...
+    uint256 relevantDomainSkillId = motion.domainSkillId;
+
+    if (getRoleSettingSigs().includes(motion.sig)) {
+      ColonyNetworkDataTypes.Skill memory s = colonyNetwork.getSkill(motion.domainSkillId);
+      if (s.parents.length > 0) {
```
kronosapiens, line 325: Will root architects be able to set root permissions according to this logic?

contracts/extensions/MultisigPermissions.sol
area: We do, and should skip over them entirely if they're not needed for the motion; this `require` shouldn't be being hit if this function is called from other contracts, but the function is public. You're proposing returning 0 in this case?

contracts/extensions/MultisigPermissions.sol
area: No, but I'm not really sure where you're coming from with this question - can you elaborate, and explain why you think this addition might allow that?

I've added a couple of tests around this, but as ever with anything to do with permissions, keen to know where you're coming from, so if a call will make that easier, let me know.

contracts/extensions/MultisigPermissions.sol
kronosapiens: That was my thought, but if the unnecessary permissions are being skipped elsewhere then I'm happy with the require as it is.

contracts/extensions/MultisigPermissions.sol
kronosapiens: - If memory serves, we don't allow architects to assign permissions in the root domain.

- In this function, it looks as though we are testing to see if the function is a role-setting function, and if so we check the permissions in the parent domain -- unless the domain is root, in which case we check permissions in root.

- If a user had architecture permission in root, would this function allow that permission to vote on a root-level permission change? That is my question.

contracts/common/GetActionSummary.sol
```diff
@@ -0,0 +1,203 @@
...
+        summary.requiredPermissions = permissionMask;
+      }
+    } else {
+      summary.domainSkillId = getActionDomainSkillId(_action, colonyNetworkAddress, colonyAddress);
+      // A special case for permission-setting functions, which can be called by root (in root) and
+      // by architecture (if being used in a child domain of where you have the permission)
+      if (permissionMask == (ONLY_ARCHITECTURE_ROLE_MASK | ONLY_ROOT_ROLE_MASK)) {
+        // slither-disable-next-line incorrect-equality
+        if (summary.domainSkillId == IColony(colonyAddress).getDomain(1).skillId) {
+          permissionMask = ONLY_ROOT_ROLE_MASK;
```
kronosapiens, line 101: Ah, I see that the issue I raised elsewhere is being handled here -- for the role-setting functions, only the root role is allowed when calling the function in the root domain.

contracts/common/GetActionSummary.sol
```diff
@@ -0,0 +1,203 @@
...
+    address target = getTarget(_altTarget, colonyAddress);
+    bytes32 permissionMask = ColonyRoles(target).getCapabilityRoles(sig);
+
+    ActionSummary memory summary;
+    summary.sig = sig;
+
+    if (isExpenditureSig(sig)) {
+      summary.domainSkillId = getActionDomainSkillId(_action, colonyNetworkAddress, colonyAddress);
+      summary.expenditureId = getExpenditureId(_action);
+    } else if (permissionMask | ROOT_ROLES == ROOT_ROLES) {
```
kronosapiens, line 87: Can you add a comment clarifying the purpose of this branch? It's escaping me -- a function with no permission mask is being assigned the root role.

area: Given it a crack. It's not quite as simple as I made the comments sound though - you need the arguments to whatever function being called (e.g. `setExpenditureMetadata` to also be valid arguments to `getActionDomainSkillId` and still have everything match.

Hopefully the motivation behind the block is clear, though, even if I haven't been able to write the test 😅

kronosapiens (approved):

arrenv:
Nice one!!! [🎉](https://emojipedia.org/party-popper)

PR #1153, opened 2023-08-09
Fix Metatransaction broadcaster timeouts

area:
Similar to a few issues we've had on the frontend in the last couple of weeks when it comes to querying large block ranges, the request was timing out and it meant that we were refusing to broadcast metatransactions that set token authorities. 

This fixes that behaviour in what should be a robust fashion.

This is already in production, so this PR is to 'catch up' the repository.


arrenv (dismissed):
Working for me, nice work!

PR #1154, opened 2023-08-11
Fix setExpenditurePayouts logic

area:
The `setExpenditurePayouts` logic is incorrect in the core contracts. When multiple values were set with the same call, the funding pot `payouts` property would not be updated / set correctly. 

The updated/added test fail without the contracts changes, demonstrating the issue.

I don't believe we need to take any actions in terms of checking for 'wrong' expenditures - they're not used in the frontend (yet - development of expenditures surfaced this issue), and the consequences are 'only' that in the case of expenditures with multiple payouts, some payouts will need a new expenditure made. Funds are incorrectly returned to the domain containing the expenditure, so they can be used to pay out again via another expenditure.


kronosapiens (approved):
Curiously enough, I had just caught this over here: https://github.com/JoinColony/colonyNetwork/pull/1150/files#diff-ff53e85cbf7115cdaa7bd0dba0cb20f911b4cd56e6ba1fa58d35ea6dc2655a35R323

PR #1157, opened 2023-08-21
Add ExpenditurePayoutClaimed event

area:
From @jakubcolony 

> I think I'll need to store whether a payout has been already claimed or not, to be able to determine whether there's any more payouts to claim from the UI. I thought about listening to the PayoutClaimed event in block-ingestor and then checking whether the funding pot ID belongs to an expenditure. However, this event doesn't contain information on which slot has been claimed, just the token address and the amount, which I don't think is enough to uniquely identify the claimed payout.
>
> I suppose I could call the getExpenditureSlotPayout every time I want to check their claimed status but caching it is a more preferable solution. Do you have any ideas on how that could be done? Could we possibly have another expenditure specific event, say ExpenditurePayoutClaimed, that would contain the slot info?

I agreed with Jakub's conclusion that there is no way to uniquely identify a claimed payout (especially if there are multiple payouts for the same amount / destination), short of tracing the transaction which we _really_ don't want to expect people to to, and agree that the easiest way to solve this is to add the `ExpenditurePayoutClaimed` event.

There's some crossover here with #1150; when that is complete there will only be one type of payout on the core contracts, so perhaps with this merged, the even there could go back to `PayoutClaimed` with an adjusted signature; alternatively, we could start this event named PayoutClaimed which would be a little confusing to begin with but correct moving forward.

### DIFF
--- a/contracts/colony/ColonyDataTypes.sol
+++ b/contracts/colony/ColonyDataTypes.sol
@@ -350,13 +350,14 @@ interface ColonyDataTypes {
   event ArbitraryTransaction(address target, bytes data, bool success);
 
   /// @notice Event logged when an expenditure payout is claimed.
-  /// @dev This is emitted in addition to PayoutClaimed event
+  /// @dev This is emitted in addition to the other PayoutClaimed
+  /// event. The other will be removed soon.
   /// @param agent The address that is responsible for triggering this event
   /// @param id Id of the expenditure
   /// @param slot Expenditure slot of the payout claimed
   /// @param token Token of the payout claim
   /// @param tokenPayout Amount of the payout claimed, after network fee was deducted
-  event ExpenditurePayoutClaimed(address agent, uint256 id, uint256 slot, address token, uint256 tokenPayout);
+  event PayoutClaimed(address agent, uint256 id, uint256 slot, address token, uint256 tokenPayout);
 
   // Structs
 
--- a/contracts/colony/ColonyDataTypes.sol
+++ b/contracts/colony/ColonyDataTypes.sol
@@ -349,6 +349,17 @@ interface ColonyDataTypes {
 
   event ArbitraryTransaction(address target, bytes data, bool success);
 
+  /// @notice Event logged when an expenditure payout is claimed.
+  /// @dev This is emitted in addition to PayoutClaimed event
+  /// @param agent The address that is responsible for triggering this event
+  /// @param id Id of the expenditure
+  /// @param slot Expenditure slot of the payout claimed
+  /// @param token Token of the payout claim
+  /// @param tokenPayout Amount of the payout claimed, after network fee was deducted
+  event ExpenditurePayoutClaimed(address agent, uint256 id, uint256 slot, address token, uint256 tokenPayout);
+
+  // Structs
+
   struct RewardPayoutCycle {
     // Reputation root hash at the time of reward payout creation
     bytes32 reputationState;
--- a/contracts/colony/ColonyFunding.sol
+++ b/contracts/colony/ColonyFunding.sol
@@ -233,6 +233,7 @@ contract ColonyFunding is ColonyStorage { // ignore-swc-123
 
     // Finish the payout
     processPayout(expenditure.fundingPotId, _token, tokenPayout, slot.recipient);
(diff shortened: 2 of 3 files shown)


contracts/colony/ColonyDataTypes.sol
```diff
@@ -349,6 +349,17 @@ interface ColonyDataTypes {
...
   event ArbitraryTransaction(address target, bytes data, bool success);
 
+  /// @notice Event logged when an expenditure payout is claimed.
+  /// @dev This is emitted in addition to PayoutClaimed event
+  /// @param agent The address that is responsible for triggering this event
+  /// @param id Id of the expenditure
+  /// @param slot Expenditure slot of the payout claimed
+  /// @param token Token of the payout claim
+  /// @param tokenPayout Amount of the payout claimed, after network fee was deducted
+  event ExpenditurePayoutClaimed(address agent, uint256 id, uint256 slot, address token, uint256 tokenPayout);
```
kronosapiens, line 359: I'm happy to simply call this `payoutClaimed`

contracts/colony/ColonyFunding.sol
```diff
@@ -233,6 +233,7 @@ contract ColonyFunding is ColonyStorage { // ignore-swc-123
 
     // Finish the payout
     processPayout(expenditure.fundingPotId, _token, tokenPayout, slot.recipient);
+    emit ExpenditurePayoutClaimed(msgSender(), _id, _slot, _token, tokenPayout);
```
kronosapiens, line 236: Newline before event emission

area:
I will note that, the way that this is currently structured, the payout values is different for these two events (see the test). One includes the network fee, and one does not. Should the new event use the same value as the old one?

kronosapiens:
@area hmm... that's an interesting question. I'm inclined to say the event should be the larger number (i.e. include the fee), since that would make it possible to use the event to track current balances and will probably be more useful for a UI.

kronosapiens:
@area sure, but if the intention is to reduce the need for queries then being able to prepare a UI using the event log seems advantageous. Also, you can use the underlying ERC20 transfer event if you need to get the actual amount the user received. Emitting the value prior to the network key will keep the expenditure-related events to be internally consistent, i.e. agnostic to the network fee.

contracts/colony/ColonyDataTypes.sol
```diff
@@ -349,6 +349,18 @@ interface ColonyDataTypes {
...
 
+  /// @notice Event logged when an expenditure payout is claimed.
+  /// @dev This is emitted in addition to the other PayoutClaimed
+  /// event. The other will be removed soon.
+  /// @param agent The address that is responsible for triggering this event
+  /// @param id Id of the expenditure
+  /// @param slot Expenditure slot of the payout claimed
+  /// @param token Token of the payout claim
+  /// @param tokenPayout Amount of the payout claimed, after network fee was deducted
+  event PayoutClaimed(address agent, uint256 id, uint256 slot, address token, uint256 tokenPayout);
```
kronosapiens, line 360: IIRC we discussed emitting the payout before the network fee is deducted. Is that still the plan?

area: That's the existing implementation, it's just that this comment is wrong from my copy/pasting. But I want to know why we did it the other way in the original implementation, plus whether this is used in the existing stack.

kronosapiens: Ok, standing by

kronosapiens: Any update?

area: It is indeed [used](https://github.com/JoinColony/block-ingestor/blob/0a4a9db0f0241be6e8d0855e7954d17fe614dc06/src/handlers/actions/payment.ts#L71-L72) [in](https://github.com/JoinColony/colonyCDapp/blob/d10adabb5986edb0fd6deb481a7fd007ae640d3e/src/i18n/en-events.ts#L48) the existing stack, so I think we need to deduct the fee.

In terms of why we did it this way, I have not been able to dig it up. My best guess is we decided the amount being actually received was the amount most naturally of interest, but I do agree that is already available via the ERC20 events.

kronosapiens (approved):

PR #1158, opened 2023-08-28
Upgrade node to 14.21.3

chmanie:
This PR upgrades node to the latest 14.x release, incorporating important bugfixes and CVE fixes.


area:
Can you test this with [the dapp repo](https://github.com/JoinColony/colonyDapp) ? I recall us having issues with mismatched versions in the past, though I admit that was probably a major version difference.

chmanie:
Yeah I really don't think this should impact the dApp. But I can test I guess

area:
Having tried it with both the Dapp and the CDapp, both will require changes for this to work without throwing incompatible version errors.

chmanie:
I guess we stay on the old version then?

area:
Superseded by #1242

PR #1160, opened 2023-08-30
Improve reputation miner user experience

chmanie:
tbd


area:
The commit I've added pages the previous mining cycles on syncing. This is exactly the same issue that we were facing in the frontend in a few different areas that Raul fixed. The same solution is done here - instead of asking for all events in one big request, we split it up in to multiple requests and join all the responses together.

PR #1164, opened 2023-09-04
Voting expenditures continuation

### LINKED ISSUE #1163: Improvements to VotingReputation
See here: https://github.com/JoinColony/colonyNetwork/pull/1135#pullrequestreview-1607124642

> I am approving, but leaving it up to you, @kronosapiens, whether you want to merge or not based on the following:
> 
> * I think the upgrade tests can be better as discussed (and also as discussed, am not suggesting this would hold up merging).
> * I think `unlockExpenditure` could be split in to two functions, one for <=v9 and one for >=v10 - it's called from two different branches, based on whether it's an old expenditure or not, and then `unlockExpenditure` does the same branch again, which seems redundant.
> * I [still](https://github.com/JoinColony/colonyNetwork/pull/1135#discussion_r1286876895) think that some function names are pretty bad - in addition to what I mention there, consider the functions `getExpenditureAction` and `createExpenditureAction`, and how different they are.
> * I think the behaviour of a motion created in v9 that is subsequently deemed an invalid multicall is weird. It can be staked against, voted on, and escalated, but then just jumps to a `finalized` state? Why are we not allowing it to be finalized?

area:
Closes #1163

Continuing #1135 

Following up on [my review](https://github.com/JoinColony/colonyNetwork/pull/1135#pullrequestreview-1607124642) of #1135, I've implemented the first three changes I suggested as they seemed likely to be uncontroversial.

How do you think we should approach my last point?


kronosapiens:
Regarding the treatment of newly-invalid motions, I believe the rationale was something along the lines of "if an invalid multicall has been staked, we cannot make any guarantees about the internal consistency of the contents. So, we don't allow it to be executed". There may also have been some consideration of unlocking an expenditure that hadn't initially been locked (as the expenditure action was hidden in the multicall). So on balance I don't think that's bad behavior, and I think that spending additional time developing logic for parsing out the contents and forcing them into consistency is a good use of energy. I would be open to perhaps halting the motion elsewhere in the lifecycle if you think that would be preferable.

contracts/extensions/votingReputation/VotingReputation.sol
```diff
@@ -285,7 +285,10 @@ contract VotingReputation is VotingReputationStorage {
     } else { // Backwards compatibility for versions 9 and below
       ActionSummary memory actionSummary = getActionSummary(motion.action, motion.altTarget);
       if (isExpenditureSig(actionSummary.sig) && getTarget(motion.altTarget) == address(colony)) {
-        unlockV9Expenditure(_motionId);
+        if (!(getSig(motion.action) == MULTICALL)) {
```
area, line 288: `if (getSig(motion.action) != MULTICALL)`?

contracts/extensions/votingReputation/VotingReputationStorage.sol
```diff
@@ -214,14 +213,11 @@ contract VotingReputationStorage is ColonyExtension, BasicMetaTransaction, Votin
       return MotionState.Finalized;
     } else if (_motionId <= motionCountV10 && getSig(motion.action) == MULTICALL) {
```
area, line 214: Is there something stopping us making this

```
} else if (_motionId <= motionCountV10 && (getSig(motion.action) == NO_ACTION || getSig(motion.action) == OLD_MOVE_FUNDS)) {
   return MotionState.Finalized;
} else {
  return MotionState.Finalizable;
}
```

area: Oh, I guess we want to guard against Multicall hiding a forbidden `OLD_MOVE_FUNDS` going forward, even if we can't do anything about it.

area:
After a review:

* I spent some time worrying about nested multicalls, but I suppose that the motion could be directed at a contract that does allow nested multicalls, and just because we don't allow them doesn't mean others might not have a reason for them.
* Not introduced by this PR, but I think I've run across a (minor) issue around a user who votes, does not reveal, and then the motion gets escalated; if the user votes in the second vote, they are not counted towards reputation that has voted in that second round. If you agree this exists, worth opening an issue for to track.
* The new way the global delays work / get reset is what was spec'd, but I don't think we've considered what we should do when the global claim delay that has been set on a colony is so large that it causes overflows when attempting to lock an expenditure? 
* Similarly, someone with arbitration permission can edit the global claim delay while an expenditure is locked, causing underflows when unlocking (and/or, indeed, effectively unlock it).

kronosapiens:
- Regarding nested multicalls, I would say let's leave as-is and if in the future there's user demand for that functionality we can revisit it
- Let me take a look
- Agree that this is possible. I believe an arbitrator _should_ be allowed to circumvent a motion, but adding logic which checks the existing claim delays to avoid over/underflows is probably wise

kronosapiens (approved):

PR #1168, opened 2023-09-13
Reputation Miner Improvements

area:
@chmanie is working on mining documentation and aids for people who wish to reputation mine. This PR is going to capture changes to the miner that are revealed by that work. Currently there is one change:

* Page the previous mining cycles on syncing. This is exactly the same issue that we were facing in the frontend in a few different areas that Raul fixed. The same solution is done here - instead of asking for all events in one big request, we split it up in to multiple requests and join all the responses together.


area:
Subsuming in to #1160

PR #1179, opened 2023-10-23
Add test showing releasing staged expenditure via motion

area:
Wrote this to help Armando, but may as well go in.

### DIFF
--- a/test/extensions/staged-expenditure.js
+++ b/test/extensions/staged-expenditure.js
@@ -151,6 +151,56 @@ contract("Staged Expenditure", (accounts) => {
       await colony.claimExpenditurePayout(expenditureId, 0, token.address);
     });
 
+    it("can release a staged payment with the arbitration permission", async () => {
+      await fundColonyWithTokens(colony, token, WAD.muln(10));
+
+      await colony.makeExpenditure(1, UINT256_MAX, 1, { from: USER0 });
+      const expenditureId = await colony.getExpenditureCount();
+      const expenditure = await colony.getExpenditure(expenditureId);
+      const domain1 = await colony.getDomain(1);
+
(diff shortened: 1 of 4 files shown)


test/extensions/staged-expenditure.js
```diff
@@ -323,4 +337,227 @@ contract("Staged Expenditure", (accounts) => {
       );
     });
   });
+
+  describe("Interactions with other extensions", async () => {
```
kronosapiens, line 341: Looking at this implementation, I'm inclined to argue that we should put these types of tests in the motions test suite, rather than in the suite associated with the extension. The argument is that we already having a mining process set up in that test suite, so it would be marginally easier to incorporate the test.

area:
Added a mechanism to the extension to allow arbitration permissions to release a staged expenditure. They can do this anyway, and the 'proper' mechanism to do this in the frontend, ideally, should emit the event to make life easier. This also should be easier to develop against, not having to use `setExpenditureState`.

test/extensions/voting-rep.js
```diff
@@ -3159,4 +3161,165 @@ contract("Voting Reputation", (accounts) => {
       expect(motion.altTarget).to.equal(oneTxPayment.address);
     });
   });
+
+  describe("interaction with other extensions", async () => {
+    it("can release a staked expenditure via a motion, even if owner has no permissions", async () => {
```
kronosapiens, line 3166: I don't think this test is particularly necessary, the second test makes this one redundant.

kronosapiens (approved):

test/extensions/voting-rep.js
area: To a large extent, I do agree. It's still legitimate behaviour though which e.g. the block ingestor will need to support, so I'm somewhat inclined to leave it in as a demonstration of that.

kronosapiens (approved):

PR #1181, opened 2023-11-02
Add strict rules about upgrading to next versions of colony/OneTxPayment

area:
With the merging of #1150, we are presented with a bit of a dilemma in terms of the user experience we want and the practicalities on the contracts, at least in combination with the restriction of putting as little extra dev work on the frontend team as possible. This comes around because the next version of OneTxPayment needs an additional permission to work, and the next version of Colony won't support old OneTxPayment versions.

The top-level requirement is that users of the frontend shouldn't be able to (with regular usage) get in to a state where OneTxPayment doesn't work, with as little work on the frontend as possible. Some work will be required (to award the new permission required) but that should be a trivial change in both frontends, and so is acceptable.

The approach I've taken here is to (if OneTxPayment is installed)

* Do not allow OneTxPayments to be upgraded 5->6 with the usual flow (i.e. by calling `upgradeExtension`).
* Upgrade OneTxPayments from 5->6 when the colony is upgraded 13->14, and award the new permission at the same time (if appropriate).
* Require OneTxPayments to be v5 when upgrading the colony from 13->14

If OneTxPayments is not installed, the colony can upgrade 13->14 as normal. If OneTxPayments is installed, but doesn't have permissions in the root domain (the implication being non-standard use exclusively in subdomains), then the same restrictions apply to the upgrade, but the extension won't be awarded the arbitration permission in the root domain. In this scenario, the user will have to award the new permission manually to OneTxPayments - but to get in to that state, they've had to do that anyway, so it has been deemed to be acceptable.

Placing these restrictions in the contracts means that (hopefully) helpful messages will bubble-up to the frontend with no extra work from them.

This PR also changes the upgrade tests that were in #1150 to use the deploy-old-colony-version functionality that I had to add here to test this functionality as I wanted.


test/contracts-network/colony.js
```diff
@@ -419,22 +430,87 @@ contract("Colony", (accounts) => {
...
+
+      OldInterface = truffleContract({
+        abi: formattedOldABI,
+      });
 
-      const tasksPaymentsContract = await TasksPayments.new();
-      await colonyResolver.register("makeTask(uint256,uint256,bytes32,uint256,uint256,uint256)", tasksPaymentsContract.address);
-      await colonyResolver.register("addPayment(uint256,uint256,address,address,uint256,uint256,uint256)", tasksPaymentsContract.address);
-      tasksPayments = await TasksPayments.at(colony.address);
+      OldInterface.setProvider(web3.currentProvider);
```
kronosapiens, line 465: I don't really understand what all of this stuff is. It looks like it should be encapsulated in a function kept in `scripts/deployOldUpgradeableVersion.js` and called with context-specific arguemnts.

test/contracts-network/colony.js
```diff
@@ -419,22 +430,87 @@ contract("Colony", (accounts) => {
...
     });
 
-    it("should be able to query for a task", async () => {
-      await tasksPayments.makeTask(1, UINT256_MAX, SPECIFICATION_HASH, 1, GLOBAL_SKILL_ID, 0, { from: USER0 });
+    beforeEach(async () => {
+      await colonyNetwork.createColony(token.address, 13, "", "");
```
kronosapiens, line 469: Have you thought about extending `setupColony` to take a version?

test/contracts-network/colony.js
```diff
@@ -419,22 +430,87 @@ contract("Colony", (accounts) => {
...
+      const action2 = await encodeTxData(authority, "setRoleCapability", [
+        ADMINISTRATION_ROLE,
+        colony.address,
+        web3.utils.soliditySha3("addPayment(uint256,uint256,address,address,uint256,uint256,uint256)").slice(0, 10),
+        true,
+      ]);
+
+      await colony.makeArbitraryTransactions([colonyAuthorityAddress, colonyAuthorityAddress], [action1, action2], true);
+
+      oldColony = await OldInterface.at(colony.address);
```
kronosapiens, line 499: Ditto all of this stuff, should be in a function in `scripts/deployOldUpgradeableVersion.js`

test/contracts-network/colony.js
area: Yeah, good call. I'm now returning the old interface out of the setup script.

test/contracts-network/colony.js
area: Thinking about this some more, I've realised we can't actually remove old functions from `ColonyAuthority`, otherwise we break new deployments of old version of colonies (which, fundamentally, is what this extra setup was fixing). This is something that we should fix in the future for gas reasons, but for now I think we just restore the old functions in ColonyAuthority. Note that this will apply to #1173 as well.

test/extensions/one-tx-payment.js
```diff
@@ -528,4 +529,73 @@ contract("One transaction payments", (accounts) => {
...
+          // eslint-disable-next-line max-len
+          "Colony,ColonyDomains,ColonyExpenditure,ColonyFunding,ColonyPayment,ColonyRewards,ColonyRoles,ColonyTask,ContractRecovery,ColonyArbitraryTransaction",
+        ],
+        "glwss4",
+        colonyNetwork,
+      );
+    });
+
+    beforeEach(async () => {
+      await colonyNetwork.createColony(token.address, 13, "", "");
```
kronosapiens, line 551: Let's use `setupColony` here as well

helpers/test-data-generator.js
```diff
@@ -289,8 +289,8 @@ exports.setupRandomColony = async function setupRandomColony(colonyNetwork, lock
   return { colony, token };
 };
 
-exports.setupColony = async function setupColony(colonyNetwork, tokenAddress) {
-  const { logs } = await colonyNetwork.createColony(tokenAddress, 0, "", "");
+exports.setupColony = async function setupColony(colonyNetwork, tokenAddress, version = 0) {
+  const { logs } = await colonyNetwork.createColony(tokenAddress, version, "", "");
```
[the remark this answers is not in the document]
area: Haven't used a `try/catch` but hopefully you're okay with the logic I've introduced.

kronosapiens (approved):

PR #1183, opened 2023-11-06
Temporarily remove skill deprecation

kronosapiens:


area:
Wanted to keep this as simple as possible, so pretending that #1173 isn't happening (which hopefully will also make the rebase easier) - i.e. still allowing global deprecations, which I don't believe there's an issue with.

I've changed how local deprecation is disabled - I think it's important the functions are still on `ColonyNetwork`, so that when old colonies try to do it, they get a meaningful error message.

kronosapiens:
@area looks good to me

area (approved):

PR #1222, opened 2024-03-06
Finalize motions changes to make estimateGas better

### LINKED ISSUE #1221: Reengineer finalizeMotion to be more intuitive to use with estimateGas
Currently thinking that `finalizeMotion` should never allow a failing action and `finalizeMotionWithFailingAction` should.

area:
Intended to close #1221


kronosapiens (approved):
lgtm

PR #1229, opened 2024-03-22
Unpin hardhat

### LINKED ISSUE #1228: Unblock Hardhat version pinning
Currently Hardhat is pinned to v2.19 due to a dependency introduced in later versions (`edr`) which breaks our build. Once #1196 is merged, we should sort out this issue so we aren't stuck on an increasingly old version.

kronosapiens:
Closes #1228


kronosapiens:
Closed in favor of #1244

PR #1234, opened 2024-03-30
Delete username/password if not set for miner

area:
These being (presumably) set in the headers in ethers requests to empty strings, rather than not being present, was causing 401 errors when making requests to Chainstack.


area:
Was included in #1241

PR #1235, opened 2024-03-30
Try building docker images on CI

area:
We keep having issues with `package-lock.json`, this helps guards against this a bit. Note we don't do anything with the resulting images, we just check they build okay.

### DIFF
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -23,7 +23,6 @@ const {
 
 const IColony = artifacts.require("IColony");
 const IMetaColony = artifacts.require("IMetaColony");
-const IColonyNetwork = artifacts.require("IColonyNetwork");
 const ITokenLocking = artifacts.require("ITokenLocking");
 const Token = artifacts.require("Token");
 const IReputationMiningCycle = artifacts.require("IReputationMiningCycle");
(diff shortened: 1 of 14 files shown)


area:
Looks like we weren't out of the woods with spotty tests, as a build here [threw](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6511/workflows/b31db264-230c-4554-917b-160a99a2cf74/jobs/43226) with the `other side closed` error I was seeing a lot before [monkeypatching `truffle-contract`](https://github.com/JoinColony/colonyNetwork/pull/1237#issuecomment-2085541186).

We thought we solved that when we stopped the node being spammed, so I've taken a similar approach here and replaced our use of `JsonRPCProvider` from ethers with [`StaticJsonRPCProvider`](https://docs.ethers.org/v5/api/providers/jsonrpc-provider/#StaticJsonRpcProvider) which will approximately halve the number of requests we're making again.

helpers/test-helper.js
```diff
@@ -1144,8 +1144,6 @@ exports.getWaitForNSubmissionsPromise = async function getWaitForNSubmissionsPro
       if (nSubmissions.toNumber() >= n) {
         event.removeListener();
         resolve();
-      } else {
-        await exports.mineBlock();
```
kronosapiens, line 1148: Where did this go?

packages/reputation-miner/Dockerfile
```diff
@@ -1,12 +1,10 @@
-FROM node:14-bullseye
+FROM node:20.11.0-buster
 RUN apt-get update || : && apt-get install python -y
 COPY ./packages ./packages
 COPY ./package.json ./
 COPY ./package-lock.json ./
 COPY ./artifacts ./artifacts
 RUN git config --global url."https://".insteadOf "ssh://git@"
 RUN npm ci
-RUN cd ./packages/reputation-miner/ && npm ci
```
kronosapiens, line 9: Why is this no longer necessary?

kronosapiens (commented):
Can you summarize the changes made to `client-auto-functionality`?

helpers/test-helper.js
area: Well, it shouldn't be necessary, and extra (quasi-unexpected) blocks being mined makes it harder to pin down what exactly is going on with these tests. It's possible that removing this will have made other tests spottier, though.

packages/reputation-miner/Dockerfile
area: In the same way that the `package-lock` files in the packages were unnecessary ( 158e1464145844ed73dd1a88c653d655ca8a130b), there's no point in running `npm ci` inside the package directories separately.

area:
In terms of the changes to the tests, in these two tests we see them [fail](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6536/workflows/9d8f134b-f003-47f8-979e-deca8f26539f/jobs/43548) [intermittently](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6536/workflows/9d8f134b-f003-47f8-979e-deca8f26539f/jobs/43550) waiting for events on-chain that happen. That suggests that the event is firing before the promise is set up; as a result, I'm trying to rework these tests to use more precise timestamps so that we don't accidentally end up letting a client confirm a hash before we set up the promise for it.

I realise we could just move when we set up the promise, but I think that's just patching over the underlying problem, which I'd prefer to resolve in a manner that (hopefully) means that when writing tests like this in the future we don't stumble across this issue.

kronosapiens:
Ok, this looks good to me. Let me know when you're ready for another review.

area:
These two most recent commits:

* Set `processingDelay` to 1 everywhere in the `auto` test suite. We were already doing this a bit, and it makes sense to do it everywhere. This only exists to protect against block reorgs, which aren't happening in these tests, and it means we aren't editing the variable on the miner tracking it on the fly, which I could believe could lead to race conditions
* Changes how we set up the promises in `getMiningCycleCompletePromise` and `getWaitForNSubmissionsPromise`. I'm all but certain the timestamps are correct now, so I'm left to conclude that even if the promise is being set up, the handler is not before the event is emitted on chain. This change makes sure that the handler is added (at least, if we can trust what ethers is telling us)

I think it's unlikely the first one was much of a factor, but I'm quite optimistic that the second one here is what we've been chasing down :crossed_fingers:

area:
I've reworked the approach to `getWaitForNSubmissionsPromise` and `getMiningCycleCompletePromise` to not use the event emitters at all, but to query for the events post-hoc. I think this should be much more reliable, without jumping through hoops making sure that the event handler is set up, which would have been needed to fix my original approach. It does result in more requests to the node, but I think that's unavoidable

kronosapiens (approved):

PR #1244, opened 2024-04-24
Bump hardhat to 2.22

kronosapiens:
Closes https://github.com/JoinColony/colonyNetwork/issues/1228


scripts/setup-bridging-contracts.js
```diff
@@ -145,8 +146,8 @@ async function setupBridging(homeRpcUrl, foreignRpcUrl) {
   return { gnosisSafe, bridgeMonitor, zodiacBridge, homeBridge, foreignBridge, homeColonyBridge, foreignColonyBridge };
 }
 
-async function getSig(provider, account, dataHash) {
-  const sig = await provider.send("eth_sign", [account, dataHash]);
```
area, line 149: This has been needed to due to an issue being tracked here: https://github.com/NomicFoundation/edr/issues/455

While signing messages in the way you've done here solves the issue for our codebase, `eth_sign` is still not callable on a Hardhat node, which is required for the CDApp development environment. As a result, this PR will need to wait until Hardhat has fixed the problem, and which point I would revert the changes related to this to catch similar issues in the future (which the CDApp can't work around).

area (commented):
Sadly, can't approve for reasons beyond your control.

area:
Just noting that the `eth_sign` issue has now been fixed on the EDR side, so once the next version is released, we should be able to move to the latest hardhat.

kronosapiens:
Closing in favor of #1254

PR #1248, opened 2024-05-01
Update selfdestruct behavior

kronosapiens:
With the move to Hardhat 2.22 and the Dencun hardfork, the behavior of `selfdestruct` has changed. We should update our contracts accordingly.


area:
Presumably, we should also actually not call `selfdestruct`, and mimic the behaviour it had (i.e. send the current ether balance of `this` to the caller)

kronosapiens:
Note: do any currently deployed contracts depend on the destruction assumption?

kronosapiens:
Closed in favor of #1254

PR #1249, opened 2024-05-02
Exploration of PNPM

area:
For whatever reason ~~- possibly with the move to Node 20 -~~ we're having real issues with a single lockfile being installable on Linux and OSX. This comes down to `fsevents` being only needed on OSX (and therefore optional elsewhere) and npm having an issue with that - see https://github.com/npm/cli/issues/4828 for the ongoing issue related to this.

I've explored what it would look like if we moved to [pnpm](https://pnpm.io/). As well as the lockfile issue being resolved (I think), a couple of nice side benefits:

* Faster install times, which has been an issue with the CDapp dev environment
* A 'proper' way to patch `@nomiclabs/truffle-contract` rather than a postinstall call to `sed`

`npm i -g pnpm@8` to install, and then just use `pnpm` instead of `npm` to call everything is all that should be necessary.

~NB based on top of `maint/docker-file-builds`.~


area:
As an addendum, after speaking to Chris, the resident Node guru:

* He uses pnpm for ColonyJS, and is thinking of moving the dapp to it, so supports this 'wholeheartedly'.
* He uses v8 there (as I do here), and will only look to move to v9 once it is more mature

kronosapiens:
I dealt with this issue for a while when we were on node@14. I went down a few rabbit holes, the only fix which consistently worked was to manually edit my `package-lock.json` to add `"linux"` to the `os` field:

```
      "os": [
        "darwin",
        "linux" // Manually add this
      ],
```

This behavior could conceivably be added to a post-commit hook using a `sed` command.

package.json
```diff
@@ -135,13 +136,12 @@
     "web3-utils": "^1.7.3"
   },
   "optionalDependencies": {
-    "fsevents": "^2.3.2"
+    "fsevents": "2.3.2"
```
kronosapiens, line 139: Why the pin?

kronosapiens (commented):
Looks reasonable to me

package.json
area: I think this is a remnant of me trying things out. Removed, I think with no consequences.

area:
Don't want this merged yet (requires https://github.com/JoinColony/colonyCDapp/pull/2316), but interested in a review.

kronosapiens (commented):
Looks reasonable, the patching functionality is quite interesting

kronosapiens (dismissed):

area:
The versioning check script is broken, but not as a result of this PR so don't intend for it to hold it up.

kronosapiens (approved):

PR #1250, opened 2024-05-08
Streaming payments only support a single stream

area:
NB Based on maint/pnpm.

Multiple streams in a single streaming payment are no longer desired on the frontend, so taking the opportunity to simplify the contracts. This will also allow accurate tracking of the number of pending payments, allowing the 'prevent uninstall if pending payments' feature.

I've taken advantage of `finishUpgrade` to make sure someone can't upgrade from v4 to v5, which we are declaring is impossible due to the storage layout changes.

### DIFF
--- a/.storage-layouts-normalized/contracts/extensions/StreamingPayments.sol:StreamingPayments.json
+++ b/.storage-layouts-normalized/contracts/extensions/StreamingPayments.sol:StreamingPayments.json
@@ -511,64 +511,42 @@
                 "label": "uint256",
                 "numberOfBytes": "32"
               }
+            },
+            {
+              "contract": "contracts/extensions/StreamingPayments.sol:StreamingPayments",
+              "label": "token",
+              "offset": 0,
+              "slot": "5",
+              "type": {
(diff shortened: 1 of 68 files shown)


kronosapiens (commented):
Overall looks good. I still don't love all the extra logic about pseudo claims, etc which we added to allow admins to change payout rates, but that's not what this PR is about.

area:
NB Now based on #1262

kronosapiens:
What's with all the storage file changes? Does the PR need a rebase?

area:
It does not need rebasing. The one change in `.storage-layouts-normalized` is expected - we have changed a struct, and removed a mapping. The changes in `.storage-layouts` to other files are because of changes to the code since the last time a storage slot was changed, and so AST IDs have changed in the rest of the compiled contracts. #1258 attempted to go through this in more detail, but we can have a call to make sure we both know what's going on, if you'd prefer.

kronosapiens (commented):
Lgtm

kronosapiens (approved):

PR #1251, opened 2024-05-08
Prevent StreamingPayment uninstallation if pending streams remain

area:
The main feature here is to prevent uninstallation of the extension in the event that there are some streaming payments unclaimed. It also:

* Reworks some calculations to prevent over/underflows. This is somewhat opinionated in that the intention is to 'cap' the calculations
* Allows editing of the interval, which is a requirement. While in an ideal world, editing the interval shouldn't be necessary as you could edit the amount, rounding errors meant that this should be editable as well. This is possible as we have moved to single-stream payments.

### DIFF
--- a/contracts/extensions/StreamingPayments.sol
+++ b/contracts/extensions/StreamingPayments.sol
@@ -389,7 +389,20 @@ contract StreamingPayments is ColonyExtensionMeta {
     if (durationToClaim == 0) {
       return 0;
     }
-    return wmul(streamingPayment.amount, wdiv(durationToClaim, streamingPayment.interval));
+
+    // Guard against overflow in wdiv
+    if (durationToClaim > type(uint256).max / WAD) {
+      durationToClaim = type(uint256).max / WAD;
+    }
+
+    uint256 intervalsToClaimAsWad = wdiv(durationToClaim, streamingPayment.interval);
+
+    // Guard against overflow in wmul
+    if (streamingPayment.amount > type(uint256).max / intervalsToClaimAsWad) {
+      return type(uint256).max;
+    }
+
+    return wmul(streamingPayment.amount, intervalsToClaimAsWad);
   }
 
   // Internal
--- a/contracts/extensions/StreamingPayments.sol
+++ b/contracts/extensions/StreamingPayments.sol
@@ -178,7 +178,7 @@ contract StreamingPayments is ColonyExtensionMeta {
     );
 
     if (getAmountClaimableLifetime(numStreamingPayments) > 0) {
-      nUnresolvedStreamingPayments++;
+      nUnresolvedStreamingPayments += 1;
     }
 
     emit StreamingPaymentCreated(msgSender(), numStreamingPayments);
--- a/contracts/extensions/StreamingPayments.sol
+++ b/contracts/extensions/StreamingPayments.sol
@@ -60,6 +60,7 @@ contract StreamingPayments is ColonyExtensionMeta {
 
   uint256 numStreamingPayments;
   mapping(uint256 => StreamingPayment) streamingPayments;
+  uint256 nUnresolvedStreamingPayments;
 
   // Modifiers
 
@@ -118,7 +119,11 @@ contract StreamingPayments is ColonyExtensionMeta {
     revert("streaming-payments-not-upgradeable-from-v4");
   }
 
-  // Public
+  /// @notice Called when uninstalling the extension
+  function uninstall() public override auth {
+    require(nUnresolvedStreamingPayments == 0, "streaming-payments-unresolved-payments");
+    super.uninstall();
+  }
 
   /// @notice Creates a new streaming payment
   /// @param _fundingPermissionDomainId The domain in which the caller holds the funding permission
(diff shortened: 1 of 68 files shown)


contracts/extensions/StreamingPayments.sol
```diff
@@ -184,6 +191,10 @@ contract StreamingPayments is ColonyExtensionMeta {
       0
     );
 
+    if (getAmountClaimableLifetime(numStreamingPayments) > 0) {
```
kronosapiens, line 194: Is this check necessary? Why would we allow a streaming payment to be created which was worth nothing?

contracts/extensions/StreamingPayments.sol
```diff
@@ -184,6 +191,10 @@ contract StreamingPayments is ColonyExtensionMeta {
...
+      nUnresolvedStreamingPayments++;
```
kronosapiens, line 195: Make this `+= 1` for consistency

kronosapiens (commented):
This is a lot of new code just to implement an uninstall check, but so it goes I guess

contracts/extensions/StreamingPayments.sol
area: We already allowed it, so I had to accommodate it. The alternative would be adding a check instead of this, so it feels like much of a wash to me. In addition, there's nothing stopping you editing a stream you create down to being worth nothing, so preventing creation feels quite arbitrary?

contracts/extensions/StreamingPayments.sol
```diff
@@ -328,8 +380,23 @@ contract StreamingPayments is ColonyExtensionMeta {
...
     streamingPayment.endTime = _endTime;
 
+    uint256 newLifetimeClaimable = getAmountClaimableLifetime(_id);
+
+    // Unlike when we're setting start time, we need to compare to pseudoAmountClaimedFromStart
+    // in order to determine if the payment is resolved or not.
+    bool wasResolved = streamingPayment.pseudoAmountClaimedFromStart >= oldLifetimeClaimable;
+    bool isResolved = streamingPayment.pseudoAmountClaimedFromStart >= newLifetimeClaimable;
+
+    if (wasResolved && !isResolved) {
```
kronosapiens, line 394: This implementation looks similar to the one above, is there a way to abstract this out somehow?

contracts/extensions/StreamingPayments.sol
```diff
@@ -420,6 +506,34 @@ contract StreamingPayments is ColonyExtensionMeta {
...
   }
 
+  function getAmountClaimableInTime(
+    uint256 _id,
+    uint256 _from,
+    uint256 _until
+  ) private view returns (uint256) {
+    StreamingPayment storage streamingPayment = streamingPayments[_id];
+
+    if (_from >= _until) {
```
kronosapiens, line 516: Why not `require(_from < _until)`?

kronosapiens (commented):
You mention that changing the interval is necessary to avoid rounding errors, but realistically what are the errors we're talking about? A few wei? I would be in favor of going the opposite direction and hard-coding the interval as 24 hours. We can provide a UI which converts different intervals (hour, week month, etc) if needed.

contracts/extensions/StreamingPayments.sol
area: In general, if able to return a valid value from such a function, I think it's generally more correct to do so, rather than be opinionated about e.g. 'zero is an error', and let functions using it make that decision.

In this case specifically, if `getAmountClaimableInTime` errored when it would return 0, then you wouldn't be able to e.g. `setStartTime` if you had created the stream in a way that had a lifetime payout of zero.

area:
> You mention that changing the interval is necessary to avoid rounding errors, but realistically what are the errors we're talking about? A few wei? I would be in favor of going the opposite direction and hard-coding the interval as 24 hours. We can provide a UI which converts different intervals (hour, week month, etc) if needed.

If tokens don't have, or have set `decimals()` to 0, then 'A few wei' could easily have significant value.

In terms of a hard-coded interval, even ignore editing existing streams: if we had it at 24 hours, we couldn't e.g. pay out 1 token that had `decimals()` as 0 every 16 hours.

contracts/extensions/StreamingPayments.sol
area: Taken a shot at in a separate commit so you can see what you think :smile:

contracts/extensions/StreamingPayments.sol
kronosapiens: I hear that, although for the record we often treat 0 values as errors elsewhere in the codebase.

area: What I said above is true, but I misunderstood what you were actually suggesting here, which is only requiring that `require(_from < _until)` and still returning 0 in other, 'healthy' states. 

We could have the require, but we'd still need a special case in the event of equality... if you feel strongly about it, we can put in the require, but I'm still happy with this as-is.

> I hear that, although for the record we often treat 0 values as errors elsewhere in the codebase.

Your statement is true, but I'm not sure this is a great argument... are you suggesting that, for consistency, `balance(address)` on our token should error if the address in question has no tokens?

kronosapiens: My preference would be for the opposite, to allow for 0 arguments wherever they are no explicitly invalid values.

kronosapiens (approved):
Looks good to me

kronosapiens (approved):

PR #1252, opened 2024-05-13
Fix setExpenditureValues with metatransactions

area:
Metatransactions do not work as-is with, specifically, `setExpenditureValues`.

This is because the assumption we made when [implementing metatransactions](https://github.com/JoinColony/colonyNetwork/pull/962) to accommodate our self-calling functions is not true for one particular call to `setExpenditurePayouts` we make after introducing [setExpenditureValues](https://github.com/JoinColony/colonyNetwork/pull/1045). Specifically, the assumption that if one of our contracts is calling itself, there are no arguments in the function call (represented by the `index >= 52` [here](https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/common/MetaTransactionMsgSender.sol#L11)).

This PR makes one attempt to fix it, but there could be others I've not considered - I am very much open to other suggestions if there is a better one we can arrive at. The solution is similar to how we treat metatransactions in multicalls, however.

As with everything that touches metatransactions, a particularly careful review would be appropriate, as the failure case for metatransactions is catastrophic.

### DIFF
--- a/contracts/common/MetaTransactionMsgSender.sol
+++ b/contracts/common/MetaTransactionMsgSender.sol
@@ -5,6 +5,7 @@ import { DSMath } from "../../lib/dappsys/math.sol";
 
 abstract contract MetaTransactionMsgSender is DSMath {
   bytes32 constant METATRANSACTION_FLAG = keccak256("METATRANSACTION");
+  uint256 METATRANSACTION_DATA_MIN_LENGTH = METATRANSACTION_FLAG.length + 20; // Where 20 is the length of an address in bytes
 
   function msgSender() internal view returns (address payable sender) {
     if (isMetatransaction()) {
@@ -21,7 +22,7 @@ abstract contract MetaTransactionMsgSender is DSMath {
 
   function isMetatransaction() internal view returns (bool) {
     uint256 index = msg.data.length;
-    if (msg.sender == address(this) && index >= 52) {
+    if (msg.sender == address(this) && index >= METATRANSACTION_DATA_MIN_LENGTH) {
       bytes memory array = msg.data;
       bytes32 flag;
       assembly {
--- a/contracts/common/MetaTransactionMsgSender.sol
+++ b/contracts/common/MetaTransactionMsgSender.sol
@@ -7,16 +7,9 @@ abstract contract MetaTransactionMsgSender is DSMath {
   bytes32 constant METATRANSACTION_FLAG = keccak256("METATRANSACTION");
 
   function msgSender() internal view returns (address payable sender) {
-    uint256 index = msg.data.length;
-    if (msg.sender == address(this) && index >= 52) {
+    if (isMetatransaction()) {
       bytes memory array = msg.data;
-      bytes32 flag;
-      assembly {
-        flag := mload(add(array, sub(index, 20)))
-      }
-      if (flag != METATRANSACTION_FLAG) {
-        return payable(msg.sender);
-      }
+      uint256 index = msg.data.length;
       assembly {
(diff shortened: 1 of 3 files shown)


kronosapiens:
@area can you clarify what is meant by "there are no arguments in the function call"? Do you mean that there are no arguments _after_ the metatransaction flag? Or no arguments entirely?

kronosapiens:
Also, in a multicall world, could we not dispense with `setExpenditureValues` entirely and simply use multicall to perform multiple expenditure edits in one transaction?

contracts/common/MetaTransactionMsgSender.sol
```diff
@@ -25,4 +25,17 @@ abstract contract MetaTransactionMsgSender is DSMath {
       return payable(msg.sender);
     }
   }
+
+  function isMetatransaction() internal view returns (bool) {
```
kronosapiens, line 29: Is it possible to use this new `isMetatransaction` function as part of `msgSender()`? Otherwise there is significant code duplication in this fiel.

area: That's fair; please review that refactorisation carefully though :smile:

area:
> @area can you clarify what is meant by "there are no arguments in the function call"? Do you mean that there are no arguments _after_ the metatransaction flag? Or no arguments entirely?

No arguments to the function that the contract is calling on itself. In the original implementation, `upgrade()` was the function we were thinking about that the contract would call on itself. If `msg.data` is less than 52, the call cannot be a metatransaction (because a metatransaction will have `METATRANSACTION_FLAG` and the caller's address appended, for a total of 52 bytes), so for `upgrade` we were able to bypass this logic with the check against msg.data's length.

Unfortunately, we now have a scenario where `msg.data` might be longer than 52 bytes, `msg.sender` might be `self`, and it might or might not be a metatransaction.

> Also, in a multicall world, could we not dispense with setExpenditureValues entirely and simply use multicall to perform multiple expenditure edits in one transaction?

This is true, but not how things have been implemented on the front-end.

contracts/common/MetaTransactionMsgSender.sol
```diff
@@ -25,4 +18,17 @@ abstract contract MetaTransactionMsgSender is DSMath {
...
+    uint256 index = msg.data.length;
+    if (msg.sender == address(this) && index >= 52) {
```
kronosapiens, line 24: I would not be opposed to introducing `uint256 constant METATX_MIN_LENGTH = 52` here for further clarity

kronosapiens (commented):
Looks good to me. I don't see significant risk to the metatransaction functionality, as most of the shuffling is taking place in `ColonyExpenditure`.

I would like to deprecate `setExpenditureValues` in the future and shift to multicall for this sort of behavior, though, as ultimately this is sort of a hack built on top of a previous hack.

kronosapiens (dismissed):

kronosapiens (dismissed):

area:
Deliberately haven't squashed the last commit to serve as a reminder to us how easily critical errors can be made, and how innocuous they can look.

area:
> Also, in a multicall world, could we not dispense with setExpenditureValues entirely and simply use multicall to perform multiple expenditure edits in one transaction?

Thanks you @jakubcolony's efforts, this is now how things have been implemented on the frontend, so maybe this PR should now be replaced by one to remove `setExpenditureValues`?

kronosapiens:
Wonderful! I wouldn't mind hanging on to the new constant though, especially after risking death to introduce it.

area:
Superseded by  #1259

PR #1255, opened 2024-05-16
Add some hardhat niceties

area:
After getting lucky and being bailed out by our existing checks in #1252, I went looking for a replacement for the smoke tests removed in #1196. 

I found [this](https://www.npmjs.com/package/hardhat-storage-layout) plugin, which I have added.  I've also added two checks in the circle pipeline; the first checks to see if anything has gone badly wrong (i.e. storage slots have been moved). The second checks whether the storage layouts commited in the repository need updating. This could happen if new storage slots have been added, or if the code has been changed resulting in the AST node IDs being changed. This latter possibility is a bit annoying, but would rather be vaguely annoyed than accidentally ruin our storage layouts again!

While I was looking at hardhat plugins, I've also added [https://github.com/solidstate-network/hardhat-4byte-uploader](this one), which uploads all of our function signatures to https://www.4byte.directory/. I've had a lot of questions recently about what a particular transaction is doing and this will make it much easier for me to give a fast answer. Again, I've added doing that upload as part of the Circle build.


package.json
```diff
@@ -116,8 +117,8 @@
     "express": "^4.18.2",
     "find-in-files": "^0.5.0",
     "ganache": "7.9.2",
-    "hardhat": "~2.19.5",
```
kronosapiens, line 119: What's this about?

kronosapiens (commented):
Looks good, but where did hardhat go?

PR #1240, opened 2024-04-11
Add additional events for StreamingPayments

area:
Additional events after discussing with Jakub about requirements for the implementation of streaming payments.

Currently a draft, not to be merged, until we've straightened everything out with the Arbitrum deployment branches, but will unblock development on the CDApp side until then.

PR #1258, opened 2024-05-27
Add more robust storage slot checking

area:
We introduced [hardhat-storage-layout-changes](https://github.com/zemse/hardhat-storage-layout-changes) in #1255, but the way I've set it up is very sensitive.

This is because a lot of the information recorded in the committed files is the `astid`. This the ID of the relevant node in the [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) which can (and does) change based on the particulars of the source code; in particular, it appears to be a function of the codebase as a whole (and then possibly a function of the order in which things are compiled).

This was unexpected enough it seemed at first glance to those unfamiliar with what was going on that it seemed to be broken, so I thought I'd circle back and make it behave a bit more intuitively.

This fundamentally means removing all `astId`s from the output that is used to detect changes. These appear in two places in the [JSON Output](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#json-output) from solidity used by the hardhat plugin. The first is the property `astId`, which is easy to traverse the JSON for and remove. More subtly, however, they also appear in the types of more complicated structures (mappings, arrays). By recursively replacing the type names with the type definitions (which are helpfully also provided in the JSON output), we can eliminate all occurrences of `astId`s and end up with a structure that should be sensitive to what we actually want.

The implementation of this is not going to winning me any prizes (in particular, repeatedly calling the `normalize` function  until there are no more changes is not performant, but it's easy to write), but it does the job.

Strictly, we don't require both `.storage-layouts` directories here, but I opted to keep both. The first check in CI (which uses the plugin, and the original directory) will fail if storage slots are moved. The second check in CI (using my script and the normalized files) will fail if new variables have been added but not committed to the storage layout directories. It would also fail if storage slots had been moved, but I believe having the distinction between those two failure cases is worthwhile.


area:
[Example build of](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6634/workflows/3ba78361-aaab-4e3d-9222-d1112696617d/jobs/45010) #1240 based on this branch, as an example.

PR #1259, opened 2024-05-27
Remove setExpenditureValues

area:
The frontend no longer relies on `setExpenditureValues` (using `multicall`, instead which didn't exist when we originally implemented that function) so as agreed, we can remove `setExpenditureValues` instead of fixing its interaction with metatransactions.

This replaces #1252 , and is currently based on #1258 which should be reviewed and merged first.


area:
As requested in #1525, this also brings over the constant we introduced to make the assembly done with metatransactions easier to understand.

PR #1261, opened 2024-05-29
Use overrides to fix wrong dependency URLs

chmanie:
`eth-sig-util` in an old version (`1.4.2`) uses an ssh URL for a dependency in its package.json which causes `pnpm install` to fail.

Before, this was fixed using a change in git config, but that doesn't seem to work in all environments (e.g. if users do not have ssh setup, it will fail for them without the aforementioned git config, or, as experienced by me, on Github actions).

This now sets the http git URL to the commit that was referenced in the SSH link in the pnpm overrides.


area (approved):
Approving, with the agreed proviso that if this issue with URLs crops up again in this repo we are reverting to this fix (which is general and works for this repo) and other repositories have to solve it for themselves.

PR #1264, opened 2024-06-11
Update versioning check scripts

area:
One of the pieces of cruft we hadn't cleaned up from the Arbitrum release was that we hadn't actually released `imwss` on GitHub, which meant that our versioning check was doing so against `hmwss`. The two main consequences of that were:

* Version bumping scripts were being overly generous (i.e. changes since `imwss` were passed if the version was bumped in `hmwss`
* With the release, our checks broke because the latest release has now moved from `truffle` to `hardhat`

As a result of 1, we missed a required version bump in #1259, but it doesn't appear any others.

This PR intends to fix these issues. Arguably the most meaningful change is an indirect result of the release though, which is how I'm treating upgrades in tests. For historical behaviour, I'm assuming that doing an upgrade from e.g. 12 to 13 to 16 is sufficient to test, rather than having to deploy 14 and 15 as well. This seems to work for all of our tests so far; it will need to be revisited if that assumption no longer holds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,6 +130,10 @@ jobs:
           name: "Upload function selectors to 4byte.directory"
           command: npx hardhat upload-selectors
       - run:
+          name: "Install nvm"
+          command: |
+            curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
+      - run:
           name: "Checking contract versions"
           command: pnpm run check:versioning
   check-docker-files-build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,11 @@ step_setup_slither: &step_setup_slither
         pip install solc-select
         solc-select install 0.8.25
         solc-select use 0.8.25
+step_install_nvm: &step_install_nvm
+    run:
+      name: "Install nvm"
+      command: |
+        curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
 
 jobs:
   unit-test-reputation:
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -52,7 +52,13 @@ step_setup_slither: &step_setup_slither
         pip install solc-select
         solc-select install 0.8.25
         solc-select use 0.8.25
-
+step_install_nvm: &step_install_nvm
```
kronosapiens, line 55: I wouldn't be opposed to adding a `step_install_lsof` as well, since we do it frequently

helpers/test-helper.js
```diff
@@ -1272,13 +1274,25 @@ exports.getMultichainSkillId = function getMultichainSkillId(chainId, skillId) {
...
 };
 
-exports.upgradeColonyTo = async function (colony, _version) {
-  const version = new BN(_version);
-  let currentVersion = await colony.version();
-  while (currentVersion.ltn(version)) {
-    await colony.upgrade(currentVersion.addn(1));
-    currentVersion = await colony.version();
-  }
+exports.upgradeColonyOnceThenToLatest = async function (colony) {
```
kronosapiens, line 1277: I would elaborate more here. It seems like the idea is that we want to test upgrading from a specific version, since that version had some bug or something we want to make sure we recover from. But aside from the specific version we want to test, we don't care about other versions, so we can just "fast forward" through them.

scripts/versioningCheck.sh
```diff
@@ -6,15 +6,34 @@ CURRENT_BRANCH=`git branch --show-current`
...
+else
+  echo "No node manager found"
+  exit 1
+fi
+
 # Compile release
-npm ci --force && npx truffle compile
-rm -rf build-$LATEST_RELEASE || true
-mv build build-$LATEST_RELEASE
+$NODE_MANAGER install
```
kronosapiens, line 25: Do all node managers have the same interface?

area: I have tested it with `nvm` (on Circle) and `fnm` (locally). I haven't tested with `n`, but the docs seemed to suggest that was the case.

kronosapiens (approved):

PR #1270, opened 2024-06-20
Introduce hardhat_setStorageAt

kronosapiens:
Replace the custom storage editing via `ContractEditing` with calls to ` hardhat_setStorageAt`. We are still dependent on `ContractEditing` functionality in the Ganache-based tests, as the special opcode is not available.

### DIFF
--- a/helpers/test-helper.js
+++ b/helpers/test-helper.js
@@ -1,4 +1,4 @@
-/* globals artifacts, hre */
+/* globals artifacts */
 
 const fs = require("fs");
 const shortid = require("shortid");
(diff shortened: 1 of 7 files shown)


area:
I've updated the provider to be more in line with what I wanted. Specifically, I do not want helpers going forward to make assumptions about the network the contract is on, and so by using the provider on the provided `contract` object, we know we're always setting the storage slot on the right chain.

kronosapiens:
@area your changes look good to me

area (approved):

PR #1271, opened 2024-06-26
Introduce Shell Colonies

kronosapiens:
WIP


kronosapiens:
IMG_4465

arrenv:
Commenting to clarify that we should not stick with the name "shells". Can we instead call it "proxy".

area:
Replaced by #1285

PR #1273, opened 2024-07-03
Improve test reliability

area:
After the merge of #1254, we got a big improvement in the time our tests take to run, but that improvement came at a cost of the suite as a whole becoming unreliable, as race conditions that could previously be ignored reared their ugly head. This PR fixes two tests, and seems to get us back to stability - four out of five builds on Circle of this branch were green, with the fifth failing due to a 502 HTTP error from Github during the build (see those builds [here](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork?branch=maint%2Fspotty-tests)), which hardly seems like something we can blame ourselves for.

Two tests were touched here. The first was the test around the oracle syncing with the reputation state correctly. With a faster hardhat, sometimes the `oracleUpdated` promise resolved at the first check, not hitting the `else` branch in the promise to `forwardTime(1, this)`. In that scenario, without that block being mined, the reputation miner client was still waiting for another block before updating, and so the oracle ended up ahead of the miner and the states did not match at the end of the test. 

The second test touched here was the classic "Losing a race shouldn't prevent a miner from continuing". I'm less certain about what was happening here, but I think it was to do with the miner still processing when a `mineBlock` was called, and then there not being another block to trigger the updates. This is the sort of thing that _should_ be handled by `endDoBlockChecks` calling `doBlockChecks` when another block has been seen, however, so not entirely sure what to make of this. Perhaps I've stumbled on to a correct solution for whatever the 'real' underlying problem is; I'd be interested in some insight on that front.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ job_python: &job_python
 step_setup_pnpm: &step_setup_pnpm #Only used in python image, as node image comes with it
   run:
     name: "Install pnpm"
-    command: npm i -g pnpm@8.14.1 #Same version as in cimg/node:20.11.0
+    command: sudo npm i -g pnpm@8.14.1 #Same version as in cimg/node:20.11.0
 step_save_cache: &step_save_cache
   save_cache:
     name: Save pnpm Package Cache
--- a/contracts/reputationMiningCycle/ReputationMiningCycle.sol
+++ b/contracts/reputationMiningCycle/ReputationMiningCycle.sol
@@ -466,7 +466,7 @@ contract ReputationMiningCycle is ReputationMiningCycleCommon {
     );
     require(
       responsePossible(
-        DisputeStages.ConfirmNewHash,
+        DisputeStages.ConfirmJRH,
         disputeRounds[_round][_index].lastResponseTimestamp
       ),
       "colony-reputation-mining-user-ineligible-to-respond"
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "@nomicfoundation/hardhat-network-helpers": "^1.0.10",
     "@nomiclabs/hardhat-ethers": "^2.2.3",
     "@nomiclabs/hardhat-truffle5": "^2.0.7",
-    "@nomiclabs/hardhat-web3": "^2.0.0",
     "@openzeppelin/contracts": "^4.9.6",
     "@solidity-parser/parser": "^0.14.2",
     "@solidstate/hardhat-4byte-uploader": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "@nomicfoundation/hardhat-network-helpers": "^1.0.10",
     "@nomiclabs/hardhat-ethers": "^2.2.3",
     "@nomiclabs/hardhat-truffle5": "^2.0.7",
-    "@nomiclabs/hardhat-waffle": "^2.0.6",
     "@nomiclabs/hardhat-web3": "^2.0.0",
     "@openzeppelin/contracts": "^4.9.6",
     "@solidity-parser/parser": "^0.14.2",
--- a/packages/package-utils/package.json
+++ b/packages/package-utils/package.json
@@ -12,6 +12,7 @@
     "axios": "^0.28.0",
     "discord.js": "^12.5.3",
     "ethers": "^5.6.9",
+    "exponential-backoff": "^3.1.0",
     "jsonfile": "^6.1.0",
     "slack": "^11.0.2"
   }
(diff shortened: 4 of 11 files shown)


area:
Jumped the gun here, back to draft for now.

area:
Still not done, but while I remember:

* When a miner is working out whether it can confirm a hash or otherwise respond to a dispute, it's possible for the cycle to be confirmed 'out from under it' - in that scenario, a call to `getMySubmissionRoundAndIndex` would return with `[-1, -1]`, which are not valid arguments to various other functions expecting `uint256`. As it was, the miner would crash; I've changed it so it will account for that possibility, which is entirely possible.
* ECONNRESET can happen if there's a network hiccup even with a good endpoint, so the miner should not crash in such a scenario and continue mining. Even if this happens during e.g. syncing, we should be able to recover.
* We were using the wrong stage in the `require` for `confirmJustificationRootHash` in the contracts, whereas the miningclients were using the correct one.

area:
screenshot-2024-07-08_09-36-17

Each of those jobs represents 13 runs of the `client-auto-functionality` tests, which were the only ones that were consistently failing. This collection of changes demonstrably represents a massive improvement in test reliability compared to the current develop, but even develop prior to #1254 being merged:

Previous develop: ~15% failure rate based on last 20 builds on `develop` before we merged #1254
Current develop: ~60% error rate based on 5 builds on `develop` and `hardhat20` branch (large error bars)
This branch:  ~1% failure rate based on 643 builds(!) (i.e. more than in that screenshot).

In addition to the changes described above, there are the following:

* If an error occurred in `doBlockChecks` before `repCycle` had been set, the error handler would fail, as it was trying to access `.address` on `undefined` when attempting to get the code at the repCycle address (NB that this behaviour is now outdated with the `selfdestruct` changes, and should be updated).

* Updated the 'forward to' timestamps in the dispute resolution test; if we go to a timestamp that's the exact start of a window, nothing is allowed to respond; I think in the past, when we were slower, the subsequent `mineBlocks` ended up causing the timestamp to tick forward anyway, but that no longer seems to be the case, consistently.
* Correctly deal with removing listeners in the auto-functionality tests; `removeListener` on an event is an `async` function, and so must be awaited, and I also added some additional cleanup where other listeners were being used.
* Reworked the logic in the test to check that we don't invalidate our own submission; it definitely had some at least suboptimal logic (`await forwardTimeTo(parseInt(goodEntry.lastResponseTimestamp, 10));` being a especially large offender).
* Wait for ReputationMinerClient2 to finish what it's doing in losing-race-continue test before killing it, so we do so cleanly
* Use the Ethers reputation mining cycle everywhere where I can in those tests.

The last point is related to some investigation that I did while trying to squeeze out that last 1% of flakiness, but the rest of the changes I explored aren't included here. That last 1% (aside from [random connectivity issues](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6885/workflows/288cb7d7-3d40-475a-b529-dcdafd64beae/jobs/53316)) seems to be due to [ECONNRESET](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6883/workflows/c7c1f46a-4318-471d-a2fb-930ab8b15cfc/jobs/53197) and [other side closed](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6883/workflows/0fb3319d-1079-4fb7-9a1e-66769359b83a/jobs/53116) errors, which are all coming out of the same package.

`pnpm why web3-core-requestmanager` says we have that package because of `@truffle/contract` and `@nomiclabs/hardhat-truffle5`, so it's possible that 'truly' moving to hardhat would solve this problem, and/or moving to Ethers contracts everywhere. It's also entirely possible that the fault lies with `hardhat node`; there are definitely issues with `undici` outstanding (see https://github.com/NomicFoundation/hardhat/issues/3136), which is where `other side closed` is coming from. In either case, I feel like they are outside the scope of what I want to get in to with this PR, and I haven't figured out why it's only these tests that seem to cause them.

The biggest thing I tried to solve this, not included here, was using [`bun`](https://bun.sh/) as the runner for these tests, but I still got the issues. But it's good to know that it works, at least, if something incredibly bad crops up in Node?

Unrelated to spottiness, this PR also includes: 

* It would seem that after an image update on Circle's side, our [`security` job fails](https://app.circleci.com/pipelines/github/JoinColony/colonyNetwork/6787/workflows/f0889a02-8b7b-4bfb-9cc0-70c61415da74/jobs/50706). I've fixed it by adding a `sudo` to the command to install `pnpm`, though obviously not wildly happy about that solution.
* We had `hardhat-waffle` installed for... no reason? So I've removed it. Same with `hardhat-web3`.

kronosapiens (approved):

PR #1275, opened 2024-07-12
Add a function to deploy IMWSS

area:
This has been requested for frontend dev in the CDapp. Unfortunately, not a simple copy-paste job, as this is the first version since we moved to `hardhat`.

I've not done a full impelmentation here - only as far as the frontend dev experience required.  The remaining implementation can come as-needed, as far as I am concerned, once we have tests or similar that require a version on hardhat.


area:
NB I've cherry-picked these in to #1144 so it can be used by people working on multisig, which is where the request came from.

kronosapiens (approved):
There may be some opportunities for refactoring but fine for now

PR #1282, opened 2024-07-30
Add nvm/fnm support for deployOldUpgradeableVersion

kronosapiens:
Does what it says on the tin

### DIFF
--- a/scripts/deployOldUpgradeableVersion.js
+++ b/scripts/deployOldUpgradeableVersion.js
@@ -1,6 +1,7 @@
 /* globals artifacts, hre */
 
 const fs = require("fs");
+const os = require("os");
 
 const Promise = require("bluebird");
 const exec = Promise.promisify(require("child_process").exec);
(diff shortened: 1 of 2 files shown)


scripts/deployOldUpgradeableVersion.js
```diff
@@ -315,17 +316,25 @@ module.exports.deployOldUpgradeableVersion = async (contractName, interfaceName,
     await exec(`rm -rf colonyNetwork-${versionTag}`);
     await exec(`git clone --depth 1 --branch ${versionTag} https://github.com/JoinColony/colonyNetwork.git colonyNetwork-${versionTag}`);
     await exec(`cd colonyNetwork-${versionTag} && git submodule update --init --recursive`);
-    const nodeVersion = fs.readFileSync(`colonyNetwork-${versionTag}/.nvmrc`);
-    await exec(`cd colonyNetwork-${versionTag} && npm install node@${nodeVersion}`);
+
+    let cmdBase;
+    if (os.arch() === "arm64") {
+      const nodeVersion = fs.readFileSync(`colonyNetwork-${versionTag}/.nvmrc`).toString().trim();
+      cmdBase = `cd colonyNetwork-${versionTag} && source $HOME/.nvm/nvm.sh && nvm use ${nodeVersion}`;
```
area, line 323: I think something like this should work on all OSes, right, so why limit it to Apple Silicon? `cmdBase` should also presumably be used through the file (so in e.g. `deployViaTruffle`)

I think doing it properly, you'll hit a 'fun' quirk that hasn't really been considered for a while, which was the `glwss4` release (at least, maybe others?) had an out-of-date `.nvmrc` asking for Node 14 which `truffle` doesn't work with. So if you get `v14.x` out of the `nvmrc` you should use v16 instead.

area: I would also like to accommodate `fnm` as well as `nvm`. For the equivalent of `source $HOME/.nvm/nvm.sh` for `fnm`, you can use `source <(fnm env)`.

area (commented):
Yeah, none of this has _really_ worked properly for a while. I support fixing it up.

scripts/deployOldUpgradeableVersion.js
kronosapiens: My hope was to limit the changes to as-needed so as not to go down too much of a rabbit hole; I implemented it this way to keep things as similar as possible in the Circle environment. Would you be open to merging this as is (with the inclusion of `cmdBase` everywhere) and adding more support as-needed down the road? Currently everything is building so I don't want to spend much more time on it right now.

area: Wasn't too much of a rabbit hole 😉 - let me know [what you think](https://github.com/JoinColony/colonyNetwork/compare/maint/upgradable...maint/upgradable-alex), and cherry-pick in to here if you're (broadly) happy with it.

area (approved):

PR #1283, opened 2024-08-06
Support forked deployments

area:
To improve the accuracy and throughput of QA, there's a desire for the ability to have an entire environment that is forked from the QA environment. That includes the database and so-forth, which is out of scope for here, but this PR includes the elements that are required on our side.

There are two main features here:

* Deploying the latest versions of any contracts that have been updated since the deployment to QA. This is done via the `deployToForkedChain` script (that's called from outside this repo). This script also downloads the latest reputation state, which the reputation miner will use.
* Allowing historical querying of reputation; I've added a new `forked.js` to the miner which is only intended to be used as the miner in this circumstance. It allows mining with an address with an unknown private key (via the [`getImpersonatedSigner`](https://hardhat.org/hardhat-network/docs/guides/forking-other-networks#impersonating-accounts) helper, as well as querying the oracle we've forked from in the case of an unknown (and therefore presumed historical) reputation state. This will over-query the oracle (as any reputations queried that don't exist will also be tried upstream), but without having the forked miner sync the whole of history (time-consuming and/or expensive), I think it's the best that we can do.

Leaving in draft for now, as the feature as a whole is still in the testing phase, but we're getting close to this being reviewed and merged.


packages/reputation-miner/bin/forked.js
```diff
@@ -0,0 +1,106 @@
+const hre = require("hardhat");
```
kronosapiens, line 1: Any reason you don't use `globals` here?

scripts/deployToForkedChain.js
```diff
@@ -0,0 +1,138 @@
...
+    events = await provider.getLogs(filter);
+    if (events.length > 0) {
+      foundEvent = true;
+    }
+  }
+  return events[0];
+}
+
+async function main() {
+  signer = await ethers.getImpersonatedSigner("0x56a9212f7f495fadce1f27967bef5158199b36c7");
```
kronosapiens, line 61: Add comment clarifying what this address is. Random?

scripts/deployToForkedChain.js
```diff
@@ -0,0 +1,138 @@
...
+  const forkedBlock = await provider.getBlockNumber();
+
+  await hre.run("deploy");
+
+  const DEPLOYED_NETWORK_ADDRESS = "0x777760996135F0791E2e1a74aFAa060711197777";
```
kronosapiens, line 67: Ditto this -- where is this coming from?

kronosapiens (commented):
Looks fine to me, a few comments here and there

packages/reputation-miner/bin/forked.js
area: This isn't being run from inside hardhat, so I don't think it exists?

scripts/deployToForkedChain.js
area: It's the address in control of the production network, which has to be the one to update resolvers etc.

scripts/deployToForkedChain.js
area: It's the address `ColonyNetwork` is deployed to in production.

kronosapiens:
Marking this `on-hold` as there does not seem to be an urgent need for it elsewhere

PR #1284, opened 2024-08-06
Split out Wormhole relayer

area:
With the current multichain functionality, we focussed on the contracts and not on the additional infrastructure around it. This PR is intended to at least create the scaffolding for that infrastructure, which we can hopefully build on.

As it stands, the `mockBridgeMonitor` script represents three components of the Wormhole system:

1. [The Guardian Network](https://docs.wormhole.com/wormhole/explore-wormhole/guardian) - these are the trusted 'oracles' of the network, who sign attestations that messages have been emitted on a particular chain
2. [The Spy](https://docs.wormhole.com/wormhole/explore-wormhole/spy) - this is something that monitors the guardian network for messages of interest
3. [The Relayer](https://docs.wormhole.com/wormhole/explore-wormhole/relayer) - this is something that listens for messages from the spy and relays them to the other chain(s)

We certainly aren't going to be responsible for the Guardian Network. We _might_ end up running our own spy. However, we will certainly need our own relayer to deliver the desired experience for people - they sign a message on one chain and it appears on the other chain, with no further interaction from them. This ends up very similar to the Metatransaction Broadcaster - they execute a transaction that emits an event on one chain, and the relayer ends up executing a transaction on the other chain, paying for gas.

This PR splits out the relayer from the `mockBridgeMonitor` script, and has it running in a separate process, communicating with the spy via the protocol that it would use in a real deployment. This give us a platform to build a robust relayer on as we go forward. The relayer is _extremely_ simple at the moment. One thing that we had avoided doing 'properly' previously was considering [VAAs](https://docs.wormhole.com/wormhole/explore-wormhole/vaa), but given the libraries being used expect to see VAAs at various points, I've had to do those much more 'properly'.

Some of the changes here we just... stumbled across being necessary. Changing providers to use `RetryProvider` where I could is certainly something that coul be in a separate PR (and will move if desired), but splitting this relayer out made us more sensitive to connection errors during builds, somehow? We also started to reliably run across issues with `hardhat`'s [memory usage](https://github.com/NomicFoundation/hardhat/issues/3471) during the cross-chain tests, which we've avoided for now by deleting the `build-info` between the two sets of tests. It feels like that's not going to be able to be a long-term solution, but it's what we've got for now.

There's a lot here, so keen to answer questions and get this in to a mergeable (and understandable) state.


scripts/setup-bridging-contracts.js
```diff
@@ -126,14 +128,69 @@ async function setupBridging(homeRpcUrl, foreignRpcUrl) {
...
     foreignBridge.address,
     homeColonyBridge.address,
     foreignColonyBridge.address,
   ); // eslint-disable-line no-unused-vars
+
+  // TODO: Start the bridge monitor
+  console.log("Starting bridge monitor");
+
+  // Write config file
+  const config = `
```
kronosapiens, line 144: Can we not import a config from somewhere?

kronosapiens (commented):
Nice. Why write it in typescript? Was it needed to use the libraries, or just a good opportunity to try it out?

.circleci/config.yml
```diff
@@ -60,18 +68,27 @@ jobs:
...
       - run:
           name: "Check git hooks run"
           command: pnpm run check:gitchanges
+      - run:
+          name: "Checking contract storage layout hasn't broken existing slots"
+          command: npx hardhat storage-layout --check
+      - run:
+          name: "Checking contract storage layout hasn't had additions made"
+          command: npx hardhat storage-layout --update && npm run normalize:storageSlots && git checkout ./.storage-layouts && git diff --exit-code
+      - <<: *step_compile_mock_spy
```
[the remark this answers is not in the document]
area: https://github.com/JoinColony/colonyNetwork/pull/1284/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R71-R72

scripts/setup-bridging-contracts.js
area: I think that's tricky because of our desire to use the `[wormhole.CHAIN_ID_ARBITRUM_SEPOLIA]` syntax. We can't import and `JSON` parse it (because it's not valid JSON), and if we imported it as a string, we'd have to meticulously replace the keys based on line / character number.

If you have a suggestion here, I'm all ears!

area: I guess regex?

contracts/testHelpers/WormholeMock.sol
```diff
@@ -95,7 +87,68 @@ contract WormholeMock is IWormhole {
     GuardianSet memory guardianSet
   ) external pure returns (bool valid, string memory reason) {}
 
-  function parseVM(bytes memory encodedVM) external pure returns (VM memory vm) {}
+  function parseVM(bytes memory encodedVM) public pure virtual returns (VM memory vm) {
+    uint index = 0;
+
+    vm.version = encodedVM.toUint8(index);
+    index += 1;
```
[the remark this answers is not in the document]
area: Yeah, that's a good idea :+1:

packages/metatransaction-broadcaster/ExtendedNonceManager.js
```diff
@@ -23,17 +24,22 @@ class ExtendedNonceManager extends NonceManager {
   async sendTransaction(transactionRequest) {
     // What nonce are we going to attach to this?
     // Definitely not any we've sent and are pending
-    const pendingNonces = Object.keys(this.signedTransactions).map((txhash) => this.signedTransactions[txhash].nonce);
-    // At least whatever the endpoint says
-    let nonce = await this.signer.getTransactionCount();
+    // const pendingNonces = Object.keys(this.signedTransactions).map((txhash) => this.signedTransactions[txhash].nonce);
```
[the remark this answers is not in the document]
area: Temporarily, it was giving me grief at some point (I can't remember why). I've cleaned it up a bit though, maybe?

packages/metatransaction-broadcaster/ExtendedNonceManager.js
```diff
@@ -23,17 +24,22 @@ class ExtendedNonceManager extends NonceManager {
...
+    // At least whatever the endpoint says, or whatever we've already sent if higher
+    const endpointNonce = await this.signer.getTransactionCount();
+    let nonce = Math.max(endpointNonce);
```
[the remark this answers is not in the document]
area: Another casualty of the grief this manager was causing me at some interim stage.

packages/wormhole-relayer/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    sourceTx,
+    providers
+  } from "@wormhole-foundation/relayer-engine";
+  import { CHAIN_ID_ARBITRUM_SEPOLIA, CHAIN_ID_SEPOLIA, SignedVaa } from "@certusone/wormhole-sdk";
+
+  const NonceManager = require("./../metatransaction-broadcaster/ExtendedNonceManager");
+  const { RetryProvider } = require("../package-utils")
+
+  export class TestStorage extends RedisStorage {
+    // startWorker(cb: onJobHandler): void {
```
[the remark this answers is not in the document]
area: This was from the original tutorial I was following, and me working out how much I could remove and still have everything work. I haven't figured out if permanent storage will be useful for us once in production, so haven't totally removed it.

area:
> Nice. Why write it in typescript? Was it needed to use the libraries, or just a good opportunity to try it out?

It wasn't needed, but was certainly helpful for some of this. I also think it's just better for things being deployed to production.

scripts/setup-bridging-contracts.js
kronosapiens: Ok, I'm not too broken up about it if there's good reasons for doing it this way.

packages/wormhole-relayer/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    // add a filter with a callback that will be
+    // invoked on finding a VAA that matches the filter
+
+    const colonyBridgeAddresses: {
+      [chainid: string]: string
+    } = {};
+
+    Object.keys(config.chains).forEach((chainid) => colonyBridgeAddresses[chainid] = config.chains[chainid].colonyBridgeAddress);
+    app.multiple( colonyBridgeAddresses,
```
kronosapiens, line 120: Extra space

packages/wormhole-relayer/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    async (ctx, next) => {
+        // console.log('callback');
```
kronosapiens, line 122: Delete, etc

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
+import {
+    Server,
+    ServerCredentials,
+} from '@grpc/grpc-js';
+import {MockGuardians} from "@certusone/wormhole-sdk/lib/cjs/mock";
+const GUARDIAN_PRIVATE_KEY = '0x5756c5018454b83225f8e8264651c3a8032c4ac931e1d310d81239754f1bb791';
```
kronosapiens, line 6: Don't like this const mixed up with the imports.

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
...
+const guardians = new MockGuardians(0, [
+    GUARDIAN_PRIVATE_KEY,
+  ]);
+
+const RetryProvider = require("../packages/package-utils").RetryProvider;
```
kronosapiens, line 11: Didn't I see this imported via `index` elsewhere? Can we be consistent?

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
...
+
+
+const ethers = require("ethers");
+
+// eslint-disable-next-line import/no-unresolved
+const bridgeAbi = require("../artifacts/contracts/testHelpers/WormholeMock.sol/WormholeMock.json").abi;
+// eslint-disable-next-line import/no-unresolved
+const wormholeBridgeForColonyAbi = require("../artifacts/contracts/bridging/WormholeBridgeForColony.sol/WormholeBridgeForColony.json").abi;
+
+function ethereumAddressToWormholeAddress(address: string) {
```
kronosapiens, line 26: I wouldn't be opposed to calling this `e2w` for brevity but I have a feeling you won't like that

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
...
+          // setTimeout(() => {
+          //     subscription.write({vaaBytes: Buffer.from(vaa, 'base64')});
+          // }, 1000);
+        }
+    });
+
+    this.relayerAddress = "0x770656997AE1C8AB0250571ca1dff5c5A3C37700";
+
+
+    this.server.bindAsync('0.0.0.0:7073', ServerCredentials.createInsecure(), () => {
```
kronosapiens, line 93: Would it make sense to define the wormhole port as a constant and import it throughout?

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
...
+    throw new Error("Unable to find transaction with provided nonce");
+  }
+
+  // VAAs are the primitives used on wormhole (Verified Action Approvals)
+  // See https://docs.wormhole.com/wormhole/explore-wormhole/vaa for more details
+  // Note that the documentation sometimes also calls them VMs (as does IWormhole)
+  // I believe VM stands for 'Verified Message'
+  async encodeMockVAA(sender: string, sequence: number, nonce: number, payload: string, consistencyLevel:number, chainId: number) {
+    const version = 1;
+    const timestamp = Math.floor(Date.now() / 1000);
```
kronosapiens, line 125: I'd add a comment along the lines of `// ms to s`

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,347 @@
...
+    });
+
+    this.foreignBridge.on("LogMessagePublished", async (sender, sequence, nonce, payload, consistencyLevel) => {
+      const { chainId } = await this.signerForeign.provider.getNetwork();
+      // Due to limitations, for local testing, our wormhole chainIDs have to be 'real' wormhole chainids.
+      // So I've decreed that for chainId 256669100, we use 10003 (which is really arbitrum sepolia)
+      // and for chainId 256669101, we use 10002 (which is really sepolia).
+      // This isn't ideal, but it's the best solution I have for now
+      let wormholeChainId;
+      if (chainId === 265669100) {
```
kronosapiens, line 222: Aren't these defined as constants somewhere? Can we import those?

.gitignore
```diff
@@ -34,3 +34,4 @@ ganache-chain-db-2/
 build-coverage/
 coverage-extensions/
 .pnpm-store
+scripts/mockGuardianSpy.js
```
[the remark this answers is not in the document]
area: It's a build artifact, so should not be in the repo.

packages/wormhole-relayer/index.ts
area: I've configured ESLint to apply to .ts files so hopefully less of this slipping through.

scripts/mockGuardianSpy.ts
area: Made this and the instance of `index` being used (which is in `packages/reputation-miner/ReputationMiner.js` consistent (but both have changed 😛 )

scripts/mockGuardianSpy.ts
area: Your feeling is correct 😆 

I'll see if I can come up with something else, but open to other suggestions. Some things just have long names though, and brevity should not come at the cost of intelligibility.

scripts/mockGuardianSpy.ts
area: I guess we could... it's only used in three places though, and this is the port it's expected to be on (from the examples / docs).

scripts/mockGuardianSpy.ts
area: Isn't actually needed, so removed.

.gitignore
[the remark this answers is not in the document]
kronosapiens: Not following -- isn't this a whole file that we're reviewing?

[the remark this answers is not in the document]
area: We are reviewing `mockGuardianSpy.ts`, which is compiled to `mockGuardianSpy.js` to be run during the tests.

In terms of the names, they can be brought in line.

area:
> We also started to reliably run across issues with hardhat's https://github.com/NomicFoundation/hardhat/issues/3471 during the cross-chain tests, which we've avoided for now by deleting the build-info between the two sets of tests. It feels like that's not going to be able to be a long-term solution, but it's what we've got for now.

With these last changes, we've started running in to this even for the first run of the tests, so deleting the `build-info` was no longer sufficient. I don't like it, but I've upped the resource class for the `coverage-test-bridging` job to `large`, which has resolved the issue.

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,356 @@
...
+
+function ethereumAddressToWormholeAddress(address: string) {
+  return ethers.utils.hexZeroPad(ethers.utils.hexStripZeros(ethers.utils.hexlify(address)), 32);
+}
+
+type QueueEntry = [ethers.Contract, string, number, number, string, number, number];
+
+class MockGuardianSpy {
+  homeRpc: string;
+
```
kronosapiens, line 30: Are these blank lines necessary? It's a lot of extra vertical space

scripts/mockGuardianSpy.ts
```diff
@@ -0,0 +1,356 @@
...
+      // So I've decreed that for chainId 256669100, we use 10003 (which is really arbitrum sepolia)
+      // and for chainId 256669101, we use 10002 (which is really sepolia).
+      // This isn't ideal, but it's the best solution I have for now
+      let wormholeChainId;
+      if (chainId === FORKED_XDAI_CHAINID) {
+        wormholeChainId = 10003;
+      } else if (chainId === FORKED_XDAI_CHAINID + 1) {
+        wormholeChainId = 10002;
+      } else {
+        throw new Error("Unsupported chainId");
```
kronosapiens, line 237: Can this duplicate code be turned into a helper function?

test/cross-chain/cross-chain.js
```diff
@@ -123,19 +124,31 @@ contract("Cross-chain", (accounts) => {
...
+    if (homeChainId === "0xfd5c9ec") {
+      wormholeHomeChainId = 10003;
+    } else {
+      wormholeHomeChainId = 10002;
+    }
 
     // const foreignNetworkId = await ethersForeignSigner.provider.send("net_version", []);
     foreignChainId = await ethersForeignSigner.provider.send("eth_chainId", []);
-    wormholeForeignChainId = BigNumber.from(foreignChainId).mod(265669).mul(2);
+    if (foreignChainId === "0xfd5c9ec") {
```
kronosapiens, line 147: Ditto code duplication -- can this be pulled out into a helper?

kronosapiens (commented):
Looks pretty good! A few minor comments.

scripts/mockGuardianSpy.ts
area: It's being enforced by https://eslint.org/docs/latest/rules/lines-between-class-members which is coming from [eslint-config-colony](https://www.npmjs.com/package/@colony/eslint-config-colony).

kronosapiens (approved):

PR #1285, opened 2024-08-09
Introduce Proxy Colonies

area:
Had to make a new PR to rename the branch.


area:
There is now work on this branch that shows what needs to be done from the perspective of extensions, as I have added functionality to the `OneTxPayment` extension. Any function that takes a token (or multiples) now needs to take both `chainId` and `token`. I've done this in this first instance by making the original function signature call the new function with `chainId` populated with `block.chainid`.

This is a lot of functions, however, and I would only want two instances of every such function on the contracts for a transitional period at most. Ideally, we could just do version-guarded upgrades for everything (which we have [done before](https://github.com/JoinColony/colonyNetwork/commit/1f29992250a08da862cf20910db1ad7123bef6d7#diff-1b3fed3c1d86a176ebfac07111aabe7a9c941b75fd01af23e56b5c6d7e270a45R72), if more limited in scope).

In the past, we've added these 'passthrough' functions for the old arguments to make the life of frontend easier (so changes in the contracts don't necessitate frontend work), but given every piece of functionality here is already going to require touching in the frontend to add multichain support, I'm inclined to just bite the bullet and avoid having to come back through later to remove all the deprecated functions.

contracts/bridging/ProxyColony.sol
```diff
@@ -0,0 +1,152 @@
...
+  mapping(address => uint256) tokenBalances;
+
+  // Events
+
+  modifier onlyColonyBridge() {
+    require(ProxyColonyNetwork(owner).colonyBridgeAddress() == msgSender(), "colony-only-bridge");
+    _;
+  }
+
+  event DomainFundsClaimed(address token, uint256 _domainId, uint256 balance);
```
kronosapiens: Remove underscore

contracts/bridging/ProxyColony.sol
```diff
@@ -0,0 +1,152 @@
...
+  event TransferMade(address token, address user, uint256 amount);
+
+  // Public functions
+
+  function claimTokens(address _token) public {
+    uint256 balance = (_token == address(0x0))
+      ? address(this).balance
+      : ERC20Extended(_token).balanceOf(address(this));
+
+    require(balance >= tokenBalances[_token], "colony-shell-token-bookkeeping-error");
```
kronosapiens, line 57: Should this be an assert? Should this ever be reachable due to user input?

contracts/bridging/ProxyColony.sol
```diff
@@ -0,0 +1,152 @@
...
+    uint256 difference = balance - tokenBalances[_token];
+
+    tokenBalances[_token] = balance;
+
+    bytes memory payload = abi.encodeWithSignature(
```
kronosapiens, line 62: Remind me why we have to send a message confirming the claim? Can we add a comment explaining that for posterity?

contracts/bridging/ProxyColony.sol
```diff
@@ -0,0 +1,152 @@
...
+    address domainTokenReceiverAddress = ProxyColonyNetwork(owner).checkDomainTokenReceiverDeployed(
+      _domainId
+    );
+
+    uint256 balance = (_token == address(0x0))
+      ? address(domainTokenReceiverAddress).balance
+      : ERC20Extended(_token).balanceOf(address(domainTokenReceiverAddress));
+
+    if (_token == address(0x0)) {
+      DomainTokenReceiver(domainTokenReceiverAddress).transferNativeToColony();
```
kronosapiens: As elsewhere, can we call this `transferChainNativeToColony()`?

contracts/bridging/ProxyColony.sol
```diff
@@ -0,0 +1,152 @@
...
+    address _token,
+    address _recipient,
+    uint256 _amount
+  ) public onlyColonyBridge {
+    tokenBalances[_token] -= _amount;
+
+    if (_token == address(0x0)) {
+      payable(_recipient).transfer(_amount);
+    } else {
+      require(ERC20Extended(_token).transfer(_recipient, _amount), "colony-shell-transfer-failed");
```
kronosapiens: "Shell" terminology is out-of-date

contracts/bridging/ProxyColonyNetwork.sol
```diff
@@ -0,0 +1,128 @@
...
+import { IColonyBridge } from "./IColonyBridge.sol";
+import { DomainReceiverManagement } from "./../common/DomainReceiverManagement.sol";
+
+contract ProxyColonyNetwork is DSAuth, Multicall, CallWithGuards, DomainReceiverManagement {
+  address resolver; // Storage slot 2 (from DSAuth there is authority and owner at storage slots 0 and 1 respectively)
+
+  address public colonyBridgeAddress;
+  uint256 public homeChainId;
+  address public proxyColonyResolverAddress;
+  mapping(address => bool) public shellColonies;
```
kronosapiens: Terminology out-of-date

contracts/bridging/ProxyColonyNetwork.sol
```diff
@@ -0,0 +1,128 @@
...
+  function setHomeChainId(uint256 _homeChainId) public auth {
+    homeChainId = _homeChainId;
+  }
+
+  function getDomainTokenReceiverResolver() public view override returns (address) {
+    return domainTokenReceiverResolver;
+  }
+
+  function msgSenderIsColony() internal view override returns (bool) {
+    return shellColonies[msgSender()];
```
kronosapiens: Ditto

contracts/bridging/ProxyColonyNetwork.sol
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  function setDomainTokenReceiverResolver(address _resolver) public auth {
+    domainTokenReceiverResolver = _resolver;
+  }
+
+  function isStopped() internal pure override returns (bool) {
```
kronosapiens, line 99: If this is meant to be re-implemented later, please add a comment explaining the reasoning

contracts/colony/Colony.sol
```diff
@@ -317,11 +317,45 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
...
-    // sig = bytes4(keccak256("cancelExpenditureViaArbitration(uint256,uint256,uint256)"));
-    // colonyAuthority.setRoleCapability(uint8(ColonyRole.Arbitration), address(this), sig, true);
+    ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
+    bytes4 sig;
+
+    sig = bytes4(keccak256("editAllowedDomainReputationReceipt(uint256,uint256,bool)"));
+    colonyAuthority.setRoleCapability(uint8(ColonyRole.Root), address(this), sig, true);
+
+    sig = bytes4(
+      keccak256("setExpenditurePayout(uint256,uint256,uint256,uint256,uint256,address,uint256)")
```
kronosapiens, line 327: Does this belong in this PR?

contracts/colony/ColonyArbitraryTransaction.sol
```diff
@@ -44,6 +44,33 @@ contract ColonyArbitraryTransaction is ColonyStorage {
...
+      bytes memory payload = abi.encodeWithSignature(
+        "makeArbitraryTransaction(address,bytes)",
+        _destination,
+        _action
+      );
+      IColonyNetwork(colonyNetworkAddress).bridgeMessage(_chainId, payload);
+    }
+  }
+
+  function callProxyNetwork(
```
kronosapiens: Can we call this `multicallProxyNetwork`?

contracts/colony/ColonyFunding.sol
```diff
@@ -100,12 +130,212 @@ contract ColonyFunding is
...
+    // able to do with recovery mode.
+    uint256 remainder = claimAmount - feeToPay;
+    nonRewardPotsTotal[_token] += remainder;
+
+    // fundingPots[0].balance[_token] += feeToPay;
+    incrementFundingPotBalance(0, block.chainid, _token, feeToPay);
+
+    uint256 approvedAmount = domainReputationApproval[_domainId];
+
+    if (tokenEarnsReputationOnPayout(_token)) {
```
kronosapiens: Have comments but they are duplicates of #1291

contracts/colony/ColonyFunding.sol
```diff
@@ -100,12 +130,212 @@ contract ColonyFunding is
...
+
+    // Check the domain has enough for what is
+    if (_token == address(0x0)) {
+      require(
+        _value + _amount <= getFundingPotBalance(domain.fundingPotId, _token),
+        "colony-insufficient-funds"
+      );
+    } else {
+      require(
+        _amount <= getFundingPotBalance(domain.fundingPotId, _token),
```
kronosapiens: Unless I'm misunderstanding, can this be one require with a `max(_amount, _value)`?

contracts/colony/ColonyFunding.sol
```diff
@@ -100,12 +130,212 @@ contract ColonyFunding is
...
+    // If the LiFi transaction didn't use all the tokens, reduce the allowance back to what it was before
+    if (postApproval > priorApproval) {
+      require(
+        ERC20Extended(_token).approve(LIFI_ADDRESS, priorApproval),
+        "colony-post-exchange-approve-failed"
+      );
+    }
+  }
+
+  function exchangeProxyHeldTokensViaLiFi(
```
kronosapiens: Can these two functions be combined? With perhaps some branching behavior based on whether `block.chainId == _chainId`?

contracts/colony/ColonyFunding.sol
```diff
@@ -100,12 +130,212 @@ contract ColonyFunding is
...
+    require(
+      ERC20Extended(_token).approve(LIFI_ADDRESS, _amount + priorApproval),
+      "colony-approve-failed"
+    );
+    (bool success, ) = LIFI_ADDRESS.call{ value: _value }(_txdata);
+
+    require(success, "colony-exchange-tokens-failed");
+
+    // Check the allowances afterwards, ensuring that at least some of the tokens we approved were spent by
+    // the LiFi transaction - if not, a different token might have been spent, so revert.
```
kronosapiens: What is the reasoning here? Under what circumstances would a different token have been spent?

contracts/colony/ColonyFunding.sol
```diff
@@ -233,11 +513,31 @@ contract ColonyFunding is
...
+
+  function getFundingPotProxyBalance(
+    uint256 _potId,
+    uint256 _chainId,
+    address _token
+  ) public view returns (uint256) {
+    return fundingPots[_potId].chainBalances[_chainId][_token];
+  }
+
+  function recordClaimedFundsFromBridge(
```
kronosapiens, line 551: Doesn't this function need some sort of guard?

contracts/colony/ColonyFunding.sol
```diff
@@ -284,19 +584,20 @@ contract ColonyFunding is
...
+    decrementFundingPotBalance(_fromPot, _chainId, _token, _amount);
+    incrementFundingPotBalance(_toPot, _chainId, _token, _amount);
 
-    if (_fromPot == 1) {
+    if (_fromPot == 1 && _chainId == block.chainid) {
       // If we're moving from the root pot, then check we haven't dropped below what we need
       // to cover any approvals that we've made.
       require(
-        fromPot.balance[_token] >= tokenApprovalTotals[_token],
+        getFundingPotBalance(_fromPot, _chainId, _token) >= tokenApprovalTotals[_token],
```
kronosapiens: I would use `block.chainId` here to emphasize the case we're in

contracts/colony/ColonyFunding.sol
```diff
@@ -381,61 +698,85 @@ contract ColonyFunding is
...
+    uint256 previousTotal = getFundingPotPayout(expenditures[_id].fundingPotId, _chainId, _token);
+    uint256 runningTotal = previousTotal;
 
     for (uint256 i; i < _slots.length; i++) {
       require(_amounts[i] <= MAX_PAYOUT, "colony-payout-too-large");
-      uint256 currentPayout = expenditureSlotPayouts[_id][_slots[i]][_token];
-
-      expenditureSlotPayouts[_id][_slots[i]][_token] = _amounts[i];
+      uint256 currentPayout = getExpenditureSlotPayout(_id, _slots[i], _chainId, _token);
+      // uint256 currentPayout = expenditureSlotPayouts[_id][_slots[i]][_token];
```
kronosapiens: Delete unused lines

contracts/colony/ColonyFunding.sol
```diff
@@ -381,61 +698,85 @@ contract ColonyFunding is
...
+    refundDomain(_fundingPotId, _chainId, _token);
 
     IColonyNetwork colonyNetworkContract = IColonyNetwork(colonyNetworkAddress);
     address payable metaColonyAddress = colonyNetworkContract.getMetaColony();
 
-    fundingPots[_fundingPotId].balance[_token] -= _payout;
-    fundingPots[_fundingPotId].payouts[_token] -= _payout;
-    nonRewardPotsTotal[_token] -= _payout;
+    decrementFundingPotBalance(_fundingPotId, _chainId, _token, _payout);
+    setFundingPotPayout(
```
kronosapiens, line 771: Why is this not implemented as `decrement` also?

contracts/colony/IColony.sol
```diff
@@ -556,6 +567,22 @@ interface IColony is IDSAuth, ColonyDataTypes, IRecovery, IBasicMetaTransaction,
     uint256 _amount
   ) external;
 
+  /// @notice @deprecated
```
kronosapiens: Is it?

area:
Something that might be a bit surprising is introduction of `CommonTypes`, but this needed so that `ColonyStorage` does not need to import `ColonyNetworkDataTypes`. Without this, the `ProxyColonyRequested` event ends up being defined twice from a Solidity perspective, but the arguments mean different things so we do actually want it defined in two places.

contracts/colony/IColony.sol
```diff
@@ -858,6 +942,73 @@ interface IColony is IDSAuth, ColonyDataTypes, IRecovery, IBasicMetaTransaction,
...
 
+  /// @notice Exchange funds between two tokens, potentially between chains
+  /// @param _permissionDomainId The domainId in which I have the permission to take this action
+  /// @param _childSkillIndex The child index in `_permissionDomainId` where we can find `_domainId`
+  /// @param _domainId Id of the domain
+  /// @param _txdata Transaction data for the exchange
+  /// @param _value Value of the transaction
+  /// @param _token Address of the token, `0x0` value indicates Ether
+  /// @param _amount Amount of tokens to exchange
+  function exchangeTokensViaLiFi(
```
kronosapiens: In the spirit of minimizing the amount of new code, I'd be in favor of scrapping this function and using the more general one (with a `chainId`) for home-chain cases.

contracts/colonyNetwork/ColonyNetwork.sol
```diff
@@ -111,6 +112,23 @@ contract ColonyNetwork is BasicMetaTransaction, ColonyNetworkStorage, Multicall
...
   }
 
+  function bridgeMessage(uint256 _chainId, bytes memory _payload) public stoppable calledByColony {
+    require(
+      IColonyBridge(colonyBridgeAddress).sendMessage(_chainId, msg.sender, _payload),
+      "colony-network-bridge-message-failed"
+    );
+  }
+
+  function bridgeMessageToNetwork(
```
kronosapiens, line 122: Shouldn't this be `bridgeMessageAsNetwork`?

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -220,37 +237,14 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
-    uint256 _domainId
-  ) public view returns (address) {
-    bytes32 salt = getDomainTokenReceiverDeploySalt(_colony, _domainId);
-
-    // To get the correct address, we have to mimic the _guard functionality of CreateX
-    bytes32 guardedSalt = keccak256(abi.encode(bytes32(uint256(uint160(address(this)))), salt));
-    return ICreateX(CREATEX_ADDRESS).computeCreate3Address(guardedSalt);
+  function msgSenderIsColony() internal view override returns (bool) {
+    require(_isColony[msgSender()], "colony-caller-must-be-colony");
+    assert(msgSender() == msg.sender);
```
kronosapiens: Why not `return msgSender() == msg.sender`?

contracts/common/DomainReceiverManagement.sol
```diff
@@ -0,0 +1,112 @@
...
+pragma solidity 0.8.27;
+pragma experimental "ABIEncoderV2";
+
+import { MetaTransactionMsgSender } from "./MetaTransactionMsgSender.sol";
+import { DomainTokenReceiver } from "./DomainTokenReceiver.sol";
+import { ICreateX } from "./../../lib/createx/src/ICreateX.sol";
+import { EtherRouterCreate3 } from "./EtherRouterCreate3.sol";
+import { EtherRouter } from "./EtherRouter.sol";
+import { Resolver } from "./Resolver.sol";
+import { IsContract } from "./IsContract.sol";
```
kronosapiens: I'd be in favor of a `mixins` directory specifically for these small function contracts

contracts/common/DomainReceiverManagement.sol
```diff
@@ -0,0 +1,112 @@
...
+
+abstract contract DomainReceiverManagement is MetaTransactionMsgSender, IsContract {
+  address constant CREATEX_ADDRESS = 0xba5Ed099633D3B313e4D5F7bdc1305d3c28ba5Ed;
+
+  function getDomainTokenReceiverResolver() public view virtual returns (address);
+  function msgSenderIsColony() internal view virtual returns (bool);
+  function isStopped() internal view virtual returns (bool);
+
+  function checkDomainTokenReceiverDeployed(
```
kronosapiens: Isn't this the function we were calling `idempotentDeployDomainTokenReceiver`?

contracts/common/DomainReceiverManagement.sol
```diff
@@ -0,0 +1,112 @@
...
+      if (address(resolver) != getDomainTokenReceiverResolver()) {
+        EtherRouter(payable(domainTokenReceiverAddress)).setResolver(
+          getDomainTokenReceiverResolver()
+        );
+      }
+    } catch {
+      revert("colony-network-domain-receiver-not-etherrouter");
+    }
+
+    // Check it's set up correctly
```
kronosapiens: Didn't we drop this section?

hardhat.config.js
```diff
@@ -79,6 +79,12 @@ task("test", "Run tests").setAction(async () => {
   }
 });
 
+task("ensureCreateXDeployed", "Ensure CreateX is deployed").setAction(async () => {
```
kronosapiens: Task name should be kebab-case

helpers/wormholescanMock/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "wormholescanmock",
```
kronosapiens: Rename `wormholescan-mock` to make consistent with directory name

helpers/wormholescanMock/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+
+function wormholeAddressToEvmAddress(address: string): string {
+  return ethers.utils.getAddress(`0x${address.slice(26, 66)}`);
+}
+
+function evmChainIdToWormholeChainId(chainId: number) {
+  // So I've decreed that for chainId 265669100, we use 10003 (which is really arbitrum sepolia)
+  // and for chainId 265669101, we use 10002 (which is really sepolia).
+
+  switch (chainId) {
```
kronosapiens, line 21: Nice switch statement!

test/contracts-network/colony-network-recovery.js
```diff
@@ -193,14 +193,13 @@ contract("Colony Network Recovery", (accounts) => {
...
       await checkErrorRevert(colonyNetwork.startTokenAuction(ADDRESS_ZERO), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.bridgeSkillIfNotMiningChain(1), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.bridgePendingReputationUpdate(ADDRESS_ZERO, 0), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.bridgeCurrentRootHash(ADDRESS_ZERO), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.addReputationUpdateLogFromBridge(ADDRESS_ZERO, ADDRESS_ZERO, 0, 0, 0), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.addPendingReputationUpdate(0, ADDRESS_ZERO), "colony-in-recovery-mode");
-      await checkErrorRevert(colonyNetwork.setReputationRootHashFromBridge(HASHZERO, 0, 0), "colony-in-recovery-mode");
       await checkErrorRevert(colonyNetwork.setDomainTokenReceiverResolver(ADDRESS_ZERO), "colony-in-recovery-mode");
       await checkErrorRevert(colonyNetwork.idempotentDeployDomainTokenReceiver(ADDRESS_ZERO), "colony-in-recovery-mode");
+      await checkErrorRevert(colonyNetwork.bridgeMessage(1, "0x00000000"), "colony-in-recovery-mode");
```
kronosapiens, line 198: I see, so all of the skill functionality is being deprecated in favor of a more general messaging implementation? I was wondering where they all went.

test/cross-chain/cross-chain.js
```diff
@@ -139,29 +145,71 @@ contract("Cross-chain", (accounts) => {
...
-    // Deploy colonyNetwork to whichever chain truffle hasn't already deployed to.
-    try {
-      const nonHardhatChainId = process.env.HARDHAT_FOREIGN === "true" ? homeChainId : foreignChainId;
-
-      await exec(`CHAIN_ID=${parseInt(nonHardhatChainId, 16)} npx hardhat deploy --network development2`);
-    } catch (err) {
-      console.log(err);
-      process.exit(1);
-    }
+    // foreignChainId = await ethersForeignSigner.provider.send("eth_chainId", []);
```
kronosapiens: Why add all these commented-out tests?

test/extensions/staked-expenditure.js
```diff
@@ -330,6 +330,26 @@ contract("StakedExpenditure", (accounts) => {
       );
     });
 
+    it("cannot slash the stake if the expenditure is already cancelled", async () => {
```
kronosapiens, line 333: Are these changes related to the rest of the PR, or did you just notice they were missing as you went along?

kronosapiens (commented):
Wow, this is a big PR! A lot going on.

My biggest feedback now is that I am wary of all of the parallel functionality being introduced. I think I'd prefer moving everything over to the chainId-aware functions and deprecating everything else, so we don't end up having to support parallel implementations. The current functions can be marked deprecated and re-implemented to call the chainId-aware versions while passing in `chainId()`

contracts/bridging/ProxyColony.sol
area: It is currently reachable if a proxy colony gives approval to spend a token to another address.

contracts/bridging/ProxyColony.sol
area: So that the colony knows the funds that it has.

contracts/bridging/ProxyColonyNetwork.sol
area: This is the implementation, it's required to exist for DomainReceiverManagement.

contracts/colony/ColonyFunding.sol
area: I think it would have to be the sum, but that's not right either, because value should be being checked against the ether balance, not the token balance.

contracts/colony/ColonyFunding.sol
area: If the supplied `_txData` spent a different token.

contracts/colony/ColonyFunding.sol
area: Yikes, sure does!

contracts/colony/ColonyFunding.sol
area: I think just because it's not used nearly as much.

contracts/colony/IColony.sol
area: Hm, yeah, I don't think it should be. Maybe I thought I was going to impelement a chainid aware `setExpenditurePayouts` until I remembered `multicall` exists.

contracts/colonyNetwork/ColonyNetwork.sol
area: I think either a probably defensible. I prefer the current name, though as I think it is more informative (we are in the network contract, so who else would we be bridging as if it's called `bridgeMessageAsNetwork`?)

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: Absolutely no reason at all, good point.

test/contracts-network/colony-network-recovery.js
area: The main place they went was 'in the bin' - skills only exist on the main chain now, and the only cross-chain stuff that exists is to do with receiving/paying funds out.

test/cross-chain/cross-chain.js
area: They're not tests, just setup that ultimately got moved in to `deploy-proxy-network-fixture` so it could be reused - but should have been deleted here.

test/extensions/staked-expenditure.js
area: I assume I found them as low-hanging fruit when trying to get coverage up.

contracts/bridging/ProxyColony.sol
area: I guess this comment predated the rebase you asked for (-:

contracts/colony/ColonyFunding.sol
area: Check if any are still valid post-rebase :)

contracts/colony/Colony.sol
area: It does. Everything that previously took a `token` as a parameter now takes a `chainId` and a `token` as a parameter, and this is the new version of `setExpenditurePayout`.

contracts/bridging/ProxyColony.sol
area: (Which we should decide on how to deal with)

contracts/colony/ColonyFunding.sol
area: Done, and the combination highlights that we don't have the equivalent checks for the cross-chain requests (which is bad).

contracts/common/DomainReceiverManagement.sol
area: Yep. The issue is that the function was in `ColonyNetworkDeployer`, but it's been moved here because we need it on both proxy- and non-proxy-chains, and so the changes didn't make it across when the rebase happened.

contracts/common/DomainReceiverManagement.sol
area: Should be sorted with the fix of `idempotentDeployDomainTokenReceiver`

contracts/common/DomainReceiverManagement.sol
area: Added. Let me know if you think anything else should be in there.

PR #1291, opened 2024-09-13
Allow funds to be sent directly to domains

area:
This is work that I've done while working on #1285 as a prerequisite for the required functionality of domains to be able to manage and exchange tokens cross-chain. For that functionality to work, I believe this functionality (or something equivalent) is required; where such an exchange is happening cross-chain, and we can't directly interrogate balances before and after the exchange on the other chain, we need a more generic way to send tokens to a domain directly in order to do the bookkeeping correctly. 

Because we can't send extra data with an ERC20 token transfer that's executed by other parties, I believe the _only_ way to do this is have a system where each domain has a unique address that can have tokens swept from it in to the corresponding domain.

Fortunately, we can (re-)use CreateX for this. By using a contract deployment salt that's a function of the colony address and domain id, we can deploy a helper contract (`DomainTokenReceiver`) to a fixed address for each domain (even across chains) with simple functionality to allow the sweep to occur.

I've tried to make this as transparent as possible - the deployment, upgrading, and management of the receiving contract is intended to not require any explicit interaction from the user. The experience should be that the users send tokens to a domain-specific address (which they can get from `getDomainTokenReceiverAddress` on `ColonyNetwork`) and then claim them (via `claimDomainFunds` on `Colony`) without having to check whether the contract is already deployed and/or up-to-date.

Tagging this as `ready-for-review` because I want feedback, but leaving as draft because I don't want it merged yet. I think the biggest question I have hanging over this implementation is what the split between the `Colony` and `ColonyNetwork` should be for the implementation but open to feedback from any and all directions.


area:
An open question here is what should happen in terms of the colony reward pot. I think I am of the opinion that whatever it is set to should be respected.

arrenv:
So, funds sent to the main colony address or specific domain addresses will also automatically go to the respective teams set by the rewards pot? Regardless if the funds are only sent to a specific team?

area:
> split amongst the respective teams set by the rewards pot

This isn't really describing what the reward pot does. It is a single pot, and the only configuration available related to it is the fraction of incoming funds that gets put in to it, and those funds will eventually be divided among all people who have both reputation and native tokens.

It's possible this is functionality that should be removed, but until that decision is made, anything other than respecting that configuration means that it can be totally bypassed (as instead of sending funds to a colony directly, you can send funds directly to the top-level domain).

arrenv:
> This isn't really describing what the reward pot does. It is a single pot, and the only configuration available related to it is the fraction of incoming funds that gets put in to it, and those funds will eventually be divided among all people who have both reputation and native tokens.
> 
> It's possible this is functionality that should be removed, but until that decision is made, anything other than respecting that configuration means that it can be totally bypassed (as instead of sending funds to a colony directly, you can send funds directly to the top-level domain).

Haha, yeah, just crisscrossed myself with the talk of teams. 

I agree with you on both. Perhaps in needs to be rethought a little or removed, but until that time, if it is configured, it seems to make sense that it can't be bypassed.

It just does not seem intuitive that it would be the case when a colony is transferring funds to itself, but, I understand that this is how this functionality would essentially work.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -175,6 +178,94 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
+  function checkDomainTokenReceiverDeployed(
+    uint256 _domainId
+  ) public stoppable calledByColony returns (address domainTokenReceiverAddress) {
+    // Calculate the address the domain should be receiving funds at
+    domainTokenReceiverAddress = getDomainTokenReceiverAddress(msgSender(), _domainId);
+
+    if (!isContract(domainTokenReceiverAddress)) {
+      // Then deploy the contract
+      bytes32 salt = getDomainTokenReceiverDeploySalt(msgSender(), _domainId);
+      address deployedAddress = ICreateX(CREATEX_ADDRESS).deployCreate3AndInit(
```
kronosapiens, line 198: Can we pull this into a helper function which only takes the salt? I've seen this code repeated a few times.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -175,6 +178,94 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
 
+  function setDomainTokenReceiverResolver(address _resolver) public stoppable auth {
+    domainReceiverResolverAddress = _resolver;
+  }
+
+  function getDomainTokenReceiverResolver() public view returns (address) {
+    return domainReceiverResolverAddress;
+  }
+
+  function checkDomainTokenReceiverDeployed(
```
kronosapiens, line 189: Can we choose a different naming convention for this type of idempotent deployment? I feel the same about `deployCreateXIfNeeded`. I'd be happy rolling this up into `getDomainTokenReceiverAddress` with a dev note that it deploys the contract idempotently if needed. `idempotentDeployDomainTokenReceiver` goes in that direction but is just as wordy so I don't know if that's a huge win. Open to your thoughts on this.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -175,6 +178,94 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
+    // Check it's got the right resolver
+    try EtherRouter(payable(domainTokenReceiverAddress)).resolver() returns (Resolver resolver) {
+      if (address(resolver) != domainReceiverResolverAddress) {
+        EtherRouter(payable(domainTokenReceiverAddress)).setResolver(domainReceiverResolverAddress);
+      }
+    } catch {
+      revert("colony-network-domain-receiver-not-etherrouter");
+    }
+
+    // Check it's set up correctly
```
kronosapiens, line 219: I'd feel comfortable rolling this up into the deploy block earlier. Realistically will this value ever be changed?

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,57 @@ contract ColonyFunding is
...
+    address domainTokenReceiverAddress = IColonyNetwork(colonyNetworkAddress)
+      .checkDomainTokenReceiverDeployed(_domainId);
+    uint256 fundingPotId = domains[_domainId].fundingPotId;
+    // It's deployed, so check current balance of pot
+    uint256 balanceBefore = getFundingPotBalance(fundingPotId, _token);
+
+    uint256 claimAmount;
+    uint256 thisBalanceBefore;
+
+    if (_token == address(0x0)) {
```
kronosapiens, line 121: Why not just have `transferToColony` return the token value? That would save you this before/after accounting.

kronosapiens (commented):
Looks good overall, will probably have some more comments on my next go-round but here are my first reactions.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: > I'd be happy rolling this up into getDomainTokenReceiverAddress with a dev note that it deploys the contract idempotently if needed.

I'm quite against this, as it makes it significantly less intuitive to use from the frontend side (as it's no longer a view function).

In terms of naming... I agree that it's much of a muchness. I will say I'm completely unfussed about `deployCreateXIfNeeded` though, as it's a non-contract function only used in tests.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: It shouldn't be, but arbitrary transactions can do arbitrary things. In general, however, I've been taking a bit more of a belt-and-braces approach when contracts at deterministic addresses are involved, hence leaving it as a separate block.

contracts/colony/ColonyFunding.sol
area: To be honest, the asserts post-calling `transferToColony` was mostly just me being careful / checking how things were going during the implementation - it's not like we do it in `claimColonyFunds`.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
kronosapiens: Ok, how about `idempotentDeployCONTRACTNAME` for these types of things? So `idempotentDeployCreateX` and `idempotentDeployDomainTokenReceiver`?

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,38 @@ contract ColonyFunding is
     emit ColonyFundsClaimed(msgSender(), _token, feeToPay, remainder);
   }
 
+  function claimDomainFunds(address _token, uint256 _domainId) public stoppable {
+    require(domainExists(_domainId), "colony-funding-domain-does-not-exist");
+    address domainTokenReceiverAddress = IColonyNetwork(colonyNetworkAddress)
```
kronosapiens, line 112: Thinking more about whether this deployment should go on the network or the colony, while putting it on the colony seems preferable to avoid the external call, if we're going to check the resolver each time then the external call will be inevitable, and if so probably best to keep all the deployment logic in one place.

area:
`isContract` is now being duplicated and could be refactored out.

EDIT: Sort-of on hold until Arren and Jack have thought more about what compromise is acceptable in terms of someone looping payments to create reputation and take over a colony.

kronosapiens:
Could say that internal tokens cannot be sent directly to domains?

area:
> Could say that internal tokens cannot be sent directly to domains?

Essentially, yes, this is what I was going to implement (or, more strictly, when such tokens are claimed, they go to the root domain) but there are concerns it will render the functionality less useful than they would like.

area:
The solution that has been decided on is to let root approve a certain amount of reputation-earning tokens to enter a domain (like ERC20). Any above the approved amount are sent to root.

The intention is that the typical in-domain swap (through the UI, at least) that would result in a domain receiving reputation-earning tokens would be a motion that requires root that makes the approval for an appropriate amount and then does the swap.

Motions being approved at the same time should be supported (or at least not break in a surprising way).

contracts/colony/Colony.sol
```diff
@@ -317,11 +317,11 @@ contract Colony is BasicMetaTransaction, Multicall, ColonyStorage, PatriciaTreeP
...
   function finishUpgrade() public always {
-    // Leaving as example for what is typically done here
-    // ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
-    // bytes4 sig;
-    // sig = bytes4(keccak256("cancelExpenditureViaArbitration(uint256,uint256,uint256)"));
-    // colonyAuthority.setRoleCapability(uint8(ColonyRole.Arbitration), address(this), sig, true);
+    ColonyAuthority colonyAuthority = ColonyAuthority(address(authority));
+    bytes4 sig;
+
+    sig = bytes4(keccak256("editAllowedDomainTokenReceipt(uint256,address,uint256,bool)"));
```
kronosapiens: I'd call this `approveDomainTokenReceipt` to keep in in line with the related ERC20 function

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,84 @@ contract ColonyFunding is
...
+      emit DomainFundsClaimed(msgSender(), _token, _domainId, feeToPay, approvedAmount);
+      emit ColonyFundsClaimed(msgSender(), _token, 0, remainder - approvedAmount);
+    }
+
+    // Claim funds
+
+    DomainTokenReceiver(domainTokenReceiverAddress).transferToColony(_token);
+  }
+
+  function tokenEarnsReputationOnPayout(address _token) internal view returns (bool) {
```
kronosapiens, line 166: I assume your plan is to generalize this once support for more reputation-bearing tokens is added?

Also, I'm wondering if it would be better to frame the approval not as an amount of tokens, but as an amount of reputation -- so an approval of 10k (for instance) would cover 100k of tokens if the reputation fraction was 1/10.

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,84 @@ contract ColonyFunding is
...
+    // initialised to MAX_UINT, and cannot be set to zero via setRewardInverse, so this is a false positive. It *can* be set
+    // to 0 via recovery mode, but a) That's not why MythX is balking here and b) There's only so much we can stop people being
+    // able to do with recovery mode.
+    uint256 remainder = claimAmount - feeToPay;
+    nonRewardPotsTotal[_token] += remainder;
+
+    fundingPots[0].balance[_token] += feeToPay;
+
+    uint256 approvedAmount = domainReputationTokenApprovals[block.chainid][_domainId][_token];
+    if (!tokenEarnsReputationOnPayout(_token) || approvedAmount >= remainder) {
```
kronosapiens: I took a stab at refactoring this, let me know what you think:

```
if (tokenEarnsReputationOnPayout(_token))
  const transferrableAmount = min(approvedAmount, remainder);
  const untransferrableAmount = remainder - transferrableAmount;

  fundingPots[fundingPotId].balance[_token] += transferrableAmount;
  domainReputationTokenApprovals[block.chainid][_domainId][_token] -= transferrableAmount;
  fundingPots[domains[1].fundingPotId].balance[_token] += untransferrableAmount;
} else {
  fundingPots[fundingPotId].balance[_token] += remainder;
}
```

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,84 @@ contract ColonyFunding is
...
+    // Claim funds
+
+    DomainTokenReceiver(domainTokenReceiverAddress).transferToColony(_token);
+  }
+
+  function tokenEarnsReputationOnPayout(address _token) internal view returns (bool) {
+    return _token == token;
+  }
+
+  function editAllowedDomainTokenReceipt(
```
kronosapiens: Why not mimic the ERC20 `approve` behavior and simply update the current value? It would make this function simpler. The front-end could easily render the current value and provide a helpful UX.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -175,6 +178,89 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
+    uint256 _domainId
+  ) public stoppable calledByColony returns (address domainTokenReceiverAddress) {
+    // Calculate the address the domain should be receiving funds at
+    domainTokenReceiverAddress = getDomainTokenReceiverAddress(msgSender(), _domainId);
+
+    if (!isContract(domainTokenReceiverAddress)) {
+      // Then deploy the contract
+      bytes32 salt = getDomainTokenReceiverDeploySalt(msgSender(), _domainId);
+      address deployedAddress = deployEtherRouterViaCreateX(salt);
+      require(
```
kronosapiens, line 199: Could make this an `assert` as it doesn't depend on user input

contracts/colonyNetwork/ColonyNetworkDeployer.sol
```diff
@@ -175,6 +178,89 @@ contract ColonyNetworkDeployer is ColonyNetworkStorage {
...
+    // Check it's set up correctly
+
+    if (DomainTokenReceiver(domainTokenReceiverAddress).getColonyAddress() != msgSender()) {
+      DomainTokenReceiver(domainTokenReceiverAddress).setColonyAddress(msgSender());
+    }
+
+    return domainTokenReceiverAddress;
+  }
+
+  function isContract(address addr) internal view returns (bool) {
```
kronosapiens: I'd move this lower down in the contract

contracts/common/DomainTokenReceiver.sol
```diff
@@ -0,0 +1,52 @@
...
+contract DomainTokenReceiver is DSAuth {
+  address resolver; // Storage slot 2 (from DSAuth there is authority and owner at storage slots 0 and 1 respectively)
+
+  address colony;
+
+  function getColonyAddress() public view returns (address) {
+    return colony;
+  }
+
+  function setColonyAddress(address _colony) public auth {
```
kronosapiens: Is there some way to make this immutable / set on construction / etc to avoid the check in the `Deployer` contract? It's a lot of code to handle a variable which will never be changed.

contracts/common/DomainTokenReceiver.sol
```diff
@@ -0,0 +1,52 @@
...
+    // Transfer the token to the colony.
+    if (tokenAddress == address(0)) {
+      // slither-disable-next-line arbitrary-send-eth
+      payable(colony).transfer(address(this).balance);
+      return;
+    } else {
+      require(
+        ERC20Extended(tokenAddress).transfer(
+          colony,
+          ERC20Extended(tokenAddress).balanceOf(address(this))
```
kronosapiens: I'd do this as

```
const balance = ERC20Extended(tokenAddress).balanceOf(address(this));
require(ERC20Extended(tokenAddress).transfer(colony, balance), "domain-token-receiver-transfer-failed")
```

test/contracts-network/colony-funding.js
```diff
@@ -549,5 +562,264 @@ contract("Colony Funding", (accounts) => {
...
+    it("when receiving native (reputation-earning) token, if no approval present for domain, all are received by root domain", async () => {
+      // Get address for domain 2
+      await colony.addDomain(1, UINT256_MAX, 1);
+      const receiverAddress = await colonyNetwork.getDomainTokenReceiverAddress(colony.address, 2);
+      await colony.mintTokens(WAD.muln(100));
+      await colony.claimColonyFunds(token.address);
+      const domain1 = await colony.getDomain(1);
+
+      // Pay the tokens to the domain
+      await setupClaimedExpenditure({
```
kronosapiens: Is this necessary? Cant you just give the user some native tokens directly?

contracts/colony/ColonyFunding.sol
area: > I assume your plan is to generalize this once support for more reputation-bearing tokens is added?

Yep

> Also, I'm wondering if it would be better to frame the approval not as an amount of tokens, but as an amount of reputation -- so an approval of 10k (for instance) would cover 100k of tokens if the reputation fraction was 1/10.

This feels like a good idea, we should run it past Arren/Jack but I assume they would be on board.

contracts/colony/ColonyFunding.sol
area: This was part of the original discussion; the worry with mimicking the approve functionality and there being two motions around the same domain where the second would overwrite the first approval before it had been used.

contracts/colony/Colony.sol
area: See later, but this was somewhat deliberate because it's not mimicking the `approve` functionality.

contracts/colony/ColonyFunding.sol
area: Hah, wow, I clearly completely spaced on `min` existing at all. This is much nicer, though will tweak it for event emission to be correct.

contracts/colonyNetwork/ColonyNetworkDeployer.sol
area: I disagree - while it doesn't depend on the user input, it does depend on the CreateX contract.

contracts/common/DomainTokenReceiver.sol
area: I don't believe so? It's an `EtherRouter`, so we don't have a constructor as an option, unless we make it its own special `EtherRouter` which I wouldn't be wild about doing. I'm open to suggestions though.

test/contracts-network/colony-funding.js
area: Just let me know how and I will :smile:

contracts/colony/ColonyFunding.sol
kronosapiens: @arrenv do you want to weigh in on this? The idea is that instead of approving a number of tokens to be sent to the domain, you approve an amount of reputation. The effect is that for tokens with only a small reputation factor (i.e. 10% or so) the approval goes further. It also means that for the user, approvals are always framed in terms of "how much reputation do I want to be able to send to this domain" regardless of the rate.

contracts/common/DomainTokenReceiver.sol
kronosapiens: What about making this callable once? I.e. adding a `require(address(colony) == address(0x0), 'error')`?

You could call it on initialisation and then skip the check on subsequent calls.

test/contracts-network/colony-funding.js
kronosapiens: Would `colony.mintTokensFor` not work? IIRC we've discussed this in the past and came up with an alternative to using expenditures, but maybe I'm hallucinating.

contracts/colony/ColonyFunding.sol
area: Yeah, we discussed this in the product call yesterday, and they agree it should be a reputation limit.

contracts/common/DomainTokenReceiver.sol
area: I like streamlining those checks 👍

test/contracts-network/colony-funding.js
area: Hah, fair play! I had no recollection of that function existing. It seems as if it was introduced for CoinMachine, which a) we're no longer really using, and b) no longer actually uses that function. I would like to propose deprecating and removing it, rather than starting to use it 😅

kronosapiens: I can live with that.

How about using `giveUserCLNYTokens` instead? We've definitely been here before...

I don't mean to be pedantic, it's just unnecessary overhead to have to parse an expenditure which is really just sending someone tokens.

contracts/colony/ColonyFunding.sol
kronosapiens: Once we add support for more reputation-bearing tokens this will need to be re-implemented along the lines of

```
reputationRemainder = wmul(remainder, reputationRate[token])

...
```

test/contracts-network/colony-funding.js
area: We _could_... it would mean that we'd have to switch the relevant tests to using the metacolony, though, which is somewhat against the intention of the `beforeEach`. 

We could do an arbitrary transaction to call `mint` on the token? (EDIT: I went ahead and did this, let me know what you think!)

I welcome the pedantry, please continue 😄

kronosapiens: Happy with your solution! Thanks for entertaining me.

kronosapiens (dismissed):
Looks good!

area:
As-is, this means domains can send money between each other arbitrarily (while paying a network fee), which we have prevented in the past. I don't believe it is possible to prevent this, but would like to be proven wrong.

kronosapiens:
From WP page 9: 

```
Only accounts holding the funding permission may move tokens; the rule is that they may move
tokens between any two pots in the subtree rooted in the domain in which they hold the permission.
It is the expectation that this permission will in many cases be given to an extension contract
implementing a specialized decision-making mechanism, such as the funding queue described in
Section 3.2.
```

The current rule is that the funding permission lets users move funds **bidirectionally** between any two domains in the subtree.

The new behavior would allow users to move funds **unidirectionally** between the domain holding the funds and any other domain in the colony.

Given that **intended** behaviors include a) paying _arbitrary addresses_ out of the domain and b) allowing _arbitrary addresses_ to send funds to any domain, I don't see how preventing a user from sending funds from a domain to any other domain is possible.

That said, I don't see this as a major issue. Given that the new affordance is only **unidirectional** and requires funding to have been placed in the domain by someone with the authority to do so, I think it's fine, even if unintended.

area:
Should this functionality work with locked tokens, @arrenv? As it stands, it does not.

arrenv:
> Should this functionality work with locked tokens, @arrenv? As it stands, it does not.

Good pickup, yes, it should work with locked tokens.

Although, I am curious how it would not be supported. It makes sense that anyone outside of the colony would not be able to transfer direct to a domain with a locked token, but that is consistent with locked token behavior and a colony could still send direct to it's own domains.

Or, is this specifically related to bridging, where the bridging contract is not able to transfer a locked token?

area:
> a colony could still send direct to it's own domains. 

It could transfer internally, but if funds were sent to a domain receving address, those addresses are not allowed to transfer tokens, and so the tokens could not be forwarded on to the colony.

arrenv:
> It could transfer internally, but if funds were sent to a domain receving address, those addresses are not allowed to transfer tokens, and so the tokens could not be forwarded on to the colony.

I see, well it would certainly be good for that to not be the case and for those domain addresses to be able to transfer locked tokens.

area:
I've changed the implemention, but it does unfortunately rely on updating `TokenAuthority`, where incredibly we've not given the colony permissions to call `transferFrom` on a locked token, so a new authority would need to be deployed for each of our tokens. This could potentially be done transparently when upgrading a colony, but I would caution against that because of the unreliability of determining whether a token is one of ours or not, and an incorrect match would at best cause this functionality to not work with locked tokens, and at worst would forever prevent a colony from upgrading.

Is there a reason we didn't give a colony those permissions, or was this an oversight? I could imagine that the answer to that question is legal, and not technical...

kronosapiens:
I don't remember how that decision was made -- most likely it predates me -- but I would guess it just wasn't a use-case that was thought about.

Would it be possible to do a "lazy updating" of authorities based on a colony's request?

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
...
+    uint256 feeToPay = claimAmount / getRewardInverse(); // ignore-swc-110 . This variable is set when the colony is
+    // initialised to MAX_UINT, and cannot be set to zero via setRewardInverse, so this is a false positive. It *can* be set
+    // to 0 via recovery mode, but a) That's not why MythX is balking here and b) There's only so much we can stop people being
+    // able to do with recovery mode.
+    uint256 remainder = claimAmount - feeToPay;
+    nonRewardPotsTotal[_token] += remainder;
+
+    fundingPots[0].balance[_token] += feeToPay;
+
+    uint256 approvedAmount = domainReputationApproval[_domainId];
```
kronosapiens, line 133: I'd move `uint256 fundingPotId = domains[_domainId].fundingPotId;` here

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
     emit ColonyFundsClaimed(msgSender(), _token, feeToPay, remainder);
   }
 
+  function claimDomainFunds(address _token, uint256 _domainId) public stoppable {
+    require(domainExists(_domainId), "colony-funding-domain-does-not-exist");
+    address domainTokenReceiverAddress = IColonyNetwork(colonyNetworkAddress)
+      .idempotentDeployDomainTokenReceiver(_domainId);
+    uint256 fundingPotId = domains[_domainId].fundingPotId;
+    // It's deployed, so check current balance of pot
```
kronosapiens: Can this comment be deleted?

contracts/colony/ColonyFunding.sol
kronosapiens: Can we add a comment to this function noting that it will be re-implemented to add support for multiple tokens at some point?

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
...
+
+    uint256 approvedAmount = domainReputationApproval[_domainId];
+
+    if (tokenEarnsReputationOnPayout(_token)) {
+      uint256 transferrableAmount = min(approvedAmount, remainder);
+      uint256 untransferrableAmount = remainder - transferrableAmount;
+
+      fundingPots[fundingPotId].balance[_token] += transferrableAmount;
+      domainReputationApproval[_domainId] -= transferrableAmount;
+      emit DomainFundsClaimed(msgSender(), _token, _domainId, feeToPay, transferrableAmount);
```
kronosapiens, line 141: Add newlines around event

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
...
+      if (untransferrableAmount > 0) {
+        fundingPots[domains[1].fundingPotId].balance[_token] += untransferrableAmount;
+        emit ColonyFundsClaimed(msgSender(), _token, 0, untransferrableAmount);
```
kronosapiens, line 145: Newlines before event emissions

kronosapiens, line 145: Shouldn't this be 1 here?

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
...
+      }
+    } else {
+      fundingPots[fundingPotId].balance[_token] += remainder;
```
kronosapiens, line 148: Is this branch redundant? I.e. if `untransferrableAmount == 0` then `remainder == transferrableAmount`

contracts/colony/ColonyFunding.sol
```diff
@@ -106,6 +107,83 @@ contract ColonyFunding is
...
+  }
+
+  function editAllowedDomainReputationReceipt(
+    uint256 _domainId,
+    uint256 _amount,
+    bool _add
+  ) public stoppable auth {
+    require(domainExists(_domainId), "colony-funding-domain-does-not-exist");
+    require(_domainId > 1, "colony-funding-root-domain");
+    if (_add) {
```
kronosapiens, line 178: I wouldn't be opposed to a comment here explaining the logic for this implementation. i.e. "We did it this way to allow for a sequence of independent motions to aggregate"

test/contracts-network/colony-funding.js
```diff
@@ -549,5 +563,283 @@ contract("Colony Funding", (accounts) => {
...
+      await colony.editAllowedDomainReputationReceipt(2, 20, false);
+      allowedReceipt = await colony.getAllowedDomainReputationReceipt(2);
+      expect(allowedReceipt).to.eq.BN(50);
+    });
+
+    it(`cannot editAllowedDomainReputationReceipt for the root domain`, async () => {
+      await checkErrorRevert(colony.editAllowedDomainReputationReceipt(1, 70, true), "colony-funding-root-domain");
+    });
+
+    it(`when receiving native (reputation-earning) token, if full approval present for domain,
```
kronosapiens: Kind of a nitpick but I would put this test before the partial case

contracts/common/DomainTokenReceiver.sol
```diff
@@ -0,0 +1,53 @@
...
+  function getColonyAddress() public view returns (address) {
+    return colony;
+  }
+
+  function setColonyAddress(address _colony) public auth {
+    require(colony == address(0), "domain-token-receiver-colony-already-set");
+    colony = _colony;
+  }
+
+  function transferNativeToColony() public onlyColony {
```
kronosapiens: Can we call this `transferChainNativeToColony`? I keep getting mixed up between the two senses of "native token"

kronosapiens (commented):
Looking good!

contracts/colony/ColonyFunding.sol
area: Have you mistaken which `if` statement this is the `else` for? If not, please explain in more detail :smile:

contracts/colony/ColonyFunding.sol
area: The third parameter for this event is the `fee`, whereas I think your suggestion stems from thinking it is `domainId`? If not, I'm not understanding what you're saying :sweat:

kronosapiens: Nope, you are correct, I thought it was the other event.

contracts/colony/ColonyFunding.sol
kronosapiens: Nope, you are correct

kronosapiens (approved):
Nicely done!

area:
Intending to review this again before merging it.

PR #1296, opened 2024-09-25
Move experimental ethers dependency with ExtendedNonceManager

area:
When ExtendedNonceManager moved, the dependency should have moved with it.

Not sure why this didn't get caught by linting, as it was flagged in my editor once I opened it, but that's not a question for now.


area:
Merging with superpowers.

PR #1298, opened 2024-10-01
Package upgrades, plus solidity upgrade

area:
This was just the result of a find/replace for the Solidity version and a `pnpm upgrade`, to see what broke. It seems fine?


kronosapiens (dismissed):
lgtm

kronosapiens (approved):

PR #1306, opened 2024-10-18
Base documentation off of compiled interface, not IContract file

### LINKED ISSUE #1272: Doc generation should result in all functions that are on contract
This issue has gotten worse now #1244 has been merged, but wasn't perfect before, either. The docs currently show functions that are explicitly defined in the interface file; they do not show functions that are present on the interface due to inheritance. This doesn't really obstruct the usefulness of the docs, as it's the explicitly defined functions that are usually of interest, but for completeness I think all functions should be present.

I would expect a solution to this to involve some combination of flattening the contract in question (`npx hardhat flatten`), and using the compiled ABI to work out which functions are actually visible on the contract, and then stitching together the comments from the source file as is currently done. I would find any solution that resulted in all functions that are callable on the contract being visible acceptable, though.

area:
Closes #1272.

Primarily this is a slight rework of the `docgen.js` script. Instead of just using the interface file for a contract to work out which functions should be in the docs, it uses the ABI and a flattened version of the source to generate the documentation of every function that is _actually_ on the contracts. In the event of a function being documented in multiple places, it picks the documentation that is the longest.

To get this to work cleanly, I've ended up introducing e.g.  `BasicMetaTransaction is IBasicMetaTransaction` and equivalents in various places, which seems like something we should have been doing anyway to ensure consistency. 

_NB Currently based on #1305_


area:
The main decision I've taken that I could have gone the other way on is how I've handled `getMetaTransactionNonce`. I've added `(IBasicMetaTransaction, BasicMetaTransaction)` to the `override` where it's relevant. This is because `IBasicMetaTransaction` defines `executeMetaTransaction` as an `external` function (implicitly, because it's an interface), whereas `BasicMetaTransaction` defines it as `public` (because it's used in `executeMetaTransaction` and we don't want an external call). Therefore, if both are inherited, we need to explicitly override both.

The alternative, I think, would be to move (most of) our `interface`s to `abstract contracts` and do the override ourselves in `BasicMetaTransaction`.

contracts/extensions/ColonyExtensionMeta.sol
```diff
@@ -19,13 +19,16 @@
 pragma solidity 0.8.27;
 pragma experimental ABIEncoderV2;
 
-import { BasicMetaTransaction } from "./../common/BasicMetaTransaction.sol";
+import { IBasicMetaTransaction } from "./../common/IBasicMetaTransaction.sol";
+import { IBasicMetaTransaction, BasicMetaTransaction } from "./../common/BasicMetaTransaction.sol";
```
kronosapiens, line 22: Duplicate import

contracts/extensions/IColonyExtension.sol
```diff
@@ -19,6 +19,7 @@
 pragma solidity 0.8.27;
 pragma experimental ABIEncoderV2;
 import { IBasicMetaTransaction } from "./../common/IBasicMetaTransaction.sol";
+import { IDSAuth } from "./../common/IDSAuth.sol";
```
kronosapiens: Is this meant to be used for something?

scripts/docgen.js
```diff
@@ -118,164 +159,183 @@ const generateMarkdown = ({ contractFile, templateFile, outputFile }) => {
...
-        console.error(`Warning: ${method.name} is missing a natspec @notice`);
-        foundError = true;
-      }
 
-      // Push the method and append natspec
-      methods.push({ ...method, natspec });
-    });
+        methods.push({ ...method, natspec });
+      });
+  }
```
kronosapiens, line 265: Is there any way to generalize all this repetitive logic? Helper functions? Another level of abstraction?

kronosapiens (commented):
Looks good!

contracts/extensions/IColonyExtension.sol
area: It's being used to make sure that the functions implemented via `DSAuth` are commented - by including this, when the contract is flattened, the interface is included, which is what has the comments.

scripts/docgen.js
area: I've given it a go, though I will note this was not a result of my changes! (The lines are flagged as new because they're indented compared to how they were before).

contracts/extensions/IColonyExtension.sol
kronosapiens: Can you add a comment to that effect? Otherwise it just looks like a redundant import

kronosapiens (approved):
📚

PR #1309, opened 2024-11-13
Create patch Selemela07 devcontainer.json

Selemela07:
configure patch and diff

Closes #[GH issue]
Implements [X] part for #[GH issue]


area:
Thanks for this PR, but are you able to explain what it does, and justify why it is something we should include in our codebase?

PR #1310, opened 2024-11-14
Add agent to arbitrary transaction event

### LINKED ISSUE #1210: The Arbitrary Transaction event should have an _agent

area:
Closes #1210

Not wild about the impelmentation I've ended up going with here, but in order to add the parameter to the event without changing other behaviour, I couldn't see a better way of doing it. Very open to suggestions that would be up to and including a complete do-over.


kronosapiens (approved):
lgtm

PR #1313, opened 2024-12-10
Add support for Blockscout 

### LINKED ISSUE #1286: Explore chain visualizers
We want to improve the development experience by making it easier to understand what is happening on-chain. We should explore some options and integrate the best ones.

### Suggestions

- [ ] [Ethernal](https://tryethernal.com/hardhat-block-explorer)
- [ ] [Tenderly](https://tenderly.co/)

kronosapiens:
Closes #1286

# Instructions:

## Starting Blockscout (1 tab)

1. Clone the Blockscout repo (https://github.com/blockscout/blockscout)
2. Cd into `blockscout/docker-compose` and run `docker compose -f hardhat-network.yml up` (`-d` to daemonize)

Blockscout will start running on `localhost:80` by default

## Starting Hardhat (2 tabs)

1. Run `npx hardhat node` as per usual
2. Run `npx hardhat deploy --network development` as per usual

At this point Blockscout will start picking up the transactions

# Todo:

Currently, contract verification is not working. This is as far as I've gotten:

```
>> npx hardhat verify --network development 0x777760996135F0791E2e1a74aFAa060711197777  --contract contracts/common/EtherRouter.sol:EtherRouter        
Successfully submitted source code for contract
contracts/common/EtherRouter.sol:EtherRouter at 0x777760996135F0791E2e1a74aFAa060711197777
for verification on the block explorer. Waiting for verification result...

We tried verifying your contract EtherRouter without including any unrelated one, but it failed.
Trying again with the full solc input used to compile and deploy it.
This means that unrelated contracts may be displayed on Etherscan...

hardhat-verify found one or more errors during the verification process:

Etherscan:
A network request failed. This is an error from the block explorer, not Hardhat. Error: Unexpected token '<', "<html>
<h"... is not valid JSON
```

I've debugged enough to confirm that the API is live, but for some reason the 2nd step (verifying with the full solc input) returns an HTML response, which breaks the verifier.

## Resources:

https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify

https://docs.blockscout.com/devs/verification/hardhat-verification-plugin

[Blockscout's issue tracking proxy contract support](https://github.com/blockscout/blockscout/issues/8758)


area:
The error in the second step is
```
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx/1.27.3</center>
</body>
</html>
```

i.e. the concatenation of all the solidity files is too big to be posted to the API (no idea who's enforcing that limit, though).

kronosapiens:
Suggestion is to try and validate contracts through the Blockscout UI directly

area:
> Suggestion is to try and validate contracts through the Blockscout UI directly

Well, to start there and if that works then potentially write our own script to validate automatically in a more judicious way than the plugin tries to.

area:
For anyone else testing this, if you run the `docker-compose` command, the containers have a restart policy of 'always', which means they will start whenever docker does. I only discovered this when something got in to a crash loop when I restarted my machine and started using all my CPU.

```
grep -irl "restart: always" | xargs sed -i.bak "s/restart: always/restart: on-failure/g"
```

run in the directory with the docker files will cause them to only restart on failure.

area:
https://medium.com/@shaiv/testnet-faucets-sucks-try-this-block-explorer-with-your-local-hardhat-f434384d5244

This is a lot lighter weight, and supports at least some of what we want... maybe it does enough? I did have to edit it to ignore warnings when verifying a contract 

```
-    if (compiledOutput.errors) {
+    if (compiledOutput.errors && compiledOutput.errors.filter(e => !e.formattedMessage.startsWith('Warning:')).length > 0) {
```

kronosapiens:
Noting that the Standard JSON Input for a contract build can be pulled out of Hardhat via:

```
> buildInfo = await hre.artifacts.getBuildInfo("contracts/colony/Colony.sol:Colony")
> require('fs').writeFileSync('input.json', JSON.stringify(buildInfo.input, null, 2));
```

kronosapiens:
Testing for contract-level verification, using the dependency contract at `contracts/testHelpers/ChainId.sol`

```
const ChainId = await ethers.getContractFactory("ChainId")
const chainId = await ChainId.deploy()
await chainId.deployed()
console.log("Contract deployed to:", chainId.address)
>>> Contract deployed to: 0x78876Ba649623A5cC3a370509fe4AD5aCa26dfB4
```

Screenshot 2025-01-22 at 4 00 36 PM

Was able to verify this contract, under the `paris` EVM version. For some reason, it gave a "partial match" whatever that means.

@area any idea why we're compiling for paris? I don't see that as an explicit setting in our hardhat config.

kronosapiens:
Now trying with Standard JSON Input:

```
buildInfo = await hre.artifacts.getBuildInfo("contracts/testHelpers/ChainId.sol:ChainId")
require('fs').writeFileSync('input.json', JSON.stringify(buildInfo.input, null, 2));
```

Then manually going into `input.json` and deleting all the sources except for 

```
    "contracts/testHelpers/ChainId.sol": {
      "content": "// SPDX-License-Identifier: GPL-3.0-or-later\n/*\n  This file is part of The Colony Network.\n\n  The Colony Network is free software: you can redistribute it and/or modify\n  it under the terms of the GNU General Public License as published by\n  the Free Software Foundation, either version 3 of the License, or\n  (at your option) any later version.\n\n  The Colony Network is distributed in the hope that it will be useful,\n  but WITHOUT ANY WARRANTY; without even the implied warranty of\n  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the\n  GNU General Public License for more details.\n\n  You should have received a copy of the GNU General Public License\n  along with The Colony Network. If not, see <http://www.gnu.org/licenses/>.\n*/\n\npragma solidity 0.8.27;\n\ncontract ChainId {\n  function getChainId() public view returns (uint256) {\n    return block.chainid;\n  }\n}\n"
    }
```

And then uploading the `input.json` to Blockscout. Now we have an exact match!

Screenshot 2025-01-22 at 4 16 56 PM

area:
> @area any idea why we're compiling for paris? I don't see that as an explicit setting in our hardhat config.

It's the [hardhat default](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/core/config/config-resolution.ts#L408). The setting is conservative to stop people accidentally deploying contracts compiled to chains that they do not run on, as adding new opcodes to chains requires a hardfork and happens slowly. CreateX, for example, is [only compiled for Paris](https://github.com/pcaversaccio/createx?tab=readme-ov-file#security-considerations).

kronosapiens:
Tried using `hardhat-verify` again. This time, I deleted every contract in `contracts/` _except_ for `ChainId.sol` and re-compiled:

```
const ChainId = await ethers.getContractFactory("ChainId")
const chainId = await ChainId.deploy()
await chainId.deployed()
console.log("Contract deployed to:", chainId.address)
>>> Contract deployed to: 0xefC85459223d1a30eFC4cEb01Af496A976A7D5b5
```

Then attempted to verify using command line, **which works:**

```
$ npx hardhat verify --network localhost 0xefC85459223d1a30eFC4cEb01Af496A976A7D5b5  --contract contracts/ChainId.sol:ChainId      
Successfully submitted source code for contract
contracts/ChainId.sol:ChainId at 0xefC85459223d1a30eFC4cEb01Af496A976A7D5b5
for verification on the block explorer. Waiting for verification result...

Successfully verified contract ChainId on the block explorer.
http://localhost:80/address/0xefC85459223d1a30eFC4cEb01Af496A976A7D5b5#code
```

Screenshot 2025-01-24 at 11 40 32 AM

kronosapiens:
Currently working on isolating the various contract subdirectories so we can iteratively compile and verify the contracts without hitting the file upload limit. The following subdirectories have been isolated (can compile repo without them):

- [x] bridging
- [ ] colony - used by `ColonyNetworkDeployer.sol`
- [x] colonyNetwork
- [ ] ~common~
- [ ] ~dataTypes (new)~
- [x] ens
- [x] extensions
- [ ] ~interfaces (new)~
- [ ] metaTxToken - used by `ColonyNetworkDeployer.sol`
- [ ] patriciaTree - used by `ColonyRewards.sol`
- [x] reputationMiningCycle
- [x] testHelpers
- [x] tokenLocking

PR #1194, opened 2023-12-04
Migrate test framework to Hardhat, cont.

kronosapiens:
Continuation of #1196 which upgrades the codebase to Ethers 6 and migrates to the use of an in-process node for testing

WIP

To do:

- Test reversions: https://hardhat.org/hardhat-chai-matchers/docs/reference#reverted-transactions
- Test events: https://hardhat.org/hardhat-chai-matchers/docs/reference#events
- Balance changes: https://hardhat.org/hardhat-chai-matchers/docs/reference#balance-change

PR #1316, opened 2024-12-19
Use in-hardhat process where possible

area:
Conceptually, the bulk of this work was done #1196 when we migrated to `hardhat`. As part of those changes (because of how `solidity-coverage` works with `hardhat`), to make our tests work where an external element had to connect to a node (such as a reputation miner) we exposed an `express` server on `8545`, and passed through any requests to the endpoint to the internal provider (an example of [standards](https://eips.ethereum.org/EIPS/eip-1193) working heavily in our favour!).

We only did this for the coverage tests, however - and there's no reason why it couldn't be done for the non-coverage tests. There are two incentives - speed (`pnpm run test:contracts` goes from about 2m00s to 1m45s locally for me), but also having access to the full hardhat integrations going forward. A small example of that is `console.log` in contracts, which I have been using intermittently debugging tests since the move to `hardhat`, but have always had to use the `:coverage` version of tests to use it, which obviously comes with a performance hit.

With no other changes, that would come with a cost - `hardhat` would exit immediately on a test failure, whereas with an external node, you can interrogate the chain state post-failure as part of your investigation. As part of our custom `test` task, I've hooked in to the return value which turns non-zero on a failure, and wait for a promise that never returns to allow such investigation to take place. 

I've keyed this exposure of the internal provider to whether the `hardhat` network is being used (and only exposing it if it is - if we're not using the `hardhat` network, we probably are using another provider, and don't want the ports to clash).

Finally, which checking all of the testing commands in `package.json` worked, I stumbled across the E2E tests which were very out-of-date and not run anywhere. These were run nightly, and disabled in the Circle `config.yml` at some point, and then removed from it entirely. I've spent some time resurrecting them. The comments seemed to suggest that they were helpful for testing dispute resolution for multiple types of additions / subtractions of reputation, but as far as I could tell they were primarily testing dispute resolution of decays. I've made it so that they're doing the former (which I think is more useful), and updated the various comments around `skillId`s (to what should be now accurate). I've also massively improved their performance, so they can be run with every build, as they are now quicker than some of the coverage tests.

Supercedes #1194


area:
How do we feel about merging our tests on CI that are quick? I'm looking at basically all of the non-coverage tests.

hardhat.config.js
```diff
@@ -35,6 +35,50 @@ task("compile", "Compile Colony contracts with pinned Token").setAction(async ()
...
+  for (const account of config.networks.hardhat.accounts) {
+    const { privateKey } = account;
+    const publicAddress = ethers.utils.computeAddress(privateKey);
+    ganacheAccounts.addresses[publicAddress] = publicAddress;
+    ganacheAccounts.private_keys[publicAddress] = privateKey;
+  }
+
+  fs.writeFileSync("ganache-accounts.json", JSON.stringify(ganacheAccounts, null, 2));
+
+  const nFails = await runSuper();
```
kronosapiens, line 75: Does `runSuper()` return the number of failures? Or is it more like an exit code?

hardhat.config.js
```diff
@@ -35,6 +35,50 @@ task("compile", "Compile Colony contracts with pinned Token").setAction(async ()
...
+  if (nFails > 0 && !process.env.CI && hre.network.name === "hardhat") {
```
kronosapiens, line 76: What do you think about adding an additional command-line argument (i.e. `stay-alive`) to toggle this behavior?

Otherwise I'm assuming people will have to manually kill the process once they're done investigating.

hardhat.config.js
```diff
@@ -56,26 +100,6 @@ task("deploy", "Deploy Colony Network as per truffle-fixture.js").setAction(asyn
   await deployNetwork();
 });
```
kronosapiens, line 102: Why remove the `coverage` task? Is it using the `test` task under-the-hood?

package.json
```diff
@@ -36,26 +36,26 @@
...
-    "test:contracts:bridging:1": "CHAIN_ID_1=265669100 CHAIN_ID_2=265669101 pnpm run start:blockchain:client:both && HARDHAT_FOREIGN=false MINING_CHAIN_ID=265669100 npx hardhat test ./test/cross-chain/* --network development",
-    "test:contracts:bridging:2": "CHAIN_ID_1=265669101 CHAIN_ID_2=265669100 pnpm run start:blockchain:client:both && HARDHAT_FOREIGN=true MINING_CHAIN_ID=265669100 CHAIN_ID=265669101 npx hardhat test ./test/cross-chain/* --network development",
-    "test:contracts:chainid": "pnpm run start:blockchain:client && npx hardhat test ./test/misc/chainid.js",
-    "test:contracts:e2e": "pnpm run start:blockchain:client && npx hardhat test test/misc/end-to-end.js --network development",
+    "test:contracts:upgrade:ganache": "pnpm run clean:test:contracts | pnpm run start:blockchain:client ganache & pnpm run generate:test:contracts && npx hardhat test ./test/upgrade/* --network integration",
+    "test:contracts:gas": "npx hardhat test test/misc/gas-costs.js",
+    "test:contracts:patricia": "pnpm run start:blockchain:client ganache && hardhat test packages/reputation-miner/patricia-test.js --network integration",
```
kronosapiens: Why ditch the `npx` here?

test/misc/end-to-end.js
```diff
@@ -280,98 +294,112 @@ contract("End to end Colony network and Reputation mining testing", function (ac
...
-        { id: 20, colonyIdx: 100, skillId: 3, account: WORKER, value: 1633 },
-        { id: 21, colonyIdx: 100, skillId: 504, account: WORKER, value: 1633 },
-        { id: 22, colonyIdx: 100, skillId: 505, account: WORKER, value: 797 },
-        { id: 23, colonyIdx: 100, skillId: 506, account: WORKER, value: 837 },
-        { id: 24, colonyIdx: 100, skillId: 507, account: WORKER, value: 0 },
-        { id: 25, colonyIdx: 100, skillId: 508, account: WORKER, value: 767 },
-        { id: 26, colonyIdx: 100, skillId: 509, account: WORKER, value: 70 },
-        { id: 27, colonyIdx: 100, skillId: 510, account: WORKER, value: 0 },
-        { id: 28, colonyIdx: 100, skillId: 511, account: WORKER, value: 0 },
+        { id: 1, colonyIdx: 100, skillId: 706, account: undefined, value: 1633 },
```
kronosapiens, line 325: These updates can't have been easy!

kronosapiens (commented):
Looking good

hardhat.config.js
area: I've not actually tried (we bail on the first failure, so it's only ever going to be 0 or 1) but I did [look at their code](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/builtin-tasks/test.ts#L141) which suggests it is the number of failures.

hardhat.config.js
area: They will, but that was the case before as well and hasn't been an issue?

hardhat.config.js
area: It does ([solidity-coverage call](https://github.com/sc-forks/solidity-coverage/blob/master/plugins/hardhat.plugin.js#L271), [hardhat definition](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/builtin-tasks/task-names.ts#L86)).

package.json
area: Don't remember it being deliberate, not sure what happened. Have restored.

hardhat.config.js
kronosapiens: What happens in this situation?

1. Run `npm run test:contracts`
2. Test fails, process keeps running
3. Edit some code
4. Run `npm run test:contracts`

Will the tests run again, or will we have to manually kill the process before re-running? In the previous setup, it was ok leaving the node running since the tests would just start over. But if the test run kicks off the process internally, that might change things.

kronosapiens: Tested this locally, confirmed that the test process doesn't stop, but it can be manually killed with Ctl+C once inspection is done.

kronosapiens (approved):

PR #1318, opened 2025-01-14
Update Dockerfiles for pnpm (probably)

area:
While trying to get #1312 deployed to QA for Raul, I discovered that (at least some of) our `Dockerfile`s were broken, and were seemingly refusing to install dependencies correctly for our packages. 

Couldn't really figure out what was going on, but putting everything in its own directory (following [the docs](https://pnpm.io/docker) more closely) seems to have solved it.


area:
Perhaps we should work on tests being able to be run using built docker images? (Not in this PR)

kronosapiens (approved):

PR #1321, opened 2025-01-30
Adjust robustness of MTX broadcaster

area:
While investigating issues that users were having with the metatransaction broadcaster, this is where we've arrived, fixing two issues:

* An error that would result in trying to hexlify a transaction response object, rather than a signed transaction. This was due to a confusion between `signer.sendTransaction` and `provider.sendTransaction` in the rebroadcasting logic.
* Some sensitivity to accidentally re-using nonces. We weren't really using the NonceManager functionality, so I've reworked the ExtendedNonceManager a fair bit to use it, and it's now much closer to the default NonceManager but with rebroadcasting logic. 

Reproducing issues around this determinsitically is hard; I've done my best with some tests, but I couldn't get the _exact_ errors we saw in production. These changes are currently live on QA, and will hit production before the next release (or maybe even before this is merged), so it's going to be an ongoing battle, but let's still review / improve these changes as normal!


hardhat.config.js
```diff
@@ -44,16 +44,23 @@ task("test", "Run tests").setAction(async () => {
...
       res.send({ jsonrpc: "2.0", result: response, id: req.body.id });
     } catch (error) {
-      const abiCoder = new ethers.utils.AbiCoder();
-      const decoded = abiCoder.decode(["string"], `0x${error.data.slice(10)}`);
-      res.send({
-        jsonrpc: "2.0",
-        error: { message: `Error: VM Exception while processing transaction: reverted with reason string '${decoded[0]}'` },
-        id: req.body.id,
-      });
+      console.log(error);
```
kronosapiens, line 52: Is this meant to be part of the final implementation?

hardhat.config.js
```diff
@@ -44,16 +44,23 @@ task("test", "Run tests").setAction(async () => {
...
+      try {
+        const abiCoder = new ethers.utils.AbiCoder();
+        const decoded = abiCoder.decode(["string"], `0x${error.data.slice(10)}`);
+        res.send({
```
kronosapiens: Minor point, but visually the two cases would read more similarly if you did it like:

```
const message = `Error: VM Exception while processing transaction: reverted with reason string '${decoded[0]}'`;
res.send({ jsonrpc: "2.0", error: { message }, id: req.body.id });
```

Vs. the current implementation where one is multi-line and the other is single-line

packages/package-utils/ExtendedNonceManager.js
```diff
@@ -9,8 +10,23 @@ class ExtendedNonceManager extends NonceManager {
       Object.keys(this.signedTransactions).map(async (txHash) => {
         const nodeTx = await this.signer.provider.getTransaction(txHash);
         if (!nodeTx) {
-          this.signer.provider.sendTransaction(this.signedTransactions[txHash]);
-          return;
+          const txCount = await this.signer.getTransactionCount("pending");
+          const parsedTransaction = parseTransaction(this.signedTransactions[txHash]);
+          if (parsedTransaction.nonce < txCount) {
+            // It's not been mined, but it's been replaced by another tx.
```
kronosapiens: Is there a chance there are simply multiple unmined transactions? Is it necessarily true that it's a duplicate tx? 

Or is the idea that since nonces must be processed in-order, the presence of a higher nonce means this transaction is no longer guaranteed to be mined, and so we replace it?

Might this create a condition where two transactions are continually leap-frogged past each other?

packages/package-utils/ExtendedNonceManager.js
```diff
@@ -9,8 +10,23 @@ class ExtendedNonceManager extends NonceManager {
...
+            delete this.signedTransactions[txHash];
+            this.sendTransaction({
+              from: parsedTransaction.from,
+              to: parsedTransaction.to,
+              value: parsedTransaction.value,
+              data: parsedTransaction.data,
+            });
+          } else {
+            // No reason to think it's been replaced, so rebroadcast.
+            this.signer.provider.sendTransaction(this.signedTransactions[txHash]);
```
kronosapiens: Is this for redundancy, or is there a chance the transaction was never broadcast in the first place?

packages/package-utils/ExtendedNonceManager.js
```diff
@@ -22,25 +38,23 @@ class ExtendedNonceManager extends NonceManager {
...
-
-    // At least whatever the endpoint says, or whatever we've already sent if higher
-    let nonce = await this.signer.getTransactionCount();
-
-    // Note the order we did the above two lines in - if a tx is mined between these two lines,
-    // and got removed by the `on block` handler above, by doing it in this order we won't be tripped up
-    // And we'll skip any nonces we've already used
-    while (pendingNonces.includes(nonce)) {
-      nonce += 1;
+    try {
```
kronosapiens, line 5: This is a pretty different implementation. One big difference is that we're no longer setting nonces unless there is a failure -- are we assuming that the `transactionRequest` comes with a nonce already set?

hardhat.config.js
area: Yeah, I guess this can be removed now that we have the try/catch.

packages/package-utils/ExtendedNonceManager.js
area: No, it's set by the NonceManager that we're extending during the `populateTransaction` call.

packages/package-utils/ExtendedNonceManager.js
area: I think we should be using `latest` and not `pending` here - does that change your assessment?

packages/package-utils/ExtendedNonceManager.js
area: We know that the node we're making the RPC call to doesn't know the transaction, so should be rebroadcast, however we arrived at that state.

area:
A demoralising update.

From the frontend side, we need to guarantee, as strongly as possible, that the hash returned is the hash of the transaction that does the metatransaction. The previous implementation, rebroadcasting the transaction if it did not succeed, does not guarantee that (as if we have to change the nonce, the hash will change). 

I've ended up at this much simpler implementation that refreshes the nonce every request (as a typical signer would), but retrying if an error relating to the nonce is seen when doing so, and only returning the transaction hash when that doesn't happen (which will be when the transaction is mined).

For cases where the broadcaster ends up behind, it'll catch-up, and for cases where the broadcaster ends up ahead it'll reset itself. The latter can't be tested properly locally currently, as `hardhat` returns an error if you use a too-big nonce, which is not typical (usually, such a transaction would sit in the mempool of a node until dropped or all the in-between nonces are filled in). That would all work with a typical signer - the custom retry functionality deals with situations where multiple requests are made simultaneously and we end up with nonce collisions.

The (previous) 'clever' version might be appropriate to use with the wormhole relayer - where we don't care about the transaction hash of something happening, we just care about it happening, but it can use this provider for now, as well, as far as I'm concerned.

kronosapiens (approved):
Looks good to me

